PR #1, opened 2025-03-07
feat: First interface draft

dianacarvalho1:
image


src/modules/market_graph.rs
```diff
@@ -0,0 +1,27 @@
+#[derive(Clone)]
+pub struct MarketGraph {
+    max_hops: usize,
+    // TODO: decide how the graph should look like
+    // vertices, edges, pool states, etc.
```
tamaralipows: These are questions I thought it would be nice to talk to Carlos about

src/modules/gas_price.rs
```diff
@@ -0,0 +1,5 @@
+pub async fn get_gas_price_task() {
+    // This needs to run in a loop.
+    // TODO: how often should the loop run? we will have very fast chains :/
```
tamaralipows, line 3: I would say to make this dependent on chain time. We also probably need to change the size of the market graph depending on chains

src/worker.rs
```diff
@@ -0,0 +1,43 @@
...
+pub struct TychoWorker {
+    protocols: Vec<String>,
+    tokens: Option<String>,
+    tvl_filter: f64,
+
+    graph: MarketGraph,
+    algorithm: Box<dyn Algorithm>,
+}
+
+impl TychoWorker {
```
tamaralipows: Do we know yet if it would be more beneficial to create a builder pattern here? Asking because we always end up doing that eventually it seems

src/modules/algorithm.rs
```diff
@@ -0,0 +1,105 @@
+use crate::modules::market_graph::MarketGraph;
+use num_bigint::BigUint;
+
+pub trait Algorithm {
+    fn get_route(&self);
+    // Sometimes this returns exactly 1 route (solver and TQ) and sometimes it returns multiple routes (MM)
+    fn get_route_amount_out(&self);
+    // maybe this could be a method of the Route object as well
+    // it should take an optional token price parameter. if it's passed, calculate the gas price in buy token and deduct from the amount_out
```
tamaralipows: Totally agree with this being in the route object!

src/modules/algorithm.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// One algorithm per use case :|
```
tamaralipows:
```suggestion
// One algorithm per use case :)
```

src/modules/algorithm.rs
```diff
@@ -0,0 +1,105 @@
...
+
+impl Algorithm for BalancePoolAlgorithm {
+    fn get_route(&self) {
+        println!(
+            "Getting best route from MarketGraph using MidPriceAlgorithm between a token pair"
+        );
+        // 1. Get routes between the two tokens
+        // 2. Calculate spot price of all pools
+        // 3. Return the pools where the price is out of balance
+        // TODO: is this algo even necessary? it isn't doing anything at this point tbh
```
tamaralipows: Also not totally sure this should be here. I would think this is in the highest level (guides?), and we use the `MidPriceAlgorithm`

src/guides.rs
```diff
@@ -0,0 +1,104 @@
...
+        println!("Findings pools that are out of balance and stabilizing them");
+        // 1. get all unbalanced routes
+        // 2. Compute the necessary swap to move the price back TODO: how can we do this?
+        // 3. encode/execute the trades self.executor(solution)
+    }
+}
+
+// TODO: who needs to know about gas?
+// TODO: Do the algorithms need direct access to the graph? this would mean having the share the graph with the worker (cloning like I'm doing won't work)
+// TODO: where should the caches be??
```
tamaralipows: caches in the market graph? I remember at least in python we had caches at every level, kindof horrific, but I guess it served a purpose?

src/guides.rs
```diff
@@ -0,0 +1,104 @@
...
+    pub fn stabilize_market() {
+        // infinite loop
+        println!("Findings pools that are out of balance and stabilizing them");
+        // 1. get all unbalanced routes
+        // 2. Compute the necessary swap to move the price back TODO: how can we do this?
+        // 3. encode/execute the trades self.executor(solution)
+    }
+}
+
+// TODO: who needs to know about gas?
```
tamaralipows: I would say in the `find_route` in the algorithm, right? After finding a route, subtract gas from the price?

src/modules/algorithm.rs
```diff
@@ -0,0 +1,105 @@
...
+    }
+}
+
+impl Algorithm for MidPriceAlgorithm {
+    fn get_route(&self) {
+        println!(
+            "Getting best route from MarketGraph using MidPriceAlgorithm between a token pair"
+        );
+        // 1. get routes between two tokens
+        // 2. simulate get_amount_out on the routes and get_amount_out_reverse
```
tamaralipows: subtract gas?

src/guides.rs
```diff
@@ -0,0 +1,104 @@
...
+// TODO: Do the algorithms need direct access to the graph? this would mean having the share the graph with the worker (cloning like I'm doing won't work)
```
tamaralipows: Why do the workers need the market graph?

tamaralipows (commented):
Looking good! Added some comments to discuss your questions!

src/modules/algorithm.rs
```diff
@@ -0,0 +1,105 @@
+use crate::modules::market_graph::MarketGraph;
+use num_bigint::BigUint;
+
+pub trait Algorithm {
```
kayibal: What is the point of formalizing the algorithm? Shouldn't we formalize the things the algorithm uses instead?

src/worker.rs
```diff
@@ -0,0 +1,43 @@
+use crate::modules::algorithm::Algorithm;
+use crate::modules::market_graph::MarketGraph;
+
+pub struct TychoWorker {
+    protocols: Vec<String>,
+    tokens: Option<String>,
+    tvl_filter: f64,
+
+    graph: MarketGraph,
+    algorithm: Box<dyn Algorithm>,
```
kayibal: I have the feeling not all Algos will need a MarketGraph

src/worker.rs
dianacarvalho1: yeah most likely that would make sense. but I think for now I won't add them so that it doesn't clutter the design. Will think about this in one of the last iterations

src/guides.rs
dianacarvalho1: yeah that's what I don't know yet. Haven't thought about it though. will do so today

src/guides.rs
dianacarvalho1: to update it with the messages from tycho indexer 😕

src/modules/algorithm.rs
dianacarvalho1: what do you mean?

src/modules/algorithm.rs
dianacarvalho1: I would actually go further and say we don't need any algorithm at all.. 🤔

src/modules/gas_price.rs
dianacarvalho1: > I would say to make this dependent on chain time

yeah but then the tricky thing is again synchronising our loop with the chain. for example for ethereum, sure we can run this every 12s but if it's not synced then 🤷🏼 
also I think that for faster chains, gas doesn't change soooo often, so maybe we can get away with not querying this too much

src/worker.rs
dianacarvalho1: Can you give an example please? or what makes you think that? (asking just to have more context)

src/modules/algorithm.rs
tamaralipows: not sure what you mean by this

dianacarvalho1: that for this use case we only need to retrieve all the 1 hop routes between the two tokens.
then checking if the price is out of balance is already the "guide" responsibility.. not really an algorithm I would say 🤔

src/guides.rs
```diff
@@ -0,0 +1,110 @@
...
+    algorithm: Box<dyn Algorithm>,
+    executor: Executor,
+}
+
+impl Solver {
+    pub fn new(
+        protocols: Vec<String>,
+        tokens: Option<String>,
+        tvl_filter: f64,
+        max_hops: usize,
```
Haikane: some algorithms might now have the concept of "hops" - e.g. a convex flow network optimiser. I'm not sure we should specify hops here. But instead leave it up to the algorithm. What we are trying to control with hops, I assume is search time (by controlling complexity), and that we already have as a parameter - it is then up to the algorithm how to achieve that search time (e.g. by restricting the hops it searches).

dianacarvalho1: Huumm I was thinking that hops would be relevant only to construct the routes in the market graph level. That the algorithm itself wouldn't be able to control the number of hops depending on time

src/guides.rs
```diff
@@ -0,0 +1,111 @@
...
+        let algorithm = SimpleAlgorithm::new(max_search_time);
+        let solver = TychoSolver::new(protocols, tokens, tvl_filter, graph, algorithm);
+        let executor = Executor {};
+        Solver { solver, executor }
+    }
+    pub fn solve(token_in: String, token_out: String, amount_in: BigUint) {
+        println!("Solving the trade");
+        // 1. get routes self.solver.get_route(token_in, token_out, amount_in)
+        // 2. encode/execute the trade self.executor(solution)
+    }
```
tamaralipows: I'm wondering if we even need `TychoSolver` if this method in `Solver` is basically passthrough at this point

dianacarvalho1: we need to maintain the market graph and connect to tycho indexer.. or do you think that should be done inside the algorithm?

dianacarvalho1:
closed in favour of https://github.com/propeller-heads/tycho-router/pull/2

PR #3, opened 2026-01-07
feat: Initial interfaces sketch

tvinagre:


docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,383 @@
...
+                                   ▼
+┌─────────────────────────────────────────────────────────────────────────────┐
+│                         SharedMarketData (Arc<RwLock<>>)                    │
+│  ┌────────────────────────────────────────────────────────────────────┐    │
+│  │  pools: HashMap<PoolId, PoolData>                                   │    │
+│  │    └── component: ProtocolComponent                                 │    │
+│  │    └── state: Box<dyn ProtocolSim>    ← Heavy data, never cloned   │    │
+│  │    └── tokens: Vec<Token>                                           │    │
+│  │  tokens: HashMap<Address, Token>                                    │    │
+│  │  route_graph: RouteGraph              ← Lightweight, clonable      │    │
```
louise-poole: Is this just another name for MarketGraph?

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,383 @@
...
+**Responsibility:** Single source of truth for all market data. Updated by TychoFeed only.
+
+```rust
+pub struct SharedMarketData {
+    pools: HashMap<PoolId, PoolData>,
+    tokens: HashMap<Address, Token>,
+    route_graph: RouteGraph,
+    gas_price: GasPrice,
+    gas_constants: HashMap<ProtocolSystem, u64>,
+    last_updated: Instant,
```
louise-poole: Do we maybe also want the block it was last updated at? Might be more useful than a timestamp (vulnerable to network delays etc making it appear as if we've recently updated, but we're actually a block behind?)

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,383 @@
...
+**File:** `src/route_graph.rs`
+
+**Responsibility:** Graph topology only. Cloned by Solvers for local optimization.
+
+```rust
+#[derive(Clone)]
+pub struct RouteGraph {
+    adjacency: HashMap<Address, Vec<Edge>>,
+    pool_tokens: HashMap<PoolId, Vec<Address>>,
+}
```
louise-poole: I assume this is MarketGraph and isn't limited to one route/token pair (RouteGraph sounds like it is imo)
Should we maybe use some existing graph crate here? Like UnGraph? (we use this in the NTQ)

EDIT: I think the shared state should be something simple like what you have here and inside the solver/algorithm they can build the graph as they wish with the crates they wish. E.g. MostLiquid would likely use UnGraph (undirected graph), but a cleverer algorithm might want to use a directed graph with spot prices as directed edge weights for optimising route searches.

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,383 @@
...
+```
+
+---
+
+### 6. TychoFeed
+
+**File:** `src/tycho_feed.rs`
+
+**Responsibility:** Connect to Tycho WebSocket, update SharedMarketData, broadcast events.
```
louise-poole: Technically it's not just the websocket we connect to/interact with. 
```suggestion
**Responsibility:** Connect to Tycho Client, update SharedMarketData, broadcast events.
```

docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,383 @@
...
+│
+├── types/                    # Shared type definitions
+│   ├── mod.rs
+│   ├── api.rs                # Request/Response types
+│   ├── solution.rs           # Solution, Route, Swap
+│   ├── internal.rs           # SolveTask, SolveError
+│   └── primitives.rs         # PoolId, Address, etc.
+│
+├── market_data.rs            # SharedMarketData
+├── route_graph.rs            # RouteGraph (clonable)
```
louise-poole: I think the lightweight market graph can be inside the shared market data, no? And then it's used by solvers to build their internal graphs. This makes it flexible so that solvers/algorithms can design their internal graphs however they want.

src/solver.rs
```diff
@@ -0,0 +1,264 @@
...
+/// A solver instance that processes solve requests.
+///
+/// Each worker thread owns one Solver. The solver maintains its own
+/// copy of the RouteGraph which it keeps synchronized with market
+/// events from the indexer.
+pub struct Solver {
+    /// Local copy of the route graph (can be pruned/optimized).
+    local_graph: RouteGraph,
+    /// Algorithm used for route finding.
+    algorithm: Box<dyn Algorithm>,
```
louise-poole: I have come fully circle here and think it now makes sense for the graph to be inside the Algorithm. Because different algorithms might want to use different graph structures.

src/solver.rs
```diff
@@ -0,0 +1,264 @@
...
+            orders: order_solutions,
+            total_gas_estimate,
+            solve_time_ms,
+        })
+    }
+
+    /// Returns the algorithm name.
+    pub fn algorithm_name(&self) -> &str {
+        self.algorithm.name()
+    }
```
louise-poole, line 257: Should get this from the internal algorithm and not from the solver level property

docs/ARCHITECTURE.md
tvinagre: Agreed, I'll replace this for Block struct

src/solver.rs
tvinagre: As discussed in a call today, Algorithm will be inside the Solver so I'm marking this as resolved

PR #9, opened 2026-01-09
refactor: add tycho-simulation dependency

zizou0x:
- Add tycho-simulation as dependency
- Use re-export instead of tycho-common
- Add cargo-audit ignore case


louise-poole (approved):
Yaaaas

PR #10, opened 2026-01-09
feat: implement GraphManager for petgraph's UnGraph

louise-poole:
Implement the GraphManager trait for petgraph's undirected graph.

### DIFF
--- a/src/feed/events.rs
+++ b/src/feed/events.rs
@@ -5,9 +5,13 @@
 
 use std::collections::HashMap;
 
+use thiserror::Error;
 use tycho_simulation::tycho_core::models::Address;
 
-use crate::types::{ComponentId, GasPrice};
+use crate::{
+    graph::GraphError,
+    types::{ComponentId, GasPrice},
+};
 
 /// Events broadcast by the indexer when market data changes.
 #[derive(Debug, Clone)]
@@ -23,8 +27,23 @@ pub enum MarketEvent {
     GasPriceUpdated { gas_price: GasPrice },
 }
 
+/// Errors that can occur when handling market events.
+#[derive(Error, Debug)]
+pub enum EventError {
+    /// Graph-related errors
+    #[error("graph errors: {0:?}")]
+    GraphErrors(Vec<GraphError>),
+    /// Invalid event data.
+    #[error("invalid event: {0}")]
+    InvalidEvent(String),
+}
+
 /// Trait for components that can receive market events.
 pub trait MarketEventHandler {
     /// Handle a market event.
-    fn handle_event(&mut self, event: &MarketEvent);
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the event could not be processed.
+    fn handle_event(&mut self, event: &MarketEvent) -> Result<(), EventError>;
 }
--- a/src/feed/events.rs
+++ b/src/feed/events.rs
@@ -3,47 +3,24 @@
 //! The indexer broadcasts these events when market data changes.
 //! Solvers subscribe to these events to keep their local graph in sync.
 
+use std::collections::HashMap;
+
 use tycho_simulation::tycho_core::models::Address;
 
-use crate::types::{ComponentId, GasPrice, ProtocolSystem};
+use crate::types::{ComponentId, GasPrice};
 
 /// Events broadcast by the indexer when market data changes.
 #[derive(Debug, Clone)]
 pub enum MarketEvent {
-    /// A new component was added to the market.
-    ComponentAdded {
-        component_id: ComponentId,
-        tokens: Vec<Address>,
-        protocol_system: ProtocolSystem,
+    /// Market was updated.
+    MarketUpdated {
+        added_components: HashMap<ComponentId, Vec<Address>>,
+        removed_components: Vec<ComponentId>,
+        updated_components: Vec<ComponentId>,
     },
-
-    /// A component was removed from the market.
-    ComponentRemoved { component_id: ComponentId },
-
-    /// A component's state was updated (reserves changed, etc.).
-    /// Solvers should re-read the state from SharedMarketData if needed.
-    StateUpdated { component_id: ComponentId },
 
     /// Gas price was updated.
     GasPriceUpdated { gas_price: GasPrice },
-
-    /// Full market snapshot.
-    /// Sent to new subscribers for initial synchronization.
-    Snapshot { components: Vec<ComponentSummary>, gas_price: GasPrice },
-}
-
-/// Summary of a component for snapshot events.
-#[derive(Debug, Clone)]
-pub struct ComponentSummary {
-    pub id: ComponentId,
-    pub tokens: Vec<Address>,
-    pub protocol_system: ProtocolSystem,
-}
-
-impl ComponentSummary {
-    pub fn new(id: ComponentId, tokens: Vec<Address>, protocol_system: ProtocolSystem) -> Self {
-        Self { id, tokens, protocol_system }
-    }
 }
 
 /// Trait for components that can receive market events.
--- a/src/graph/petgraph.rs
+++ b/src/graph/petgraph.rs
@@ -78,8 +78,9 @@ impl GraphManager<UnGraph<Address, ComponentId>> for PetgraphUnGraphManager {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
     use std::str::FromStr;
+
+    use super::*;
 
     /// Helper function to create a test address from a hex string.
     fn addr(s: &str) -> Address {
--- a/src/graph/petgraph.rs
+++ b/src/graph/petgraph.rs
@@ -104,6 +104,19 @@ impl PetgraphGraphManager {
     /// Helper function to add edges for all token pairs in a component.
     /// Takes a slice of node indices corresponding to the tokens.
     fn add_component_edges(&mut self, component_id: &ComponentId, node_indices: &[NodeIndex]) {
+        // Special case: if only 1 node, create a self-loop edge
+        if node_indices.len() == 1 {
+            let node_idx = node_indices[0];
+            let edge_idx =
+                self.graph
+                    .add_edge(node_idx, node_idx, EdgeData::new(component_id.clone()));
+            self.edge_map
+                .entry(component_id.clone())
+                .or_default()
+                .push(edge_idx);
+            return;
+        }
+
         // Create bidirectional edges for each token pair
         for (i, &from_idx) in node_indices.iter().enumerate() {
             for &to_idx in node_indices.iter().skip(i + 1) {
@@ -128,13 +141,16 @@ impl PetgraphGraphManager {
         }
     }
 
+    /// Adds components to the graph.
+    ///
+    /// If a component has no tokens, it is skipped.
+    ///
+    /// Arguments:
+    /// - components: A map of component IDs to their tokens.
     fn add_components(&mut self, components: &HashMap<ComponentId, Vec<Address>>) {
         for (comp_id, tokens) in components {
-            if tokens.len() < 2 {
-                warn!(
-                    "Skipping component {} with too few tokens (need at least 2 for edges)",
-                    comp_id
-                );
+            if tokens.is_empty() {
+                warn!("Skipping component {} with no tokens (need at least 1)", comp_id);
                 continue;
             }
             // Ensure all tokens are added as nodes (or get existing ones) and collect their indices
--- a/src/graph/petgraph.rs
+++ b/src/graph/petgraph.rs
@@ -1,11 +1,19 @@
-//! Petgraph's UnGraph implementation of GraphManager.
+//! Petgraph's StableDiGraph implementation of GraphManager.
 //!
-//! This module provides PetgraphUnGraphManager, which implements GraphManager for
-//! petgraph::graph::UnGraph, providing a reusable implementation for algorithms that use petgraph.
+//! This module provides PetgraphStableDiGraphManager, which implements GraphManager for
+//! petgraph::stable_graph::StableDiGraph, providing a reusable implementation for algorithms that
+//! use petgraph.
+//!
+//! A stable graph is a graph that maintains the indices of its edges even after removals. This is
+//! useful for optimising the graph manager's performance by allowing for O(1) edge and node
+//! lookups.
 
 use std::collections::{HashMap, HashSet};
 
-use petgraph::graph::{DiGraph, EdgeIndex, NodeIndex};
+use petgraph::{
+    graph::{EdgeIndex, NodeIndex},
+    stable_graph::StableDiGraph,
+};
 use tycho_simulation::tycho_core::models::Address;
 
 use super::GraphManager;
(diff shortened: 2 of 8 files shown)


src/graph/petgraph.rs
```diff
@@ -3,49 +3,581 @@
...
+        // Check if node already exists
+        if let Some(node_idx) = self
+            .graph
+            .node_indices()
+            .find(|&idx| self.graph[idx] == *addr)
+        {
+            node_idx
+        } else {
+            // Create new node if it doesn't exist
+            self.graph.add_node(addr.clone())
```
zizou0x: How does this internal method work if the node already exist?

src/graph/petgraph.rs
```diff
@@ -3,49 +3,581 @@
...
+                        .add_edge(to_idx, from_idx, EdgeData::new(component_id.clone()));
+                self.edge_map
+                    .entry(component_id.clone())
+                    .or_default()
+                    .push(edge_idx_ba);
+            }
+        }
+    }
+
+    fn add_components(&mut self, components: &HashMap<ComponentId, Vec<Address>>) {
```
zizou0x: Can we have docs here and below even if it's simple please?

src/graph/petgraph.rs
```diff
@@ -3,49 +3,581 @@
...
+        for (comp_id, tokens) in components {
+            if tokens.len() < 2 {
+                warn!(
```
zizou0x: This is really unexpected no? Would it make sense to return a Result and let the caller decide what to do with it? (panic or just log)

src/graph/petgraph.rs
```diff
@@ -3,49 +3,581 @@
...
         Self::new()
     }
 }
 
-#[allow(unused_variables)]
-impl GraphManager<petgraph::graph::UnGraph<Address, ComponentId>> for PetgraphUnGraphManager {
-    fn initialize_graph(&mut self, components: &HashMap<ComponentId, Vec<Address>>) {
-        unimplemented!("initialize_graph is not implemented for PetgraphGraphManager");
+impl GraphManager<DiGraph<Address, EdgeData>> for PetgraphGraphManager {
+    // TODO: make a type alias for the component topology
```
zizou0x: Is it still needed? What was the motivation behind it? I think a HashMap is simple enough no?

src/graph/petgraph.rs
```diff
@@ -3,49 +3,581 @@
...
+        }
     }
 
-    fn graph(&self) -> &petgraph::graph::UnGraph<Address, ComponentId> {
+    fn graph(&self) -> &DiGraph<Address, EdgeData> {
         &self.graph
     }
+}
 
+impl MarketEventHandler for PetgraphGraphManager {
```
zizou0x: Don't you think we should return a `Result<>` here to let the "solver engine" or "worker" (not sure about the name for this component yet) handle it?

zizou0x (commented):
Just a few comments, otherwise looks good!

src/graph/petgraph.rs
louise-poole: `add_node` always creates a new node. So if we call it for a token already in the graph it'll add a new node with the same token address label.

src/graph/petgraph.rs
louise-poole: Hmmm... I can add logic to handle a loop edge (to and from the same node). And then we only warn for components with 0 tokens? 
I didn't want to return an error in this case as this function handles multiple components and returning an error indicates the whole operation failed instead of succeeding with just one skipped component. In the past we've used this practice of just logging a warning for skipped operations. But I'm open to changing this if we feel strongly?

src/graph/petgraph.rs
louise-poole: Motivation was that I was typing out the full type multiple times and got annoyed 😂

src/graph/petgraph.rs
louise-poole: Sure, we can do that. Currently the only error we get is if we try remove a component that doesn't exist. It made sense to log and ignore this case as we're wanting to remove it anyway and it won't break the continued operation of the system. But we can for sure propogate the error and leave it to the solver to decide to ignore this.

src/graph/petgraph.rs
zizou0x: A component in Tycho should always have two different tokens AFAIK, else it doesn't allow for any route which is the purpose of it. 

So I think as it is now we should `warn!` (maybe even `error!`) if we have a component with `HashSet<Address>::len()<2`

I don't mind logging and skipping for now if it makes the interfaces cleaner

src/feed/events.rs
```diff
@@ -3,47 +3,24 @@
...
 /// Events broadcast by the indexer when market data changes.
 #[derive(Debug, Clone)]
 pub enum MarketEvent {
-    /// A new component was added to the market.
-    ComponentAdded {
-        component_id: ComponentId,
-        tokens: Vec<Address>,
-        protocol_system: ProtocolSystem,
+    /// Market was updated.
+    MarketUpdated {
```
tvinagre, line 20: I was going to propose this change - way better and scalable :)

src/graph/petgraph.rs
```diff
@@ -3,49 +3,581 @@
...
+    feed::events::{MarketEvent, MarketEventHandler},
+    graph::GraphError,
+    types::ComponentId,
+};
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum EdgeWeight {
+    Depth(f64),
+    SpotPrice(f64),
+}
```
kayibal, line 31: Why not have both?

src/graph/petgraph.rs
```diff
@@ -3,49 +3,581 @@
...
 ///
-/// This struct implements GraphManager for petgraph::graph::UnGraph, providing
-/// a reusable implementation for algorithms that use petgraph.
+/// This struct implements GraphManager for petgraph::graph::UnGraph.
 ///
 /// The graph manager maintains the graph internally and updates it based on market events.
-pub struct PetgraphUnGraphManager {
-    // Undirected graph with token addresses as nodes and edges as possible swaps.
-    graph: petgraph::graph::UnGraph<Address, ComponentId>,
+pub struct PetgraphGraphManager {
```
kayibal: I don't like this name, can we have something more expressive?

src/graph/petgraph.rs
```diff
@@ -3,49 +3,581 @@
...
+use crate::{
+    feed::events::{MarketEvent, MarketEventHandler},
+    graph::GraphError,
+    types::ComponentId,
+};
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum EdgeWeight {
+    Depth(f64),
+    SpotPrice(f64),
```
tvinagre, line 29: Does it make sense to use SpotPrice as weight, since it's not normalized for all the routes?
Also, should we specify the WeightType here or just make the weights an arbitrary f64 value that is not tied to an Enum??

src/graph/petgraph.rs
```diff
@@ -3,49 +3,581 @@
...
+        Self { graph: DiGraph::new(), edge_map: HashMap::new() }
+    }
+
+    /// Helper function to find a node index by address
+    fn find_node(&self, addr: &Address) -> Result<NodeIndex, GraphError> {
+        self.graph
+            .node_indices()
+            .find(|&idx| self.graph[idx] == *addr)
+            .ok_or_else(|| GraphError::TokenNotFound(addr.clone()))
+    }
```
kayibal, line 102: This is potentially expensive, no since it iterates over all tokens. We could do the same as we do with components here...

src/graph/petgraph.rs
```diff
@@ -3,49 +3,581 @@
...
+            self.graph.add_node(addr.clone())
+        }
+    }
+
+    /// Helper function to add edges for all token pairs in a component.
+    /// Takes a slice of node indices corresponding to the tokens.
+    fn add_component_edges(&mut self, component_id: &ComponentId, node_indices: &[NodeIndex]) {
+        // Create bidirectional edges for each token pair
+        for (i, &from_idx) in node_indices.iter().enumerate() {
+            for &to_idx in node_indices.iter().skip(i + 1) {
```
kayibal: Isn't there a more expressive way to do this? E.g. a combinations of 2 iterator?

src/graph/petgraph.rs
kayibal: we need to make really sure to not create loops, so we need to have a check for this here.

src/graph/petgraph.rs
```diff
@@ -3,49 +3,581 @@
...
+            .edge_map
+            .get(component_id)
+            .ok_or_else(|| GraphError::ComponentNotFound(component_id.clone()))?;
+
+        let mut updated = false;
+        for &edge_idx in edge_indices {
+            // Skip if edge endpoints are not found
+            let (edge_from, edge_to) = match self.graph.edge_endpoints(edge_idx) {
+                Some(endpoints) => endpoints,
+                None => continue,
```
kayibal, line 261: Under what conditions can this happen?

src/graph/petgraph.rs
```diff
@@ -3,49 +3,581 @@
...
+
+            // Determine if we should update this edge based on edge tokens and bidirectional flag
+            let should_update = if bidirectional {
+                // Update both directions
+                (edge_from == from_idx && edge_to == to_idx) ||
+                    (edge_from == to_idx && edge_to == from_idx)
+            } else {
+                // Update only forward direction
+                edge_from == from_idx && edge_to == to_idx
+            };
```
kayibal, line 272: I'm not quite sure under what circumstances this would be the case? Are you assuming here that we can express that a component is only quotable in certain pairs?

src/graph/petgraph.rs
```diff
@@ -3,49 +3,581 @@
...
-#[allow(unused_variables)]
-impl GraphManager<petgraph::graph::UnGraph<Address, ComponentId>> for PetgraphUnGraphManager {
-    fn initialize_graph(&mut self, components: &HashMap<ComponentId, Vec<Address>>) {
-        unimplemented!("initialize_graph is not implemented for PetgraphGraphManager");
+impl GraphManager<DiGraph<Address, EdgeData>> for PetgraphGraphManager {
+    // TODO: make a type alias for the component topology
+    fn initialize_graph(&mut self, component_topology: &HashMap<ComponentId, Vec<Address>>) {
+        // Clear existing graph and component map
+        self.graph = DiGraph::new();
+        self.edge_map.clear();
```
kayibal: Isn't it an error if you call this twice?

src/graph/petgraph.rs
kayibal: Do we need this type? It requires allocating a new Vec for each component and it is less expressive than e.g. having the component.

kayibal (commented):
I wonder how we can account for the fact that some components only support a subset of trade directions. And those may potentially change at any time. I am accounting for this by adding this method to the state: `fn quotable_pairs(&self) -> Vec<(TokenAddress, TokenAddress)>;`

src/graph/petgraph.rs
tvinagre: Yeah I also dislike this silent error handling, feels anti-rust. I would rather either fail if we encounter a component with less than 2 tokens (which is unlikely) OR return a list of skipped components and the caller would decide whether to log this or do something else

tvinagre: Since we're erroring the whole function if we try to delete a component that is not found below, I'd fail here as well

src/graph/petgraph.rs
louise-poole: We can for sure! I was leaving that sort of detail for when we implement the algorithm (can easily add exactly what the algorithm needs here). For now I understand we're just using spot price in the initial implementation though.

src/graph/petgraph.rs
louise-poole: @kayibal Why do we need to avoid loops? Is it to counter for a naive algorithm that would try explore loops when they do not benefit it?
My thoughts were to leave the graph implementation basic and have a 'optimise_graph' fn on the algorithm that removes things the algorithm might not handle well (like loops for a naive algorithm etc). Other smarter algorithms know how to handle loops.
I didn't want to limit the graph itself in case we have a component one day with a single token. Like in a certain situation it's beneficial to loop through a certain contract for some reason, even if it doesn't convert any tokens for you.

src/graph/petgraph.rs
louise-poole: We wanted to keep it lightweight so that it's easily clonable. I suppose having a HashSet of ProtocolComponents works too, it'll just contain a lot more data than we need here. Which do we prefer?
@zizou0x Maybe we can remove the 'topology' field entirely and just have 2 hashmaps for the components: one for its static data and one for its dynamic state? Right now they're [combined](https://github.com/propeller-heads/tycho-solver/blob/main/src/feed/market_data.rs#L34). (also I see Token is repeated in the component and in the global shared state. It only needs to be in the shared state no?)

zizou0x: I don't have anything against it 👍

> (also I see Token is repeated in the component and in the global shared state. It only needs to be in the shared state no?)

I left both so far because I didn't know which one was the most appropriate and we have things relying on it, this is going to change with my polishing when I'll be implementing the Tycho feed logic

src/graph/petgraph.rs
louise-poole: Yes - as is the case with Lido no? 
Also want to cover the case where the graph's weight might be different depending on the direction (depth etc). So we ned to be able to update the edge weight in a specified direction in that case.

src/graph/petgraph.rs
louise-poole: If the edge was removed from the graph but not from our edge indices map (so was removed during optimisation/graph pruning and not by a market event).

src/graph/petgraph.rs
louise-poole: No - not necessarily. A solver might hyper-optimise a graph for speed, but lose some lessor connected pools in the process. Then they might want to start from scratch and reinitialise and optimise the graph every x blocks to 'refresh' including pools that previously weren't optimal. I am trying to make sure to not limit the use cases of this implementation.

src/graph/petgraph.rs
louise-poole: Does it need to be normalised? If it is a weight then the built-in algorithms can use it when finding top paths by multiplying them together. In a direct comparison of paths between 2 tokens, we will see a final price in out_token which is useful.
Also I liked the idea of the Enum as we can then define combinations of values like:
DepthAndSpotPrice(f64, f64)
Where one value might be updated independent of the other value and the combined result is used as the algorithm weight:
```
    /// Extracts the numeric weight value for use in algorithms.
    pub fn as_f64(&self) -> f64 {
        match self {
            EdgeWeight::Depth(v) | EdgeWeight::SpotPrice(v) => *v,
            EdgeWeight::DepthAndSpotPrice(depth, spot_price) => depth * spot_price,
        }
    }
```

src/graph/petgraph.rs
```diff
@@ -1,51 +1,873 @@
...
+                self.add_edge(from_idx, to_idx, component_id);
+                self.add_edge(to_idx, from_idx, component_id);
+            });
+    }
+
+    /// Adds components to the graph.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if any components have no tokens (components must have at least 1 token).
```
zizou0x: Wouldn't a component with only 1 token be a "dangling" edge?

zizou0x (approved):
LGTM

src/graph/petgraph.rs
tvinagre: Yeah I also didn't get it - used to be min 2 tokens in the past commits - why the change?

tvinagre (approved):

src/graph/petgraph.rs
louise-poole: See my comment here: https://github.com/propeller-heads/tycho-solver/pull/10#discussion_r2685515381
Though I'm happy to remove it?

And no - it's not a dangling edge... it just uses the same token as the start and end point so is a self-loop.

PR #11, opened 2026-01-09
feat: polish market data implementation

zizou0x:
Remove unnecessary fields, replace placeholders with actual tycho-simulation structs and other small improvements.

### DIFF
--- a/src/feed/market_data.rs
+++ b/src/feed/market_data.rs
@@ -64,7 +64,9 @@ impl SharedMarketData {
             tokens: HashMap::new(),
             component_topology: HashMap::new(),
             gas_price: GasPrice::default(),
-            gas_constants: GAS_COST_PER_SWAP.clone(),
+            gas_constants: GAS_COST_PER_SWAP.clone(), /* TODO: maybe we want a way to override
+                                                       * this? if not we can use the constants
+                                                       * directly in the algorithm. */
             last_updated: Block::default(),
         }
     }
--- a/src/feed/market_data.rs
+++ b/src/feed/market_data.rs
@@ -48,8 +48,6 @@ pub struct SharedMarketData {
 /// Data for a single component.
 #[derive(Debug)]
 pub struct ComponentData {
-    /// Unique identifier.
-    pub id: ComponentId,
     /// Protocol component information.
     pub component: ProtocolComponent,
     /// Protocol simulation object.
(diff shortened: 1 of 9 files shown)


src/feed/market_data.rs
```diff
@@ -34,74 +39,34 @@ pub struct SharedMarketData {
...
-        gas_constants.insert(ProtocolSystem::Balancer, 150_000);
-        gas_constants.insert(ProtocolSystem::Other, 150_000);
-
         Self {
             components: HashMap::new(),
             tokens: HashMap::new(),
             component_topology: HashMap::new(),
             gas_price: GasPrice::default(),
-            gas_constants,
+            gas_constants: GAS_COST_PER_SWAP.clone(),
```
louise-poole: If it's static does it need to be on the shared market data? Can this not just be a constant that all solvers can access directly without needing to read/touch the shared market data?

src/feed/market_data.rs
```diff
@@ -34,74 +39,34 @@ pub struct SharedMarketData {
...
     gas_constants: HashMap<ProtocolSystem, u64>,
     /// When the data was last updated.
     last_updated: Block,
 }
 
 /// Data for a single component.
+#[derive(Debug)]
 pub struct ComponentData {
     /// Unique identifier.
     pub id: ComponentId,
```
louise-poole: Is ID necessary? It's duplicated in the ProtocolComponent and in the 'components' Hashmap key above?

src/feed/market_data.rs
zizou0x: My plan was to leave it here for now in case we want to allow overwriting this through config. I'm not sure yet if it will be possible but I would like to avoid having to build a new solver image when we need to add a protocol.

If it ends up not being possible I'll clean this up 👍

src/feed/market_data.rs
zizou0x: Indeed it's duplicated! I removed it

src/feed/market_data.rs
louise-poole: Sounds good. Please maybe add a TODO for this? Or a comment suggesting allowing overwrites/configuration?

louise-poole: Though adding a new protocol will require a new build anyway for adding it to the ProtocolStream builder.

louise-poole (approved):

src/feed/market_data.rs
zizou0x: > Though adding a new protocol will require a new build anyway for adding it to the ProtocolStream builder.

This is what I'll try to find a solution for, if the protocol doesn't rely on a custom filter function I think there is a chance we can just add it dynamically with a config but I'm not sure yet

PR #12, opened 2026-01-09
feat: implement TychoFeedBuilder for TychoFeed creation and configuration

zizou0x:
Added a new builder pattern for TychoFeed to enforce required fields at compile time using the typed-builder crate.


src/feed/builder.rs
```diff
@@ -0,0 +1,96 @@
...
+/// Builder for TychoFeed using typestate pattern to enforce required fields.
+///
+/// # Required fields (enforced at compile time)
+/// - `tycho_url`
+/// - `tycho_api_key`
+/// - `rpc_url`
+/// - `protocols`
+/// - `min_tvl` and `max_tvl`
+/// - `market_data`
+/// - `health_tracker`
```
louise-poole: Maybe the required fields can be args for the builder constructor? Then it's impossible to have a builder without these set no? That's what we've been doing till now.
Or does this builder crate only do it like this? Not a huge issue, just wondering.

src/feed/builder.rs
```diff
@@ -0,0 +1,96 @@
...
+    rpc_url: String,
+
+    /// Protocols to index (required).
+    protocols: Vec<String>,
+
+    /// Minimum TVL filter in ETH (required).
+    min_tvl: f64,
+
+    /// Maximum TVL filter in ETH (required).
+    max_tvl: f64,
```
louise-poole: These should be better described. Maybe it should be worded as 'tvl_threshold' or 'tvl_buffered_threshold' and can be a tuple/other struct. The min and max are not filters, but are distinct add and removal thresholds which is unclear here.

src/feed/builder.rs
louise-poole: Or the required fields can be in a FeedConfig struct? That is passed to the builder constructor? That way we can extend the required fields without breaking the builder constructor's interface.

zizou0x: I've decided to remove the builder as it wasn't adding much value here. Instead I'm using the config directly to create a feed.

src/feed/builder.rs
zizou0x: I've change this to be `min_tvl` and optionally `tvl_buffer_multiplier`, I think it's gonna be much clearer than the tvl range for users. I've also improved comments.

louise-poole (approved):
LGTM!

src/feed/mod.rs
```diff
@@ -1,3 +1,93 @@
...
+    /// Tycho API key.
+    pub(crate) tycho_api_key: String,
+    /// Names of the protocols to index.
+    /// For example, "uniswap_v2", "uniswap_v3", "sushiswap", etc.
+    pub(crate) protocols: Vec<String>,
+    /// TVL filter in native token, usually ETH.
+    /// Components with TVL below this threshold will be ignored/removed from the market data.
+    pub(crate) min_tvl: f64,
+    /// Multiplier used to define the upper bound of the TVL filter.
+    /// The upper bound is calculated as `min_tvl * tvl_buffer_multiplier`.
```
tvinagre, line 22: Why use multiplier here instead of explicitly specify the bounds, like we do on the standard interface?

tvinagre (approved):

tvinagre (approved):
LGTM - small comment

src/feed/mod.rs
zizou0x: Because most of our users get it wrong even though it's properly documented on tycho-client interface. 

The reason is because specifying a TVL range is usually interpreted as: whatever is within this range will be indexed. So user were specifying i.e `TvlRange(10.0,100000.0)` which led to no component returned and they got confused with this multiple times.

The multiplier makes it much clearer (or at least they wouldn't know what to expect from it so they would have to read docs). IMO it should also be changed on the tycho-client interface too.

tvinagre: Hmm yeah I understand the motivation but I'm not sure I agree with the solution. If the user is misunderstanding the parameter, then I agree there is something that we need to do, but I'd like to not move away from our main interface, this just adds another parameter that the user needs to understand and that in the end does the same thing as the previous, with less precision / less control.

zizou0x: Moving away from absolute values makes it possible to set it to a default values (set to 1.1 right now) which is the main motivation here, normal users shouldn't have to deal with this anymore, they can rely on the default and only set `min_tvl`

PR #13, opened 2026-01-09
feat: Improve swap interfaces

tvinagre:


src/types/solution.rs
```diff
@@ -1,61 +1,220 @@
...
-    /// Actual output amount.
+    /// Amount of output token (in token units, as decimal string).
+    #[serde(with = "biguint_as_string")]
     pub amount_out: BigUint,
-    /// Estimated gas for this order's swaps.
+    /// Estimated gas cost for executing this route (as decimal string).
+    #[serde(with = "biguint_as_string")]
     pub gas_estimate: BigUint,
-    /// Price impact in basis points (if calculable).
+    /// Price impact in basis points (1 bp = 0.01%).
```
tvinagre:
```suggestion
    /// Price impact in basis points (1 bip = 0.01%).
```

src/types/api.rs
```diff
@@ -25,47 +26,46 @@ pub struct SolutionOptions {
...
-    /// Returns true if this is an exact-out order.
-    pub fn is_exact_out(&self) -> bool {
-        self.amount_out.is_some()
-    }
+/// Specifies what kind of order this is.
+///
+/// Currently only `Sell` is supported. `Buy` will be added in a future version.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum OrderKind {
```
louise-poole: For some reason the wording 'order kind' feels weird to me. Maybe we can call it OrderSide? Or even just OrderType if you want to be more broad?

src/types/api.rs
```diff
@@ -25,47 +26,46 @@ pub struct SolutionOptions {
...
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub slippage_bps: Option<u16>,
 }
 
-impl Order {
-    /// Returns true if this is an exact-in order.
-    pub fn is_exact_in(&self) -> bool {
-        self.amount_in.is_some()
-    }
+/// Generate a unique order ID.
```
louise-poole: Technically we can't guarantee uniqueness here... 
```suggestion
/// Generate a random order ID.
```

src/types/mod.rs
```diff
@@ -14,18 +14,25 @@ pub mod serde_helpers;
 pub mod solution;
 
 // Re-export primitive types
```
louise-poole: This comment is just hanging now?

louise-poole (approved):
A couple minor comments, otherwise looks good!

src/types/solution.rs
```diff
@@ -1,61 +1,220 @@
...
+    /// Optional solving parameters that apply to all orders.
+    #[serde(default)]
+    pub options: SolutionOptions,
+}
+
+/// Options to customize the solving behavior.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct SolutionOptions {
+    /// Timeout in milliseconds. If `None`, uses server default.
+    pub timeout_ms: Option<u64>,
```
zizou0x, line 45: Why do we allow for this? Doesn't it make whoever hosts this vulnerable to DOS?

src/types/solution.rs
```diff
@@ -1,61 +1,220 @@
...
+}
+
+/// Specifies whether an order is a sell (exact input) or buy (exact output).
+///
+/// Currently only `Sell` is supported. `Buy` will be added in a future version.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum OrderKind {
+    /// Sell exactly the specified amount of the input token.
+    Sell,
```
zizou0x, line 118: Shouldn't we also follow the in-out convention here instead of sell/buy?

zizou0x (approved):

src/types/api.rs
zizou0x: Theoretically we can have a collision but practically chances can be considered 0 IMO.

src/types/solution.rs
louise-poole: So like 'ExactIn' and 'ExactOut'? Though 'exact' implies precision and we might one day support partial fills... I think in this situation 'buy' and 'sell' nomenclature is pretty standard no?

src/types/api.rs
tvinagre: Yeah I prefer OrderKind. OrderType can be super broad e.g: "fill-or-kill" could fit under orderType - will change

src/types/solution.rs
tvinagre: How would this happen? I will add a server-side min & max_timeout so the user can't set a timeout that would either be too low or too high. DOS needs to be prevented with a rate-limiter

src/types/solution.rs
tvinagre: IMO, Sell order or Buy order is a pretty standard financial interface and it's more clear than In / Out

PR #14, opened 2026-01-13
refactor: rename Solver as SolverWorker

louise-poole:
For clarity we will refer to the entire system as a 'Solver' and the individual internal solving processes as 'SolverWorkers'.


zizou0x (approved):
Thanks!

ElusAegis (approved):
👍

PR #15, opened 2026-01-13
feat: impl most liquid algorithm

ElusAegis:


src/graph/petgraph.rs
```diff
@@ -302,6 +306,72 @@ impl PetgraphStableDiGraphManager {
...
+    /// * `to` - Destination token address
+    /// * `min_hops` - Minimum number of hops (1 = at least one swap like A→B)
+    /// * `max_hops` - Maximum number of hops (2 = up to A→B→C, 3 = up to A→B→C→D)
+    ///
+    /// # Returns
+    /// All paths with hop count in range [min_hops, max_hops], ordered by hop count (BFS).
+    ///
+    /// # Errors
+    /// Returns `InvalidHopRange` if min_hops > max_hops.
+    pub fn find_paths(
```
louise-poole: See this is algorithm specific and should be done in the algorithm, not in the graph manager. The graph manager's sole purpose is to create and maintain the graph. The algorithm should be the one to explore the graph and create the routes/paths.

src/algorithm/mod.rs
```diff
@@ -55,7 +66,7 @@ pub trait Algorithm: Send + Sync {
     /// The best route found, or an error if no route could be found.
     fn find_best_route(
         &self,
-        graph: &Self::GraphType,
+        graph: &Self::GraphManager,
```
louise-poole: This should stay as GraphType. The algorithm should get a graph and then explore it using its algorithm. The graph manager should not do any of the path finding.

src/types/solution.rs
```diff
@@ -116,7 +116,8 @@ impl Order {
 pub enum OrderSide {
     /// Sell exactly the specified amount of the input token.
     Sell,
-    // TODO: Add Buy variant for exact output orders
+    /// Buy exactly the specified amount of the output token.
+    Buy,
```
louise-poole: This is currently unsupported by execution. Do we want it here as an orderside option already?

src/algorithm/mod.rs
```diff
@@ -99,3 +114,272 @@ pub enum AlgorithmError {
     #[error("{0}")]
     Other(String),
 }
+
+/// Simulates swaps along a path using each pool's `ProtocolSim::get_amount_out`.
+/// Tracks intermediate state changes to handle routes that revisit the same pool.
+fn simulate_path(
+    path: Path,
+    graph: &StableDiGraph,
```
louise-poole: We need to keep this `algorithm/mod.rs` file generic... any of this graph type specific fns should be inside the specific implementations (most_liquid). Actually it looks like all this info you read from the graph was included in the original `Path` struct. Why did you decide to change the struct? If you keep the old `Path` struct you can drop this graph arg and keep this fn in our generic mod file here.

louise-poole (commented):
Didn't have much time so just did a quick first review. Will come back and review properly tomorrow!

src/algorithm/test_utils.rs
```diff
@@ -0,0 +1,303 @@
...
+
+/// Mock ProtocolSim that multiplies input by a configurable factor.
+///
+/// Each call to `get_amount_out` returns a new state with an incremented multiplier,
+/// simulating liquidity changes after a swap. This allows testing state override logic
+/// when the same pool is used multiple times in a path.
+///
+/// Optionally supports liquidity limits - if `liquidity` is set, swaps exceeding it fail.
+#[derive(Debug, Clone)]
+pub struct MockProtocolSim {
```
zizou0x, line 38: It seems we double implemented this, we should centralise a version of this somewhere

src/algorithm/mod.rs
```diff
@@ -87,6 +98,10 @@ pub enum AlgorithmError {
     #[error("insufficient liquidity on all paths")]
     InsufficientLiquidity,
 
+    /// Gas cost exceeds output amount.
+    #[error("gas cost exceeds output amount")]
+    GasExceedsOutput,
```
zizou0x: Why are we considering this as an error? From a user perspective there are possible cases where I'd be fine with this

src/algorithm/mod.rs
```diff
@@ -99,3 +114,272 @@ pub enum AlgorithmError {
     #[error("{0}")]
     Other(String),
 }
+
+/// Simulates swaps along a path using each pool's `ProtocolSim::get_amount_out`.
+/// Tracks intermediate state changes to handle routes that revisit the same pool.
+fn simulate_path(
```
zizou0x: Would be nice to add some spans here

src/algorithm/mod.rs
```diff
@@ -99,3 +114,272 @@ pub enum AlgorithmError {
...
+        let token_out = market
+            .get_token(address_out)
+            .ok_or_else(|| AlgorithmError::Other(format!("token not found: {:?}", address_out)))?;
+
+        let component_id = &edge_data.component_id;
+        let component_data = market
+            .get_component(component_id)
+            .ok_or_else(|| {
+                AlgorithmError::Other(format!("component not found: {}", component_id))
+            })?;
```
zizou0x: We keep the lock of `market` during the full execution only for these?

As it seems that we are anyway doing the full iteration, I think we could first get all the data that we need, then we release the lock and then do our simulations.

src/algorithm/most_liquid.rs
```diff
@@ -121,84 +74,124 @@ impl Default for MostLiquidAlgorithm {
...
-            // Update best if this is better
-            let is_better = best_route
+            // Calculate net output (output - gas cost in wei)
+            // TODO: Convert gas cost to output token terms for proper ranking.
+            // Currently subtracting raw wei from output amount, which is incorrect when
+            // token_out != ETH. Need to:
+            // 1. Store ETH price per token (token/ETH rate) - likely in SharedMarketData
+            // 2. Look up ETH price for the output token of this path
+            // 3. Convert: gas_cost_in_token_out = gas_cost_wei * eth_price_in_token_out
+            let gas_price = market.gas_price().effective_gas_price();
```
zizou0x: As we hold the lock, this won't change during execution. So we can only pull it once instead of for each path

src/algorithm/most_liquid.rs
```diff
@@ -121,84 +74,124 @@ impl Default for MostLiquidAlgorithm {
...
-                }
+                best = Some((route, net_output));
             }
         }
 
-        best_route
-            .map(|(route, _)| route)
-            .ok_or(AlgorithmError::InsufficientLiquidity)
+        // Log solve result
+        stats.log_result(self.name(), best.as_ref(), market, &amount_in);
```
zizou0x: Same idea than in `simulate_path`, would be nice if we could release the lock as early as possible. 

That being said, here it's less straightforward, so if it makes the code too complex, we can start like this, leave a TODO and improve later once we have some metrics

src/algorithm/most_liquid.rs
```diff
@@ -35,81 +36,33 @@ impl MostLiquidAlgorithm {
...
+    /// This accounts:
+    /// - Spot price: the theoretical exchange rate along the path not accounting for slippage
+    /// - Fees: deducted per hop as `price *= (1 - fee)`
+    /// - Depth (inertia): minimum depth acts as a liquidity bottleneck indicator
+    ///
+    /// Returns `None` if edge weights are missing required data (spot_price, depth).
+    /// Higher score = better path candidate. Paths through deeper, lower-fee pools rank higher.
+    fn score_path(path: &Path, graph: &StableDiGraph) -> Option<f64> {
+        let mut price = 1.0;
+        // If path is empty, return max score
```
zizou0x: What's the reasoning behind this? Why an empty path would get max score?

src/algorithm/mod.rs
```diff
@@ -99,3 +103,112 @@ pub enum AlgorithmError {
...
+
+        // Select the correct state for simulation, using override if we've swapped through this
+        // pool, and otherwise the original state stored in market data
+        // TODO - is this stable for the VM states?
+        let state = state_overrides
+            .get(component_id.as_str())
```
tvinagre: There could be components that can interfere with state from other components (like components that share a token pool).
This could lead to a case where a swap on component A could change a further swap on component B.

In this case, I think it's safer to always apply all the state overrides.

src/algorithm/most_liquid.rs
```diff
@@ -35,81 +36,33 @@ impl MostLiquidAlgorithm {
...
-                protocol_system,
-                token_in,
-                token_out,
-                current_amount.clone(),
-                amount_out.clone(),
-            ));
-
-            current_amount = amount_out;
+    /// Formula: `weight = (product of all [spot_price × (1 - fee)]) × min(depths)`
+    ///
```
tvinagre, line 201: Didn't we update spot_price to always account for fees?

src/algorithm/mod.rs
tvinagre: I was reviewing a previous version of this PR - I've updated and this comment is outdated now

src/algorithm/most_liquid.rs
ElusAegis: Right, thank you for pointing to this!

src/algorithm/mod.rs
ElusAegis: The reason I switched to `GraphManager` is that when we do `find_paths` we need to find the start and end nodes in the graph, and the `GraphManager` has extra information that could do it quicker. However, I agree that it is not worth breaking the agreement that the `GraphManager` should only handle updates. So I switched to the Graph and just left a TODO to consider how we can optimize this.

src/algorithm/test_utils.rs
ElusAegis: Yes, ideally we could move it to tycho core, or decide to use one of the real protocol sims instead

src/algorithm/mod.rs
ElusAegis: Got rid of that, thank you! Instead, the net amount out could be negative (amount out - gas fees), but the user will never see this

src/algorithm/mod.rs
ElusAegis: I have neglected the spans so far, will do this in the next commit 👍

src/algorithm/mod.rs
ElusAegis: Sounds good. We would need to also change the Algorithm to not recieve the market as a reference, but take ownership of the lock, to allow dropping it and freeing up the writes. Will address this in the next revision of the Algo part

src/algorithm/most_liquid.rs
ElusAegis: Initially I thought to support empty path and add them a max score as not doing anything is always a good place to start (especially for arbitrage), but since then I have made the empty path impossible to be generated. And if they were, the score will filter them out.

src/algorithm/most_liquid.rs
ElusAegis: Makes sense, thank you!

src/algorithm/most_liquid.rs
ElusAegis: The use of the market here is minimal, so I assume that should be quite possible

src/algorithm/most_liquid.rs
```diff
@@ -1,115 +1,322 @@
...
+                let mut new_path = current_path.clone();
+                new_path.add_hop(&graph[current_node], edge.weight(), &graph[next_node]);
+
+                if next_node == to_idx && new_path.len() >= min_hops {
+                    paths.push(new_path.clone());
+                }
+
+                queue.push_back((next_node, new_path));
+            }
+        }
```
louise-poole, line 193: Can't we use `graph.all_simple_paths(from_idx, to_idx, self.max_hops)` to simplify and optimise this code?

src/algorithm/most_liquid.rs
```diff
@@ -1,115 +1,322 @@
...
+    /// # Arguments
+    /// * `path` - The edge path to simulate
+    /// * `graph` - The graph containing edge and node data
+    /// * `market` - Market data for token/component lookups and gas price
+    /// * `amount_in` - The input amount to simulate
+    fn simulate_path<D>(
+        path: Path<D>,
         market: &SharedMarketData,
         amount_in: BigUint,
     ) -> Result<Route, AlgorithmError> {
```
louise-poole, line 249: Now that this is pretty generic (not graph specific) you could have it in `mod.rs`

src/graph/petgraph.rs
```diff
@@ -23,78 +21,70 @@ use crate::{
...
-        }
-    }
-}
-
-/// Edge data containing both component ID and weight.
-#[derive(Debug, Clone)]
-pub struct EdgeData {
+/// Data stored on each edge of the graph.
+///
+/// Contains the component ID (which pool/protocol this edge represents) and
```
louise-poole:
```suggestion
/// Contains the component ID (which pool this edge represents) and
```

src/algorithm/most_liquid.rs
ElusAegis: As far as I remember, `graph.all_simple_paths` returned the node paths, and we need all edge paths. But please correct if I am wrong here

src/types/solution.rs
```diff
@@ -219,19 +220,79 @@ pub struct BlockInfo {
...
+    }
+}
+
+impl Ord for Route {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        // Higher net_amount_out is better
+        self.net_amount_out
+            .cmp(&other.net_amount_out)
     }
+}
```
zizou0x: I hesitant regarding using std traits here, the implementations aren't expected IMO so maybe we should use custom functions that would be clearly documented instead?

src/algorithm/most_liquid.rs
```diff
@@ -1,114 +1,332 @@
...
-        Self { max_hops: 3, timeout: Duration::from_millis(50) }
+        Self { min_hops: 1, max_hops: 3, timeout: Duration::from_millis(50) }
     }
 
     /// Creates a new MostLiquidAlgorithm with custom settings.
-    pub fn with_config(max_hops: usize, timeout_ms: u64) -> Self {
-        Self { max_hops, timeout: Duration::from_millis(timeout_ms) }
+    pub fn with_config(min_hops: usize, max_hops: usize, timeout_ms: u64) -> Self {
+        Self { min_hops, max_hops, timeout: Duration::from_millis(timeout_ms) }
     }
```
zizou0x: Shouldn't we error here if `min_hops > max_hops`?

src/algorithm/most_liquid.rs
```diff
@@ -1,114 +1,332 @@
...
+        max_hops: usize,
+    ) -> Vec<Path<'a, DepthAndPrice>> {
+        use std::collections::VecDeque;
+
+        use petgraph::visit::EdgeRef;
+
+        // Early return for invalid hop range
+        if min_hops > max_hops {
+            return vec![];
+        }
```
zizou0x: This is a misconfiguration and would mean solutions will always be empty until the config changes, correct? 

If yes we should do more than silently fail here, ideally error at the configuration level IMO (see comment above). Logging an error here would also be nice for debugging if it ever happens

src/algorithm/most_liquid.rs
```diff
@@ -1,114 +1,332 @@
...
+        let from_idx = graph
+            .node_indices()
+            .find(|&n| &graph[n] == from);
+        let to_idx = graph
+            .node_indices()
+            .find(|&n| &graph[n] == to);
+
+        let (from_idx, to_idx) = match (from_idx, to_idx) {
+            (Some(f), Some(t)) => (f, t),
+            _ => return vec![],
```
zizou0x: Should we also add a log here? 

One thing that I'm noticing is that any `vec![]` is associated to a `AlgorithmError::NoPath`. I think it would be valuable to know a bit more, for example here we could accurately define which one of the two tokens wasn't found. For this you'll have to adapt the interface to return a `Result<>`

zizou0x (approved):
Looks nice! Thanks for this!

src/types/solution.rs
ElusAegis: Sounds good, have no strong opinion here, given that we only use it in one place. Also, this enforces a single ordering, which might be not quite true in the long term

UPD: Just realised that the implementation for the Patial Equal actually does not make much sense

src/algorithm/most_liquid.rs
ElusAegis: Makes sense, will use the Other error type here

src/types/solution.rs
ElusAegis: Got rid of this, to avoid confusion

src/algorithm/mod.rs
ElusAegis: Spans + metrics were added

src/algorithm/mod.rs
ElusAegis: started working on this in the #20

PR #16, opened 2026-01-14
feat: implement Tycho Feed

zizou0x:
This PR adds the core implementation of the TychoFeed component which connects to TychoStream, updates the SharedMarketData and broadcast events through the event channel.

I've left a few TODOs that are for future code improvements, happy to discuss about them

### DIFF
--- a/src/feed/events.rs
+++ b/src/feed/events.rs
@@ -6,7 +6,7 @@
 use std::collections::HashMap;
 
 use thiserror::Error;
-use tycho_simulation::tycho_core::models::Address;
+use tycho_simulation::tycho_common::models::Address;
 
 use crate::{
     graph::GraphError,
--- a/src/feed/events.rs
+++ b/src/feed/events.rs
@@ -14,7 +14,8 @@ use crate::{
 };
 
 /// Events broadcast by the indexer when market data changes.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone)]
+#[cfg_attr(test, derive(PartialEq))]
 pub enum MarketEvent {
     /// Market was updated.
     MarketUpdated {
--- a/src/feed/events.rs
+++ b/src/feed/events.rs
@@ -14,7 +14,7 @@ use crate::{
 };
 
 /// Events broadcast by the indexer when market data changes.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum MarketEvent {
     /// Market was updated.
     MarketUpdated {
--- a/src/feed/market_data.rs
+++ b/src/feed/market_data.rs
@@ -133,7 +133,7 @@ impl SharedMarketData {
         for token in tokens {
             self.tokens
                 .entry(token.address.clone())
-                .or_insert_with(|| token.clone());
+                .or_insert_with(|| token);
         }
     }
 
@@ -144,15 +144,15 @@ impl SharedMarketData {
     ) {
         for (protocol_system, status) in sync_states {
             self.protocol_sync_status
-                .insert(protocol_system.clone(), status);
+                .insert(protocol_system, status);
         }
     }
 
     /// Removes a component.
-    pub fn remove_components(&mut self, ids: impl IntoIterator<Item = String>) {
+    pub fn remove_components<'a>(&mut self, ids: impl IntoIterator<Item = &'a ComponentId>) {
         for id in ids {
-            self.components.remove(&id);
-            self.simulation_states.remove(&id);
+            self.components.remove(id);
+            self.simulation_states.remove(id);
         }
     }
 
@@ -162,8 +162,7 @@ impl SharedMarketData {
         states: impl IntoIterator<Item = (ComponentId, Box<dyn ProtocolSim>)>,
     ) {
         for (id, state) in states {
-            self.simulation_states
-                .insert(id.clone(), state);
+            self.simulation_states.insert(id, state);
         }
     }
 
--- a/src/feed/market_data.rs
+++ b/src/feed/market_data.rs
@@ -4,17 +4,21 @@
 //! It's protected by a RwLock and shared across all components:
 //! - TychoIndexer: WRITE access to update data
 //! - Solvers: READ access to query states during solving
+//!
+//! We use tokio RwLock (which is write-preferring) to avoid writer starvation.
 
 use std::{collections::HashMap, sync::Arc};
 
 use tokio::sync::RwLock;
-use tycho_simulation::tycho_core::{
-    dto::Block,
-    models::{protocol::ProtocolComponent, token::Token, Address},
-    simulation::protocol_sim::ProtocolSim,
+use tycho_simulation::{
+    tycho_client::feed::SynchronizerState,
+    tycho_core::{
+        models::{protocol::ProtocolComponent, token::Token, Address},
+        simulation::protocol_sim::ProtocolSim,
+    },
 };
 
-use crate::types::{constants::GAS_COST_PER_SWAP, ComponentId, GasPrice, ProtocolSystem};
+use crate::types::{BlockInfo, ComponentId, GasPrice};
 
 /// Thread-safe handle to shared market data.
 pub type SharedMarketDataRef = Arc<RwLock<SharedMarketData>>;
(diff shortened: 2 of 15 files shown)


src/feed/tycho_feed.rs
```diff
@@ -49,171 +75,894 @@ impl TychoFeed {
...
+    /// Handles a message from Tycho stream.
+    #[instrument(skip(self, msg))]
+    async fn handle_tycho_message(&self, msg: Update) -> Result<(), TychoFeedError> {
+        // Collect variables for market shared data update
+        let added_components = msg.new_pairs.clone();
+        let removed_components = msg.removed_pairs.clone();
+        let updated_or_new_states = msg.states.clone();
+        let updated_components_ids: HashSet<_> = updated_or_new_states
+            .keys()
+            .filter(|id| !added_components.contains_key(id.as_str())) // TODO: Should we still emit as updated if the component is new?
```
tvinagre, line 168: I'll leave this answer to @louise-poole as she's taking the graph, but if we can handle updated & inserted together would be better IMO.
Otherwise the tycho_feed would need to be aware of the market graph (an updated pool could be missing in the graph so it would be treated as update while it should be an insertion)

src/feed/tycho_feed.rs
```diff
@@ -49,171 +75,894 @@ impl TychoFeed {
...
+            .cloned()
+            .collect();
+
+        let maybe_new_tokens = msg
+            .new_pairs
+            .values()
+            .flat_map(|component| component.tokens.iter().cloned());
+        // TODO: how do we handle delayed and stale states? Should the feed or the solvers handle
```
tvinagre, line 175: It's a good question that we should discuss with everyone. I'm tending to take the optimistic approach and only remove all the pools from a protocol_system once they're stale for a long time

tvinagre (approved):
trés beau

src/feed/events.rs
```diff
@@ -6,15 +6,15 @@
...
+use tycho_simulation::tycho_common::models::Address;
 
 use crate::{
     graph::GraphError,
     types::{ComponentId, GasPrice},
 };
 
 /// Events broadcast by the indexer when market data changes.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
ElusAegis: Given the Eq seems to be only used by tests, should we make it cfg for test only? Also, afaik looks like we can only keep `PartialEq`.

src/feed/market_data.rs
```diff
@@ -31,172 +35,145 @@ pub fn new_shared_market_data() -> SharedMarketDataRef {
...
+    }
 
-        // Update tokens map
-        for token in &component_data.tokens {
+    /// Inserts or updates tokens.
+    pub fn upsert_tokens(&mut self, tokens: impl IntoIterator<Item = Token>) {
+        for token in tokens {
             self.tokens
                 .entry(token.address.clone())
                 .or_insert_with(|| token.clone());
```
ElusAegis: Curious why we handle the tokens and components upserts differently? For components, we override the previous value, for the tokens, we only insert if empty. As far as I understand, both of them seem stable and should not change after initialization.

src/feed/tycho_feed.rs
```diff
@@ -49,171 +75,894 @@ impl TychoFeed {
...
     /// Returns a new subscriber for market events.
-    pub fn subscribe(&self) -> broadcast::Receiver<MarketEvent> {
+    pub fn subscribe(&self) -> Receiver<MarketEvent> {
         self.event_tx.subscribe()
     }
 
-    /// Returns the event sender for creating additional subscribers.
-    pub fn event_sender(&self) -> broadcast::Sender<MarketEvent> {
+    /// Returns an additional event sender.
+    pub fn event_sender_clone(&self) -> Sender<MarketEvent> {
```
ElusAegis, line 97: I remember we used something like that in the solver initially, but seems it is not used any more. What would be the use case for this?

src/feed/tycho_feed.rs
```diff
@@ -49,171 +75,894 @@ impl TychoFeed {
...
-        // 3. Process messages in a loop:
-        //    - Component added -> update market_data, broadcast ComponentAdded
-        //    - Component removed -> update market_data, broadcast ComponentRemoved
-        //    - State update -> update market_data, broadcast StateUpdated
-        // 4. Handle disconnects gracefully
+    /// Handles a message from Tycho stream.
+    #[instrument(skip(self, msg))]
+    async fn handle_tycho_message(&self, msg: Update) -> Result<(), TychoFeedError> {
+        // Collect variables for market shared data update
+        let added_components = msg.new_pairs.clone();
```
ElusAegis: I created a small PR to this to clean up some cloning - #18

src/feed/tycho_feed.rs
```diff
@@ -49,171 +75,894 @@ impl TychoFeed {
...
+
+        // Verify state was updated
+        {
+            let data = market_data.read().await;
+            assert!(
+                data.get_simulation_state(component_id)
+                    .is_some(),
+                "Component state should be updated"
+            );
+            assert!(data
```
ElusAegis: This seems to be a duplicated assertion, with the same duplication for the  state update. Also, should we explicitly check that that `get_simulation_state` returns expected state?

src/feed/tycho_feed.rs
```diff
@@ -49,171 +75,894 @@ impl TychoFeed {
...
+                .get_simulation_state(component_id)
+                .is_some());
+        }
+
+        // Now update its state
+
+        // Create an update with state information
+        let update = Update::new(12345, states, HashMap::new());
```
ElusAegis: Looks like we use the same state for the update as we stored initially

ElusAegis (approved):
LGTM, all the changes will be easy to merge into the algo work I am doing. There are a few comments I left, but also I understand this is more of the early version before a proper cleanup

src/feed/events.rs
zizou0x: Yes we can put it behind a `cfg(test)`. I don't think it hurts too much to have it here but I'm down to try to keep the binary size as small as possible 👍

src/feed/market_data.rs
zizou0x: I honestly didn't put much thought, because as you said these struct are static and they will never change. I'll make everything "upsert" to stay consistent in the code

src/feed/tycho_feed.rs
zizou0x: Currently it's not used anywhere (maybe just in a test but I can easily feature gate it behind "test"). I didn't do it because I feel it's a convenient function that it doesn't cost much, also wanted to see how things turn out with the gas price fetcher regarding the events. 

I'll leave a TODO here so we can come back to it at the end and see if we want it or not

src/feed/tycho_feed.rs
zizou0x: Awesome! Wanted to do this polishing today, thanks for the help!

src/feed/tycho_feed.rs
zizou0x: True, I wanted to assert the component in the first assertion, I'll fix this.

About the exact assertion, currently the Mock that I use is so simple that all of them are the same. But I can tweak it in a way that makes it possible yes (probably using the `.fee()` function), but it will be quite shady I feel.

The best solution would be to implement a proper mock for `ProtocolSim` in the tycho_common crate behind the "test-assets" feature. Then we would use it here and we would be able to distinguish between mocks.

I didn't want to dive into this now though

src/feed/tycho_feed.rs
zizou0x: Yes, the reason is explain above, I'll implement an easy workaround for now so we can properly test this

src/feed/tycho_feed.rs
zizou0x: I decided to just cfg(test) this one too, we can always easily make it available if we need an additional sender later

src/feed/tycho_feed.rs
zizou0x: Right now each solver worker can decide to treat delayed component as they want, it's the simplest and the most flexible. I like it this way honestly.

I'm not against re-assessing in the future if we want to move this responsibility to the feed in which case it will be the same for all workers.

src/feed/tycho_feed.rs
ElusAegis: That would be very useful, as I ended up needing another mock for the testing of the Algorithm. I could first merge it with your implementation, and possibly then migrate it to the tycho-common

src/feed/tycho_feed.rs
louise-poole: Preferably nope, but isn't a trainsmash either way.

PR #17, opened 2026-01-14
feat: Implement Solver Worker

louise-poole:


src/worker.rs
```diff
@@ -76,187 +76,249 @@ where
...
-
-        // Solve each order
-        let mut order_solutions = Vec::with_capacity(request.orders.len());
-
-        for order in &request.orders {
-            // Validate order
-            if let Err(_e) = order.validate() {
-                order_solutions.push(OrderSolution {
+        // Validate order - return early if invalid (no lock needed)
+        if let Err(_e) = order.validate() {
```
kayibal: Not sure what this does, but I feel it would be better to do that before we pass the order to the worker.

src/worker.rs
```diff
@@ -76,187 +76,249 @@ where
...
-    pub async fn solve(&mut self, request: &SolutionRequest) -> Result<Solution, SolveError> {
-        let start_time = Instant::now();
+    /// Gets block info from market data.
+    fn get_block_info(market: &crate::feed::market_data::SharedMarketData) -> BlockInfo {
+        let last_block = market.last_updated();
+        BlockInfo {
+            number: last_block.number,
+            hash: format!("{:?}", last_block.hash),
+            timestamp: last_block.ts.and_utc().timestamp() as u64,
+        }
```
kayibal: This struct does nothing other than block? Why do we have it?

src/worker.rs
```diff
@@ -76,187 +76,249 @@ where
...
+                let amount_in = route
+                    .swaps
+                    .first()
+                    .map(|s| s.amount_in.clone())
+                    .unwrap_or_else(|| BigUint::ZERO);
+                let amount_out = route
+                    .swaps
+                    .last()
+                    .map(|s| s.amount_out.clone())
+                    .unwrap_or_else(|| BigUint::ZERO);
```
kayibal: I don't like that these amounts can differ from what was specified by the SolveRequest...

src/worker.rs
```diff
@@ -76,187 +76,249 @@ where
...
+        &mut self,
+        mut event_rx: broadcast::Receiver<MarketEvent>,
+        task_rx: Arc<Mutex<mpsc::Receiver<SolveTask>>>,
+        mut shutdown_rx: broadcast::Receiver<()>,
+    ) {
+        info!(self.worker_id, "worker started");
+
+        loop {
+            tokio::select! {
+                // Check for shutdown
```
kayibal, line 277: I would suggest to poll tasks in a `biased` way:   https://docs.rs/tokio/latest/tokio/macro.select.html#fairness

This way market updates take priority over solve tasks.

src/worker.rs
```diff
@@ -76,187 +76,249 @@ where
...
-    /// Errors are logged but do not stop processing of subsequent events.
-    pub fn process_events(&mut self) {
-        while let Ok(event) = self.event_rx.try_recv() {
-            if let Err(e) = self.graph_manager.handle_event(&event) {
-                warn!("Warning: Error handling market event: {:?}", e);
+    /// Processes a single market event.
+    pub fn process_event(&mut self, event: MarketEvent) {
+        match event {
+            MarketEvent::MarketUpdated { .. } => {
+                if let Err(e) = self.graph_manager.handle_event(&event) {
```
kayibal, line 114: It would be nice to measure the time these take as well and expose them as a metric.

src/worker.rs
```diff
@@ -76,187 +76,249 @@ where
...
+                    match task {
+                        Some(task) => {
+                            let task_id = task.id;
+                            let _wait_time = task.wait_time();
+                            let task_response_tx = task.response_tx;
+
+                            // Process any pending events first to guarantee graph is up-to-date (alleviate race condition)
+                            while let Ok(event) = event_rx.try_recv() {
+                                self.process_event(event);
+                            }
```
kayibal: If you make the select loop biased, I think you don't need this...

src/worker.rs
```diff
@@ -76,187 +76,249 @@ where
...
+    ///
+    /// # Arguments
+    ///
+    /// * `event_rx` - Receiver for market events
+    /// * `task_rx` - Shared receiver for solve tasks
+    /// * `shutdown_rx` - Receiver for shutdown signals
+    pub async fn run(
+        &mut self,
+        mut event_rx: broadcast::Receiver<MarketEvent>,
+        task_rx: Arc<Mutex<mpsc::Receiver<SolveTask>>>,
```
kayibal: Why is this in a Mutex, imo this is a red flag that something is off. Are we sharing this queue with multiple Workers? https://users.rust-lang.org/t/arc-mutex-mpsc-receiver-vs-async-channel/70401

kayibal (commented):
Added some comments. There are quite a few configurations that don't fit into SolverWorker e.g. the size of the broadcast channel. Should we create a SolverWorkerPool struct for these?

src/worker_pool.rs
```diff
@@ -86,58 +86,17 @@ impl WorkerPool {
...
                         );
 
-                        // Create solver (graph type and manager are automatically inferred from
-                        // algorithm)
+                        // Create solver
                         let mut worker =
-                            SolverWorker::new(market_data, event_rx, algorithm, worker_config);
+                            SolverWorker::new(market_data, algorithm, worker_config, worker_id);
 
                         // Initialize solver graph
```
ElusAegis: Just curious if we should move this into the `run` method or `new` method, such that not to forget to do this accidentally when we start the run

src/worker.rs
tvinagre, line 277: I think we should add biased here to process conflicting msgs in the order that is stated here (shutdown > market event > solution request)

tvinagre (approved):

src/worker.rs
tvinagre: Yeah we are already validating the order on the OrderManager, no need to validate here

src/worker.rs
tvinagre: lol just saw this, commented the same thing - I agree

src/worker_pool.rs
louise-poole: Which part are you referring to? I moved around 2 things in these lines...

src/worker.rs
louise-poole: Do you mean we should use the tycho_common's Block struct? Not sure why a new one was created, but I am happy to remove it 👍

src/worker.rs
louise-poole: How do we want to handle metrics in this project? I think adding metrics should be in a separate PR.

src/worker.rs
```diff
@@ -76,187 +73,225 @@ where
...
-    /// Call this on startup or when recovering from missed events.
+    /// Call this on startup or to recreate the graph from the latest market topology.
     /// Gets the market topology from SharedMarketData and uses it to build the graph.
     pub async fn initialize_graph(&mut self) {
-        let market = self.market_data.read().await;
-        let topology = market.component_topology();
+        let topology = {
+            // read lock on market data
+            let market = self.market_data.read().await;
+            market.component_topology().clone() // clone to avoid holding the lock
```
zizou0x, line 102: No need to clone, it's already giving an owned HashMap

src/worker.rs
```diff
@@ -76,187 +73,225 @@ where
...
-                    amount_out: BigUint::ZERO,
-                    gas_estimate: BigUint::ZERO,
-                    price_impact_bps: None,
-                    block: block_info.clone(),
-                    algorithm: String::new(),
-                });
-                continue;
+        // Get block info
+        // maybe the algorithm should return the block info with the route? The block might update
+        // while solving and the route returned might be for the newer block.
```
zizou0x: Yes, this will most likely happen because tokio RwLock is write-preferring (which is required else we would get writer starvation)

As they are sequential maybe we should take the lock, get the block then pass the owned lock to `find_best_route`. Then each algo is responsible for dropping this ASAP. This way we ensure the market data stays the same during solving (except for vm protocols but we don't have a hand on it anyway)

src/worker.rs
```diff
@@ -76,187 +73,225 @@ where
...
-                    let amount_out = route
+                        .unwrap_or_else(|| BigUint::ZERO)
+                };
+                let amount_out = if order.is_sell() {
+                    route
                         .swaps
                         .last()
                         .map(|s| s.amount_out.clone())
-                        .unwrap_or_else(|| BigUint::ZERO);
+                        .unwrap_or_else(|| BigUint::ZERO)
```
zizou0x: I wonder if it makes sense to return a `OrderSolution` if we haven't found a solution. Wouldn't it make more sense to return an error here?

If I'm not wrong it would also mean that the algo didn't return a `NoPath` error which means a bug in one of the implementation. If that's correct then I think it's worth logging an `error!`.

Cargo.toml
```diff
@@ -41,6 +41,7 @@ num-traits = "0.2"
 num_cpus = "1.16"
 lazy_static = { version = "1.5.0", default-features = false }
 uuid = { version = "1", features = ["v4", "serde"] }
+async-channel = "2.5.0"
```
zizou0x, line 48: As we aren't planing to be runtime agnostic can't we use tokio channels instead? I think there are more optimized for tokio

zizou0x (commented):
Nice work!

src/algorithm/most_liquid.rs
```diff
@@ -151,6 +151,9 @@ impl Algorithm for MostLiquidAlgorithm {
             });
         }
 
+        // Acquire lock to read market data
+        let market = market.read().await;
```
ElusAegis, line 418: That will be super useful for me to have locks in the algo 👍

src/types/solution.rs
```diff
@@ -147,6 +149,19 @@ pub struct Solution {
     pub solve_time_ms: u64,
 }
 
+/// Solution for a single [`Order`] with timing information.
+///
+/// Contains the order solution along with the time taken to compute it.
+/// This is useful when solving a single order and you need both the solution
+/// details and the solve time.
```
ElusAegis: Just thinking about this, and feels like we could add the solve time to the OrderSolution structure, to eliminate this wrapper. This would also enable per-order timing analysis in batches (identifying slow orders)

ElusAegis: `enqueue(&self, order: Order) -> Result<SingleOrderSolution, SolveError>` just feels more organic to have a signature `-> Result<OrderSolution, SolveError>`

src/worker.rs
```diff
@@ -76,187 +73,225 @@ where
...
+                let amount_in = if order.is_sell() {
+                    order.amount.clone()
+                } else {
+                    route
                         .swaps
                         .first()
                         .map(|s| s.amount_in.clone())
-                        .unwrap_or_else(|| BigUint::ZERO);
-                    let amount_out = route
+                        .unwrap_or_else(|| BigUint::ZERO)
```
ElusAegis:
```suggestion
                          .unwrap_or_default()
  ```

File: src/worker.rs (hunk 11 above)
Comment (ElusAegis) [inline src/worker.rs]:
```suggestion
                          .unwrap_or_default()
  ```

File: src/worker.rs [hunk 16]
```diff
@@ -76,187 +73,225 @@ where
...
+    /// stays up-to-date while processing solve tasks.
+    ///
+    /// # Arguments
+    ///
+    /// * `event_rx` - Receiver for market events
+    /// * `task_rx` - Shared receiver for solve tasks
+    /// * `shutdown_rx` - Receiver for shutdown signals
+    pub async fn run(
+        &mut self,
+        mut event_rx: broadcast::Receiver<MarketEvent>,
```
ElusAegis, line 267: For personal understanding, what was the reasoning for moving the `event_rx` from a inner feild of a worker into a function parameter?

ElusAegis (commented):
Looks great! Just some comments for personal undestanding

src/worker.rs
louise-poole: @ElusAegis can I leave this change to you with your Algo implementation?

ElusAegis: FYI just merged the algo impl, but this should be easy to do as a second return arg for the happy path sure. but I can see that if we get an error, we still need the block info here, which is harder to think of conceptually, as reading the block info twice as mentioned can give different results

louise-poole: Hmmm... yeah let's leave this for now. We need to carefully think about locks when we refactor the SharedMarketData to use multiple locks instead of one giant lock.

Cargo.toml
louise-poole: What tokio channel do you recommend for this situation? We basically want shared consumers (the first one to get the task from the channel solves it). Originally I had a arc mutex around a mpsc channel but that's pretty ugly.

src/types/solution.rs
louise-poole: I think we return the OrderSolution struct to the user... not sure if we want solve time in there? I'll leave this to @tvinagre as he's dealing with the task queue and api side of things.

src/worker.rs
louise-poole: No real reason other than it felt cleaner to pass the channels we want to run on when we call run. Didn't feel necessary to have it as a inner field when it's only ever used in this one fn.

Cargo.toml
zizou0x: I think https://docs.rs/tokio/latest/tokio/sync/broadcast/fn.channel.html could work for this case?

src/worker.rs
ElusAegis: Makes sense, thank you!

PR #19, opened 2026-01-15
feat: Implement order manager

tvinagre:


docs/ARCHITECTURE.md
```diff
@@ -103,11 +116,60 @@ Tycho Solver is a high-performance solver built on Tycho for finding optimal swa
...
+pub struct SolverPoolHandle {
+    pub name: String,
+    pub algorithm: String,
+    pub queue: TaskQueueHandle,
+}
+```
+
+**Key Features:**
+
+1. **Fan-out**: Sends each order to ALL solver pools in parallel
```
louise-poole: This won't always be the case no? Just is easiest to implement for v1?

docs/ARCHITECTURE.md
```diff
@@ -103,11 +116,60 @@ Tycho Solver is a high-performance solver built on Tycho for finding optimal swa
...
+```rust
+// Solutions are ranked by amount_out_net_gas (amount_out minus gas cost)
+// Constraints like max_gas can filter out expensive solutions
+fn select_best(&self, responses: &OrderResponses, options: &SolutionOptions) -> OrderSolution {
+    responses.solutions.iter()
+        .filter(|(_, sol)| sol.status == SolutionStatus::Success)
+        .filter(|(_, sol)| options.max_gas.map(|max| sol.gas_estimate <= max).unwrap_or(true))
+        .max_by_key(|(_, sol)| &sol.amount_out_net_gas)
+}
+```
```
louise-poole: Is this necessary in the docs? I think the line `**Best Selection**: Chooses solution with highest `amount_out_net_gas`` is descriptive enough for this.

docs/ARCHITECTURE.md
```diff
@@ -121,22 +183,38 @@ impl TaskQueueHandle {
 
 ---
 
-### 3. WorkerPool
+### 4. WorkerPool
 
 **File:** `src/worker_pool.rs`
 
-**Responsibility:** Manage dedicated compute threads, each owning a SolverWorker instance.
+**Responsibility:** Manage dedicated compute threads for a single algorithm type, each owning a SolverWorker instance. Each pool has its own TaskQueue and workers.
```
louise-poole:
```suggestion
**Responsibility:** Manage dedicated compute threads for a single algorithm type. Each pool has its own TaskQueue and SolverWorkers.
```

docs/ARCHITECTURE.md
```diff
@@ -121,22 +183,38 @@ impl TaskQueueHandle {
...
 
 ```rust
 pub struct WorkerPool {
+    name: String,
+    algorithm_type: AlgorithmType,
```
louise-poole: Can the name not just be the AlgorithmType as a string? We don't need a separate name var no?

docs/ARCHITECTURE.md
```diff
@@ -121,22 +183,38 @@ impl TaskQueueHandle {
...
 ```
 
+**Design Rationale (Queue per Pool):**
+
+| Aspect | Benefit |
+|--------|---------|
+| Independent backpressure | Slow algorithm doesn't block fast ones |
+| Independent scaling | Can have 8 workers for expensive algo, 2 for fast algo |
+| Clean isolation | Algorithm bugs don't affect other pools |
+| Easy extensibility | Add new algorithm = add new pool |
```
louise-poole, line 200: Ewwwwww... can this be prettier? 🥺  the table 'walls' are funky

docs/ARCHITECTURE.md
```diff
@@ -418,18 +546,32 @@ src/
...
 │
+<<<<<<< HEAD
+=======
+├── feed/                     # Market data feed
+│   ├── mod.rs
+│   ├── market_data.rs        # SharedMarketData
+│   ├── events.rs             # MarketEvent enum
+│   └── tycho_feed.rs         # TychoFeed (WebSocket client)
+│
+>>>>>>> a8eb622 (feat: Implement OrderManager)
```
louise-poole: Looks like some merge notation was included in what you pushed?

src/order_manager/mod.rs
```diff
@@ -0,0 +1,689 @@
...
+    },
+};
+
+/// Handle to a solver pool for dispatching orders.
+#[derive(Clone)]
+pub struct SolverPoolHandle {
+    /// Human-readable name for logging & metrics.
+    pub name: String,
+    /// Algorithm identifier.
+    pub algorithm: String,
```
louise-poole: Same question here: would the name ever differ from the algorithm name? I guess maybe we'd have 2 pools with the same algorithm and different config (one is 2 hop, the other is 3 hop)?

src/types/solution.rs
```diff
@@ -41,15 +39,25 @@ pub struct SolutionRequest {
...
 /// Options to customize the solving behavior.
+#[serde_as]
 #[derive(Debug, Clone, Default, Serialize, Deserialize)]
 pub struct SolutionOptions {
     /// Timeout in milliseconds. If `None`, uses server default.
     pub timeout_ms: Option<u64>,
+    /// Minimum number of solver responses to wait for before returning.
+    /// If `None` or `0`, waits for all solvers to respond (or timeout).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub min_responses: Option<usize>,
```
louise-poole, line 55: We should document somewhere that they should use the '/health' endpoint to find out how many solvers are active before deciding to set this value here. Also do we validate this (check if they don't set it to more solvers than we have)?

src/types/solution.rs
```diff
@@ -62,7 +70,7 @@ pub struct Order {
     /// Output token address (the token being bought).
     pub token_out: Address,
     /// Amount to swap, interpreted according to `side` (in token units, as decimal string).
-    #[serde(with = "biguint_as_string")]
+    #[serde_as(as = "DisplayFromStr")]
```
louise-poole, line 78: FMI: what's the difference here?

src/main.rs
```diff
@@ -26,7 +27,7 @@ use tycho_solver::{
 struct Config {
     http_host: String,
     http_port: u16,
-    num_workers: usize,
+    num_workers_per_pool: usize,
```
louise-poole, line 32: Would be better if we could configure this per pool...

src/task_queue.rs
```diff
@@ -34,6 +34,13 @@ pub struct TaskQueueHandle {
 }
 
 impl TaskQueueHandle {
+    /// Creates a TaskQueueHandle from an existing sender.
+    ///
+    /// This is primarily useful for testing or custom queue implementations.
+    pub fn from_sender(sender: async_channel::Sender<SolveTask>) -> Self {
+        Self { sender }
+    }
```
louise-poole, line 43: If it's just used for testing for now should we use `#[cfg(test)]`?

src/task_queue.rs
```diff
@@ -109,14 +116,397 @@ impl TaskQueue {
...
+        let result = handle.enqueue(make_order()).await;
+
+        worker
+            .await
+            .expect("worker should complete");
+        assert!(matches!(result, Err(SolveError::NoRouteFound { .. })));
+    }
+
+    #[tokio::test]
+    async fn test_enqueue_returns_internal_error_when_receiver_dropped() {
```
louise-poole: Not such a fan of the test names being too verbose... like this could just be 
```suggestion
    async fn test_enqueue_error_when_receiver_dropped() {
```

src/task_queue.rs
```diff
@@ -109,14 +116,397 @@ impl TaskQueue {
...
+        let result = handle.enqueue(make_order()).await;
+
+        worker
+            .await
+            .expect("worker should complete");
+        assert!(matches!(result, Err(SolveError::Internal(_))));
+    }
+
+    #[tokio::test]
+    async fn test_enqueue_returns_queue_full_when_channel_closed() {
```
louise-poole: And 
```suggestion
    async fn test_enqueue_queue_full_error() {
```

src/worker.rs
```diff
@@ -2,7 +2,7 @@
 //!
 //! The Solver Worker:
 //! - Initializes graph from market topology (via a GraphManager)
-//! - Consumess MarketEvents to keep local topology in sync
+//! - Consumes MarketEvents to keep local topology in sync
```
louise-poole, line 5: How dare you imply I made a typo! 😱 I actually take pride in the typo coz it proves I didn't use AI no? 👀 😂

src/worker_pool.rs
```diff
@@ -1,13 +1,17 @@
 //! Worker pool for processing solve tasks.
 //!
 //! The worker pool manages dedicated OS threads for CPU-bound route finding.
-//! Each worker has its own tokio runtime and processes tasks from the queue.
+//! Each worker owns a SolverWorker instance and processes tasks from the queue.
```
louise-poole: Actually owns multiple workers...

src/worker_pool.rs
```diff
@@ -1,13 +1,17 @@
...
+//! A pool is configured with a specific algorithm type, allowing multiple
+//! pools with different algorithms to compete via the OrderManager.
```
louise-poole: Different algorithms/configs. Like we could have 2 pools:
- most liquid with 2 hops
- most liquid with 3 hops

src/worker_pool.rs
```diff
@@ -18,9 +22,38 @@ use crate::{
...
 };
 
+/// Algorithm type for the worker pool.
+///
+/// Each pool is dedicated to a single algorithm type. The OrderManager
+/// can fan out orders to multiple pools with different algorithms and
+/// select the best solution.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum AlgorithmType {
```
louise-poole: Would prefer if we has a registry with the different algorithm implementations. Also each algorithm will have a config so we need to think about how that is handled.

src/worker_pool.rs
louise-poole: See - we have it hardcoded as MostLiquidAlgorithm. Would be preferable if we could retrieve this from a registry

docs/ARCHITECTURE.md
tvinagre: Yes, this is just for V1, we can optimize in the future. There's a comment where I've implemented this, will fix this in this doc

docs/ARCHITECTURE.md
tvinagre: hahahaha yes

src/types/solution.rs
tvinagre: I'm not convinced this should be exposed to the user, but actually be set by the developer 
But yes I think we should verify this value.

src/worker_pool.rs
tvinagre: tru

src/worker_pool.rs
tvinagre: tru

src/api/handlers.rs
```diff
@@ -75,7 +69,7 @@ async fn health(state: web::Data<AppState>) -> HttpResponse {
     let status = HealthStatus {
         healthy: is_healthy,
         last_update_ms: age_ms,
-        queue_depth: state.task_queue.approximate_depth(),
+        num_solver_pools: state.order_manager.num_pools(),
```
zizou0x, line 72: Small thoughts, just to raise a discussion, no need to change it right now.

Have you considered leaving the `/health` very simple, just to know if the application is running (like it returns just bool) and then we could have another `/status` endpoint for the details (like how many solvers are running, what kind of algo they use, with what params ect...)

src/order_manager/config.rs
```diff
@@ -0,0 +1,36 @@
...
+use std::time::Duration;
+
+/// Configuration for the OrderManager.
+#[derive(Debug, Clone)]
+pub struct OrderManagerConfig {
+    /// Default timeout per order (can be overridden per-request).
+    pub default_timeout: Duration,
+    /// Minimum number of solver responses to wait for before returning.
+    /// If 0, waits for all solvers to respond (or timeout).
+    pub min_responses: usize,
```
zizou0x: Can we detail a bit more how this works please? Does it means that if the timeout is 100ms but I got `>=min_responses` already, I am returning the solution?

Also small detail but I wonder if an Option or a custom config enum wouldn't be clearer than 0 for all solvers

src/api/mod.rs
```diff
@@ -73,16 +73,16 @@ impl Default for HealthTracker {
...
     /// Health tracker for monitoring data freshness.
     pub health_tracker: HealthTracker,
 }
 
 impl AppState {
     /// Creates new application state.
-    pub fn new(task_queue: TaskQueueHandle, health_tracker: HealthTracker) -> Self {
-        Self { task_queue, health_tracker }
+    pub fn new(order_manager: OrderManager, health_tracker: HealthTracker) -> Self {
+        Self { order_manager: Arc::new(order_manager), health_tracker }
```
zizou0x, line 85: What's the motivation for Arcing it?

src/main.rs
```diff
@@ -38,7 +39,7 @@ impl Default for Config {
         Self {
             http_host: "0.0.0.0".to_string(),
             http_port: 3000,
-            num_workers: num_cpus::get(),
+            num_workers_per_pool: num_cpus::get(),
```
zizou0x, line 45: Shouldn't this be `num_cpus/number_of_pools`?

zizou0x (approved):

docs/ARCHITECTURE.md
tvinagre: F

src/order_manager/mod.rs
tvinagre: yeah - we could have smt like "most_liquid_2hop" vs "most_liquid_3hop"

src/types/solution.rs
tvinagre: It's a crate that simplifies serde - In this case, DisplayFromStr is an ready-made adapter. It automatically serializes using the type's Display impl and deserializes using its FromStr impl.

More here: https://docs.rs/serde_with/latest/serde_with/attr.serde_as.html

src/main.rs
tvinagre: It is configurable per pool but this is not exposed yet. I'll leave it for when we actually load config from file

src/api/handlers.rs
tvinagre: Yeah, we can extend that later but I think a lot of clients would like to hide this info from the users

src/api/mod.rs
tvinagre: I think it's because AppState is cloned for each Actix worker thread so I had to Arc it

src/main.rs
tvinagre: Nop - pools are independent, and each should maximize CPU usage for its algorithm.

src/types/primitives.rs
```diff
@@ -70,13 +69,14 @@ impl fmt::Display for ProtocolSystem {
 }
 
 /// Gas price information for transaction cost estimation.
+#[serde_as]
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct GasPrice {
     /// Base fee per gas (EIP-1559)
-    #[serde(with = "biguint_as_string")]
+    #[serde_as(as = "DisplayFromStr")]
```
ElusAegis, line 77: Thats nice! 👀

PR #22, opened 2026-01-20
feat: implement gas price feed

zizou0x:


src/feed/tycho_feed.rs
```diff
@@ -376,622 +411,623 @@ fn register_exchanges(
...
-            if message_count == 0 {
-                break;
-            }
-            dbg!(&event);
-        }
-
-        feed_handle.abort();
-    }
-}
+// #[cfg(test)]
```
zizou0x: I had to comment these out because the new GasPrice isn't PartialEq. I'll open a PR in tycho-ethereum for this.

src/feed/gas.rs
```diff
@@ -0,0 +1,47 @@
...
+        client: C,
+        shared_market_data: SharedMarketDataRef,
+    ) -> (Self, mpsc::Sender<oneshot::Sender<()>>) {
+        let (signal_tx, signal_rx) = mpsc::channel(5);
+        (Self { client, signal_rx, shared_market_data }, signal_tx)
+    }
+
+    pub async fn run(&mut self) -> Result<(), DataFeedError> {
+        loop {
+            let update_tx = self
```
tvinagre, line 26: IIUC, we have extended latency here because we only start fetching for gas price once we get a block from Tycho, right? So, assuming Tycho has 200ms latency, gas price will have 200ms + gas fetching time. 
We can make this better in the future (not prio for V1), would be nice to have metrics for all of these

src/feed/gas.rs
```diff
@@ -0,0 +1,47 @@
...
+                .signal_rx
+                .recv()
+                .await
+                .ok_or(DataFeedError::GasPriceFetcherError("Trigger channel closed".to_string()))?;
+
+            let fee_price = self
+                .client
+                .get_latest_fee_price()
```
tvinagre, line 34: We could potentially pass the block here. If the node is lagging, we'd get outdate gas prices

src/feed/gas.rs
zizou0x: This is correct, latency on gas price will be tycho-latency + client-latency, we could try to syncronize with the block time but at the ms scale, block production also fluctuates from a block to another

zizou0x: I've added the metrics for the lag

tvinagre: > we could try to syncronize with the block time

Yeah but sometimes Tycho blocks are ahead of nodes, so unless we consume the block msg straight from Substreams is very hard for us to fully eliminate this. We should just to be aware of this limitation but I think it's completely fine for now

tvinagre (approved):

PR #23, opened 2026-01-20
feat: Define data producer interfaces

tvinagre:


ElusAegis (approved):
Looks good for me, we could clarify & adjust the details on the go

src/derived/store.rs
```diff
@@ -0,0 +1,116 @@
...
+///
+/// Provides typed access to previously computed derived data.
+/// Each field is `Option` to indicate whether the computation has run.
+#[derive(Debug, Default)]
+pub struct DerivedDataStore {
+    token_prices: Option<TokenPrices>,
+    pool_depths: Option<PoolDepths>,
+    spot_prices: Option<SpotPrices>,
+    /// Block number at which data was last computed.
+    last_block: u64,
```
louise-poole: Would we not encounter a situation where the depths were updated recently, but the token_prices weren't? Then this 'last_block' would be valid for depth only no?

ElusAegis: Also, to this note, should it be Option to handle the case when it is not initialized?

src/derived/computation.rs
```diff
@@ -0,0 +1,107 @@
...
+    ///
+    /// * `market` - Read access to raw market data (components, tokens, topology)
+    /// * `store` - Read access to previously computed derived data
+    ///
+    /// # Returns
+    ///
+    /// The computed output, or an error if computation failed.
+    fn compute(
+        &self,
+        market: &SharedMarketData,
```
louise-poole, line 104: Should this not take SharedMarketDataRef? Then the lock can be handled internally and be released quicker?

louise-poole (commented):
Where is the process/struct in charge of running these computations?

src/derived/computation.rs
ElusAegis: Agree, I made this adjustment in the #27, however, thinking about this, is it optimal that each computation aquires a lock for itself and we allow for a situation where the `market` changes between two different computations, leading to a potentially inconsistent state?

louise-poole: Yes I think so. Some computations will be pretty slow (like the depth calculator) and the idea behind having an independent process for calculating it is so that it'll be non-blocking. If we hold the lock the whole time that makes it blocking.

src/derived/computation.rs
```diff
@@ -0,0 +1,107 @@
...
+        store: &DerivedDataStore,
```
louise-poole, line 105: Also instead of passing the store (it hasn't been used by any of the computation implementations anyway), why not pass the components that were updated? Then we don't need to recompute these values for the entire market graph and only need to do it for the ones that changed?

tvinagre:
Since @ElusAegis has already addressed these comments on his PR, I'll merge this one and we take it from there

PR #24, opened 2026-01-20
feat: add pool depth, spot price and gas token price computation modules with error handling

ElusAegis:


src/derived/computations/spot_price.rs
```diff
@@ -0,0 +1,178 @@
...
+    ) -> Result<Self::Output, ComputationError> {
+        let mut spot_prices = SpotPrices::new();
+        let mut error_count = 0usize;
+
+        let topology = market.component_topology();
+        let tokens = market.token_registry_ref();
+
+        for (component_id, token_addresses) in topology.iter() {
+            let Some(sim_state) = market.get_simulation_state(component_id) else {
+                trace!(component_id = %component_id, "skipping: no simulation state");
```
louise-poole: This is actually an error no? We should not have a component in our topology that doesn't have a state in the shared market data.

src/derived/computations/spot_price.rs
```diff
@@ -0,0 +1,178 @@
...
+impl DerivedComputation for SpotPriceComputation {
+    type Output = SpotPrices;
+
+    const ID: ComputationId = "spot_prices";
+
+    fn compute(
+        &self,
+        market: &SharedMarketData,
+        _store: &DerivedDataStore,
+    ) -> Result<Self::Output, ComputationError> {
```
louise-poole, line 45: Do we ever want to handle the case where some components failed and others succeeded? Like it might be useful to return which failed? Or do we feel the fact that they're missing from the output results is evident enough that they failed?

src/derived/computations/spot_price.rs
```diff
@@ -0,0 +1,178 @@
...
+}
+
+impl DerivedComputation for SpotPriceComputation {
+    type Output = SpotPrices;
+
+    const ID: ComputationId = "spot_prices";
+
+    fn compute(
+        &self,
+        market: &SharedMarketData,
```
louise-poole, line 43: I suspect this should rather be SharedMarketDataRef and then you can handle the lock internally and release it as soon as possible.

src/derived/computations/spot_price.rs
```diff
@@ -0,0 +1,178 @@
...
+
+            let pool_tokens: Vec<_> = token_addresses
+                .iter()
+                .filter_map(|addr| tokens.get(addr))
+                .collect();
+
+            if pool_tokens.len() != token_addresses.len() {
+                trace!(component_id = %component_id, "skipping: missing token metadata");
+                continue;
+            }
```
louise-poole: Again, this shouldn't happen as we should not have been able to receive a component that has tokens we do not have. It is a requirement of tycho-simulation's decoder: the tokens need to be known before we can decode components with them.
I would rather error here.

src/derived/computations/spot_price.rs
```diff
@@ -0,0 +1,178 @@
...
+                }
+            }
+        }
+
+        if error_count > 0 {
+            debug!(
+                error_count,
+                success_count = spot_prices.len(),
+                "spot price computation completed with some errors"
+            );
```
louise-poole: I feel like this is an important log. We should either error or warn here to start. If we decide it's ignorable/spammy we can lower the log level later.

src/derived/computations/spot_price.rs
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        if spot_prices.is_empty() && error_count > 0 {
```
louise-poole: You don't increment the error count on missing tokens/states though. I think a more robust check here would be:
```suggestion
        if spot_prices.is_empty() && !topology.is_empty() {
```

src/derived/computations/spot_price.rs
```diff
@@ -0,0 +1,178 @@
...
+            warn!(error_count, "spot price computation failed for all pairs");
+            return Err(ComputationError::NoValidResult {
+                reason: format!("spot price computation failed for all {error_count} pairs"),
```
louise-poole:
```suggestion
                reason: format!("spot price computation failed for all {} pairs", topology.len()),
```

src/derived/computations/spot_price.rs
```diff
@@ -0,0 +1,178 @@
...
+    feed::market_data::SharedMarketData,
+    types::ComponentId,
+};
+
+/// Key for spot price lookups: (component_id, token_in, token_out).
+pub type SpotPriceKey = (ComponentId, Address, Address);
+
+/// Spot prices map: key → spot price as f64.
+/// Represents: 1 token_in = spot_price token_out.
+pub type SpotPrices = HashMap<SpotPriceKey, f64>;
```
louise-poole: Maybe these can be in the `types.rs` file?

louise-poole (commented):
Reviewing in parts so that you can start addressing them. Reviewed spot prices first (looked like the easiest to me).

src/derived/computations/pool_depth.rs
```diff
@@ -0,0 +1,297 @@
...
+        let mut best_valid = BigUint::zero();
+
+        for _ in 0..MAX_BINARY_SEARCH_ITERATIONS {
+            if high <= &low + BigUint::one() {
+                break;
+            }
+
+            let mid = (&low + &high) / 2u32;
+
+            match sim_state.get_amount_out(mid.clone(), token_in, token_out) {
```
louise-poole, line 111: Hmmm... I'm wondering that instead of repeated calls to `get_amount_out`, why not use the new `query_pool_swap`? There you can specify the target price and it'll give you the max swap amount that respects the target price.

src/derived/computations/pool_depth.rs
```diff
@@ -0,0 +1,297 @@
...
+    pub fn new(slippage_threshold: f64) -> Result<Self, ComputationError> {
+        if !(slippage_threshold > 0.0 && slippage_threshold < 1.0) {
+            return Err(ComputationError::InvalidConfiguration(format!(
+                "slippage_threshold must be between 0 and 1 exclusive, got {slippage_threshold}"
+            )));
+        }
+        Ok(Self { slippage_threshold })
+    }
+
+    /// Binary search to find the maximum amount_in with acceptable slippage.
```
louise-poole, line 70: Might be nice to have a look at what we used to do on our python solvers. The depth calculator was called the [InertiaCalculator](https://github.com/propeller-heads/defibot/blob/da4e29b9c0c1b4f1090ebc10f3a14d7537669a21/defibot/data_pipeline/jobs/inertia_calculator.py#L17). Though I think our approach there might be negated by the use of `query_pool_swap`.

src/derived/computations/pool_depth.rs
```diff
@@ -0,0 +1,297 @@
...
+
+                    let spot_price = match sim_state.spot_price(token_in, token_out) {
+                        Ok(p) if p.is_finite() && p > 0.0 => p,
+                        _ => {
+                            error_count += 1;
+                            continue;
+                        }
+                    };
+
+                    match self.find_depth_binary_search(sim_state, token_in, token_out, spot_price)
```
louise-poole: If you're passing the whole sim_state, why add complexity by getting the swap price here and passing it as an extra param instead of just getting it within `find_depth_binary_search`?

louise-poole (commented):
I think we can make this depth search a lot more efficient through the use of query_pool_swap on the ProtocolSim trait. Have a look into it?

src/derived/computations/token_gas_price.rs
```diff
@@ -0,0 +1,514 @@
...
+
+                    // Check if this is better than what we had
+                    let dominated_by_existing = best_states
+                        .get(neighbor_addr)
+                        .is_some_and(|existing| !new_state.is_better_than(existing));
+
+                    if dominated_by_existing {
+                        // This path is not better, don't explore further
+                        continue;
+                    }
```
louise-poole: If we only take the best path, we're vulnerable to outliers that skew our pricing. On the NTQ we calculated the [mid_price](https://github.com/propeller-heads/token-quoter/blob/main/src/searchers/token_quoter_v2.rs#L323).

src/derived/computations/token_gas_price.rs
```diff
@@ -0,0 +1,514 @@
...
+        // Build adjacency for quick lookup: token -> [(component_id, pool_tokens)]
+        let mut adjacency: HashMap<Address, Vec<(ComponentId, Vec<Address>)>> = HashMap::new();
+        for (component_id, token_addresses) in topology.iter() {
+            for addr in token_addresses {
+                adjacency
+                    .entry(addr.clone())
+                    .or_default()
+                    .push((component_id.clone(), token_addresses.clone()));
+            }
+        }
```
louise-poole: Could potentially cache this? Maybe we should add a span here to see how long it takes and if it's notably slow we can look to improvements. If we adapt the `calculate` interface to pass the changed components only that will speed things up here significantly.

src/derived/types.rs
louise-poole, line 1: FMI - why move all of this out of here?

src/derived/computations/token_gas_price.rs
```diff
@@ -0,0 +1,514 @@
...
+
+        while let Some(current) = stack.pop() {
+            let Some(current_token_info) = tokens.get(&current.token) else {
+                continue;
+            };
+
+            // Explore all pools containing this token
+            let Some(pools) = adjacency.get(&current.token) else {
+                continue;
+            };
```
louise-poole: I think we should maybe make use of the graph's built-in 'find paths' fns? And limit the number of hops to keep this efficient.

src/derived/computations/pool_depth.rs
tvinagre: Is `query_pool_swap` implemented for all pools? I'm not sure it's reliable atm

src/derived/computations/token_gas_price.rs
```diff
@@ -0,0 +1,514 @@
...
+            Some(num / den)
+        }
+    }
+}
+
+/// Token prices map: token address → price ratio.
+pub type TokenGasPrices = HashMap<TokenGasPriceKey, TokenGasPrice>;
+
+/// 1 ETH in wei.
+fn one_eth() -> BigUint {
```
tvinagre: this breaks the assumption that we can run this on any-chain. 
We should have a map of chain to gas token. Or, at least, leave a TODO here so we don't forget

src/derived/computations/token_gas_price.rs
```diff
@@ -0,0 +1,514 @@
...
+    BigUint::from(10u64).pow(18)
+}
+
+/// State tracked for each token during price discovery.
+#[derive(Debug, Clone)]
+struct TokenState {
+    /// Amount of this token received from 1 ETH input.
```
tvinagre: If we're fixing the implementation to use a fixed denominator, then we need to update the interface to remove the denominator

src/derived/computations/spot_price.rs
```diff
@@ -0,0 +1,178 @@
...
+                continue;
+            }
+
+            for (i, token_in) in pool_tokens.iter().enumerate() {
+                for (j, token_out) in pool_tokens.iter().enumerate() {
+                    if i == j {
+                        continue;
+                    }
+
+                    match sim_state.spot_price(token_out, token_in) {
```
tvinagre: did you double check if this is the correct interface, not `token_in`, `token_out`?

src/derived/computations/pool_depth.rs
```diff
@@ -0,0 +1,297 @@
...
+
+                    if effective_price >= min_acceptable_price {
+                        best_valid = mid.clone();
+                        low = mid;
+                    } else {
+                        high = mid;
+                    }
+                }
+                Err(_) => {
+                    // Simulation failed - below protocol minimum as we maintain an invariant that
```
tvinagre: Is this true, can't we fail for other reasons?

src/derived/computations/token_gas_price.rs
```diff
@@ -0,0 +1,514 @@
...
+
+            for (component_id, pool_tokens) in pools {
+                let Some(sim_state) = market.get_simulation_state(component_id) else {
+                    continue;
+                };
+
+                // Try swapping to each other token in the pool
+                for neighbor_addr in pool_tokens {
+                    // Prevent revisiting tokens in current path (cycle detection)
+                    if current.path.contains(neighbor_addr) {
```
tvinagre: For deep paths, this becomes O(depth ^2). Maybe it's better to use a HashSet for the visited path?
`path_set: HashSet<Address>,`

src/derived/computations/token_gas_price.rs
tvinagre: also, if this is static we don't need to recreate every time it is called

src/derived/computations/token_gas_price.rs
```diff
@@ -0,0 +1,514 @@
...
+            .compute(&market, &store)
+            .unwrap();
+
+        let one = one_eth();
+
+        // Should pick path through A: 1 ETH -> 2 A -> 10 C
+        let price_c = prices.get(&token_c.address).unwrap();
+        assert_eq!(price_c.numerator, &one * 10u32);
+    }
+}
```
tvinagre, line 891: What happens if the gas token has no pools? I'd add a test for that

src/derived/computations/spot_price.rs
ElusAegis: Agree! This is adjusted in #27

src/derived/computations/spot_price.rs
ElusAegis: @tvinagre do you have an opinion on this? Should we prefer reselience to some errors or correctness ?

src/derived/computations/spot_price.rs
ElusAegis: I will simplify to just print a warning when we have `error_count > 0`, while also hard failing for the missing state. Should be easier to reason about

src/derived/computations/spot_price.rs
ElusAegis: I guess the right approach is to currently fail on any simulation error, and if we ever get too many of them, handle this more gracefully, allowing for parcial failures

src/derived/computations/spot_price.rs
ElusAegis: Also, I for now will fail on any simulation error. If it will be too much - we can switch to the partial failure support as originally here

src/derived/computations/spot_price.rs
ElusAegis: Would it not be cleaner to have each compuation define all of its assets in its own module? I thuoght that would make it a lot easier to extend in the future, where you can look into a single file to understand everything

src/derived/computations/pool_depth.rs
ElusAegis: I was not aware of this fn, I will call it first, and on err try to use the `get_amount_out` method

src/derived/computations/pool_depth.rs
ElusAegis: Good point 👍 Though as we now also use the `query_pool_swap` which also depends on the spot price, I will keep it in the main iteration loop

src/derived/computations/token_gas_price.rs
ElusAegis: Added spans for all computations, but yeah, this looks to be the most concerning one

src/derived/computations/pool_depth.rs
ElusAegis: As far as I understood , the [InertiaCalculator](https://github.com/propeller-heads/defibot/blob/da4e29b9c0c1b4f1090ebc10f3a14d7537669a21/defibot/data_pipeline/jobs/inertia_calculator.py#L17) used so called Golden Section, which is useful for convex / concave functions to find their local minima/maxima. But as far as I understand the `get_amount_out` is monotonical in the `input_amount`, with a monotonic valid/invalid boundary, hence binary search will be faster as it halves the interval, while golden section only reduces by ~60%

ElusAegis: Is there anything else we can take away from the original implementation?

src/derived/types.rs
ElusAegis: Felt to me like a lot more modular and cleaner way to navigate especially when there will be many computations. Such that all computation data is in one place, just like with ProtocolSim - where a module exists for each of them, and contains all of the data

src/derived/computations/token_gas_price.rs
ElusAegis: Unfortunately the BigUint does not have const constructors and is also not Copy, so even if we used Lazy, we would have had to clone it on every usage. Is there perhaps some other way I am missing here?

ElusAegis: I have adjusted the code otherwise to make it more suitable for multiple chains

src/derived/computations/token_gas_price.rs
ElusAegis: Done🫡 Took some redesigning, so would be good to review the code again

src/derived/computations/token_gas_price.rs
ElusAegis: Now it is no longer fixed

src/derived/computations/spot_price.rs
ElusAegis: So the `spot_price` definition says:

```
fn spot_price(&self, base: &Token, quote: &Token) -> Result<f64, SimulationError>
 
Returns the protocol's current spot buy price for base in units of quote.
```

Which makes it seem that it should be `spot_price(token_out, token_in)`, but that also seemed off to me (and I repeatedly made an inverse mistake in other places. Can we check what is the right one?

src/derived/computations/token_gas_price.rs
ElusAegis: As far as I know the graph does not provide a function to enumerate all paths through the edges, only through the tokens (same as why we do not use it in `find_paths` in `most_liquide`

src/derived/computations/pool_depth.rs
ElusAegis: Added this assumption explicitly:

```rust
    /// As we never exceed the upper bound, we assume that if the simulation errors, it's because
    /// we are below the lower bound of valid amounts, and thus should increase the lower bound.
    /// This assumes that the simulation should not have errors in the valid range.
    ```

File: src/derived/computations/token_gas_price.rs (hunk 19 above)
Comment (ElusAegis) [reply src/derived/computations/token_gas_price.rs]: Agree, though I completely adjusted the logic here now

File: src/derived/computations/token_gas_price.rs (hunk 20 above)
Comment (ElusAegis) [reply src/derived/computations/token_gas_price.rs:891]: That caught an edge case, thank you for suggesting!

File: src/derived/computations/spot_price.rs (hunk 17 above)
Comment (ElusAegis) [reply src/derived/computations/spot_price.rs]: Even prior to my changes to the doc of the `spot_price`, it implied the reverse order (though none of the actual implementations follow it:


```rust
    /// Returns the protocol's current spot price of two tokens.
    ///
    /// Current spot price is defined as the amount of the quote token required to buy one unit of
    /// the base token, accounting for protocol fees.
    ///
    /// E.g. if BTC/USDT is trading at 1000 with a 20% fee, you would need 1,250 USDT (1000 / 0.8)
    /// to buy 1 BTC (base currency).
    ///
    /// # Arguments
    ///
    /// * `base` - the token being priced in a pair. For the symbol BTC/USDT, BTC would be the base
    ///   token.
    /// * `quote` - the token used to price the base token. For the symbol BTC/USDT, USDT would be
    ///   the quote token.
    ```
    
    Should the doc comment be changed to:
    
    ```rust
          /// Returns the protocol's current spot sell price for `base` in units of `quote`.
      ///
      /// The returned price is the amount of `quote` received when selling exactly 1 unit of `base`,
      /// accounting for the protocol fee (i.e. `price = pre_fee_price * (1.0 - fee)`)
      /// and assuming zero slippage (i.e., a negligibly small trade size).
      ///
      /// # Arguments
      /// * `base` - the token being priced (what you sell). For BTC/USDT, BTC is the base token.
      /// * `quote` - the token received in exchange for `base`. For BTC/USDT, USDT is the quote token.
      ///
      /// # Examples
      /// If BTC/USDT is trading at 1000 with a 20% fee, this returns `1000 * (1.0 - 0.20) = 800`
    ```

File: src/derived/computations/spot_price.rs (hunk 2 above)
Comment (tvinagre) [reply src/derived/computations/spot_price.rs:45]: Yes, I think the best would be to return the ones that fail. This way we can handle partial failures, and also remove the ones that failed from the graph.

File: src/derived/computations/spot_price.rs (hunk 8 above)
Comment (tvinagre) [reply src/derived/computations/spot_price.rs]: What if we take the middle approach and create a types.rs under`computations` ?
I do prefer having a centralized place to look for interfaces / types

File: src/derived/computations/token_gas_price.rs (hunk 13 above)
Comment (tvinagre) [reply src/derived/computations/token_gas_price.rs]: Fine to merge is as is so we can improve later

File: src/derived/computations/spot_price.rs (hunk 17 above)
Comment (tvinagre) [reply src/derived/computations/spot_price.rs]: @ElusAegis just to be clear, I didn't check Tycho Sim implementation - I just commented bc usually one of the biggest and most-common issues that we have (and will have) when testing is data ordering / correctly interpreting the data and conversions - since those are not caught by the compiler and can easily slip on unittests. 
If our interface is different from Tycho Sim's interface, please adjust - Tycho Sim is our source of truth in this case.

File: src/derived/computations/pool_depth.rs [hunk 21]
```diff
@@ -0,0 +1,411 @@
+//! Pool depth computation.
+//!
+//! Computes liquidity depths for all pools using `query_pool_swap` when available,
```
tvinagre, line 3: excited to see how this plays out in our tests

src/derived/computations/spot_price.rs
tvinagre: We should support partial failures for every computation actually

ElusAegis: So adjust what the Self::Output is to a tuple of updated + failed IDs?

src/derived/computations/spot_price.rs
ElusAegis: As far as I understand that's more of the tycho-common question?

tvinagre (approved):

PR #25, opened 2026-01-21
feat: add RFQ stream

zizou0x:
This PR adds support for RFQs in TychoFeed. Protocol related config have been moved to a dedicated file to make it easier to maintain.

### DIFF
--- a/src/feed/tycho_feed.rs
+++ b/src/feed/tycho_feed.rs
@@ -7,37 +7,31 @@
 
 use std::{collections::HashSet, sync::Arc};
 
-use tokio::sync::{broadcast, mpsc, oneshot, RwLock};
+use tokio::{
+    sync::{broadcast, mpsc, oneshot, RwLock},
+    task::JoinHandle,
+};
 use tokio_stream::StreamExt;
 use tracing::{debug, info, instrument, span, trace, Instrument, Level};
 use tycho_simulation::{
-    evm::{
-        engine_db::tycho_db::PreCachedDB,
-        protocol::{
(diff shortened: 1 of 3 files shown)


src/feed/tycho_feed.rs
```diff
@@ -994,4 +947,44 @@ mod tests {
...
+    #[ignore]
+    async fn test_real_combined_feed() {
+        let rpc_url = env::var("RPC_URL").expect("RPC_URL must be set");
+        let tycho_api_key = env::var("TYCHO_API_KEY").expect("TYCHO_API_KEY must be set");
+        let tycho_url = env::var("TYCHO_URL").expect("TYCHO_URL must be set");
+        let config = TychoFeedConfig::new(
+            tycho_url,
+            Chain::Ethereum,
+            tycho_api_key,
+            vec!["rfq:bebop".to_string(), "rfq:hashflow".to_string(), "uniswap_v2".to_string()],
```
ElusAegis, line 1079: Should it be seperately tested that `rfq` streem works? Just that this test will still pass if the `rfq` streem never produces events, as the `"uniswap_v2"` will still produce 5 messages.

ElusAegis (approved):
Quickly looked through this - Good work!

src/feed/tycho_feed.rs
zizou0x: I've added a test for this, I originally didn't do it because protocol streams was breaking if it was empty. I now added conditions so we run a stream only when it's needed

tvinagre (approved):

PR #26, opened 2026-01-21
feat: Implement Solver Builder and CLI interface

louise-poole:
The tycho-solver binary has the following interface:
CLI: with various args that relate to the solver as a whole
Config TOML: with worker-pool specific configs

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -38,6 +38,7 @@ pub struct TychoSolverBuilder {
     rpc_url: String,
     protocols: Vec<String>,
     min_tvl: f64,
+    min_token_quality: i32,
     tvl_buffer_multiplier: f64,
     gas_refresh_interval: Duration,
     reconnect_delay: Duration,
@@ -65,6 +66,7 @@ impl TychoSolverBuilder {
             rpc_url,
             protocols,
             min_tvl: defaults::MIN_TVL,
+            min_token_quality: defaults::MIN_TOKEN_QUALITY,
             tvl_buffer_multiplier: defaults::TVL_BUFFER_MULTIPLIER,
             gas_refresh_interval: Duration::from_secs(defaults::GAS_REFRESH_INTERVAL_SECS),
             reconnect_delay: Duration::from_secs(defaults::RECONNECT_DELAY_SECS),
@@ -88,6 +90,12 @@ impl TychoSolverBuilder {
     /// Sets the minimum TVL filter (default: 10.0).
     pub fn min_tvl(mut self, min_tvl: f64) -> Self {
         self.min_tvl = min_tvl;
+        self
+    }
+
+    /// Sets the minimum token quality filter.
+    pub fn min_token_quality(mut self, min_token_quality: i32) -> Self {
+        self.min_token_quality = min_token_quality;
         self
     }
 
@@ -157,7 +165,8 @@ impl TychoSolverBuilder {
         )
         .tvl_buffer_multiplier(self.tvl_buffer_multiplier)
         .gas_refresh_interval(self.gas_refresh_interval)
-        .reconnect_delay(self.reconnect_delay);
+        .reconnect_delay(self.reconnect_delay)
+        .min_token_quality(self.min_token_quality);
 
         let (tycho_feed, event_rx) =
             TychoFeed::new(tycho_feed_config, Arc::clone(&market_data), health_tracker.clone());
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,0 +1,301 @@
+use std::{collections::HashMap, sync::Arc, time::Duration};
+
+use actix_web::{dev::ServerHandle, App, HttpServer};
+use anyhow::{Context, Result};
+use futures::future::Either;
+use tokio::{sync::RwLock, task::JoinHandle};
+use tracing::{error, info};
+use tycho_simulation::tycho_common::models::Chain;
+
+use crate::{
+    api::{configure_app, AppState, HealthTracker},
+    config::{defaults, PoolConfig},
+    feed::market_data::SharedMarketData,
+    order_manager::{OrderManager, OrderManagerConfig, SolverPoolHandle},
+    task_queue::{TaskQueue, TaskQueueConfig},
+    worker::WorkerConfig,
+    worker_pool::{WorkerPool, WorkerPoolBuilder},
+    TychoFeed, TychoFeedConfig,
+};
+
+/// Builder that assembles the Tycho solver and returns a running server handle.
+///
+/// The builder does the following:
+/// - Creates a new Tycho feed
+/// - Creates worker pools (one task queue per pool)
+/// - Creates a new order manager
+/// - Creates a new HTTP server
+/// - Returns a running server handle
+pub struct TychoSolverBuilder {
+    chain: Chain,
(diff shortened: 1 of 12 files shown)


src/feed/tycho_feed.rs
ElusAegis, line 478: Should we add perhaps that this is test only?
```suggestion
    [cfg(test)]
    fn create_test_config() -> TychoFeedConfig {
```

ElusAegis, line 504:
```suggestion
    [cfg(test)]
    fn create_test_component(id: &str, tokens: Vec<Token>) -> ProtocolComponent {
```

src/builder.rs
```diff
@@ -0,0 +1,315 @@
...
+/// - Creates a new HTTP server
+/// - Returns a running server handle
+pub struct TychoSolverBuilder {
+    chain: Chain,
+    http_host: String,
+    http_port: u16,
+    pools: HashMap<String, PoolConfig>,
+    tycho_url: String,
+    tycho_api_key: Option<String>,
+    use_tls: bool,
```
ElusAegis: should it be renamed to `tycho_use_tls` or something similar to be clear what this is about? Or add a doc comment, as we have for all other use cases of `use_tls`

src/builder.rs
```diff
@@ -0,0 +1,315 @@
...
+                if let Err(e) = server_result {
+                    tracing::error!(error = %e, "Server task error");
+                }
+                info!("shutting down...");
+                feed_handle.abort();
+                for pool in worker_pools {
+                    pool.shutdown();
+                }
+            }
+            Either::Right((_, _server_task)) => {
```
ElusAegis: Was not aware of `futures::future::select`, that's a nice one!  `_server_task` probably should be without an underscore and we could also move the `for pool in worker_pools {` outside of the select, as both sides will do this

src/builder.rs
```diff
@@ -0,0 +1,315 @@
...
+                }
+            }
+        }
+
+        info!("shutdown complete");
+        Ok(())
+    }
+}
+
+pub fn parse_chain(chain: &str) -> Result<Chain> {
```
ElusAegis, line 298: Given that the `serde_json::from_str::<Chain>` accepts lower case, could we turn chain to lowercase and run the `from_str`?

src/config.rs
```diff
@@ -0,0 +1,79 @@
...
+/// Worker pools configuration loaded from TOML file.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct WorkerPoolsConfig {
+    /// Pool configurations (at least one pool must be specified)
+    pub pools: HashMap<String, PoolConfig>,
+}
+
+/// Per-pool configuration.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct PoolConfig {
```
ElusAegis, line 16: I might have spent too much time on this, but there is a way to write defaults to serde without declaring a unique fn for each field, which I explained below. Not sure that this is 100% worth the adjustment, but perhaps it will look more straight forward:

```rust
pub struct PoolConfig {
    /// Algorithm name for this pool (e.g., "most_liquid", "dijkstra")
    pub algorithm: String,
    /// Number of worker threads for this pool
    #[serde(default = "num_cpus::get")]
    pub num_workers: usize,
    /// Task queue capacity for this pool
    #[serde(default = "usize_val::<1000>")]
    pub task_queue_capacity: usize,
    /// Minimum hops to search (must be >= 1)
    #[serde(default = "usize_val::<1>")]
    pub min_hops: usize,
    /// Maximum hops to search
    #[serde(default = "usize_val::<3>")]
    pub max_hops: usize,
    /// Timeout for solving in milliseconds
    #[serde(default = "u64_val::<100>")]
    pub timeout_ms: u64,
}

fn usize_val<const V: usize>() -> usize {
    V
}

fn u64_val<const V: u64>() -> u64 {
    V
}
```

ElusAegis: Also, the CPU default can be just inlined

worker_pools.toml
```diff
@@ -0,0 +1,23 @@
+# Worker pools configuration file
+
+# Worker pools
+[pools.most_liquid_2_hops]
+algorithm = "most_liquid"
+num_workers = 1
+task_queue_capacity = 1000
+max_hops = 2
```
ElusAegis, line 8: We should also probably provide `min_hops`, as otherwise it will be by defualt 1, which means that both solvers will simulate the 1 and 2 hop paths

ElusAegis (approved):
Looks good!

src/feed/tycho_feed.rs
zizou0x: The full module is already test only

src/cli.rs
```diff
@@ -0,0 +1,167 @@
...
+use crate::config::defaults;
+
+/// Tycho Solver - High-performance DEX router built on Tycho
+///
+/// Finds optimal swap routes across multiple protocols using real-time market data.
+#[derive(Parser, PartialEq, Debug)]
+#[command(name = "tycho-solver", version, about, long_about = None)]
+pub struct Cli {
+    /// Target chain (e.g. Ethereum)
+    #[arg(short, long, default_value = "Ethereum")]
```
zizou0x, line 14: Would it work also with lower case?

src/builder.rs
```diff
@@ -0,0 +1,315 @@
...
+    http_host: String,
+    http_port: u16,
+    pools: HashMap<String, PoolConfig>,
+    tycho_url: String,
+    tycho_api_key: Option<String>,
+    use_tls: bool,
+    rpc_url: String,
+    protocols: Vec<String>,
+    // Optional fields with defaults
+    min_tvl: f64,
```
tvinagre, line 40: we should also expose token min_quality

src/lib.rs
```diff
@@ -73,6 +73,9 @@
 
 pub mod algorithm;
 pub mod api;
+pub mod builder;
+pub mod cli;
```
zizou0x, line 77: I haven't put much thoughts here but should CLI be part of the lib or just use in the binary?

worker_pools.toml
louise-poole: Can do 👍 This is just an example config - will get overwritten upon deployment. I don't think in all cases we would want them to be mutually exclusive. I'm picturing:
OrderManager gets an order that has a very short timeout: sends it to the 2 hop pool only since it's ranked a 'faster solver'.
OrderManager then gets an order with a more patient timeout: sends it to the 3 hop pool as it covers more options but might be a bit slower.

src/cli.rs
louise-poole: Yes

src/lib.rs
louise-poole: Hmmmm - true. I'm inclined to think just the binary. Will think on it a bit more and decide.

tvinagre (approved):

src/config.rs
louise-poole: Niiiiiiiice!!!! I like this a lot!

src/lib.rs
louise-poole: Removing this makes the import of cli in the main file awkward and ugly so I'll just keep it. I'll remove the re-export of the cli structs though.

PR #27, opened 2026-01-21
feat: implement computation manager

ElusAegis:


src/derived/manager.rs
```diff
@@ -0,0 +1,418 @@
...
+    error::ComputationError,
+    store::DerivedData,
+};
+use crate::{
+    feed::events::{EventError, MarketEvent, MarketEventHandler},
+    SharedMarketDataRef,
+};
+
+/// Thread-safe handle to shared derived data store.
+pub type SharedDerivedDataRef = Arc<RwLock<DerivedData>>;
```
tvinagre, line 28: Not sure I like this type, I think I prefer having a explicit Arc<RwLock - IMO changing the type makes it easier for someone to overlook this when reviewing or implementing.

src/derived/manager.rs
```diff
@@ -0,0 +1,418 @@
...
+                .last_updated()
+                .map(|b| b.number);
+            if block.is_none() {
+                warn!("computing derived data without block info - market data may not be initialized");
+            }
+            block
+        };
+
+        // Run all computations (each acquires locks as needed)
+        let token_prices_result = self
```
tvinagre: These three computations are independent - they could run concurrently:

``` rust 
  let (token_prices_result, spot_prices_result, pool_depths_result) = tokio::join!(
      self.token_price_computation.compute(&self.market_data, &self.store),
      self.spot_price_computation.compute(&self.market_data, &self.store),
      self.pool_depth_computation.compute(&self.market_data, &self.store),
  );
```

src/derived/manager.rs
```diff
@@ -0,0 +1,418 @@
...
+                    added = added_components.len(),
+                    removed = removed_components.len(),
+                    updated = updated_components.len(),
+                    "market updated, running all computations"
+                );
+
+                self.compute_all().await;
+            }
+            MarketEvent::GasPriceUpdated { .. } => {
+                trace!("gas price updated, recomputing token prices");
```
tvinagre: We will end up computing twice on every block - I think we can skip this event for now

src/derived/manager.rs
```diff
@@ -0,0 +1,418 @@
...
+    pub fn store(&self) -> SharedDerivedDataRef {
+        Arc::clone(&self.store)
+    }
+
+    /// Runs the computation manager's main loop.
+    ///
+    /// Processes market events and updates the derived data store until
+    /// shutdown is signaled or the event channel closes.
+    pub async fn run(
+        mut self,
```
tvinagre, line 130: By having the mut ref you can't access the manager after calling run(). The store() method must be called before run(). This is fine but could be surprising so please document this behavior.

src/derived/manager.rs
```diff
@@ -0,0 +1,418 @@
...
+
+#[async_trait]
+impl MarketEventHandler for ComputationManager {
+    async fn handle_event(&mut self, event: &MarketEvent) -> Result<(), EventError> {
+        match event {
+            MarketEvent::MarketUpdated {
+                added_components,
+                removed_components,
+                updated_components,
+            } if !added_components.is_empty() ||
```
tvinagre, line 257: we could 
```rust
  impl MarketEvent {
      fn has_changes(&self) -> bool { ... }
  }
```

src/derived/store.rs
```diff
@@ -78,7 +78,7 @@ impl DerivedDataStore {
     }
 
     /// Sets spot prices.
-    pub fn set_spot_prices(&mut self, prices: SpotPrices, block: u64) {
+    pub fn set_spot_prices(&mut self, prices: SpotPrices, block: Option<u64>) {
         self.spot_prices = Some(prices);
         self.last_block = block;
```
tvinagre: Do we want to update `last_block` to `None` if we have `prices`?

src/derived/computation.rs
```diff
@@ -93,15 +101,20 @@ pub trait DerivedComputation: Send + Sync + 'static {
...
     /// # Returns
     ///
     /// The computed output, or an error if computation failed.
-    fn compute(
+    ///
+    /// # Lock Management
+    ///
+    /// Computations should acquire locks only when needed and release them as early
+    /// as possible to minimize contention. Use `.read().await` for async lock acquisition.
+    async fn compute(
```
tvinagre, line 116: Can you add metrics for all the computations, pls? Either here or on the other PR

src/derived/manager.rs
```diff
@@ -0,0 +1,420 @@
...
+        let spot_prices_result = self
+            .spot_price_computation
+            .compute(&market_data, &store_read);
+        let pool_depths_result = self
+            .pool_depth_computation
+            .compute(&market_data, &store_read);
+
+        // Release read locks
+        drop(store_read);
+        drop(market_data);
```
louise-poole: Personal preference: I tend to like to scope locks within code blocks. So just open a `{`, get the lock, use it, then close `}`. To me this makes it clear where locks exists when skimming the code and I don't need to read actual words to see where it's dropped.

src/derived/manager.rs
```diff
@@ -0,0 +1,420 @@
...
+                    !updated_components.is_empty() =>
+            {
+                trace!(
+                    added = added_components.len(),
+                    removed = removed_components.len(),
+                    updated = updated_components.len(),
+                    "market updated, running all computations"
+                );
+
+                self.compute_all().await;
```
louise-poole, line 269: I'd prefer we only ran computations on the components listed in the event...

src/derived/manager.rs
louise-poole: Agreed

src/derived/store.rs
```diff
@@ -38,7 +38,7 @@ impl DerivedDataStore {
     }
 
     /// Sets token prices.
-    pub fn set_token_prices(&mut self, prices: TokenGasPrices, block: u64) {
+    pub fn set_token_prices(&mut self, prices: TokenGasPrices, block: Option<u64>) {
```
louise-poole: Surely it can't return a None block here no? Unless I guess the prices failed to calculate? Can we document this use case?

src/derived/manager.rs
ElusAegis: I have added for consistency with the Market. Should we remove them?

src/derived/manager.rs
ElusAegis: Nice! Will do! Only that now the `token_gas_price_computation` and  `self.pool_depth_computation` depend on the `spot_price_computation`, so I will only run them two in parallel

src/derived/manager.rs
ElusAegis: Agree, that would be very useful!

src/derived/store.rs
ElusAegis: Would switching to the follwoing make sense? If `self.market_data` does not have a block set yet, skip computations, otherwise, compute? I do not feel like throwing an error is worth a while here, so we could just warn and skip until we get a block number

src/derived/computation.rs
ElusAegis: Added for each computation to also report some specific info

src/derived/manager.rs
ElusAegis: Sounds good! I have one place where I could not use it, but I just inlined the lock aquisition and dropping. Will only leave the `drop(...)` for tests

src/derived/manager.rs
ElusAegis: Should we try to do it right now, or keep until further iteration?

src/derived/manager.rs
tvinagre: Fine to leave it as it is, as we discussed within the team and it doesn't bother everyone - we can adjust later

src/derived/store.rs
tvinagre: I just think it's odd that we have market data and not block. Can that happen?

src/derived/manager.rs
tvinagre: Can you do this in a separate PR please?

tvinagre (approved):

src/derived/store.rs
ElusAegis: Agree :) Sorry, I fixed this in this [commit](https://github.com/propeller-heads/tycho-solver/pull/27/commits/3b4410f28cc4e6e472c233c3ff9fc8ca942f099a) and forgot to reply here

PR #29, opened 2026-01-26
feat: add example to benchmark solving time

zizou0x:


worker_pools.toml
```diff
@@ -3,15 +3,16 @@
...
+num_workers = 10
 task_queue_capacity = 1000
 max_hops = 2
 timeout_ms = 100
 
-[pools.most_liquid_3_hops]
-algorithm = "most_liquid"
-num_workers = 1
-max_hops = 3
+# [pools.most_liquid_3_hops]
```
tvinagre: removed because it's slow?

Cargo.toml
```diff
@@ -57,6 +57,8 @@ chrono = "0.4"
 criterion = "0.5"
 rstest = "0.18"
 chrono = "0.4"
+reqwest = { version = "0.11", features = ["json"] }
```
tvinagre: Why use this version, since we already use 0.12.28 via tycho-sim?

worker_pools.toml
tvinagre: This file should be the "production" config so I wouldn't commit this if it's only because of the benchmark.

examples/benchmark/config.rs
```diff
@@ -0,0 +1,145 @@
...
+        println!("Using {} different request templates (randomized)", requests.len());
+        println!("First template example:");
+        println!("{}", serde_json::to_string_pretty(&requests[0])?);
+    }
+    println!();
+
+    Ok((requests, requests_file.map(|s| s.to_string())))
+}
+
+fn create_default_request() -> Result<SolutionRequest, Box<dyn std::error::Error>> {
```
tvinagre, line 177: in the future, we could expose "number_of_requests" to the config and use this number as a seed for generating random requests. This way we could have more than one, with reproducibility

examples/benchmark/main.rs
```diff
@@ -0,0 +1,237 @@
...
+        pools_config.pools,
+        cli.tycho_url.clone(),
+        cli.rpc_url.clone(),
+        protocols.clone(),
+    )
+    .http_port(cli.http_port)
+    .build()?;
+
+    let server_handle = solver.server_handle();
+    let solver_task = tokio::spawn(async move {
```
tvinagre: By doing this, both Solver and Bench share the same tokio thread pool, cpu cores & mem heap. This is not very troublesome if we're doing the sequential benchmark, but for the other ones, we'd be competing with the solver when doing serde ops, collecting statistics etc. Even the tokio's scheduler overhead would count as latency, inflating the metrics. In this case, I think it'd be safer to run the solver as a subprocess.

examples/benchmark/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Parallelization Modes
+
+- **sequential** - Wait for each response before firing the next request
+- **fixed:N** - Maintain exactly N concurrent requests (e.g., `fixed:5`)
+- **rate:Nms** - Fire requests every N milliseconds (e.g., `rate:100`)
+
+## Request Templates
+
+By default, uses a WETH→USDC swap. To use custom requests, create a JSON file with an array of `SolutionRequest` objects and use `--requests-file` to specify it. The benchmark will randomly select from your templates for each request. See `example_requests.json` for the format.
```
tvinagre: The file is `requests_set.json`

examples/benchmark/exporter.rs
```diff
@@ -0,0 +1,165 @@
...
+
+        let mut sorted = times.to_vec();
+        sorted.sort();
+
+        let min = sorted[0];
+        let max = sorted[sorted.len() - 1];
+        let sum: u64 = sorted.iter().sum();
+        let mean = sum / sorted.len() as u64;
+        let median = sorted[sorted.len() / 2];
+        let p95 = sorted[(sorted.len() as f64 * 0.95).min((sorted.len() - 1) as f64) as usize];
```
tvinagre, line 26: For very small sample sizes (like 2 requests) percentile calculations may produce misleading results. I'd add a minimum sample size check or raise a warning.

examples/benchmark/exporter.rs
```diff
@@ -0,0 +1,165 @@
...
+
+        println!("  {:>5}-{:<5}ms [{:>4}] {}", bucket_start, bucket_end, count, bar);
+    }
+}
+
+/// Exports benchmark results to a JSON file with complete configuration and statistics.
+/// The output includes worker pool configuration content and all timing measurements.
+#[allow(clippy::too_many_arguments)]
+pub fn export_results(
+    chain_str: String,
```
tvinagre: huge params, maybe we should add a struct?

examples/benchmark/config.rs
zizou0x: We already have more than one via the requests-set but true, it's not reproducible right now it relies on RNG. 

We could use a see to produce a a semi-random sequence

worker_pools.toml
zizou0x: Yes removed because it's too slow. 

I agree with you that this file should contain a generic prod config. But I don't think this 3 hops should be part of it. 

In the end it really depends on the graph size and speed requirements. Happy to define this properly but IMO something generic for a most liquid solver is max 2 hops to keep a decent graph size for a classic config with at least all uniswaps.

zizou0x: I removed my changes, I think think the 3hops config should be in the default but I don't think it has to be changed in this PR anyway

tvinagre: The goal is for the user to define the solution timeline (difference here between we're able to provide a solution in 50ms vs we provide all or the best solutions within 50ms), so I'd keep this second option in the toml file, but we can reenable it any time, not a big deal imo. I just want to have more than 1 solver pool by default so we are always testing the full architecture by default.
The ideal would be to have a second pool with a different algo, but that's a future step

examples/benchmark/config.rs
tvinagre: yeah I'm a fan of the seed for reproducibility - specially bc we know that some pools within the same protocol have simulation times way higher than others (like pools with a lot of ticks on Univ3)

tvinagre (approved):

tvinagre:
I'd add an option to run the server from the benchmark as a subprocess, as:
```
let solver_process = Command::new("cargo")
      .args(["run", "--release", "--", ...])
      .spawn()?;
```

but if we decide to do it, I'd do in a separate PR

PR #30, opened 2026-02-02
feat: Integrate Readiness Tracker into worker

tvinagre:
This PR also includes a series of bugfixes to resolve to a final working product.


examples/benchmark/runner.rs
```diff
@@ -13,7 +13,7 @@ impl ParallelizationMode {
         solver_url: &str,
         requests: &[SolutionRequest],
         num_requests: usize,
-    ) -> (Vec<u64>, Vec<u64>, usize) {
+    ) -> (Vec<u64>, Vec<u64>, usize, usize, usize) {
```
louise-poole: Would prefer if this was a struct we return instead of a tuple. But since this is just the benchmark I'm fine if we choose to leave it like this for now.

louise-poole: Could we maybe just document what is returned?

src/derived/computation.rs
```diff
@@ -31,22 +51,32 @@ impl ComputationRequirements {
...
     }
 
-    /// Builder method to add a required computation.
-    pub fn with_required(mut self, id: ComputationId) -> Self {
-        self.required.insert(id);
+    /// Builder method to add a computation that requires fresh data (current block).
+    pub fn with_fresh(mut self, id: ComputationId) -> Self {
+        self.require_fresh.insert(id);
         self
     }
```
louise-poole, line 76: I'm not a fan of this use of the builder pattern. We have both a constructor and builder fn to do the same thing? `fresh` and `with_fresh`? When do we use which? I would prefer we just have a `new` that makes an empty requirements and then we have the two `with` fns to add requirements.

src/derived/computation.rs
```diff
@@ -31,22 +51,32 @@ impl ComputationRequirements {
...
 
-    /// Builder method to add an optional computation.
-    pub fn with_optional(mut self, id: ComputationId) -> Self {
-        self.optional.insert(id);
+    /// Builder method to add a computation that allows stale data (any past block).
+    pub fn with_stale(mut self, id: ComputationId) -> Self {
+        self.allow_stale.insert(id);
         self
     }
```
louise-poole: Also we have no check for if the same requirement was added to both the stale and the fresh lists? Would this cause issues?

src/derived/computations/pool_depth.rs
```diff
@@ -148,62 +171,98 @@ impl DerivedComputation for PoolDepthComputation {
 
     const ID: ComputationId = "pool_depths";
 
-    #[instrument(level = "debug", skip(market, store), fields(computation_id = Self::ID, updated_pool_depths))]
+    #[instrument(level = "debug", skip(market, store, changed), fields(computation_id = Self::ID, updated_pool_depths))]
     async fn compute(
         &self,
         market: &SharedMarketDataRef,
         store: &SharedDerivedDataRef,
+        changed: &ChangedComponents,
```
louise-poole, line 183: ❤️

src/derived/computations/pool_depth.rs
```diff
@@ -148,62 +171,98 @@ impl DerivedComputation for PoolDepthComputation {
...
     ) -> Result<Self::Output, ComputationError> {
         let market = market.read().await;
-        let store = store.read().await;
+        let store_guard = store.read().await;
```
louise-poole, line 186: ❤️

src/derived/computations/spot_price.rs
```diff
@@ -39,63 +39,105 @@ impl DerivedComputation for SpotPriceComputation {
...
+                .await
+                .spot_prices()
+                .cloned()
+                .unwrap_or_default()
+        };
+
+        // Remove spot prices for removed components
+        for component_id in &changed.removed {
+            spot_prices.retain(|key, _| &key.0 != component_id);
+        }
```
louise-poole: Small optimisation: We can move this into the let if block above: we only need to perform it if we have fetched existing spot prices from the store.

src/derived/computations/spot_price.rs
```diff
@@ -39,63 +39,105 @@ impl DerivedComputation for SpotPriceComputation {
...
-                        )))?;
+                        warn!(
+                            component_id,
+                            token_in = %token_in.address,
+                            token_out = %token_out.address,
+                            error = %e,
+                            "spot price failed, skipping pair"
+                        );
+                        spot_prices.remove(&key);
+                        failed += 1;
```
louise-poole, line 132: Should we maybe error if all spot price calculations fail?

src/derived/computations/pool_depth.rs
```diff
@@ -148,62 +171,98 @@ impl DerivedComputation for PoolDepthComputation {
...
+                .keys()
+                .chain(changed.updated.iter())
+                .collect()
+        };
+
+        let mut succeeded = 0usize;
+        let mut failed = 0usize;
+
+        for component_id in components_to_compute {
+            // Get token addresses - from changed.added for new components, from topology for updates
```
louise-poole: Not fully sure why, but I got a bit tripped up on this sentence. I think this sounds clearer:
```suggestion
            // Get token addresses: changed.added for new components, topology for existing
```

src/derived/computations/spot_price.rs
```diff
@@ -39,63 +39,105 @@ impl DerivedComputation for SpotPriceComputation {
...
+                .keys()
+                .chain(changed.updated.iter())
+                .collect()
+        };
+
+        let mut succeeded = 0usize;
+        let mut failed = 0usize;
+
+        for component_id in components_to_compute {
+            // Get token addresses - from changed.added for new components, from topology for updates
```
louise-poole:
```suggestion
            // Get token addresses: changed.added for new components, topology for existing
```

src/derived/computations/token_gas_price.rs
```diff
@@ -281,121 +286,238 @@ impl TokenGasPriceComputation {
...
-            .into_iter()
-            .map(|(k, (_, price))| (k, price))
+        Ok(best_prices)
+    }
+
+    /// Attempts incremental computation for state-only changes.
+    ///
+    /// Returns `Ok(Some(prices))` if incremental computation succeeded,
+    /// `Ok(None)` if we should fall through to full computation (e.g., no deps stored yet),
+    /// or `Err` on failure.
```
louise-poole:
```suggestion
    /// Attempts incremental recomputation for state-only changes.
    ///
    /// Only recomputes token prices whose dependency paths intersect with changed components.
    /// Returns `Ok(Some(prices))` if incremental recomputation succeeded,
    /// `Ok(None)` if full recomputation is needed (e.g., no dependencies stored yet),
    /// or `Err` if computation failed.
```

src/derived/computation.rs
```diff
@@ -14,15 +14,35 @@ pub type ComputationId = &'static str;
...
-/// Each algorithm declares which computations it needs:
-/// - `required`: Must wait for these before solving (blocks)
-/// - `optional`: Best-effort, use if available (non-blocking)
+/// Each algorithm declares which computations it needs and their freshness requirements:
+///
+/// - `require_fresh`: Data must be from the current block (same block as SharedMarketData).
+///   Workers wait for these computations to complete for the current block before solving.
+///
+/// - `allow_stale`: Data can be from any past block, as long as it has been computed at least once.
+///   Workers only check that the data exists, not that it's from the current block.
```
louise-poole: I like this wording a lot more than 'with_fresh' and 'with_stale'. The 'with' terminology sounds like we're passing it the stale/fresh data and not that we're setting an expectation.
Can we update the actual fn names to be these? Right now there is a mismatch between them and these docs

src/derived/events.rs
```diff
@@ -0,0 +1,40 @@
...
+    },
+
+    /// All computations completed for a block.
+    ///
+    /// This is a convenience event - workers tracking individual computations
+    /// don't need to wait for this, but it can be useful for diagnostics.
+    AllComplete {
+        /// Block number all computations completed for.
+        block: u64,
+    },
```
louise-poole: Is this needed? It seems frivolous but I don't feel too strongly about it either way.

src/derived/events.rs
```diff
@@ -0,0 +1,40 @@
...
+/// when required computations are ready.
+#[derive(Debug, Clone)]
+pub enum DerivedDataEvent {
+    /// A new block has started, clearing previous readiness state.
+    ///
+    /// Workers should clear their ready set when receiving this event.
+    NewBlock {
+        /// The new block number.
+        block: u64,
+    },
```
louise-poole, line 20: Surely a worker will get a new block message from the Feed? Is it necessary to get a new block message here too?

src/feed/tycho_feed.rs
```diff
@@ -343,9 +346,9 @@ impl TychoFeed {
         trace!("Market data updated");
 
         // Only broadcast event if there are actual changes
-        if !added_components.is_empty() ||
-            !removed_components.is_empty() ||
-            !updated_components_ids.is_empty()
+        if !added_components.is_empty()
+            || !removed_components.is_empty()
+            || !updated_components_ids.is_empty()
```
louise-poole: Formatting 👀

src/graph/petgraph.rs
```diff
@@ -8,16 +8,36 @@
...
 use petgraph::{graph::NodeIndex, stable_graph};
+use tracing::debug;
 use tycho_simulation::tycho_common::models::Address;
 
+/// Components that are blacklisted from routing due to simulation issues.
+///
+/// These pools are excluded from the routing graph because they have known
+/// issues with the simulation (e.g., rebasing tokens that don't work correctly
+/// with certain protocols).
+static BLACKLISTED_COMPONENTS: LazyLock<HashSet<&'static str>> = LazyLock::new(|| {
```
louise-poole: Would be nice if this was configurable - maybe a blacklist toml?

src/graph/petgraph.rs
```diff
@@ -302,9 +397,21 @@ impl<D: Clone + Send + Sync> GraphManager<StableDiGraph<D>> for PetgraphStableDi
         self.edge_map.clear();
         self.node_map.clear();
 
-        let unique_tokens: HashSet<Address> = component_topology
+        // Filter out blacklisted components
+        let filtered_topology: HashMap<_, _> = component_topology
```
louise-poole: I would prefer we filtered out blacklist components on the feed side and not on the graph. The graph should try stay pure to graph logic only. 
If we blacklist on the feed side we avoid triggering recomputation attemps etc for these unusable components.

src/derived/tracker.rs
```diff
@@ -0,0 +1,408 @@
...
+///     block: 99, // stale is fine
+/// });
+///
+/// // Check if ready to solve
+/// if tracker.is_ready() {
+///     // Safe to acquire DerivedData lock and solve
+/// }
+/// ```
+#[derive(Debug)]
+pub struct ReadinessTracker {
```
louise-poole, line 50: I don't see this being used by the worker yet though?

src/builder.rs
```diff
@@ -273,6 +306,9 @@ pub struct TychoSolver {
     worker_pools: Vec<WorkerPool>,
     feed_handle: JoinHandle<()>,
     gas_price_worker_handle: JoinHandle<()>,
+    computation_manager_handle: JoinHandle<()>,
+    #[allow(dead_code)]
```
louise-poole: Is this still dead code?

src/derived/events.rs
tvinagre: I've left it because I thought it could be good for diagnostics. The reasoning is that in the future, if the system needs to scale horizontally, the derived data computation can be done in a separate worker and, since we can't guarantee the message delivery, one worker could receive this message and know that the computations are done even if it has missed a previous message. Anyways, I agree it's frivolous - fine to remove it but I think it's harmless

src/derived/events.rs
tvinagre: Serves as a liveness check for the derived computation. Also, the ReadinessTracker does not consume from the dataFeed, only from the derived computation events feed. I'd rather keep this one

src/builder.rs
tvinagre: I naively expected the compiler to warn me about this 😭

src/derived/types.rs
```diff
@@ -46,3 +46,23 @@ pub type TokenGasPriceKey = Address;
...
+///
+/// Tracks which components (pools) were used in the selected path,
+/// enabling selective recomputation when only specific pools change.
+#[derive(Debug, Clone)]
+pub struct TokenPriceEntry {
+    /// The computed mid-price relative to gas token.
+    pub price: Price,
+    /// Components (pool IDs) used in the selected path.
+    ///
+    /// Used for invalidation: if any of these components change,
```
ElusAegis, line 60: As far as I understand `path_components` only track the current best price ComponentId. So if we ever get a new best price option, we will not discover it until the current best price has any component change in it

src/derived/computation.rs
```diff
@@ -14,15 +17,35 @@ pub type ComputationId = &'static str;
...
-    pub required: HashSet<ComputationId>,
-    /// Computations that are useful but not required.
-    pub optional: HashSet<ComputationId>,
+    /// Computations that must be from the current block.
+    pub require_fresh: HashSet<ComputationId>,
+    /// Computations that can use data from any past block.
+    ///
+    /// TODO: Stale data can be dangerous if stale for too long. In the future, associate staleness
+    /// to a block limit might be implemented.
+    pub allow_stale: HashSet<ComputationId>,
```
ElusAegis, line 56: This is a lot cleaner for the naming 🤗And probably when there is a parameter for how long data is valid for, `allow_stale` could be merged with `require_fresh` and just specify `staleness: 0`

src/derived/manager.rs
```diff
@@ -6,17 +6,68 @@
...
 use tycho_simulation::tycho_common::models::Address;
 
+use crate::{feed::market_data::SharedMarketData, types::ComponentId};
+
+/// Information about which components changed in a market update.
+///
+/// Used to enable incremental computation - only recomputing derived data
+/// for components that actually changed.
+#[derive(Debug, Clone, Default)]
+pub struct ChangedComponents {
```
ElusAegis, line 27: I guess eventually it would be nice to pass the events here directly, such as MarketUpdated, and allow the DerivedData to handle them individually. This would also allow creating an event CleanStart to represent the `is_full_recompute`

src/derived/manager.rs
```diff
@@ -6,17 +6,68 @@
...
+    pub is_full_recompute: bool,
+}
+
+impl ChangedComponents {
+    /// Creates a marker for full recompute where all components are considered changed.
+    ///
+    /// Used for startup and lag recovery scenarios.
+    pub fn all(market: &SharedMarketData) -> Self {
+        Self {
+            added: market.component_topology().clone(),
```
ElusAegis, line 44: Should `added` be also empty, as the computation could access this data from the passed market data parameter?

src/derived/computations/pool_depth.rs
```diff
@@ -102,25 +104,50 @@ impl PoolDepthComputation {
...
+            .spot_price(token_in, token_out)
+            .map_err(|e| {
+                ComputationError::SimulationFailed(format!(
+                    "spot_price failed for pool {component_id} {}/{}: {e}",
+                    token_in.address, token_out.address
+                ))
+            })?;
+
+        // Check if the limit itself doesn't exceed slippage — if so, depth is the limit
+        if let Ok(result) = sim_state.get_amount_out(max_input.clone(), token_in, token_out) {
```
ElusAegis, line 117: Absolutely missed that 👍

src/derived/computations/pool_depth.rs
```diff
@@ -148,62 +175,99 @@ impl DerivedComputation for PoolDepthComputation {
...
-                    .get(&spot_price_key)
-                    .ok_or_else(|| ComputationError::InvalidDependencyData {
-                        dependency: SpotPriceComputation::ID,
-                        reason: format!(
-                            "missing spot price for pool {} {}/{}",
-                            component_id, token_in.address, token_out.address
-                        ),
-                    })?;
+                let spot_price_key: SpotPriceKey = key.clone();
+                let Some(spot_price) = spot_prices.get(&spot_price_key) else {
```
ElusAegis:
```suggestion
                let Some(spot_price) = spot_prices.get(&key) else {
```

[STYLING: OPTIONAL] This would also require cleaning an unused `SpotPriceKey` import and removing the `spot_price_key` as unused

src/feed/tycho_feed.rs
```diff
@@ -295,7 +335,13 @@ impl TychoFeed {
             })
             .max_by_key(|b| b.number);
 
-        debug!("Received message from with {} new components, {} removed components, and {} updated components", added_components.len(), removed_components.len(), updated_or_new_states.len());
+        debug!(
+            "Received message with {} new components ({} after blacklist filter), {} removed, {} updated",
```
ElusAegis, line 339: Just to check if this is intentional, we do not report `after blacklist filter` for the `removed` and `updated`

ElusAegis, line 339: Just to check if this is intentional, we do not report `after blacklist filter` for the `removed` and `updated`

src/types/constants.rs
```diff
@@ -15,4 +16,51 @@ lazy_static! {
         map.insert(ProtocolSystem::Balancer, 150_000);
         map
     };
+
```
ElusAegis: Not related to this PR, but it seems we have never used the `GAS_COST_PER_SWAP`, as we have access to the gas cost per component. Should we axe this?

src/builder.rs
```diff
@@ -181,10 +193,25 @@ impl TychoSolverBuilder {
 
         tycho_feed = tycho_feed.with_gas_price_worker_signal_tx(gas_price_worker_signal_tx);
 
+        // Computation manager for derived data (token prices, pool depths)
+        let gas_token = native_token(&self.chain).context("gas token not configured for chain")?;
+        let computation_config = ComputationManagerConfig::new()
+            .with_gas_token(gas_token)
+            .with_depth_slippage_threshold(defaults::DEPTH_SLIPPAGE_THRESHOLD);
+        let (computation_manager, _derived_event_rx) =
```
ElusAegis, line 201: Should we use the `params.derived_event_rx.resubscribe()` instead of `params.derived_event_tx.subscribe();` in the worker registry? Seems they would work identically

src/worker_pool/registry.rs
```diff
@@ -95,15 +101,17 @@ fn spawn_workers_generic<A, F>(
 ) -> Vec<JoinHandle<()>>
 where
     A: crate::algorithm::Algorithm + 'static,
-    A::GraphManager: crate::feed::events::MarketEventHandler,
+    A::GraphManager: crate::feed::events::MarketEventHandler + EdgeWeightUpdaterWithDepths,
```
ElusAegis: Would be nice to make it generic, and not depend on the  `GraphManger` impl `EdgeWeightUpdaterWithDepths`

src/worker_pool/worker.rs
```diff
@@ -116,6 +140,18 @@ where
             "processing order"
         );
 
+        // Check readiness before solving
```
ElusAegis, line 143: That is more of  a safety check, right? As we already await for the Dervied Data to be ready in the worker. And I also wonder if it would help  to check in the `self.readiness_tracker.is_ready()` or in the `solve` if the latest block in the `market_data` equals the latest block in the `readiness_tracker`, for consistency.

src/worker_pool/worker.rs
```diff
@@ -290,6 +447,19 @@ where
                             let _wait_time = task.wait_time();
                             let task_response_tx = task.response_tx;
 
+                            // Wait for derived data readiness before solving
+                            // Use algorithm timeout as the max wait time
+                            if let Err(e) = self.wait_until_ready(self.algorithm.timeout()).await {
```
ElusAegis, line 448: Would we then be effectively waiting for `2*timeout` at the worst case (the derived data waiting + algo timout)? Should the timeout be measured on the worker level then? And be a parameter to the `solve`?

src/algorithm/most_liquid.rs
```diff
@@ -94,10 +95,30 @@ impl DepthAndPrice {
...
+impl crate::graph::EdgeWeightFromSimAndDepths for DepthAndPrice {
+    fn from_sim_and_depths(
+        sim: &dyn ProtocolSim,
+        component_id: &ComponentId,
+        token_in: &Token,
+        token_out: &Token,
+        pool_depths: &crate::derived::PoolDepths,
+    ) -> Option<Self> {
+        let spot_price = sim
+            .spot_price(token_in, token_out)
```
ElusAegis: Should we use the derived data for spot prices as well? And perhaps then we could pass the full derived data read ref (as we already hold a read lock), instead of parts of it by reference. This would also allow having a general trait `EdgeWeightFromSimAndDervied`, and only have specific impl for particular types of weights defined by the Algo

src/algorithm/most_liquid.rs
```diff
@@ -323,33 +346,45 @@ impl MostLiquidAlgorithm {
...
-        let gas_cost_wei = total_gas * gas_price;
-        let gas_cost_out = gas_cost_wei * 1u32; // Placeholder until conversion is implemented
-
-        // Use BigInt to allow negative values when gas exceeds output
-        // (can happen due to inaccurate gas estimation or price conversion)
-        let net_amount_out = BigInt::from(output_amount) - BigInt::from(gas_cost_out);
-
-        Ok(Route::new(swaps, net_amount_out))
+
+        let net_amount_out = if let Some(last_swap) = swaps.last() {
```
ElusAegis: We could use the `Route::total_gas` method here

src/algorithm/most_liquid.rs
```diff
@@ -373,15 +408,27 @@ impl Algorithm for MostLiquidAlgorithm {
         &self,
         graph: &Self::GraphType,
         market: SharedMarketDataRef,
+        derived: Option<SharedDerivedDataRef>,
```
ElusAegis, line 415: Why do we want to make the derived data optional?

src/derived/types.rs
tvinagre: Yes, well raised - it's an optimization decision that I think we should address in the future. I've added a TODO and will add a task for it

src/derived/computation.rs
tvinagre: Yes :)

src/derived/manager.rs
tvinagre: I agree - I think this is quite unnecessary repetition - it made sense on the first design but it became a repetitive design. I'll see if I can remove it. If it becomes too cumbersome, I'll add a task to clean this up in the future.

src/feed/tycho_feed.rs
tvinagre: yeah - was intentional. I think it would be too much.

src/worker_pool/registry.rs
tvinagre: Yes, I had added a TODO raising this point, as now the "agnosticity" of the WorkerPool is broken with this requirement. We should do this in another PR though

src/worker_pool/worker.rs
tvinagre: Yes, safety net for direct solve() calls. I think block consistency check is a good future enhancement, will add to the improvements list

src/worker_pool/worker.rs
tvinagre: Yeah, valid, currently, the `timeout_ms` is tied to a SolutionRequest, not an individual `Order`. I think it makes sense to expose these to the solver and the worker would wait until the smallest value of (`algo timeout` or `solution timeout`). 
I don't understand the 2* though

src/algorithm/most_liquid.rs
tvinagre: Because of the possibility of an algo not needing any of it

src/worker_pool/worker.rs
ElusAegis: Just that as far as I understand we might wait up to `solution timeout` until the dervied data is ready, and then again for `solution timeout` until the solution is ready, resulting in `2x`

src/algorithm/most_liquid.rs
ElusAegis: Just that if it is a ref without a lock, it seems not to hurt to pass it anyway. But I see the other side as well

ElusAegis (approved):
Saw that most of the issues were resolved

src/derived/computation.rs
```diff
@@ -54,26 +62,30 @@ impl ComputationRequirements {
...
-        Self { require_fresh: HashSet::new(), allow_stale: ids.into_iter().collect() }
-    }
-
     /// Builder method to add a computation that requires fresh data (current block).
-    pub fn with_fresh(mut self, id: ComputationId) -> Self {
+    ///
+    /// # Errors
+    ///
+    /// Returns `RequirementConflict` if the same ID is already in `allow_stale`.
+    pub fn expect_fresh(mut self, id: ComputationId) -> Result<Self, RequirementConflict> {
```
louise-poole: Can't this be named the same as the internal list? I prefer `require_fresh` tbh, but `expect_fresh` is ok enough.

src/derived/computation.rs
```diff
@@ -54,26 +62,30 @@ impl ComputationRequirements {
...
+        Ok(self)
     }
 
     /// Builder method to add a computation that allows stale data (any past block).
-    pub fn with_stale(mut self, id: ComputationId) -> Self {
+    ///
+    /// # Errors
+    ///
+    /// Returns `RequirementConflict` if the same ID is already in `require_fresh`.
+    pub fn expect_stale(mut self, id: ComputationId) -> Result<Self, RequirementConflict> {
```
louise-poole: Some here - I prefer `allow_stale`. 'Expect stale' implies I **need** it to be stale which is not true. It's merely **ok** for it to be stale.

src/cli.rs
```diff
@@ -78,6 +78,10 @@ pub struct Cli {
     /// Path to worker pools TOML config file
     #[arg(short, long, env, default_value = "worker_pools.toml")]
     pub worker_pools_config: PathBuf,
+
+    /// Path to blacklist TOML config file (optional)
+    #[arg(long, env)]
+    pub blacklist_config: Option<PathBuf>,
```
louise-poole: Reminder: default to the blacklist.toml file pls

src/config.rs
```diff
@@ -44,6 +48,41 @@ impl WorkerPoolsConfig {
...
+}
+
+impl BlacklistConfig {
+    /// Load blacklist configuration from a TOML file.
+    ///
+    /// The TOML file should have a `[blacklist]` section:
+    /// ```toml
+    /// [blacklist]
+    /// components = ["0x86d257cdb7bc9c0df10e84c8709697f92770b335"]
+    /// protocol_systems = []
```
louise-poole: Are we blacklisting whole protocol systems? I assume this isn't meant to be here...

louise-poole (commented):
Some final small comments, otherwise I'm happy with this! Approved in the meantime (trusting you'll address my comments)

src/config.rs
tvinagre: Noo, this was not to be here, fixing. Thanks!

PR #32, opened 2026-02-06
feat: Create Quickstart example

tvinagre:


Cargo.toml
zizou0x: Let's remove this as we decided to not directly use rust benchmarking and use a benchmark example script instead

examples/quickstart/main.rs
```diff
@@ -0,0 +1,1250 @@
...
+    #[arg(long, default_value_t = 50)]
+    slippage_bps: u32,
+
+    /// Protocol systems to use (comma-separated, e.g., "uniswap_v2,uniswap_v3")
+    #[arg(
+        long,
+        default_value = "uniswap_v2,uniswap_v3,uniswap_v4,ekubo_v2,vm:curve,rocketpool,pancakeswap_v3,\
+        vm:maverick_v2,sushiswap_v2,erc4626,uniswap_v4_hooks,fluid_v1,pancakeswap_v2,vm:balancer_v2"
+    )]
+    protocols: String,
```
louise-poole: Ewwwwwww.... no better way to do this?

examples/quickstart/main.rs
```diff
@@ -0,0 +1,1250 @@
...
+    )
+    .await?;
+
+    // Display quote
+    display_quote(&quote, &sell_token, &buy_token, &amount_in)?;
+
+    // Determine if we can proceed with simulation/execution
+    let can_simulate_only = cli.simulate_only && cli.sender.is_some();
+    if private_key.is_none() && !can_simulate_only {
+        println!("\nNo PRIVATE_KEY set. Set it to simulate or execute swaps.");
```
zizou0x: Shouldn't this be returned instead of printed and return Ok()

examples/quickstart/main.rs
```diff
@@ -0,0 +1,1250 @@
...
+
+    if !resp.status().is_success() {
+        return Err(format!("Health check failed: {}", resp.status()).into());
+    }
+
+    Ok(resp.json().await?)
+}
+
+/// Fetch AMM protocol components via Tycho REST API (paginated).
+async fn fetch_amm_components(
```
zizou0x, line 474: We should use tycho-client lib here

louise-poole (approved):
LGTM!

examples/quickstart/main.rs
```diff
@@ -0,0 +1,1250 @@
...
+            .get(&solver_swap.component_id)
+            .cloned()
+            .unwrap_or_else(|| {
+                // If not found, create a minimal component
+                create_minimal_component(
+                    &solver_swap.component_id,
+                    &solver_swap.protocol.to_string(),
+                    chain,
+                )
+            });
```
zizou0x: What does this cover? I don't think this is expected to ever happen? If not we should probably just panic

examples/quickstart/types.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub tvl_gt: Option<f64>,
+    pub pagination: PaginationRequest,
+}
+
+/// Pagination parameters for REST API requests.
+#[derive(Debug, Serialize)]
+pub struct PaginationRequest {
+    pub page: usize,
+    pub page_size: usize,
+}
```
zizou0x: Can't we use tycho-common DTO instead?

examples/quickstart/Readme.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Quick Start
+
+### 1. Start the solver
+
+In one terminal:
+```bash
+export TYCHO_URL=tycho-beta.propellerheads.xyz
+export TYCHO_API_KEY=your_api_key
+export RPC_URL=https://eth-mainnet.g.alchemy.com/v2/your_key
```
zizou0x: Better to stay neutral on the provider here

examples/quickstart/main.rs
```diff
@@ -0,0 +1,1250 @@
...
+            run_eth_simulation(
+                &provider,
+                &tx,
+                &sell_token_address,
+                &amount_in,
+                signer.address(),
+                chain.id(),
+            )
+            .await?;
+        }
```
zizou0x: Seems to be repeated in case 0 below, we should probably make it better? Maybe using a simulation handler object with config instead of raw functions for each case

examples/quickstart/main.rs
```diff
@@ -0,0 +1,1250 @@
...
+                break;
+            }
+            page += 1;
+        }
+    }
+
+    Ok(all_components)
+}
+
+async fn get_solver_quote(
```
zizou0x, line 524: Maybe we don't want to maintain this ourselves but it would be nice to provide a tycho-pathfinder-client lib that makes all these query functions easy to use

examples/quickstart/main.rs
```diff
@@ -0,0 +1,1250 @@
...
+        if let Some(ref err) = sim_result.simulation.error_message {
+            println!("    Error: {}", err);
+        }
+    }
+
+    Ok(())
+}
+
+fn build_approval_calldata(amount: &BigUint) -> Vec<u8> {
+    let permit2_address = Address::from_str("0x000000000022D473030F116dDEE9F6B43aC78BA3").unwrap();
```
zizou0x, line 1214: Should we make it configurable or it's fine to assume this address is the same on all chains (I think this assumption is valid for all supported chains so far)

zizou0x (approved):
Mostly looks good, added some comments for improvements

examples/quickstart/main.rs
tvinagre: Actuallyyyyyy, I think I'll default to `None` and try to get all them from the API

examples/quickstart/main.rs
tvinagre: true

examples/quickstart/types.rs
tvinagre: I forgot we had them 😅

examples/quickstart/main.rs
tvinagre: Yeah I'd rather to that in the future though

PR #33, opened 2026-02-09
feat: Add README.md

tvinagre:


README.md
```diff
@@ -1 +1,195 @@
...
+## Features
+
+- **Multi-protocol routing** - Routes through your favorite on-chain liquidity protocol, like Uniswap, Balancer, Curve, RFQ protocols, or any other protocol supported by [Tycho](https://docs.propellerheads.xyz/tycho/for-solvers/supported-protocols).
+- **Real-time market data** - Tycho Stream keeps all liquidity states synchronized every block
+- **Multi-algorithm competition** - Multiple solver pools run different algorithm configurations in parallel; the best result wins
+- **Gas-aware ranking** - Solutions are ranked by net output after gas costs, not just raw output
+- **Sub-100ms solves** - Dedicated OS threads for CPU-bound route finding, separate from the async I/O runtime
+- **Production-ready** - Prometheus metrics, structured logging, health endpoints, graceful shutdown
+- **Extensible** - Implement the `Algorithm` trait to add new routing strategies with zero framework changes
+- **Multi-chain** - Supports Ethereum, Base, and Unichain
```
dianacarvalho1: ambitious to claim this when we haven't tested in other chains 😕 should we (purpellers) test it in base and unichain?

README.md
```diff
@@ -1 +1,195 @@
...
+
+## Quick Start
+
+### Prerequisites
+
+- Rust 1.75+ (2021 edition)
+- A Tycho API key ([get one here](https://www.propellerheads.xyz/tycho))
+- An Ethereum RPC endpoint
```
dianacarvalho1: depends on the chain

README.md
```diff
@@ -1 +1,195 @@
...
+
+Returns service health status. HTTP 200 if healthy, 503 if stale.
+
+## Configuration
+
+### CLI / Environment Variables
+
+| Flag | Env Var | Default | Description |
+|------|---------|---------|-------------|
+| `--rpc-url` | `RPC_URL` | (required) | Ethereum RPC endpoint |
```
dianacarvalho1: chain

README.md
```diff
@@ -1 +1,195 @@
...
+| `-w, --worker-pools-config` | `WORKER_POOLS_CONFIG` | `worker_pools.toml` | Worker pools config |
+| `--blacklist-config` | `BLACKLIST_CONFIG` | `blacklist.toml` | Blacklist config |
+
+See `--help` for the full list.
+
+Find the list of all available protocols on Tycho [here](https://docs.propellerheads.xyz/tycho/for-solvers/supported-protocols)
+
+### Worker Pools (worker_pools.toml)
+
+Configure solver pools with different algorithms and parameters:
```
dianacarvalho1, line 169: I think a bigger explanation of the worker pools is warranted here

README.md
```diff
@@ -1 +1,195 @@
...
+
+```bash
+# Clone and build
+git clone <repo-url>
+cd tycho-solver
+cargo build --release
+
+# Set required environment variables
+export TYCHO_API_KEY=your-api-key
+export RPC_URL=https://eth.llamarpc.com
```
dianacarvalho1, line 33:
```
export RUST_LOG=tycho_solver=info
```
otherwise we don't see anything when running

dianacarvalho1 (commented):
Thank you @tvinagre! 🙏🏼 I have a few comments

README.md
```diff
@@ -1 +1,195 @@
...
+
+# Run
+cargo run --release -- \
+  --tycho-url tycho.propellerheads.xyz \
```
dianacarvalho1: this url doesn't even exist

tvinagre: copy pasta went wrong - thanks!

README.md
tvinagre: here I treated Ethereum RPC as the EVM-compatible Node, not Ethereum Mainnet RPC. But will add this to the docs

README.md
tvinagre: Removed multichain mentioning

README.md
tvinagre: This was going on the gitbook docs, but I'll add a brief explanation

README.md
```diff
@@ -1 +1,203 @@
...
+
+- Rust 1.75+ (2021 edition)
+- A Tycho API key ([get one here](https://www.propellerheads.xyz/tycho))
+- An Ethereum RPC endpoint for the target chain
+
+### Run the Solver
+
+```bash
+# Clone and build
+git clone <repo-url>
```
zizou0x: Shouldn't we add the real url here?

zizou0x (approved):

README.md
```diff
@@ -1 +1,203 @@
...
+export RUST_LOG=info
+
+# Run
+cargo run --release -- \
+  --tycho-url tycho-beta.propellerheads.xyz \
+  --rpc-url $RPC_URL \
+  --protocols uniswap_v2,uniswap_v3
+```
+
+The solver starts on `http://localhost:3000` by default.
```
dianacarvalho1, line 43: can we add a small section for RFQs? to say that:
-  you can not run them alone (without onchain protocols) 
- you need to set some env vars
- that here we only deal with indicative prices and the longer you take to encode+execute the bigger might be the difference between the price on the solution we give and the actual executed one (this is actually valid for all protocols 🤔 but because of slippage)

README.md
tvinagre: this repo will change name so I was waiting for the final naming definition but will add it here for now

README.md
tvinagre: > that here we only deal with indicative prices and the longer you take to encode+execute the bigger might be the difference between the price on the solution we give and the actual executed one (this is actually valid for all protocols 🤔 but because of slippage)

I think this belongs on the gitbook, not on the README

ElusAegis (approved):

PR #39, opened 2026-02-16
refactor: conditionally enable OTEL

zizou0x:
If the env variable isn't set, we don't start OTEL anymore.


dianacarvalho1 (approved):
thank you @zizou0x 🙏🏼

PR #40, opened 2026-02-19
fix: reduce lock hold duration on the depth calculator

louise-poole:
Problem: currently the depth calculation holds the shared market state lock for the entire duration of its calculations. On chains whose block time is faster than the depth calculator, this causes us to fall behind on consuming blocks as the tycho-feed cannot get a write lock until the depth calculations complete.

Solution: on the depth calculator we instead hold the lock long enough to clone any data relevant to the calculations it needs to do. Thereafter we drop the lock and proceed with the slow calculations in the background.


dianacarvalho1 (approved):
Thank you @louise-poole ! 🙏🏼 It looks good to me!

PR #41, opened 2026-02-20
feat: Split the monolithic Fynd codebase into two libraries

dianacarvalho1:
New Structure:
- fynd-core: Pure solving logic with no HTTP dependencies:
  - Algorithm implementations (MostLiquidAlgorithm)
  - Graph management (petgraph integration)
  - Derived data computations (spot prices, pool depths, gas prices)
  - Feed management (Tycho integration, market data)
  - Worker pool with multi-threaded solver
  - Core types (Order, Route, Swap, OrderSolution)
  - Order manager for solution coordination
- fynd-rpc: HTTP server and API layer:
  - REST API handlers and error types
  - Server builder and configuration
  - Application state and health tracking

Let me know if you think that something should be somewhere else!

Also done:
- Moved API specific structs from types.rs to api/types.rs
- HealthTracker now depends on SharedMarketDataRef instead of being inside TychoFeed
- Removed rocketpool
- Added solve_order.rs example for fynd-core. The output looks like this
- Make some structs private (that don't need to be public)
```

=== Fynd Core: Solve Order Example ===

Loading market data... done. Waiting for fresh data... done
Computing derived data... (waiting 1 minute) done
Solving: 1000 USDC → WBTC ... done (1275ms)

Result: 0.02 WBTC
Gas:    264000

Route (2 hops):
  1. 1000.000000 USDC → 1000.118841 USDT (uniswap_v3)
  2. 1000.118841 USDT → 0.015861 WBTC (uniswap_v3)

```
TODO in another PR:
- Are the current public structs nicely designed?


docs/ARCHITECTURE.md
```diff
@@ -133,7 +149,8 @@ Actix Web HTTP handlers. Validates requests, delegates to OrderManager, returns
 
 ### 2. OrderManager
 
-**Location:** `src/order_manager/`
+**Crate:** `fynd-rpc`
+**Location:** `fynd-rpc/src/order_manager/`
```
kayibal: I'm surprised that this lives in the RPC layer.

examples/tutorial/types.rs
```diff
@@ -79,7 +79,7 @@ pub trait SwapToExecution {
     ) -> tycho_execution::encoding::models::Swap;
 }
 
-impl SwapToExecution for fynd::Swap {
+impl SwapToExecution for fynd_core::Swap {
```
kayibal, line 82: Should we implement `Into<tycho_execution::Swap> for fynd::Swap`, instead?

fynd-core/src/feed/events.rs
kayibal, line 1: These types seem to be part of the derived data computation framework. I wonder if we could keep them private for now? I don't feel confident yet about this part so if we can keep it private I'd prefer it.

fynd-core/src/worker_pool/task_queue.rs
kayibal, line 1: Do we need this one to be pub? Or can we configure this via the WorkerPoolConfig?

fynd-core/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+lazy_static.workspace = true
+uuid.workspace = true
+chrono.workspace = true
+thiserror.workspace = true
+anyhow.workspace = true
+tracing.workspace = true
+metrics.workspace = true
+typetag.workspace = true
+# OpenAPI (only for schema generation, no runtime overhead)
+utoipa.workspace = true
```
kayibal: Do we need this in core? I'd expect this one to live in the rpc crate only.

fynd-rpc/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+uuid.workspace = true
+chrono.workspace = true
+itertools.workspace = true
+lazy_static.workspace = true
+thiserror.workspace = true
+anyhow.workspace = true
+tracing.workspace = true
+tracing-subscriber.workspace = true
+tracing-actix-web.workspace = true
+metrics.workspace = true
```
kayibal, line 27: Do we really need all of those?

kayibal (approved):
Thanks this looks mostly good. Added comments / questions to a few sections that we could have a closer look at.

docs/ARCHITECTURE.md
dianacarvalho1: you would expect it to be in core? The thing is that the return types of OrderManager are already the API expected types 😕 That's why I put it here. I was hesitant though - I'm open for discusion

examples/tutorial/types.rs
dianacarvalho1: yes, there is also a separate TODO for this mission to include encoding in Fynd so that would be covered there

fynd-core/Cargo.toml
dianacarvalho1: yeah initially I did that same thing.. but in the end I added it here so that the models could derive a `ToSchema` automatically and we can set an example value. See [here](https://github.com/propeller-heads/fynd/blob/d13f631478b76bea19945bc4a271980d42bcdfef/fynd-core/src/types/solution.rs#L34). Do you think it is worth it to do this only inside the rpc?

fynd-rpc/Cargo.toml
dianacarvalho1: I think yes but I will double check!

docs/ARCHITECTURE.md
dianacarvalho1: after thinking more about this, yes, order manager should be in fynd-core! moving it now!

fynd-rpc/Cargo.toml
dianacarvalho1: cleaned this up!

tamaralipows (approved):
Thank you!

PR #42, opened 2026-02-20
ci: Add API breaking change detection workflow

kayibal:
## Summary

- Adds `.github/workflows/semver-check.yaml` that runs `cargo-semver-checks` on every PR targeting `main`
- Posts a PR comment with the full `semver-checks` output when breaking public API changes are detected, then fails the job
- Automatically handles new crates: uses `cargo metadata` to enumerate all library crates, then `git cat-file` to skip any that don't exist in `origin/main` — avoiding errors when a PR introduces a brand-new crate alongside changes to existing ones

### DIFF
--- a/.github/workflows/semver-check.yaml
+++ b/.github/workflows/semver-check.yaml
@@ -1,0 +1,104 @@
+name: API Semver Check
+
+on:
+  pull_request:
+    branches:
+      - main
+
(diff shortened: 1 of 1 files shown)


dianacarvalho1 (approved):
🙈 ✅

PR #43, opened 2026-02-21
feat: add simulation-driven Bellman-Ford routing algorithm

Haikane:
## Summary

- Add a new `BellmanFordAlgorithm` that runs actual simulations (`get_amount_out()`) during edge relaxation instead of scoring paths with heuristic spot prices
- SPFA optimization: active-nodes set replaces O(max_idx) per-layer scan, early termination when no distances improve
- Top-N re-simulation: re-simulates top 3 candidate layers instead of single best, addressing re-simulation divergence
- Subgraph extraction via BFS prunes the graph from ~10K edges to a few hundred before relaxation
- Runs as a separate worker configuration alongside MostLiquid; the existing OrderManager picks the best result across all configurations

## Benchmark Results (2026-03-16)

All benchmarks use real trades from Dune `dex.trades` on Ethereum. All comparisons are **net of gas** (BF uses gas-aware relaxation since commit `9a9e0e9`). "5-hop BF" means the algorithm explores up to 5 hops; actual routes may be shorter.

### Coverage

| Solver | Routes Found (of 10K) | % |
|------|---:|---:|
| BF (2-5 hop) | 7,569 | 75.7% |
| ML (2-3 hop) | 4,756 | 47.6% |

BF finds routes for ~59% more trades than ML. ML requires pre-computed edge weights (spot price + pool depth); if any edge lacks this data, ML drops the entire path. BF simulates directly via `get_amount_out()` and is unaffected.

### Best solver per trade (10K trades, all solvers concurrent)

| Solver | Best | % |
|------|---:|---:|
| BF 4-hop | 3,082 | 40.7% |
| BF 5-hop | 2,236 | 29.5% |
| BF 3-hop | 1,328 | 17.5% |
| ML 3-hop | 604 | 8.0% |
| BF 2-hop | 236 | 3.1% |
| ML 2-hop | 83 | 1.1% |

**BF is best for 90.8% of trades.**

### Head-to-head quality (contested trades only)

| Matchup | BF Wins | ML Wins | Ties |
|---------|---:|---:|---:|
| BF 2-hop vs ML 2-hop | 11.4% | 2.8% | 85.9% |
| BF 3-hop vs ML 3-hop | 84.9% | 0.9% | 14.2% |
| BF 5-hop vs ML 3-hop | 85.0% | 0.9% | 14.2% |

### Solve times (isolated, 1K trades)

| Algorithm | p50 | p95 | p99 |
|-----------|----:|----:|----:|
| ML 2-hop | 0ms | 1ms | 2ms |
| BF 2-hop | 4ms | 6ms | 16ms |
| BF 3-hop | 4ms | 7ms | 26ms |
| BF 5-hop | 5ms | 8ms | 37ms |
| ML 3-hop | 27ms | 391ms | 432ms |

BF 5-hop (5ms p50) finishes before ML 3-hop (27ms p50).

### Gas-aware relaxation (1K trades, v2+v3+v4, 6 native protocols)

| Pool | Wins | % |
|------|------|---|
| BF gas-aware (5-hop) | 470 | 61.0% |
| BF gross (5-hop) | 57 | 7.4% |

Gas-aware BF wins 96% of head-to-head vs gross BF.

### Key findings

1. BF covers 59% more trades (architectural advantage: no edge weight dependency)
2. At equal depth (3 vs 3), BF wins 85% of contested trades
3. BF 5-hop is 5x faster than ML 3-hop
4. Gas-aware relaxation adds another 61% vs 7.4% win rate over gross BF
5. Both should run concurrently; ML still wins ~8% of trades (mostly v4 routes)

## Design

Based on [Janos Tapolcai's simulation-based BF arbitrage searcher](https://github.com/jtapolcai/tycho-searcher), adapted from cycle detection to A→B routing.

**Key architectural decisions:**
- **Flat SPFA**: `distance[node]` tracks the best amount reachable at each token. Active set (SPFA queuing) ensures only tokens that improved get re-relaxed.
- **Forbid revisits**: Before relaxing an edge, the predecessor chain is walked to check that neither the destination token nor the pool already appears in the path. Prevents routes that depend on arbitrage opportunities.
- **Gas-aware relaxation**: Compares net amounts (`gross_output - cumulative_gas_cost_in_token_terms`) instead of gross output during edge relaxation, so deeper routes are only chosen when they actually beat shorter ones after gas.
- **Subgraph extraction**: BFS from source prunes the graph from ~10K edges to a few hundred before relaxation starts.
- **`()` edge weights**: BF simulates directly via `get_amount_out()` during relaxation; no pre-computed edge data needed.

See `docs/bellman-ford-explainer.md` for the full algorithm walkthrough.

## Files changed

| File | Change |
|------|--------|
| `src/algorithm/bellman_ford.rs` | New algorithm (~580 lines + 570 test lines, 19 tests) |
| `src/algorithm/mod.rs` | Register module + re-export |
| `src/graph/mod.rs` | `EdgeWeightFromSimAndDerived` impl for `()` |
| `src/worker_pool/registry.rs` | `spawn_bellman_ford_workers` + match arm |
| `worker_pools.toml` | Add `bellman_ford_pool` config (2 workers, max_hops=5, 200ms timeout) |
| `blacklist.toml` | Add UniV2 AMPL/WETH pool to blacklist |
| `src/lib.rs` | Re-export `BlacklistConfig` from crate root |

## Test plan

- [x] `cargo clippy --lib --tests -- -D warnings` passes
- [x] 19 BF-specific unit tests pass (`cargo test --lib algorithm::bellman_ford`)
- [x] Full test suite passes (no regressions)
- [x] Live E2E: 10K real trades from Dune, BF is best configuration for 64-75% of trades depending on depth
- [ ] Run on staging with production traffic and compare `order_manager_best_solution_pool` metrics


examples/benchmark/README.md
```diff
@@ -79,3 +79,90 @@ cargo run --example benchmark --release -- \
...
+  solver.log \
+  bellman_ford_pool \
+  bellman_ford_v2_pool
+```
+
+Output shows wins, ties, and improvement percentages for contested trades (where both pools returned a route).
+
+### 7. Clean up
+
+After benchmarking, remove the v2 module and revert `registry.rs` / `worker_pools.toml` to their single-algorithm state. Merge the winning algorithm's code into the main module.
```
louise-poole: Was this part meant to be pushed?

Haikane: No, reverted in b324801. The A/B comparison scripts and README section were accidentally included and have been removed.

Haikane:
Addressing the two review comments:

## 1. Why does ML not find routes for some token pairs?

Investigated and confirmed experimentally. The root cause is ML's dependency on pre-computed edge weights.

ML requires `DepthAndPrice` (spot price + pool depth) on every edge along a path. If any edge lacks this data, `try_score_path` drops the entire path as unscorable. BF uses `()` edge weights and simulates directly via `get_amount_out()`, so it never hits this gate.

**Fresh BF-2 vs ML-2 run (both at max_hops=2, 1K trades):**

| | Count | % of contested |
|---|---:|---:|
| Both succeeded (contested) | 517 | |
| BF wins | 68 | 13.2% |
| ML wins | 62 | 12.0% |
| Ties | 387 | 74.9% |
| **BF only (ML returned nothing)** | **262** | |
| ML only | 0 | |

The debug log shows **36,576 "edge missing weight data, path cannot be scored"** messages from ML. The affected edges are pools where spot price is too small to compute depth (e.g., long-tail tokens with very low liquidity relative to WETH/USDC). ML's BFS finds the 2-hop paths fine, but the scoring gate rejects them all. BF walks right through because it simulates directly.

This is an architectural tradeoff, not a bug. ML trades coverage for faster heuristic scoring on well-known pairs (1ms vs 4ms at p50). Added an explanatory note to the PR description.

## 2. Solve times table updated

Added BF 2-hop to the table (was missing). Ran a solo BF-2 benchmark:

| Algorithm | p50 | p95 | p99 | mean | max |
|-----------|----:|----:|----:|-----:|----:|
| ML 2-hop | 1ms | 2ms | 3ms | 1ms | 19ms |
| **BF 2-hop** | **4ms** | **7ms** | **12ms** | **3ms** | **15ms** |
| BF 3-hop | 6ms | 11ms | 30ms | 6ms | 33ms |
| BF 5-hop | 12ms | 22ms | 78ms | 12ms | 87ms |
| ML 3-hop | 43ms | 443ms | 520ms | 145ms | 1,146ms |

BF-2 is 4x slower than ML-2 at p50 but still very fast in absolute terms. The coverage advantage (262 extra routes per 1K trades, 33.6% of solved orders) more than compensates.

Haikane:
## Updated Benchmark: Simplified BF (2026-03-13)

Fresh benchmark after the simplification commits (remove re-simulation, merge BFS+adj build, single conflict check). Same methodology: 10K real trades from Dune, UniV2+UniV3, sequential requests against live Tycho state.

### Coverage (routes found per solver, combined run)

| Solver | Routes Found | % of 10k |
|------|---:|---:|
| BF 4-hop | 7,569 | 75.7% |
| BF 5-hop | 7,569 | 75.7% |
| BF 3-hop | 7,556 | 75.6% |
| BF 2-hop | 7,526 | 75.3% |
| ML 3-hop | 4,756 | 47.6% |
| ML 2-hop | 4,746 | 47.5% |

BF finds routes for ~59% more trades than ML. This is the same architectural gap as before (ML requires pre-computed edge weights; BF simulates directly).

### Best solver (which solver produced the best result per order)

| Solver | Selected Best | % |
|------|---:|---:|
| BF 4-hop | 3,082 | 40.7% |
| BF 5-hop | 2,236 | 29.5% |
| BF 3-hop | 1,328 | 17.5% |
| ML 3-hop | 604 | 8.0% |
| BF 2-hop | 236 | 3.1% |
| ML 2-hop | 83 | 1.1% |

BF solvers are the best choice for **90.8%** of trades (up from ~64-75% in the pre-simplification benchmark).

### Head-to-head quality (contested trades where both found a route)

| Matchup | Contested | BF Wins | ML Wins | Ties |
|---------|---:|---:|---:|---:|
| BF 2-hop vs ML 2-hop | 4,746 | 539 (11.4%) | 132 (2.8%) | 4,075 (85.9%) |
| BF 2-hop vs ML 3-hop | 4,752 | 4,013 (84.4%) | 61 (1.3%) | 678 (14.3%) |
| BF 3-hop vs ML 3-hop | 4,756 | 4,039 (84.9%) | 43 (0.9%) | 674 (14.2%) |
| BF 4-hop vs ML 3-hop | 4,756 | 4,042 (85.0%) | 46 (1.0%) | 668 (14.0%) |
| BF 5-hop vs ML 3-hop | 4,756 | 4,041 (85.0%) | 42 (0.9%) | 673 (14.2%) |

At equal depth (3 vs 3), BF wins **84.9%** of contested trades (up from 71.1% pre-simplification). ML wins under 1%.

### Solve times (isolated, 1K trades each)

| Algorithm | p50 | p95 | p99 | mean | max | Solved/1K |
|-----------|----:|----:|----:|-----:|----:|---:|
| ML 2-hop | 0ms | 1ms | 2ms | 0ms | 3ms | 451 |
| **BF 2-hop** | **4ms** | **6ms** | **16ms** | **3ms** | **202ms** | **738** |
| **BF 3-hop** | **4ms** | **7ms** | **26ms** | **3ms** | **34ms** | **743** |
| **BF 4-hop** | **5ms** | **9ms** | **38ms** | **4ms** | **41ms** | **779** |
| **BF 5-hop** | **5ms** | **8ms** | **37ms** | **4ms** | **202ms** | **761** |
| ML 3-hop | 27ms | 391ms | 432ms | 117ms | 668ms | 463 |

BF 3-hop is now **4ms p50** (down from 6ms pre-simplification). BF 5-hop is **5ms p50** (down from 12ms). The simplification cut solve times roughly in half at higher hop counts.

### Comparison to pre-simplification

| Metric | Before | After | Change |
|--------|--------|-------|--------|
| BF 3-hop vs ML 3-hop win rate | 71.1% | 84.9% | +13.8pp |
| BF best solver overall | 64-75% | 90.8% | +16-27pp |
| BF 3-hop p50 | 6ms | 4ms | -33% |
| BF 5-hop p50 | 12ms | 5ms | -58% |
| BF coverage | ~75% | 75.7% | stable |

> Note: Absolute numbers vary between runs due to live market state. The relative comparisons (BF vs ML within a single run) are the meaningful signal.

Haikane:
## Gas-aware relaxation benchmark (2026-03-16)

Added gas-aware relaxation to BF: during SPFA edge relaxation, compare `amount_out - cumulative_gas_cost_in_token_terms` instead of gross `amount_out`. Configurable via `gas_aware_relaxation` in pool TOML config (defaults to true, commit `9a9e0e9`).

### BF gas-aware vs BF gross (head-to-head, 1K trades, uniswap_v2+v3)

| Pool | Wins | % |
|------|------|---|
| **bf_gas_aware** | **703** | **96.0%** |
| bf_gross | 29 | 4.0% |

### 3-way comparison (1K trades, 6 native protocols, 500ms timeout, 2 workers each, min_tvl=20)

| Pool | Wins | % |
|------|------|---|
| **bf_gas_aware** (5-hop) | **470** | **61.0%** |
| most_liquid (1-3 hop) | 243 | 31.6% |
| bf_gross (5-hop) | 57 | 7.4% |

### Deep solo analysis (500 trades, BF-only vs ML-only sequential runs)

| | Count | % |
|---|---:|---:|
| BF wins | 222 | 91.7% |
| ML wins | 18 | 7.4% |
| Ties | 2 | 0.8% |
| BF-only solved | 143 | - |

ML's 18 wins break down as:
- **9 simulation noise**: same pool, same route, tiny diffs from block state drift between sequential runs
- **4 BF picks 4-hop, ML's 3-hop is better net**: all USDT->WETH via v4-v4-v2
- **4 BF picks 1-hop, ML's 3-hop is dramatically better**: BF's greedy relaxation settles on the direct route before discovering multi-hop v4 paths

All 8 real ML wins use `uniswap_v4` routes. The earlier PR benchmarks (97.7% BF) used only v2+v3 where this gap doesn't exist.

Haikane:
## Consolidated Benchmark Results (2026-03-16)

All benchmarks use real trades from Dune `dex.trades` on Ethereum. All comparisons are **net of gas** (BF uses gas-aware relaxation since commit `9a9e0e9`). "5-hop BF" means the algorithm explores up to 5 hops; actual routes may be shorter.

### Coverage

| Solver | Routes Found (of 10K) | % |
|------|---:|---:|
| BF (2-5 hop) | 7,569 | 75.7% |
| ML (2-3 hop) | 4,756 | 47.6% |

BF finds routes for ~59% more trades than ML. ML requires pre-computed edge weights (spot price + pool depth); if any edge lacks this data, ML drops the entire path. BF simulates directly via `get_amount_out()` and is unaffected.

### Best solver per trade (10K trades, all solvers concurrent)

| Solver | Best | % |
|------|---:|---:|
| BF 4-hop | 3,082 | 40.7% |
| BF 5-hop | 2,236 | 29.5% |
| BF 3-hop | 1,328 | 17.5% |
| ML 3-hop | 604 | 8.0% |
| BF 2-hop | 236 | 3.1% |
| ML 2-hop | 83 | 1.1% |

**BF is best for 90.8% of trades.**

### Head-to-head quality (contested trades only)

| Matchup | BF Wins | ML Wins | Ties |
|---------|---:|---:|---:|
| BF 2-hop vs ML 2-hop | 11.4% | 2.8% | 85.9% |
| BF 3-hop vs ML 3-hop | 84.9% | 0.9% | 14.2% |
| BF 5-hop vs ML 3-hop | 85.0% | 0.9% | 14.2% |

### Solve times (isolated, 1K trades)

| Algorithm | p50 | p95 | p99 |
|-----------|----:|----:|----:|
| ML 2-hop | 0ms | 1ms | 2ms |
| BF 2-hop | 4ms | 6ms | 16ms |
| BF 3-hop | 4ms | 7ms | 26ms |
| BF 5-hop | 5ms | 8ms | 37ms |
| ML 3-hop | 27ms | 391ms | 432ms |

BF 5-hop (5ms p50) finishes before ML 3-hop (27ms p50).

### Gas-aware relaxation (1K trades, v2+v3+v4, 6 native protocols)

| Pool | Wins | % |
|------|------|---|
| BF gas-aware (5-hop) | 470 | 61.0% |
| ML (1-3 hop) | 243 | 31.6% |
| BF gross (5-hop) | 57 | 7.4% |

Gas-aware BF wins 96% of head-to-head vs gross BF. ML's remaining 8 real wins (out of 18 apparent) all use `uniswap_v4` routes; the rest is simulation noise from block state drift between sequential runs.

### Key takeaways

1. BF covers 59% more trades (architectural advantage: no edge weight dependency)
2. At equal depth (3 vs 3), BF wins 85% of contested trades
3. BF 5-hop is 5x faster than ML 3-hop
4. Gas-aware relaxation adds another 61% vs 7.4% win rate over gross BF
5. Both should run concurrently; ML still wins ~8% of trades (mostly v4 routes)

---

*This comment consolidates three earlier benchmark comments (Mar 10, Mar 13, Mar 16). The earlier comments are now minimized.*

docs/bellman-ford-explainer.md
```diff
@@ -0,0 +1,397 @@
...
+
+This route looks profitable on paper, but it will fail in practice. Here's why.
+
+### Token revisit implies arbitrage
+
+If the path goes ETH -> USDC -> ETH, the USDC -> ETH leg is getting more ETH back than you started with (otherwise the roundtrip would lose money and the algorithm wouldn't choose it). That means there's a price discrepancy between pool A and pool B for the ETH/USDC pair. This is an **arbitrage opportunity**.
+
+Arbitrageurs monitor these discrepancies with specialized infrastructure. They will execute the ETH -> USDC -> ETH cycle themselves, pocketing the profit. Their transaction will adjust the pool reserves, eliminating the discrepancy. By the time our transaction executes, the pools will have moved, and the route will no longer produce the expected output.
+
+Building a route that depends on an arbitrage opportunity is building on sand.
```
dianacarvalho1, line 121: not to mention that our current Router does not support these cases either 😅

docs/bellman-ford-explainer.md
```diff
@@ -0,0 +1,397 @@
...
+
+### The fix: check before relaxing
+
+Before each relaxation, we walk the predecessor chain backward from the current node to the source and check two things:
+
+1. **Token check**: does the destination token already appear in the path? If yes, skip this edge.
+2. **Pool check**: has this pool already been used in the path? If yes, skip.
```
dianacarvalho1: why are we filtering out the pools too? 🤔 so we don't need to keep track of the updated state? I understand this argument for 2 token pools, but for pools with more tokens it's not so obvious why we should remove them

docs/bellman-ford-explainer.md
```diff
@@ -0,0 +1,397 @@
...
+
+With these constraints, every path the algorithm considers visits each token at most once and uses each pool at most once. The route is a simple chain through distinct pools, which is exactly what we want for execution.
+
+---
+
+## 3. Subgraph Extraction
+
+Before relaxation even starts, we prune the graph.
+
+The full token graph has ~2,400 tokens and ~10,000 directed edges. For a trade from ETH to USDC with a 3-hop budget, most of these are irrelevant. Tokens that are 4 hops away from ETH can't appear in any 3-hop route.
```
dianacarvalho1: These numbers are for ethereum I suppose

docs/bellman-ford-explainer.md
```diff
@@ -0,0 +1,397 @@
...
+
+Validate that the order is a sell order (exact input amount, find best output). Look up the source and destination tokens in the graph. Acquire a read lock on the shared market data (pool states update every block), snapshot what we need, and release the lock. All subsequent steps are lock-free.
+
+### Step 2: Subgraph extraction
+
+BFS from the source token up to `max_hops` depth. Collect all reachable edges. Build a token map (node index to token metadata) and extract a market data subset containing only the relevant pools.
+
+### Step 3: Initialize
+
+Set `distance[source] = order_amount`. Build an adjacency list from the subgraph edges. Seed the active set with the source node.
```
dianacarvalho1: before in this doc it's been named as `amount` not `distance`

docs/bellman-ford-explainer.md
```diff
@@ -0,0 +1,397 @@
...
+
+In head-to-head benchmarks, gas-aware relaxation wins 96% of contested trades against gross-only relaxation (see Benchmarks). The improvement is most visible on routes where a cheap 2-hop path beats an expensive 4-hop path with marginally higher gross output.
+
+### Forward-only BFS vs. bidirectional
+
+Subgraph extraction uses forward BFS from the source, not bidirectional BFS (forward from source, backward from destination). Bidirectional BFS produces a tighter subgraph (edges must lie on a viable source-to-destination path), but forward-only is simpler and the active set mechanism already avoids processing nodes that don't lead anywhere useful.
+
+---
+
+## 7. Benchmarks
```
dianacarvalho1: Were the benchmark tests run against the updated main? meaning that the ML results have the latest improvements?

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -0,0 +1,1251 @@
+//! Bellman-Ford algorithm with SPFA optimization for simulation-driven routing.
```
dianacarvalho1, line 1: what does SPFA actually mean (the letters)?

fynd-core/src/algorithm/mod.rs
```diff
@@ -40,6 +42,9 @@ pub struct AlgorithmConfig {
     timeout: Duration,
     /// Maximum number of paths to simulate. `None` means no cap.
     pub max_routes: Option<usize>,
+    /// Enable gas-aware relaxation in Bellman-Ford (compares net amounts during relaxation).
+    /// Ignored by other algorithms. Defaults to true.
+    gas_aware_relaxation: bool,
```
dianacarvalho1: the `AlgorithmConfig` should be as general as possible. `relaxation` is a specific term of the Bellman-Ford algorithm. I would prefer if here we could make the attributes as general as possible. Only `gas_aware` would be enough

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -0,0 +1,1251 @@
...
+    pub(crate) fn with_config(config: AlgorithmConfig) -> Result<Self, AlgorithmError> {
+        Ok(Self {
+            max_hops: config.max_hops(),
+            timeout: config.timeout(),
+            gas_aware_relaxation: config.gas_aware_relaxation(),
+        })
+    }
+}
+
+impl Algorithm for BellmanFordAlgorithm {
```
dianacarvalho1, line 312: I would suggest also add an

```
impl BellmanFordAlgorithm {
  fn helper_methods(....){}

}
```
instead of having them defined as standalone functions at the bottom

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -0,0 +1,1251 @@
...
+        }
+
+        let (token_prices, spot_prices) = if let Some(ref derived) = derived {
+            let guard = derived.read().await;
+            (guard.token_prices().cloned(), guard.spot_prices().cloned())
+        } else {
+            (None, None)
+        };
+
+        // Acquire read lock, BFS into adjacency list, snapshot market data, release lock.
```
dianacarvalho1: this whole bit on creating the subgraph I would move to a method of `BellmanFordAlgorithm` called `get_subgraph` instead of having it all here defined in the main method

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -0,0 +1,1251 @@
...
+        // edge_gas[node] = gas estimate for the edge that last improved distance[node].
+        // cumul_gas[node] = total gas units along the best path to this node.
+        let max_idx = graph
+            .node_indices()
+            .map(|n| n.index())
+            .max()
+            .unwrap_or(0) +
+            1;
+
+        let mut distance: Vec<BigUint> = vec![BigUint::ZERO; max_idx];
```
dianacarvalho1: ah here we are naming this `distance`. Tbh I think `amount` is more obvious. Could we make this coherent with the docs please?

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -0,0 +1,1251 @@
...
+
+            if adj.is_empty() {
+                return Err(AlgorithmError::NoPath {
+                    from: order.token_in().clone(),
+                    to: order.token_out().clone(),
+                    reason: NoPathReason::NoGraphPath,
+                });
+            }
+
+            let token_map: HashMap<NodeIndex, Token> = token_nodes
```
dianacarvalho1, line 366: I believe the lock read
```
            let market = market.read().await;
```
is only needed here. no need to do it all the way up there unnecessarily

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -0,0 +1,1251 @@
...
+                edge_gas[to_node.index()].clone(),
+                component.clone(),
+                sim_state.clone_box(),
+            ));
+        }
+
+        let route = Route::new(swaps);
+        let final_amount_out = distance[out_idx].clone();
+
+        let gas_price = market_subset
```
dianacarvalho1: is there a reason why we are fetching the gas price again? We already have it in `gas_price_wei`

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -0,0 +1,1251 @@
...
+                            v_price.as_ref(),
+                        );
+                        net_candidate > net_existing
+                    } else {
+                        result.amount > distance[v_idx]
+                    };
+
+                    if is_better {
+                        // Update spot product for the new path to v
+                        spot_product[v_idx] = compute_edge_spot_product(
```
dianacarvalho1: we are calling `compute_edge_spot_product` with the same arguments twice unnecessarily (see line 260)

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -0,0 +1,1251 @@
...
+    let Some(last_swap) = route.swaps().last() else {
+        return BigInt::from(amount_out.clone());
+    };
+
+    let total_gas = route.total_gas();
+
+    if gas_price.is_zero() {
+        warn!("missing gas price, returning gross amount_out");
+        return BigInt::from(amount_out.clone());
+    }
```
dianacarvalho1: In this case, if `TokenGasPrices` is missing we are just warning and returning the gross amount with no spot price fallback. While during relaxation we are using cumulative spot price product along the path if `TokenGasPrices` is missing. Can we use it here too please?

dianacarvalho1 (commented):
Thank you @Haikane ! 🙏🏼 The doc makes the algorithm extra clear (reading english is easier than code to understand an idea 😅) I left a few comments for a preliminary review (I haven't looked at the tests though). Also CI is still failing

docs/bellman-ford-explainer.md
dianacarvalho1, line 1: I'm not sure this is the best place for this doc but don't really know where I would put it tbh. Does someone have any ideas?

docs/bellman-ford-explainer.md
```diff
@@ -0,0 +1,397 @@
...
+
+Note that the algorithm also explored A -> B -> D (6,000) and A -> C -> B -> D (7,500) but found A -> C -> D (20,000) to be the best.
+
+---
+
+## 6. Design Tradeoffs
+
+### Simulation during relaxation vs. precomputed weights
+
+The existing MostLiquid algorithm precomputes heuristic edge weights (spot prices and pool depths) and uses them to score paths. This is faster per edge but can fail: if the heuristic can't compute a weight for a pool (e.g., unusual token decimals, exotic pool types), that edge is invisible. The route that would have used it is never found.
```
kayibal: The reasons listed here for a weight calculation failing are wrong. It's most likely that the pool has very low liquidity - or it has a bug. If you use it anyway it most likely means a bad outcome for whoever is swapping.

The following conclusions to this paragraph should be adjusted.

docs/bellman-ford-explainer.md
tamaralipows: Good point, we could use a pool more than once without repeating a path!

docs/bellman-ford-explainer.md
tamaralipows: My question too, there should be a big difference since disallowing invalid cycles in the updated main

fynd-core/src/algorithm/mod.rs
tamaralipows: For me this raises the question of - what if future algorithms need some new config attr? Should we add some `algorithm_specific_config` dict here to generalize that? Wouldn't be too nice to keep putting specific things in this general config...

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -0,0 +1,1251 @@
...
+        return 0.0;
+    };
+    let key = (component_id.clone(), u.clone(), v.clone());
+    match prices.get(&key) {
+        Some(&spot) if spot > 0.0 => parent_spot * spot,
+        _ => 0.0,
+    }
+}
+
+/// Resolves the WETH-to-token conversion rate for gas cost calculation.
```
tamaralipows: This docstring mentions WETH but it seems this fn is more general, referring to `v_addr`?

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -0,0 +1,1251 @@
...
+            "bellman_ford route found"
+        );
+
+        Ok(result)
+    }
+
+    fn computation_requirements(&self) -> ComputationRequirements {
+        ComputationRequirements::none()
+            .allow_stale("token_prices")
+            .expect("Conflicting Computation Requirements")
```
tamaralipows: Shouldn't we be using proper error handling instead of `expect`s?

tamaralipows (commented):
Thanks!

docs/bellman-ford-explainer.md
Haikane: Will rerun, but much easier once we have it on main and we have the pr on compare tool also merged.

docs/bellman-ford-explainer.md
Haikane: As soon as this PR is merged. I will move this later into a separate page in the docs on algorithms, and also add 1. and overview of the routing problem 2. different approaches and limitations 3. an explainer on the most liquid algorithm and this explainer. Didn't want to overload the PR.

docs/bellman-ford-explainer.md
kayibal: We could, but most likely it won't result in good results and it would make the algorithm more complex.

fynd-core/src/algorithm/bellman_ford.rs
kayibal: Would probably call this `best_amount_at_token` or something more obvious

docs/bellman-ford-explainer.md
Haikane: True! Both are sell-only for now. Added a note in the doc.

docs/bellman-ford-explainer.md
Haikane: Good question. For multi-token pools, reusing the same pool on a different pair is theoretically possible but would add complexity to the revisit tracking (you'd need to track per-pool-per-pair usage instead of per-pool). As @kayibal noted, it's unlikely to produce better results in practice and would make the algorithm harder to reason about. Added a note in the doc explaining the rationale for multi-token pools.

docs/bellman-ford-explainer.md
Haikane: Yes, added "(on Ethereum)" qualifier in f2c6174.

docs/bellman-ford-explainer.md
Haikane: Fixed. Renamed `distance` to `amount` in both code and doc for consistency (f2c6174).

fynd-core/src/algorithm/bellman_ford.rs
Haikane: Shortest Path Faster Algorithm. Expanded the acronym in the module docstring (f2c6174).

fynd-core/src/algorithm/mod.rs
Haikane: Agreed. Renamed to `gas_aware` throughout (f2c6174). For future algorithm-specific config, we can revisit if/when it's needed.

fynd-core/src/algorithm/bellman_ford.rs
Haikane: Done. Moved all standalone helper functions into `impl BellmanFordAlgorithm` as associated functions (f2c6174).

fynd-core/src/algorithm/bellman_ford.rs
Haikane: Done. Extracted BFS + adjacency list + market subset snapshot into `fn get_subgraph()` (f2c6174).

fynd-core/src/algorithm/bellman_ford.rs
Haikane: Renamed to `amount` to match the doc and be more self-explanatory (f2c6174).

fynd-core/src/algorithm/bellman_ford.rs
Haikane: Good catch. The lock was already scoped to the BFS block (released at line 151), but the gas price was fetched again later unnecessarily. Fixed by reusing the snapshotted `gas_price_wei` (f2c6174).

fynd-core/src/algorithm/bellman_ford.rs
Haikane: Removed the redundant fetch. Now reuses `gas_price_wei` from the snapshot for `compute_net_amount_out` (f2c6174).

fynd-core/src/algorithm/bellman_ford.rs
Haikane: Fixed. Now computes the spot product once before the comparison and reuses it in the update (f2c6174).

fynd-core/src/algorithm/bellman_ford.rs
Haikane: Good point. `compute_net_amount_out` now uses `resolve_token_price` (with spot product fallback) instead of only checking `TokenGasPrices` directly (f2c6174).

docs/bellman-ford-explainer.md
Haikane: You're right. Rewrote the paragraph: a weight calculation most likely fails due to very low liquidity or a pool bug.

fynd-core/src/algorithm/bellman_ford.rs
Haikane: Fixed. Docstring now says "gas-to-token conversion rate" instead of "WETH-to-token" (f2c6174).

fynd-core/src/algorithm/bellman_ford.rs
Haikane: Improved the expect messages to be descriptive. These are in `computation_requirements()` where the string literals are compile-time constants that can't actually conflict, so a panic is appropriate (programmer error, not runtime). Made the messages clearer about what went wrong (f2c6174).

fynd-core/src/algorithm/mod.rs
Haikane: Good thought. For now, `gas_aware` is general enough that other algorithms could use it too (any algo could benefit from gas-aware scoring). If we get truly algorithm-specific config later, we can add an `extra: HashMap<String, Value>` or typed enum. Don't want to over-engineer it yet.

docs/bellman-ford-explainer.md
Haikane: And no cycles

docs/bellman-ford-explainer.md
Haikane: Removed the entire 'Simulation during relaxation vs. precomputed weights' section. The comparison was confused and the reasoning about why weight calculations fail was wrong as you pointed out. The remaining Design Tradeoffs section covers gas-aware vs. gross relaxation and forward-only BFS, which are the actually useful design decisions to document.

docs/bellman-ford-explainer.md
dianacarvalho1: I believe the compare tool PR has been merged a long time ago

tamaralipows (approved):
🙏

PR #44, opened 2026-02-23
refactor(pool-depth): replace binary search fallback with generic Brent solver

pistomat:
## Summary

- Replace the custom binary search fallback (`find_depth_binary_search`) with the generic Brent solver (`query_pool_swap` from `tycho_simulation::evm`) when a protocol's native `query_pool_swap` is missing or doesn't support `TradeLimitPrice`
- Add `query_pool_swap` to `MockProtocolSim` for test coverage
- Remove ~100 lines of binary search implementation and its dedicated tests

## Motivation

The binary search measured price impact via post-swap *spot price* changes, which is a different metric than the execution price (`amount_out / amount_in`) used by the Brent solver. The Brent solver is both more general (works on any `ProtocolSim`) and directly optimizes the execution price constraint that `TradeLimitPrice` specifies.

Beyond correctness, the Brent solver is significantly faster on EVM-simulated pools where each `get_amount_out` call requires VM execution. Brent converges in far fewer simulation calls than binary search over the full integer range. A comparison benchmark on `ConstantProductSim` is available on the reference branch [`mp/brent-comparison-tests`](https://github.com/propeller-heads/fynd/tree/mp/brent-comparison-tests) (run with `cargo test -p fynd -- bench_brent_vs_binary_search --ignored --nocapture`).

## Changes

| File | What changed |
|---|---|
| `src/derived/computations/pool_depth.rs` | Removed `find_depth_binary_search` method and its tests; fallback now calls `query_pool_swap` from `tycho_simulation::evm` |
| `src/algorithm/test_utils.rs` | Added `query_pool_swap` impl to `MockProtocolSim` (supports `TradeLimitPrice`) |

## Test plan

- [x] Existing `pool_depth` tests pass (`cargo nextest run -p fynd -- pool_depth`)
- [x] Verify depth values are reasonable on staging/testnet data
- [x] Run comparison benchmark on [`mp/brent-comparison-tests`](https://github.com/propeller-heads/fynd/tree/mp/brent-comparison-tests) to confirm Brent performance

### DIFF
--- a/src/derived/computations/pool_depth.rs
+++ b/src/derived/computations/pool_depth.rs
@@ -1,7 +1,6 @@
 //! Pool depth computation.
 //!
-//! Computes liquidity depths for all pools using `query_pool_swap` when available,
-//! falling back to binary search with `get_amount_out`.
+//! Computes liquidity depths for all pools using `query_pool_swap`.
 //! Depth represents the maximum input amount before reaching the configured slippage
 //! threshold from the spot price.
 //!
@@ -16,13 +15,10 @@ use std::collections::HashSet;
 use async_trait::async_trait;
 use itertools::Itertools;
 use num_bigint::BigUint;
-use num_traits::{One, Zero};
 use tracing::{debug, instrument, warn, Span};
 use tycho_simulation::{
-    tycho_common::{
-        models::token::Token,
-        simulation::{errors::SimulationError, protocol_sim::ProtocolSim},
-    },
+    evm::query_pool_swap::query_pool_swap,
+    tycho_common::simulation::errors::SimulationError,
     tycho_core::simulation::protocol_sim::{Price, QueryPoolSwapParams, SwapConstraint},
 };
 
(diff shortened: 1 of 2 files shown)


src/derived/computations/pool_depth.rs
```diff
@@ -318,24 +214,29 @@ impl DerivedComputation for PoolDepthComputation {
...
                         if msg == "query_pool_swap not implemented" =>
                     {
-                        self.find_depth_binary_search(sim_state, token_in, token_out, component_id)
+                        query_pool_swap(sim_state, &params)
                     }
                     Err(SimulationError::InvalidInput(msg, _))
                         if msg.contains("does not support TradeLimitPrice") =>
                     {
-                        self.find_depth_binary_search(sim_state, token_in, token_out, component_id)
+                        query_pool_swap(sim_state, &params)
```
kayibal, line 229: Should we at least have a warning here? Afaik it means we didn't properly implement the protocol, right?

kayibal (approved):

src/derived/computations/pool_depth.rs
pistomat: Yes, it means that we have not implemented the protocol.

Imho you should consider removing the default `Err(SimulationError::FatalError("query_pool_swap not implemented".into()))` from `ProtocolSim::query_pool_swap` and enforce implementation (Brent is a good default when analytical solution is not possible).

PR #50, opened 2026-02-25
DO NOT MERGE: Price guard sketch

tamaralipows:
Added examples providers:
- Binance
- Hyperliquid
- Chainlink

All providers are just roughly implemented and are not the point of this sketch. True implementation of each provider will need to be carefully double-checked. This sketch PR is rather to see how everything would fit into place. Please rather focus on the base traits than the actual provider implementations.


fynd-rpc/src/builder.rs
```diff
@@ -166,13 +181,50 @@ impl FyndBuilder {
...
+            .push(Box::new(ChainlinkProvider::new(self.rpc_url.clone())));
+        self
+    }
+
+    /// Registers a custom price provider for the price guard.
+    ///
+    /// The provider's [`start`](PriceProvider::start) method is called during
+    /// [`build`](Self::build) with the shared market data. A solution passes if
+    /// **at least one** registered provider validates within the BPS tolerance.
+    pub fn register_price_provider(mut self, provider: Box<dyn PriceProvider>) -> Self {
```
tamaralipows, line 205: Decided it might be useful if the user would like to use their own price provider instead of our defaults. Do you agree?

fynd-core/src/order_manager/mod.rs
```diff
@@ -123,11 +130,30 @@ impl OrderManager {
...
+            match &self.encoder {
+                Some(encoder) => {
+                    // Only encode solutions that are ready for on-chain execution
+                    let (to_encode, rest): (Vec<_>, Vec<_>) = order_solutions
+                        .into_iter()
+                        .partition(|s| s.status == SolutionStatus::Success);
+                    let mut encoded = encoder
+                        .encode(to_encode, request.options.slippage)
+                        .await?;
+                    encoded.extend(rest);
```
dianacarvalho1: this might be changing the order of the solutions and I think we didn't want this 😬 The solutions' order should be the same as the original orders' order

fynd-core/src/price_guard/binance_ws.rs
```diff
@@ -0,0 +1,646 @@
...
+
+/// Background task that connects to Binance WebSocket and populates the price cache.
+///
+/// On each connection, fetches the Binance exchange info to discover which trading
+/// pairs exist, then cross-references with [`SharedMarketData`] to subscribe only to
+/// relevant bookTicker streams. Periodically re-checks for new tokens and subscribes
+/// to additional streams on the live connection.
+struct BinanceWsWorker {
+    cache: PriceCache,
+    market_data: Arc<RwLock<SharedMarketData>>,
```
dianacarvalho1: do we need the full market_data? couldn't we just pass all the tokens?

dianacarvalho1: I also don't see any extra logic to add new tokens as they come up - which I think it's fine for now. but then we can simplify this

fynd-core/src/price_guard/common.rs
```diff
@@ -0,0 +1,124 @@
...
+///
+/// `price` is in human terms: how many units of `token_out` per 1 unit of `token_in`.
+/// `amount_in` is in raw units (e.g. wei). Returns raw units of `token_out`.
+pub fn compute_expected_out(
+    amount_in: &BigUint,
+    price: f64,
+    decimals_in: u32,
+    decimals_out: u32,
+) -> BigUint {
+    const SCALE: f64 = 1_000_000_000_000.0; // 10^12
```
dianacarvalho1, line 60: why only 10^12?

fynd-core/src/price_guard/mod.rs
```diff
@@ -0,0 +1,578 @@
...
+    ) -> Result<Vec<OrderSolution>, SolveError> {
+        if !self.config.enabled() {
+            return Ok(solutions);
+        }
+
+        let mut validated = Vec::with_capacity(solutions.len());
+
+        for mut solution in solutions {
+            if solution.status != SolutionStatus::Success {
+                validated.push(solution);
```
dianacarvalho1, line 70: we are changing the order here too

tamaralipows: Triple checked but we actually aren't changing the order here. It's pushed every single time, just in different branches of the if. Obv we should make this cleaner and that fact more obvious

fynd-core/src/order_manager/mod.rs
tamaralipows: @dianacarvalho1 I tried to do this while preserving the interface of the encoder and it looks quite messy.
I've decided instead of have the encoder take one solution at a time. How do you feel about this?

This way the encoder doesn't need to know anything about the SolutionStatus, and we don't need to do any awkward indexing here to preserve order.

See recent commits

dianacarvalho1: I think it should be equivalent for the encoder to take one solution at a time. At least while we don't support batching orders anyway

PR #58, opened 2026-02-27
feat: implement fynd-client Rust crate (ENG-5593)

kayibal:
## Summary

Adds `clients/rust/` — a new `fynd-client` crate providing a typed Rust interface to the Fynd DEX router RPC server.

- `FyndClientBuilder` / `FyndClient`: builder pattern with configurable base URL, router address, chain ID, RPC URL, HTTP timeout, retry policy, and optional submit URL / default sender
- `quote()`: sends orders to `/v1/solve`, deserializes responses into typed `Quote` / `OrderSolution` / `Route` / `Swap` values, and retries on transient errors with exponential backoff
- `health()`: queries `/v1/health` and returns a `HealthStatus`
- `signable_payload()`: builds an EIP-1559 `TypedTransaction` for a given `OrderSolution`, auto-resolving sender nonce and fees from the RPC provider; supports optional `eth_call` simulation via `SigningHints`
- `execute()`: broadcasts a `SignedOrder` as a raw transaction and returns an `ExecutionReceipt` future that resolves once the receipt appears on-chain, decoding ERC-20 and ERC-6909 Transfer logs to compute `settled_amount`

## Key design decisions

- No `fynd-rpc` dependency: private wire-format mirror structs in `mapping.rs` (not `pub mod`) reproduce exactly what the server serializes/deserializes, keeping the client free of server-side infrastructure (actix-web, etc.)
- `BackendKind::Fynd` is hardcoded on all solutions returned by the Fynd RPC (Turbine solutions are not yet routed through this client)
- `component_id` (server wire name) is renamed to `pool_id` in the public `Swap` type
- `amount_out_net_gas` is consumed during deserialization via `#[serde(rename = "amount_out_net_gas")]` on `_amount_out_net_gas` and not exposed; it is an internal server ranking field
- `TurbinePayload` is a stub struct with `#[allow(dead_code)]` — Turbine signing will land in a follow-up
- `ExecutionReceipt` implements `Future<Output = Result<SettledOrder, FyndError>>` directly, so callers can `.await` it or poll it manually

## Out of scope

- TypeScript client
- `LocalFyndInstance` (local dev server wrapper)
- Turbine `execute()` path — returns `FyndError::Protocol` if called with a Turbine payload

## Fixes applied during review

- Added `#[serde(rename = "amount_out_net_gas")]` to the `_amount_out_net_gas` field in `WireOrderSolution` — without this serde would look for `"_amount_out_net_gas"` in the JSON and fail every deserialization
- Removed the unused `futures` crate from `clients/rust/Cargo.toml`


Haikane:
## Extreme Code Review: PR #58

**Agents:** Claude, Codex (o3), Gemini
**Date:** 2026-03-02
**Jira:** ENG-5593 (Story 2 - Build the Rust Client)

### Summary

Substantial, well-structured new crate (~3500 lines) with good test coverage (35+ tests). Architecture is sound: clean separation of public types from wire format, proper `From`/`TryFrom` traits, `fynd-rpc-types` extraction eliminates mirror structs. One blocker and several important issues to address before merge.

### Blocker (Must Fix)

| # | Finding | Agents |
|---|---------|--------|
| 1 | **`clients/rust/src/client.rs:725-743` - `execute()` polls indefinitely for tx receipt.** The inner future loops forever with `tokio::time::sleep(2s)`. If a transaction is dropped from the mempool (nonce reused, gas too low, reorg), this future never resolves. Add a configurable max poll duration or max attempts, returning a timeout error. | Claude, Codex, Gemini |

### Important (Should Fix)

| # | Finding | Agents |
|---|---------|--------|
| 2 | **`clients/rust/src/signing.rs` - Turbine variant methods panic.** `signing_hash()` and `quote()` on `SignablePayload::Turbine` call `unimplemented!()`. Library crates should never panic on expected code paths. Change to return `Result::Err(FyndError::Protocol("Turbine signing not yet implemented"))`. Note: `into_fynd_parts()` already does this correctly. | Claude, Codex, Gemini |
| 3 | **`clients/rust/src/lib.rs:~line 1479` - Stale doc example.** The doc example shows `QuoteParams::new(vec![order], ...)` but the API now takes a single `Order` after the batch simplification in the last commit. This will fail `cargo doc --test`. | Claude |
| 4 | **`clients/rust/src/client.rs:668` - Empty calldata.** The EIP-1559 transaction has `input: AlloyBytes::new()` (no calldata). Submitting this to a router contract will revert on-chain. Add validation that rejects empty calldata, or document this limitation prominently so callers don't sign and broadcast a guaranteed-to-fail tx. | Claude |
| 5 | **`.github/workflows/ci.yaml` - CI actions not pinned to SHA hashes.** Pin `actions/checkout@v4`, `dtolnay/rust-toolchain@v1`, `Swatinem/rust-cache@v2` to commit SHAs for supply chain security. | Claude, Codex |
| 6 | **`clients/rust/Cargo.toml` - `thiserror = "1.0.69"` pins exact patch.** Use `thiserror = { workspace = true }` instead to avoid duplicate versions in the dependency tree. | Claude |

### Suggestions

| # | Finding | Agents |
|---|---------|--------|
| 7 | **`SigningHints` uses public fields** while all other public structs use private fields + getters. Consider making fields private with a builder for API consistency. | Claude, Gemini |
| 8 | **`clients/rust/src/client.rs:498-499` - Duplicate doc comment** on `new_with_providers` (same paragraph appears twice). | Claude |
| 9 | **`clients/rust/src/types.rs` - `BatchQuote` dead code.** `total_gas_estimate` and `solve_time_ms` have `#[allow(dead_code)]` but are never read after the single-order simplification. Consider removing them. | Claude |
| 10 | **`fynd-rpc-types/Cargo.toml`** - Add `publish = false` to prevent accidental crate publication. | Claude |
| 11 | **Additional test coverage** - Consider adding tests for: receipt polling timeout (once implemented), address checksum rejection, max-retries-reached edge case. | Codex |

### Resolved Concerns

- **ERC-6909 topic ordering** (raised by Codex): Verified correct against the EIP-6909 spec. Event is `Transfer(address caller, address indexed from, address indexed to, uint256 indexed id, uint256 amount)`, giving topics = [sig, from, to, id]. Code reads `topics[2]` for `to` and `data[32..64]` for amount, both correct.
- **`is_retryable()` for Http errors** (raised by Gemini): Not an issue. `FyndError::Http` only wraps transport-level `reqwest::Error` (DNS, connection reset, timeout). HTTP 4xx/5xx status codes are parsed into `FyndError::Api` before reaching this path.

### Observations

- The `fynd-rpc-types` extraction is well-executed, giving compile-time sync between server and client.
- Test suite is solid: wiremock for HTTP mocking, alloy mock transport for RPC, covering quote/health/retry/signable_payload/settlement.
- `compute_settled_amount` correctly handles both ERC-20 and ERC-6909 Transfer logs with proper topic/data decoding.
- The rename from `OrderSolution` to `Quote` improves clarity from the caller's perspective.
- Good use of `bytes::Bytes` and `BigUint` for chain-agnostic public API.

---
*Reviewed by Claude + Codex + Gemini*

kayibal:
I disagree with the blocker mentioned here. I think it is preferable if the users wraps this in their own timeout. Although it should be clearly documented, this way, they can set the timeout (mechanism) that they prefer.

kayibal:
`thiserror`, already uses `workspace=true`

https://github.com/propeller-heads/fynd/pull/58/changes#diff-1a73a3d3f9a8a57e761780a69f5b3522ce2f78d09ad12b87b45b225ecca3a29aR37

clients/rust/src/mapping.rs
```diff
@@ -0,0 +1,389 @@
...
+) -> Result<alloy::primitives::Address, FyndError> {
+    if b.len() != 20 {
+        return Err(FyndError::Protocol(format!("expected 20-byte address, got {} bytes", b.len())));
+    }
+    let arr: [u8; 20] = b
+        .as_ref()
+        .try_into()
+        .expect("length checked above");
+    Ok(alloy::primitives::Address::from(arr))
+}
```
tamaralipows, line 29: Wouldn't something like 
```
pub(crate) fn bytes_to_alloy_address(
    b: &bytes::Bytes,
) -> Result<alloy::primitives::Address, FyndError> {
    let arr: [u8; 20] = b
        .as_ref()
        .try_into()
        .map_err(|_| {
            FyndError::Protocol(format!(
                "expected 20-byte address, got {} bytes",
                b.len()
            ))
        })?;

    Ok(alloy::primitives::Address::from(arr))
```
be more idiomatic?

clients/rust/src/signing.rs
```diff
@@ -0,0 +1,515 @@
...
+    quote: Quote,
+    tx: TypedTransaction,
+}
+
+impl FyndPayload {
+    pub(crate) fn new(quote: Quote, tx: TypedTransaction) -> Self {
+        Self { quote, tx }
+    }
+
+    /// The order solution this payload was built from.
```
tamaralipows, line 31: Are we using the words order solution and quote interchangeably? Can we choose one for consistency?

clients/rust/tests/integration.rs
```diff
@@ -0,0 +1,492 @@
...
+// ============================================================================
+
+/// Build a [`FyndClient<RootProvider<Ethereum>>`] pointing at `base_url` backed by an
+/// alloy mock transport, so that Ethereum RPC calls are handled in-process.
+fn make_client(
+    base_url: String,
+    retry: RetryConfig,
+    default_sender: Option<Address>,
+) -> (FyndClient<RootProvider<Ethereum>>, alloy::providers::mock::Asserter) {
+    use alloy::providers::mock::Asserter;
```
tamaralipows, line 38: This should be imported at the top?

clients/rust/src/types.rs
```diff
@@ -0,0 +1,530 @@
...
+    pub fn receiver(&self) -> Option<&Bytes> {
+        self.receiver.as_ref()
+    }
+}
+
+/// Optional parameters that tune solving behaviour for a [`QuoteParams`] request.
+///
+/// Build via the builder methods; unset options use server defaults.
+#[derive(Default)]
+pub struct QuoteOptions {
```
dianacarvalho1, line 90: not in the scope of this task but the endpoint is called "solve". Should we rename it to "quote"? I vote yes

clients/rust/src/types.rs
```diff
@@ -0,0 +1,530 @@
...
+    /// Solutions for each order, in the same order as the request.
+    pub fn quotes(&self) -> &[Quote] {
+        &self.quotes
+    }
+
+    /// Aggregate estimated gas units for executing all solutions.
+    #[allow(dead_code)]
+    pub fn total_gas_estimate(&self) -> &BigUint {
+        &self.total_gas_estimate
+    }
```
dianacarvalho1: do we want to keep them? or are we planning to implement them at a future PR?

clients/rust/src/mapping.rs
```diff
@@ -0,0 +1,389 @@
...
+    types::{
+        BackendKind, BatchQuote, BlockInfo, HealthStatus, Order, OrderSide, Quote, QuoteOptions,
+        QuoteParams, Route, SolutionStatus, Swap,
+    },
+};
+// ============================================================================
+// ADDRESS CONVERSION HELPERS
+// ============================================================================
+
+pub(crate) fn bytes_to_alloy_address(
```
dianacarvalho1, line 21: you can use `tycho_ethereum::BytesCodec ` instead of redefining this

clients/rust/src/mapping.rs
```diff
@@ -0,0 +1,389 @@
...
+}
+
+/// Convert a tycho wire-format address to a client `bytes::Bytes` address.
+fn tycho_to_bytes(addr: TychoAddress) -> bytes::Bytes {
+    // hex_bytes::Bytes has From<Bytes> -> bytes::Bytes, and inner field .0 is pub
+    addr.0
+}
+
+// ============================================================================
+// CLIENT TYPES → WIRE FORMAT
```
dianacarvalho1: what does "wire" mean in this case? I understand we are referring to the fynd-rpc-types but why the word wire? (just trying to understand)

clients/rust/src/client.rs
```diff
@@ -0,0 +1,1309 @@
...
+/// use bytes::Bytes;
+///
+/// let mut overrides = StorageOverrides::default();
+/// let contract = Bytes::copy_from_slice(&[0xAA; 20]);
+/// let slot    = Bytes::copy_from_slice(&[0x00; 32]);
+/// let value   = Bytes::copy_from_slice(&[0x01; 32]);
+/// overrides.insert(contract, slot, value);
+/// ```
+#[derive(Clone, Default)]
+pub struct StorageOverrides {
```
dianacarvalho1, line 194: pretty sure we already have this defined somewhere else

clients/rust/src/client.rs
```diff
@@ -0,0 +1,1309 @@
...
+        let state_diff = slot_map
+            .iter()
+            .map(|(slot, val)| Ok((bytes_to_b256(slot)?, bytes_to_b256(val)?)))
+            .collect::<Result<alloy::primitives::map::B256HashMap<B256>, FyndError>>()?;
+        result.insert(addr, AccountOverride { state_diff: Some(state_diff), ..Default::default() });
+    }
+    Ok(result)
+}
+
+fn bytes_to_b256(b: &Bytes) -> Result<B256, FyndError> {
```
dianacarvalho1, line 226: again you can use `tycho_ethereum::BytesCodec `

clients/rust/src/client.rs
```diff
@@ -0,0 +1,1309 @@
...
+/// `build` performs two network calls: one to validate the RPC URL (fetching `chain_id`) and one
+/// to construct the HTTP provider. It does **not** connect to the Fynd API.
+pub struct FyndClientBuilder {
+    base_url: String,
+    timeout: Duration,
+    retry: RetryConfig,
+    /// Optional router contract address. Defaults to `Address::ZERO` if not set.
+    ///
+    /// TODO: Replace with the actual RouterV3 contract address once deployed.
+    router_address: Option<Address>,
```
dianacarvalho1, line 272: encoding gives you the address directly, I don't think we need to have this here

clients/rust/src/client.rs
```diff
@@ -0,0 +1,1309 @@
...
+                    tokio::time::sleep(delay).await;
+                    delay = (delay * 2).min(self.retry.max_backoff);
+                }
+                Err(e) => return Err(e),
+            }
+        }
+        Err(FyndError::Protocol("retry loop exhausted without result".into()))
+    }
+
+    async fn do_quote(
```
dianacarvalho1: maybe `request_quote`?

clients/rust/src/client.rs
```diff
@@ -0,0 +1,1309 @@
...
+            max_priority_fee_per_gas,
+            gas_limit,
+            to: TxKind::Call(self.router_address),
+            value: U256::ZERO,
+            input: AlloyBytes::new(),
+            access_list: AccessList::default(),
+        };
+
+        // Optionally simulate the transaction.
+        if hints.simulate {
```
dianacarvalho1: in simulation we assume that the sender has enough funds and approvals right?

dianacarvalho1 (commented):
Thank you @kayibal ! 🙏🏼  Overall it looks good to me! Nicely built public structs 👏🏼 (I didn't review the tests carefully though)
Added some comments and I have a few more:
1. Why would we want Turbine to be a possible backend for a **Fynd** client?
2. 
> component_id (server wire name) is renamed to pool_id in the public Swap type

  why? Overall there are a few changes in naming in the new types vs the rpc/core types (solve vs quote for example). Can we be consistent please? 🥺  
3. Can you add a readme or an example please? 
4. Did you try to run the whooooleeee thing? 👀

clients/rust/src/client.rs
```diff
@@ -0,0 +1,1309 @@
...
+            .map_err(|e| FyndError::Config(format!("invalid submit URL: {e}")))?;
+        let submit_provider = ProviderBuilder::default().connect_http(submit_url);
+
+        // Fetch chain_id from the RPC node.
+        let chain_id = provider
+            .get_chain_id()
+            .await
+            .map_err(|e| FyndError::Config(format!("failed to fetch chain_id from RPC: {e}")))?;
+
+        // TODO: Replace with the actual RouterV3 contract address once deployed.
```
tamaralipows, line 414: So this will be hardcoded? Or you mean it will be the default here instead of Address::ZERO?

clients/rust/src/client.rs
```diff
@@ -0,0 +1,1309 @@
...
+        self
+    }
+
+    /// Use a separate RPC URL for transaction submission and receipt polling.
+    ///
+    /// If not set, the `rpc_url` passed to [`new`](Self::new) is used for both.
+    pub fn with_submit_url(mut self, url: impl Into<String>) -> Self {
+        self.submit_url = Some(url.into());
+        self
+    }
```
tamaralipows, line 314: I'm curious for what reasons someone might want this to be different

clients/rust/src/client.rs
```diff
@@ -0,0 +1,1309 @@
...
+
+        let tx_eip1559 = TxEip1559 {
+            chain_id: self.chain_id,
+            nonce,
+            max_fee_per_gas,
+            max_priority_fee_per_gas,
+            gas_limit,
+            to: TxKind::Call(self.router_address),
+            value: U256::ZERO,
+            input: AlloyBytes::new(),
```
tamaralipows, line 651: I assume we will be using real calldata here - let's add a TODO?

clients/rust/src/client.rs
```diff
@@ -0,0 +1,1309 @@
...
+// ============================================================================
+// SIGNING HINTS
+// ============================================================================
+
+/// Optional hints to override auto-resolved transaction parameters.
+///
+/// All fields default to `None` / `false`. Unset fields are resolved automatically from the
+/// RPC node during [`FyndClient::signable_payload`].
+#[derive(Default)]
+pub struct SigningHints {
```
tamaralipows, line 92: Are we okay with some pub structs having pub fields without getters? Are we positive this won't change?

clients/rust/src/client.rs
```diff
@@ -0,0 +1,1309 @@
...
+        Self {
+            max_attempts: 3,
+            initial_backoff: Duration::from_millis(100),
+            max_backoff: Duration::from_secs(2),
+        }
+    }
+}
+
+// ============================================================================
+// SIGNING HINTS
```
tamaralipows, line 82: What is the reasoning behind calling them hints rather than overrides/params like we do in our other codebases?

clients/rust/src/client.rs
```diff
@@ -0,0 +1,1309 @@
...
+    /// For [`BackendKind::Fynd`] solutions, this resolves the sender nonce and EIP-1559 fee
+    /// parameters from the RPC node (unless overridden via `hints`), then constructs an
+    /// unsigned EIP-1559 transaction targeting the RouterV3 contract.
+    ///
+    /// [`BackendKind::Turbine`] is not yet implemented and returns
+    /// [`FyndError::Protocol`].
+    ///
+    /// `token_out` and `receiver` are read directly from the `solution` (populated during
+    /// `quote()`). Pass `&SigningHints::default()` to auto-resolve all transaction parameters.
+    pub async fn signable_payload(
```
tamaralipows, line 582: I could use some clarity on the flow here. The client has three public endpoints - in what order are they expected to be called? The user gets a `Quote`, calls `signable_payload` with it in order to get the `SignablePayload` object, and then what? Needs to combine it into a `SignedOrder` object by themselves in order to call the `execute` method?

tamaralipows (commented):
Thanks @kayibal, I mostly have some questions for now

clients/rust/src/types.rs
kayibal: Yes I think we should, I think more people understand `quote`, `solve` is too general.

clients/rust/src/types.rs
kayibal: Hm yeah it depends if we want to support, batch quotes. We currently do but from a client perspective it complicates things, especially the flow since the execution does not support batches atm.

In general the client tries to be as close to the server as possible. I guess we are keeping batch quoting for now on the server. I don't have a strong opinion, probably more flexibility if we remove them for now?

clients/rust/src/mapping.rs
kayibal: Is it worth the full `tycho-ethereum` dependency?

clients/rust/src/mapping.rs
kayibal: Wire, means network connection. Yes this is new wording from Claude. I don't mind it it's quite clear for me. We usually use the term `dto`.

clients/rust/src/client.rs
kayibal: Yes probably `tycho-common` or `tycho-ethereum` but again, it may involve an additional dependency for a single type. Most likely it is not worth it, I think this crate should have as little dependencies as possible. This type is unlikely to change a lot so I think it is ok to redefine it here.

clients/rust/src/client.rs
kayibal: If you feel strongly about this and you think it makes a major difference I'll change it, I think they are very similar in terms of their expressivenes.

clients/rust/src/client.rs
kayibal: Yes, correct. They can use the state overrides in case they don't.

clients/rust/src/client.rs
kayibal: `Address::ZERO` is a placeholder here since the router is not deployed yet.

I think as Diana mentioned we should take it from the server response directly. But I guess then we have to ask ourselves, where does the server get it from? Is it hardcoded there?

clients/rust/src/client.rs
kayibal: To submit to flashbots for example to avoid paying for reverted swaps.

clients/rust/src/client.rs
kayibal: Sure

clients/rust/src/client.rs
kayibal: No I have to fix that.

clients/rust/src/client.rs
kayibal: Did you have a look at the epic?

https://propeller-heads.atlassian.net/browse/ENG-5591

clients/rust/src/client.rs
kayibal: I liked the name, because these are all very optional. I expect this struct to also contain Turbine signing hints in the future.

clients/rust/src/client.rs
tamaralipows: Ahh I missed the `SignedOrder::assemble` method. AG thx!

kayibal:
@dianacarvalho1 Thanks for taking the time to review this.

1. Make it easy for our clients to submit to both exchange, or test the other one with a single flag
2. Sorry yeah mostly Claude lacking context, will fix, although the quote change came from me. I think it's more in line with what people are used to.
3. There are doctests and a lot of in-code docs [here](https://github.com/propeller-heads/fynd/blob/65bbb9bbdf455b286e69b207adad21b1d98820ff/clients/rust/src/lib.rs#L1). I can add more docs, although I'd like to wait until we have execution since I expect that things will change until then.
4. I have a commit here that refactors the benchmark to use the Rust client but I can't test the execution part yet since we don't have the calldata yet. I suggest to merge this soon so we don't drift too much from the changes happening in main (breaking changes will create compilation errors) and then do a proper test and release once calldata is part of the Fynd response

clients/rust/src/mapping.rs
kayibal: Yes this is more idiomatic, but both solutions compile to the same machine code. I'll change it.

clients/rust/src/signing.rs
kayibal: I prefer quote, I think it is a more common word, more people know exactly what it means.

clients/rust/src/types.rs
dianacarvalho1: added this to a cleanup task

clients/rust/src/types.rs
dianacarvalho1: yeah I would prefer to remove it. If later we want to support batch quotes we can bring this back

clients/rust/src/mapping.rs
dianacarvalho1: I think yes. But I won't insist here

clients/rust/src/mapping.rs
dianacarvalho1: then can we be consistent everywhere at least within Fynd? 😅 I never liked dto tbh

clients/rust/src/client.rs
dianacarvalho1: yes, it's hardcoded inside of tycho-execution. So the router address will match with the current's version encoding code

dianacarvalho1:
Thank you @kayibal !

> 1. Make it easy for our clients to submit to both exchange, or test the other one with a single flag

Then we shouldn't call this a Fynd Client? 👀 It feels like the scope is much wider (propeller client?)

I agree with waiting for encoding to be integrated to fully test this!

clients/rust/src/mapping.rs
dianacarvalho1: ah I see you changed wire to dto already - AG

dianacarvalho1 (approved):
Thank you @kayibal ! 🙏🏼

tamaralipows (approved):
Very nice. Thank you!

PR #59, opened 2026-03-03
feat: extend models

Troshchk:
- adds FailedEncoding variant to SolveError
- creates Transaction
- adds transaction to OrderSolution
- adds encoding_options to SolutionOptions
- adds gas_price to OrderSolution 
- sets gas_price in MostLiquid
- add the protocol component and protocol state

Note:
 **#[schema(value_type = Object)]** annotation for `ProtocolComponent` is necessary. 
 This can be removed if `tycho-commo`n is bumped to a version that uses utoipa v5

### DIFF
--- a/fynd-core/src/types/solution.rs
+++ b/fynd-core/src/types/solution.rs
@@ -68,6 +68,7 @@ pub struct SolutionOptions {
     pub encoding_options: Option<EncodingOptions>,
 }
 
+/// Options to customize the encoding behavior.
 #[serde_as]
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct EncodingOptions {
@@ -267,6 +268,7 @@ pub struct OrderSolution {
     /// Effective gas price (in wei) at the time the route was computed.
     #[serde_as(as = "Option<DisplayFromStr>")]
     pub gas_price: Option<BigUint>,
+    /// An encoded EVM transaction ready to be submitted on-chain.
     pub transaction: Option<Transaction>,
 }
 
--- a/fynd-core/src/types/solution.rs
+++ b/fynd-core/src/types/solution.rs
@@ -765,7 +765,31 @@ mod tests {
             "token_out": "0x0202020202020202020202020202020202020202",
             "amount_in": "1000000000000000000",
             "amount_out": "999000000000000000",
-            "gas_estimate": "150000"
+            "gas_estimate": "150000",
+            "protocol_component": {
+                "id": "test-pool",
+                "protocol_system": "uniswap_v2",
+                "protocol_type_name": "swap",
+                "chain": "ethereum",
+                "tokens": [
+                    "0x0101010101010101010101010101010101010101",
+                    "0x0202020202020202020202020202020202020202"
+                ],
+                "contract_addresses": [],
+                "static_attributes": {},
+                "change": "Update",
+                "creation_tx": "0x",
+                "created_at": "1970-01-01T00:00:00"
+            },
+            "protocol_state": {
+                "protocol": "MockProtocolSim",
+                "state": {
+                    "spot_price": 2,
+                    "gas": 50000,
+                    "liquidity": 340282366920938463463374607431768211455,
+                    "fee": 0.0
+                }
+            }
         }"#;
 
         let swap: Swap = serde_json::from_str(json).unwrap();
--- a/fynd-core/src/types/solution.rs
+++ b/fynd-core/src/types/solution.rs
@@ -20,7 +20,11 @@ use num_bigint::{BigInt, BigUint};
 use num_traits::Zero;
 use serde::{Deserialize, Serialize};
 use serde_with::{serde_as, DisplayFromStr};
-use tycho_simulation::{tycho_common::models::Address, tycho_core::models::token::Token};
+pub use tycho_execution::encoding::models::UserTransferType;
+use tycho_simulation::{
+    tycho_common::models::Address,
+    tycho_core::{models::token::Token, Bytes},
+};
 use uuid::Uuid;
 
 use super::primitives::ComponentId;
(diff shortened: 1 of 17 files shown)


fynd-core/src/types/solution.rs
```diff
@@ -57,6 +61,52 @@ pub struct SolutionOptions {
     #[serde_as(as = "Option<DisplayFromStr>")]
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub max_gas: Option<BigUint>,
+    pub encoding_options: Option<EncodingOptions>,
+}
+
+#[serde_as]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct EncodingOptions {
+    pub slippage: f64,
```
dianacarvalho1, line 75: can we add a comment here please?

fynd-core/src/types/solution.rs
```diff
@@ -57,6 +61,52 @@ pub struct SolutionOptions {
...
+    /// Token transfer method. Defaults to `TransferFrom`.
+    #[serde(default = "default_transfer_type")]
+    pub transfer_type: UserTransferType,
+    /// Permit2 single-token authorization. Required when using `TransferFromPermit2`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub permit: Option<PermitSingle>,
+    /// Permit2 signature (65 bytes). Required when `permit` is set.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub signature: Option<Bytes>,
```
dianacarvalho1: can we call this permit2_signature please?

fynd-core/src/types/solution.rs
```diff
@@ -210,6 +260,10 @@ pub struct OrderSolution {
     /// Algorithm that found this solution (internal use only).
     #[serde(skip)]
     pub algorithm: String,
+    /// Effective gas price (in wei) at the time the route was computed.
+    #[serde_as(as = "Option<DisplayFromStr>")]
+    pub gas_price: Option<BigUint>,
+    pub transaction: Option<Transaction>,
```
dianacarvalho1, line 272: missing doc

fynd-core/src/types/solution.rs
```diff
@@ -429,10 +485,27 @@ impl Swap {
...
+    /// ABI-encoded calldata.
+    pub data: Vec<u8>,
+}
+
 // ============================================================================
 // PRIVATE HELPERS
 // ============================================================================
 
+/// Generates a unique order ID using UUID v4.
+fn default_transfer_type() -> UserTransferType {
```
dianacarvalho1, line 530: is this necessary?? I think since we already have a default in the dto, it is redundant to have one here too

fynd-rpc/src/api/dto.rs
```diff
@@ -43,6 +43,74 @@ pub struct SolutionOptions {
...
+    /// Use funds already present in the Tycho Router (no transfer performed).
+    None,
+}
+
+#[serde_as]
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct EncodingOptions {
+    #[serde_as(as = "DisplayFromStr")]
+    #[schema(example = "0.001")]
+    pub slippage: f64,
```
dianacarvalho1, line 71: missing doc

fynd-rpc/src/api/dto.rs
```diff
@@ -161,6 +229,12 @@ pub struct OrderSolution {
     pub amount_out_net_gas: BigUint,
     /// Block at which this quote was computed.
     pub block: BlockInfo,
+    /// Effective gas price (in wei) at the time the route was computed.
+    #[serde_as(as = "Option<DisplayFromStr>")]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[schema(value_type = Option<String>, example = "20000000000")]
+    pub gas_price: Option<BigUint>,
+    pub transaction: Option<Transaction>,
```
dianacarvalho1, line 242: missing doc

dianacarvalho1 (commented):
Thank you @Troshchk 🙏🏼 
ah I noticed now I missed something in the task: to add the protocol component and protocol state to swap too (like [here](https://github.com/propeller-heads/fynd/compare/main...draft/dc/add-encoding#diff-2f0e40b9747781e9ffaf6745eec552504a5c7934573285936cd4db345b76359dR489)) can you add it please?

fynd-core/src/types/solution.rs
```diff
@@ -429,10 +485,27 @@ impl Swap {
...
+    pub value: num_bigint::BigUint,
+    /// ABI-encoded calldata.
+    pub data: Vec<u8>,
+}
+
 // ============================================================================
 // PRIVATE HELPERS
 // ============================================================================
 
+/// Generates a unique order ID using UUID v4.
```
tamaralipows: This docstring is wrong too I think (but we should anyway delete this default)

tamaralipows (commented):
Thanks! No additional comments other than what Diana mentioned

fynd-core/src/types/solution.rs
Troshchk: Here the `UserTransferType` used is from the `tycho-execution`, which does not have a default value. This is why the function is here.
I can add a default option `UserTransferType` in the `tycho-execution`, but I am not sure it is a good solution. Let me know what you think

dianacarvalho1: no, I'm wondering if we need a default here at all. In the Fynd logic we will construct this object from the dto object, which already has a default.
On the other side people might want to use Fynd only through core and for that use case, ok, having the default here would be nice. 
I'll leave it up to you!

Troshchk: I would prefer to keep it for that case, thanks!

fynd-core/src/types/solution.rs
```diff
@@ -412,10 +474,16 @@ pub struct Swap {
     /// Estimated gas cost for this swap (as decimal string).
     #[serde_as(as = "DisplayFromStr")]
     pub gas_estimate: BigUint,
+    /// Protocol component description.
```
dianacarvalho1: description?

dianacarvalho1 (approved):
Looks good! Thank you 🙏🏼

PR #60, opened 2026-03-03
feat: add @fynd/client TypeScript package

kayibal:
## Summary

- Adds `@fynd/client` — a TypeScript client for the Fynd DEX router, mirroring the Rust client (`feat/eng-5593-rust-client`)
- All wire types sourced from `@fynd/autogen` (no hand-rolled OpenAPI types)
- Integrates into the existing `clients/typescript` pnpm workspace

## What's included

**Source (`clients/typescript/client/src/`):**
- `types.ts` — client-facing types: `Order`, `Quote`, `Swap`, `Route`, `BlockInfo`, `HealthStatus`
- `error.ts` — `FyndError` with `ServerErrorCode`/`ClientErrorCode`, `isRetryable`, factory methods
- `signing.ts` — `signingHash`, `assembleSignedOrder`, `ExecutionReceipt`, `SettledOrder`
- `mapping.ts` — wire↔client conversions, `exactOptionalPropertyTypes`-compliant
- `client.ts` — `FyndClient` with `quote` (retry + ±25% jitter), `health`, `signablePayload`, `execute` (standard + dry-run), ERC-20/ERC-6909 Transfer log decoding

**Tests:** 94 tests across `error.test.ts`, `mapping.test.ts`, `signing.test.ts`, `client.test.ts`

**CI:** `check_ts_client` job added — runs `tsc --noEmit`, `oxlint`, and `vitest run`

## Relates to

[ENG-5595](https://propeller-heads.atlassian.net/browse/ENG-5595)

[ENG-5595]: https://propeller-heads.atlassian.net/browse/ENG-5595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -112,6 +112,25 @@ jobs:
           npm ci
           npx tsc --noEmit
 
+  check_ts_client:
+    name: TypeScript client typecheck + lint + test
+    needs: typescript_autogen_drift
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4
+      - uses: pnpm/action-setup@9fd676a19091d4595eefd76e4bd31c97133911f1  # v4.2.0
+        with:
+          version: "10.30.3"
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020  # v4
+        with:
+          node-version: "22"
+          cache: "pnpm"
+          cache-dependency-path: clients/typescript/pnpm-lock.yaml
+      - run: pnpm --dir clients/typescript install --frozen-lockfile
+      - run: pnpm --dir clients/typescript --filter @fynd/client run typecheck
+      - run: pnpm --dir clients/typescript --filter @fynd/client run lint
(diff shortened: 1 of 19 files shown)


mp-web3 (commented):
Nice setup, strict tsconfig, ESM-only, branded types, clean barrel exports. 

A few things I recommend fixing:

**moduleResolution: "bundler" → should be "node16"**
bundler resolution is for apps, not libraries. It doesn't enforce .js extensions on relative imports, nothing stops a future contributor from adding a bare import that compiles fine here but breaks for Node ESM consumers. node16 or nodenext enforces correct resolution at compile time.

**Test files will end up in dist/**
tsconfig.json includes src/**/* and build is just tsc. The colocated *.test.ts files will compile into the published package. Either add "exclude": ["src/**/*.test.ts"] or use a separate tsconfig.build.json that the build script uses.

**No files field in package.json**
Without "files": ["dist"], npm publish ships source, tests, and everything else.

**skipLibCheck: true can hide issues for consumers**
Declarations compile fine, but consumers with skipLibCheck: false (the default) may hit type errors you never see during development.

PR #61, opened 2026-03-05
fix: Fix Dockerfile

dianacarvalho1:
After the split into multiple crates the Dockerfile was broken Ran it locally now and it works!


Troshchk (approved):
Thank you ! 💫

PR #62, opened 2026-03-05
fix: remove protocol related data from dto Swap

Troshchk:
Remove protocol-related data.

When sending a solve request to Fynd, the payload currently includes excessive protocol information that is not required for the request.


dianacarvalho1 (approved):
🙏🏼

PR #65, opened 2026-03-06
feat: add encoder

Troshchk:
This PR implements encoder:
- converts order solution into the execution Solution model
- encodes Solution
- encodes full tycho call using `EncodingOptions`
- adds Transaction to the successful order solutions


fynd-core/src/types/solution.rs
```diff
@@ -270,6 +274,14 @@ pub struct OrderSolution {
...
     /// An encoded EVM transaction ready to be submitted on-chain.
     pub transaction: Option<Transaction>,
+    /// Address of the sender.
+    pub sender: Bytes,
+    /// Address of the receiver.
+    pub receiver: Bytes,
+    /// False if the solution is an exact input solution. Currently only exact input solutions are
+    /// supported.
+    #[serde(default)]
+    pub exact_out: bool,
```
dianacarvalho1: no need to have this here for now

fynd-core/src/types/quote.rs
```diff
@@ -239,6 +239,10 @@ pub struct SingleOrderSolution {
 pub struct OrderSolution {
     /// ID of the order this solution corresponds to.
     pub order_id: String,
+    /// Input token address (the token being sold).
+    pub token_in: Address,
```
dianacarvalho1: not sure if we need this because Route has a input_token method. can you check please?

fynd-core/src/encoding/encoder.rs
```diff
@@ -0,0 +1,613 @@
...
+                encoding_options.clone(),
+            )?;
+            solutions[idx].transaction = Some(transaction);
+        }
+
+        Ok(solutions)
+    }
+
+    /// Maps an `OrderSolution` to the `Solution` type expected by TychoEncoder.
+    fn order_solution_to_solution(
```
dianacarvalho1: maybe `order_solution_to_tycho_solution` ?

fynd-core/src/encoding/encoder.rs
```diff
@@ -0,0 +1,613 @@
...
+            .swaps;
+        let first_swap = swaps
+            .first()
+            .ok_or_else(|| SolveError::Internal("route has no swaps".to_string()))?;
+        let last_swap = swaps
+            .last()
+            .ok_or_else(|| SolveError::Internal("route has no swaps".to_string()))?;
+
+        let token_in = Bytes::from(first_swap.token_in.as_ref());
+        let token_out = Bytes::from(last_swap.token_out.as_ref());
```
dianacarvalho1: here you are also ignoring the new `token_in`/`token_out` you added. But you can just use the Route methods for that

fynd-core/src/encoding/encoder.rs
```diff
@@ -0,0 +1,613 @@
...
+
+        let amount_in = order_solution.amount_in.clone();
+        let amount_out = order_solution.amount_out.clone();
+
+        let native_action = self.native_action(order_solution, &token_in, &token_out)?;
+
+        let (given_token, checked_token, given_amount, checked_amount) = if order_solution.exact_out
```
dianacarvalho1: we only support exact in anyway now

fynd-core/src/encoding/encoder.rs
dianacarvalho1: actually.. instead of having this could we have a 
```
impl From<OrderSolution> for Solution
```
?

fynd-core/src/encoding/encoder.rs
```diff
@@ -0,0 +1,613 @@
...
+        };
+        Ok(solution_exec)
+    }
+
+    /// Determines whether a wrap or unwrap of the native token is required.
+    ///
+    /// Returns `Wrap` when the order sells the native token and the first swap expects the wrapped
+    /// token, `Unwrap` when the order buys the native token and the last swap outputs the wrapped
+    /// token, and `None` otherwise.
+    fn native_action(
```
dianacarvalho1: no need for this. Fynd doesn't support any kind of Native action as of now anyway

fynd-core/src/encoding/encoder.rs
```diff
@@ -0,0 +1,613 @@
...
+/// # Arguments
+/// * `encoded_solution` - Output of `TychoEncoder::encode_solutions` for one solution.
+/// * `solution` - Original solution providing token addresses, amounts, and native action.
+/// * `user_transfer_type` - How tokens are transferred from the user to the router.
+/// * `native_address` - Address used to identify the native token (e.g. ETH).
+/// * `encoding_options` - Slippage, permit2 data, and other encoding options.
+///
+/// # Returns
+/// A `Transaction` containing the ABI-encoded calldata and ETH value to send.
+pub fn encode_tycho_router_call(
```
dianacarvalho1: why isn't this a method of the Encoder?

fynd-core/src/encoding/encoder.rs
```diff
@@ -0,0 +1,613 @@
...
+    let value = if solution.given_token == *native_address {
+        solution.given_amount.clone()
+    } else {
+        BigUint::ZERO
+    };
+    Ok(Transaction { to: encoded_solution.interacting_with, value, data: contract_interaction })
+}
+
+/// Encodes the input data for a function call to the given function selector.
+fn encode_input(selector: &str, mut encoded_args: Vec<u8>) -> Vec<u8> {
```
dianacarvalho1: same here

fynd-core/src/encoding/encoder.rs
```diff
@@ -0,0 +1,613 @@
...
+        Bytes::from_str("0x0000000000000000000000000000000000000000").unwrap()
+    }
+
+    fn weth() -> Bytes {
+        Bytes::from_str("0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2").unwrap()
+    }
+
+    fn make_route_swap(token_in_byte: u8, token_out_byte: u8) -> crate::types::Swap {
+        make_route_swap_addrs(make_address(token_in_byte), make_address(token_out_byte))
+    }
```
dianacarvalho1: this is just a pass through method no? I would remove it

fynd-core/src/encoding/encoder.rs
```diff
@@ -0,0 +1,613 @@
...
+
+    fn make_route_swap_addrs(token_in: Address, token_out: Address) -> crate::types::Swap {
+        use tycho_simulation::tycho_core::models::{token::Token, Chain as SimChain};
+
+        use crate::algorithm::test_utils::{component, MockProtocolSim};
```
dianacarvalho1: can import this at the top of the test section

fynd-core/src/encoding/encoder.rs
```diff
@@ -0,0 +1,613 @@
...
+            address: addr,
+            symbol: "T".to_string(),
+            decimals: 18,
+            tax: Default::default(),
+            gas: vec![],
+            chain: SimChain::Ethereum,
+            quality: 100,
+        };
+        let tin = make_token(token_in.clone());
+        let tout = make_token(token_out.clone());
```
dianacarvalho1, line 350: ahahahhaha tin tout 😂 that's very funny actually

fynd-core/src/encoding/encoder.rs
```diff
@@ -0,0 +1,613 @@
...
+                .collect::<Vec<u8>>()
+    {
+        encoded_args = encoded_args[32..].to_vec();
+    }
+    call_data.extend(encoded_args);
+    call_data
+}
+
+#[cfg(test)]
+mod tests {
```
dianacarvalho1, line 321: we are actually never testing the `encode` method.. can you add a test for it please?

fynd-core/Cargo.toml
```diff
@@ -28,6 +28,7 @@ tracing.workspace = true
 metrics.workspace = true
 typetag.workspace = true
 tycho-execution.workspace = true
+alloy = { version = "1.0", default-features = false, features = ["sol-types"] }
```
dianacarvalho1: can we use the one from the workspace? if there isn't any, add it there instead of defining it here? also it's a dev dependency already

dianacarvalho1 (commented):
Thank you @Troshchk ! 🙏🏼 I have some comments

fynd-core/src/encoding/encoder.rs
```diff
@@ -0,0 +1,613 @@
...
+    fn order_solution_to_solution(
+        &self,
+        order_solution: &OrderSolution,
+    ) -> Result<Solution, SolveError> {
+        // Can use unwrap since status is successful status which means route exists
+        let swaps = &order_solution
+            .route
+            .as_ref()
+            .unwrap()
+            .swaps;
```
tamaralipows: Not sure I feel great about this unwrap - I'm worried this might be called sometime without realizing the solution has to be successful. How bad would it be to just explicitly handle the case here that there is no route?

fynd-core/src/encoding/encoder.rs
tamaralipows: We should probably error if this is passed?

tamaralipows (commented):
Thanks Ksenia, I think Diana already covered most comments!

fynd-core/src/types/quote.rs
tamaralipows: I don't think this is a good idea. Route is `Option<Route>`, so it's `None` for failed solutions. Those solutions still need `token_in`/`token_out` in the API response so the client knows which token pair failed. For example in `order_manager/mod.rs`, failed responses are constructed with `token_in`/`token_out` but route: `None`. 

This makes it far more explicit, and I'm also used to a similar paradigm in encoding for `tycho-execution`, where the `Solution` object has the tokens, as does the inner swaps.  

Furthermore if we remove this and rely on the input_token of the route instead, we won't be able to implement a `From<OrderSolution>`, since it now relies on the route existing and won't be able to just create a Solution with empty swaps. Please see my latest commit.

fynd-core/src/encoding/encoder.rs
```diff
@@ -32,6 +32,40 @@ pub struct Encoder {
     chain: Chain,
 }
 
+impl From<&OrderSolution> for Solution {
+    fn from(order_solution: &OrderSolution) -> Self {
+        let swaps = order_solution
```
dianacarvalho1: also add the split here bitte

fynd-core/src/encoding/encoder.rs
```diff
@@ -0,0 +1,517 @@
...
+        user_transfer_type: &UserTransferType,
+        encoding_options: EncodingOptions,
+    ) -> Result<Transaction, EncodingError> {
+        let (mut unwrap, mut wrap) = (false, false);
+        if let Some(action) = solution.native_action.clone() {
+            match action {
+                NativeAction::Wrap => wrap = true,
+                NativeAction::Unwrap => unwrap = true,
+            }
+        }
```
dianacarvalho1: why is this still needed? we can assume it will always be false

fynd-core/src/encoding/encoder.rs
```diff
@@ -0,0 +1,517 @@
...
+
+        let given_amount = biguint_to_u256(&solution.given_amount);
+        let precision = BigUint::from(1_000_000u64);
+        let slippage_amount = solution.checked_amount.clone() *
```
dianacarvalho1, line 173: unrelated: should we add a warning when people pass weird slippage values in the request (like 0 or too big)? 🤔

fynd-core/src/encoding/encoder.rs
```diff
@@ -0,0 +1,517 @@
...
+                    expiration: p.details.expiration,
+                    nonce: p.details.nonce,
+                },
+                spender: p.spender,
+                sig_deadline: p.sig_deadline,
+            };
+            let permit = Some(
+                PermitSingle::try_from(&models_permit)
+                    .map_err(|_| EncodingError::InvalidInput("Invalid permit".to_string()))?,
+            );
```
dianacarvalho1: why do we need to convert to `ModelsPermitDetails` and then to `PermitSingle`? couldn't we skip the ModelsPermitDetails? I think I'm missing something here

fynd-core/src/encoding/encoder.rs
```diff
@@ -0,0 +1,517 @@
...
+        }
+
+        fn validate_solution(&self, _solution: &Solution) -> Result<(), EncodingError> {
+            Ok(())
+        }
+    }
+
+    fn test_encoder(chain: Chain) -> Encoder {
+        test_encoder_with_encoded_solutions(chain, vec![])
+    }
```
dianacarvalho1: this is just passthrough

fynd-core/src/encoding/encoder.rs
```diff
@@ -0,0 +1,517 @@
...
+
+    fn test_encoder_with_encoded_solutions(
```
dianacarvalho1: moked_encoder

fynd-core/src/types/quote.rs
dianacarvalho1: > Those solutions still need token_in/token_out in the API response so the client knows which token pair failed.

This quote is related to an order, so I wouldn't say the user really neeeeeds the tokens here tbh.

> This makes it far more explicit, and I'm also used to a similar paradigm in encoding for tycho-execution, where the Solution object has the tokens, as does the inner swaps.

I feel that in execution it is different because we don't have the order concept - that's why we need everything in the Solution

> Furthermore if we remove this and rely on the input_token of the route instead, we won't be able to implement a From<OrderSolution>, since it now relies on the route existing and won't be able to just create a Solution with empty swaps. Please see my latest commit.

I don't understand why this is relevant - this case should never happen. we are only doing `Solution::from(quote))` for cases where `quote.status() == QuoteStatus::Success`

dianacarvalho1 (commented):
Thank you for taking this over @tamaralipows 🙏🏼 I have a few more comments!

fynd-core/src/types/quote.rs
tamaralipows: > This quote is related to an order, so I wouldn't say the user really neeeeeds the tokens here tbh.

So in this case we just return an empty byte string for the in and out token? And add a comment that this shouldn't happen? Else we will need to error on the case where `quote.status() != QuoteStatus::Success` which means we will have to use a `TryFrom` instead of a regular `From` - are we okay with that?

fynd-core/src/encoding/encoder.rs
tamaralipows: Mhhh I thought this was the simplest way to do this. I'll construct the PermitSingle from scratch but it's not super elegant with its type conversions (see latest comment-addressing commit)

clients/openapi.json
```diff
@@ -7,7 +7,7 @@
       "name": "MIT",
       "identifier": "MIT"
     },
-    "version": "0.3.1"
+    "version": "0.4.0"
```
dianacarvalho1, line 10: I think this should be done automatically? 🤔

dianacarvalho1 (approved):
niceeeee thank you! 🙏🏼

clients/openapi.json
tamaralipows: idk what happened. CI was failing

PR #68, opened 2026-03-09
feat: allow external algorithms in WorkerPoolBuilder

dianacarvalho1:
Add WorkerPoolBuilder::with_algorithm() so callers can pass any Algorithm implementation directly without modifying the registry.

- Add AlgorithmSpawner enum (Registry | Custom) to registry.rs, replacing the spawn_workers free function; Custom stores a type-erased factory closure that bypasses the string-based registry
- WorkerPoolConfig.algorithm: String -> pub(crate) spawner: AlgorithmSpawner; add algorithm_name() public accessor; remove Clone (boxed closures are not Clone)
- WorkerPoolBuilder gains with_algorithm<A, F>() alongside the existing algorithm() method; existing .algorithm("most_liquid") call sites are unchanged
- spawn_workers_generic now takes factory: &F (borrow not consume) so the type- erased spawner closure can call it without giving up ownership of the factory
- Export ComputationRequirements from lib.rs so external Algorithm implementors can satisfy the trait return type
- Add examples/custom_algorithm.rs showing a delegating MyAlgorithm using with_algorithm("my_custom_algo", MyAlgorithm::new)

🤖 Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/fynd-core/src/worker_pool/pool.rs
+++ b/fynd-core/src/worker_pool/pool.rs
@@ -15,20 +15,22 @@ use crate::{
     feed::{events::MarketEvent, market_data::SharedMarketDataRef},
     types::internal::SolveTask,
     worker_pool::{
-        registry::{spawn_workers, SpawnWorkersParams, UnknownAlgorithmError, DEFAULT_ALGORITHM},
+        registry::{
+            spawn_workers_generic, AlgorithmSpawner, SpawnWorkersParams, UnknownAlgorithmError,
+            DEFAULT_ALGORITHM,
+        },
         task_queue::{TaskQueue, TaskQueueConfig, TaskQueueHandle},
     },
 };
 
 /// Configuration for the worker pool.
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct WorkerPoolConfig {
     /// Human-readable name for this pool (used in logging/metrics).
     /// Can differ from algorithm to distinguish pools with same algorithm but different configs.
     name: String,
-    /// Algorithm name for this pool (e.g., "most_liquid").
-    /// Use `worker_pool::list_algorithms()` to see available options.
-    algorithm: String,
+    /// How to spawn workers — either a built-in registry lookup or a custom factory.
+    pub(crate) spawner: AlgorithmSpawner,
     /// Number of worker threads.
     num_workers: usize,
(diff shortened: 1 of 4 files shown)


fynd-core/src/worker_pool/pool.rs
```diff
@@ -15,20 +15,22 @@ use crate::{
...
+#[derive(Debug)]
 pub struct WorkerPoolConfig {
     /// Human-readable name for this pool (used in logging/metrics).
     /// Can differ from algorithm to distinguish pools with same algorithm but different configs.
     name: String,
-    /// Algorithm name for this pool (e.g., "most_liquid").
-    /// Use `worker_pool::list_algorithms()` to see available options.
-    algorithm: String,
+    /// How to spawn workers — either a built-in registry lookup or a custom factory.
+    pub(crate) spawner: AlgorithmSpawner,
```
tamaralipows: Shouldn't we make this private with a getter?

fynd-core/src/worker_pool/pool.rs
```diff
@@ -169,11 +181,36 @@ impl WorkerPoolBuilder {
...
+    /// # Example
+    ///
+    /// ```ignore
+    /// builder.with_algorithm("my_algo", |config| MyAlgorithm::new(config))
+    /// ```
+    pub fn with_algorithm<A, F>(mut self, name: impl Into<String>, factory: F) -> Self
+    where
+        A: crate::algorithm::Algorithm + 'static,
+        A::GraphManager: crate::feed::events::MarketEventHandler
+            + crate::graph::EdgeWeightUpdaterWithDerived
```
tamaralipows: Can we move these imports to the top for readability please

tamaralipows (approved):
Thank you - just minor things

fynd-core/src/worker_pool/pool.rs
dianacarvalho1: good eye! but actually this doesn't even need to be pub(crate) and there is already a setter in the builder further down
```

    pub fn algorithm(mut self, algorithm: impl Into<String>) -> Self {
        self.config.spawner = AlgorithmSpawner::Registry { algorithm: algorithm.into() };
        self
    }
```

PR #74, opened 2026-03-10
feat: include derived data in readiness in health check

louise-poole:
## Summary

  The health endpoint only checked market data freshness, so it could report healthy while workers failed with "timeout waiting for derived data: missing {"token_prices"}". Now the service is reported as unhealthy until at least one derived data computation cycle completes.

  - Adds derived_data_ready field to HealthStatus response (serde(default) for backward compat)
  - Health returns 503 if market data is stale or derived data hasn't been computed yet
  - derived_data_ready reflects overall readiness, not per-block freshness — algorithms requiring fresh data will wait for recomputation before solving

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,12 @@ Submit one or more swap orders and receive optimal routes.
 
 ### GET /v1/health
 
-Returns service health status. HTTP 200 if healthy, 503 if stale.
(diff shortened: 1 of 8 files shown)


tamaralipows (approved):
Thank you!

dianacarvalho1 (approved):
Thank you @louise-poole 🙏🏼 Looks good to me!

PR #76, opened 2026-03-10
feat(rpc): add GET /v1/prices endpoint for derived market data

Haikane:
## Summary

- Adds `GET /v1/prices` endpoint that returns derived token gas prices, with optional spot prices and pool depths via `include` query parameter
- **Feature-gated behind `experimental`** (off by default). Enable with `--features experimental` to opt in.
- Extends `AppState` with `derived_data` (SharedDerivedDataRef) and `gas_token` (Address), both gated behind the same feature flag
- Wires the existing `ComputationManager` store into the HTTP layer

## Details

By default returns token prices only. Callers opt into additional data:

- `GET /v1/prices` returns token gas prices
- `GET /v1/prices?include=spot_prices` adds spot prices
- `GET /v1/prices?include=depths` adds pool depths
- `GET /v1/prices?include=depths,spot_prices&limit=500` returns everything, capped

Key design decisions:
- RwLock read acquired, staleness checked first, data cloned, lock dropped before serialization
- Spot prices and pool depths sorted by `(component_id, token_in, token_out)` for deterministic limit truncation
- Unknown `include` values return 400; no derived data yet returns 503
- `Price` (BigUint fraction) converted to f64 with zero-denom/overflow guard

### Feature gate

The entire endpoint is behind `#[cfg(feature = "experimental")]`:
- `prices` module, route registration, handler, and OpenAPI schemas
- `AppState.derived_data` and `AppState.gas_token` fields
- Separate `ExperimentalApiDoc` merged into OpenAPI spec only when enabled
- The checked-in `openapi.json` reflects the default (non-experimental) build

## Test plan

- [x] `cargo check` passes without the feature
- [x] `cargo check --features experimental` passes
- [x] `cargo clippy -p fynd-rpc -- -D warnings` zero warnings (both configs)
- [x] `cargo test -p fynd-rpc --features experimental` 12/12 pass (10 new: 6 include-parsing, 4 price-conversion)
- [ ] Manual test against running instance with `curl /v1/prices?include=depths,spot_prices`

### DIFF
--- a/clients/openapi.json
+++ b/clients/openapi.json
@@ -42,77 +42,6 @@
         }
       }
     },
-    "/v1/prices": {
-      "get": {
-        "tags": [
-          "prices"
-        ],
-        "summary": "GET /v1/prices - Return derived token prices and optional market data.",
-        "description": "By default returns token gas prices only. Use `include` query parameter\nto add spot prices and/or pool depths.\n\n# Query Parameters\n\n- `include` - Comma-separated list: `depths`, `spot_prices`\n- `limit` - Max entries for spot_prices / pool_depths (default: 1000)",
-        "operationId": "get_prices",
-        "parameters": [
-          {
-            "name": "include",
-            "in": "query",
-            "description": "Comma-separated list of additional data to include.\nValid values: `depths`, `spot_prices`.",
-            "required": false,
-            "schema": {
-              "type": [
-                "string",
-                "null"
-              ]
-            },
-            "example": "depths,spot_prices"
-          },
-          {
-            "name": "limit",
-            "in": "query",
-            "description": "Maximum number of spot_prices and pool_depths entries (default: 1000).",
-            "required": false,
-            "schema": {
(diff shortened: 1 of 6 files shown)


Haikane:
Can you feature-gate the `/v1/prices` endpoint behind a Cargo feature flag called `experimental`?

The endpoint itself looks good, but we should gate it so it's opt-in until we're confident in the API surface. Specifically:

1. Add `experimental` to `[features]` in `fynd-rpc/Cargo.toml` (default off)
2. Gate the route registration in `configure_routes` with `#[cfg(feature = "experimental")]`
3. Gate the `prices` module, the handler, and the OpenAPI schema registrations behind the same flag
4. The `AppState` fields (`derived_data`, `gas_token`) and builder wiring can stay ungated since the computation manager is already running regardless

That way consumers only get the endpoint when they explicitly opt in with `--features experimental`.

kayibal (approved):

PR #79, opened 2026-03-10
feat: Use encoder in order manager

tamaralipows:
Note:
- Allowed the encoder to be passed to the FyndBuilder: this might be useful in case the user wants to override the executor addresses or add their own protocol's swap encoders for example.

I did test this using the tutorial and got a quote. **However**, the tutorial changes are out of scope of this PR so they are not committed. They will be added in the next PR. I have only fixed the examples that were directly initializing the `OrderManager`.

```
(base) [15:37:45] [~/Code/fynd] >>> cargo run --example tutorial
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.87s
     Running `target/debug/examples/tutorial`

========== Quote ==========
Status: Success
Swap: 1000.000000 USDC -> 0.489979 WETH
Price: 0.000490 WETH per USDC
Gas estimate: 120000

Route (1 hops):
  1. USDC -> WETH via uniswap_v2 (pool: 0xb4e16d0168e52d35cacd2c6185b44281ec28c9dc)

Encoded transaction:
  To: 0xfd0b31d2e955fa55e3fa641fe90e08b677188d35
  Value: 0
  Calldata: 0x5c4b639c000000000000000000000000000000000000000000000000000000003b9aca00000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc200000000000000000000000000000000000000000000000006c40d74e7a99286000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000001200000000000000000000000000000000000000000000000000000000000000052ae04ca7e9ed79cbd988f6c536ce11c621166f41ba0b86991c6218b36c1d19d4a2e9eb0ce3606eb48b4e16d0168e52d35cacd2c6185b44281ec28c9dc000000000000000000000000000000000000000001000000000000000000000000000000

Solve time: 68ms
Total gas: 120000
============================
```


fynd-core/src/order_manager/mod.rs
```diff
@@ -466,7 +491,8 @@ mod tests {
     async fn test_order_manager_single_pool_success() {
         let (pool, worker) = create_mock_pool("pool_a", Ok(make_single_quote(900)), 0);
 
-        let manager = OrderManager::new(vec![pool], OrderManagerConfig::default());
+        let manager =
```
dianacarvalho1, line 525: can we add some asserts to the tests that the quotes have calldata please?

fynd-core/examples/custom_algorithm.rs
```diff
@@ -170,9 +172,14 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("Worker pool algorithm: {}", worker_pool.algorithm());
 
     // 7. OrderManager
+    let swap_encoder_registry = SwapEncoderRegistry::new(chain)
+        .add_default_encoders(None)
+        .expect("Failed to create default swap encoder registry");
+    let encoder = Encoder::new(chain, swap_encoder_registry).expect("Failed to create encoder");
```
dianacarvalho1, line 178: then this is unnecessary no? if no encoder is passed, then the default one is created

dianacarvalho1 (commented):
Thank you @tamaralipows 🙏🏼 small comments only

> I think we will also need to add ability to pass encoding options to the fynd client?

Encoding options are a part of the Quote Request.. they should already be passed.. or am I missing something?

clients/openapi.json
```diff
@@ -7,7 +7,7 @@
       "name": "MIT",
       "identifier": "MIT"
     },
-    "version": "0.5.0"
+    "version": "0.6.0"
```
louise-poole: FMI - why is this version bumped? I assume this was automated, but see no changes to the api interface so this seems weird to me.

fynd-core/examples/custom_algorithm.rs
louise-poole: I think you're thinking of the FyndBuilder (a default encoder is created there and passed to the OrderManager it makes). OrderManager itself expects an encoder always.

louise-poole (approved):
Looks good to me!

fynd-core/examples/custom_algorithm.rs
dianacarvalho1: 🥴 you are right! Thank you @louise-poole 🙏🏼 resolving my comment

clients/openapi.json
tamaralipows: I don't know honestly :/ the CI was failing if I didn't do this manually actually.

tamaralipows: oh I think this explains it https://propellerswap.slack.com/archives/C05LQALE9L1/p1773245461247809

PR #82, opened 2026-03-11
fix: quote endpoint, block hash encoding, and quote example

kayibal:
## Summary

- Fixes the Rust client sending requests to `/v1/solve` instead of `/v1/quote`
- Fixes block hash being double-encoded (`"\"0x...\""`) due to `{:?}` debug formatting of a `&str` in `worker.rs`
- Adds a `quote` example that runs health check and two quote requests against a live Fynd instance

## Changes

**`fynd-core/src/worker_pool/worker.rs`**
`format!("{:?}", last_block.hash())` used the `Debug` formatter on a `&str`, which wraps the value in double-quotes. Replaced with `last_block.hash().to_string()`.

**`clients/rust/src/client.rs` + `tests/integration.rs`**
All `/v1/solve` references updated to `/v1/quote` to match the actual server route.

**`clients/rust/examples/quote.rs`**
New example: connects to `localhost:3000` via `build_quote_only`, checks health, fetches WETH→USDC and USDC→USDT quotes, and asserts status and amounts.

## Test plan

- [ ] `cargo test -p fynd-client` — all integration tests pass
- [ ] `cargo test -p fynd-core` — passes
- [ ] `cargo run --example quote -p fynd-client` against a local Fynd instance — all assertions pass

### DIFF
--- a/clients/rust/examples/quote.rs
+++ b/clients/rust/examples/quote.rs
@@ -57,14 +57,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // -----------------------------------------------------------------------
     let quote = client
         .quote(QuoteParams::new(
-            Order::new(
-                addr(WETH),
-                addr(USDC),
-                one_ether(),
-                OrderSide::Sell,
-                addr(VITALIK),
-                None,
-            ),
+            Order::new(addr(WETH), addr(USDC), one_ether(), OrderSide::Sell, addr(VITALIK), None),
             QuoteOptions::default(),
         ))
         .await?;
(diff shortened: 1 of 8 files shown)


scripts/update-openapi.sh
louise-poole, line 1: Can you update the CI check action to prompt the user to run this script instead of giving them the manual update command? At least on the openapi check (it fails first before the typescript check).

louise-poole (approved):
LGTM! 🙏

clients/rust/examples/quote.rs
```diff
@@ -0,0 +1,118 @@
+//! Example: request quotes from a local Fynd instance.
```
dianacarvalho1, line 1: I'm getting a bit confused with all the examples 😅 What should the current [tutorial](https://github.com/propeller-heads/fynd/tree/main/examples/tutorial) be like? I think ultimately it should use fynd client to actually simulate or execute

dianacarvalho1 (approved):
Thank you @kayibal 🙏🏼

clients/rust/examples/quote.rs
kayibal: Yeah I agree. This example was mainly to test if the Fynd interactions are working properly with the client. I can remove that if you prefer and we wait for the calldata to be available from the server. Then we can modify the current example to be the main global one.

dianacarvalho1: The calldata should already be available (https://github.com/propeller-heads/fynd/pull/79 was just merged!)

PR #84, opened 2026-03-11
fix(ci): strip info.version from openapi drift check

kayibal:
## Summary

- The release bot bumps `Cargo.toml` and commits it to `main`, which changes the `info.version` field in the generated OpenAPI spec
- This caused `openapi_drift` to fail on every branch created after a release, even with no real API changes
- Strip `info.version` from both sides before diffing so only actual API surface changes are caught

## Test plan

- [ ] Verify CI passes on this PR
- [ ] After next release, confirm a new branch no longer fails `openapi_drift`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -82,10 +82,15 @@ jobs:
           cache-on-failure: true
       - name: Check OpenAPI spec is up to date
         run: |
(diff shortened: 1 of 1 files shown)


tamaralipows (approved):
Ahhhhh this explains a lot! Thank you!!

PR #88, opened 2026-03-12
feat: default to all protocol systems when --protocols is omitted

louise-poole:
Fetch available onchain protocols from Tycho RPC at startup when no --protocols flag is provided or "all_onchain" keyword is used. The protocol registry now warns and skips unknown protocols instead of erroring, removing hardcoded protocol lists.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -7,8 +7,12 @@
 //! # Usage
 //!
 //! ```bash
-//! # All supported protocols are fetched from Tycho RPC by default:
+//! # All on-chain protocols are fetched from Tycho RPC by default:
 //! fynd serve --tycho-url tycho-beta.propellerheads.xyz
+//!
+//! # Combine all on-chain protocols with specific RFQ protocols:
+//! fynd serve --tycho-url tycho-beta.propellerheads.xyz \
+//!            --protocols all_onchain,rfq:bebop
 //!
 //! # Or specify protocols explicitly:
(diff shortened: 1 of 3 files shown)


tamaralipows (approved):
Thank yew!

src/main.rs
```diff
@@ -292,3 +317,43 @@ async fn run_solver(args: cli::ServeArgs) -> Result<(), SolverError> {
...
+    let mut all_protocols = Vec::new();
+    let mut page = 0;
+
+    loop {
+        let request = ProtocolSystemsRequestBody {
+            chain: chain.into(),
+            pagination: PaginationParams { page, page_size: PAGE_SIZE },
+        };
+        let response = rpc_client
+            .get_protocol_systems(&request)
```
dianacarvalho1, line 362: will this give us balancer v3 and those broken protocols?

dianacarvalho1 (commented):
Thank you @louise-poole ! One big question though: should we support all onchain protocols + RFQs? I do think that RFQs shouldn't be on by default because you need the credentials but do you think we should support just adding an RFQ protocol to all onchain ones?

src/main.rs
louise-poole: Nope - already merged the fix on the indexer side so it will only return only actively available protocols. Just need to deploy it to prod.
One step further: if the key the user is using is restricted to certain protocols, only those protocols (if active) are returned.

dianacarvalho1 (approved):
Merci 🙏🏼

PR #89, opened 2026-03-12
feat: update fynd defaults to match fynd-basic plan

zizou0x:
## Summary

Rebased on top of #88 — protocol defaults are now handled by the RPC-based fetch from that PR (no hardcoding).

- **fix: connect min_token_quality CLI arg to builder** — the `--min-token-quality` flag was accepted but silently ignored
- **feat: add traded_n_days_ago parameter** — new `--traded-n-days-ago` flag (default: 3) passed through CLI → builder → feed → `load_all_tokens`
- **fix: apply TVL buffer as lower bound for hysteresis** — the buffer ratio now defines the remove threshold (`min_tvl / ratio`), so `min_tvl` is correctly sent as `tvl_gt` to the indexer
- **refactor: rename tvl_buffer_multiplier to tvl_buffer_ratio** — better reflects its role as a hysteresis ratio
- **feat: add chain-specific defaults for tycho_url** — `--tycho-url` is now optional; defaults resolve per chain to the fynd indexer endpoints (e.g. `tycho-fynd-ethereum.propellerheads.xyz`)

With these changes, a user with a fynd-basic plan key can run `fynd serve --tycho-api-key <key>` with no extra args and get a working solver.

### DIFF
--- a/fynd-rpc/src/config.rs
+++ b/fynd-rpc/src/config.rs
@@ -170,4 +170,14 @@ pub mod defaults {
     pub const POOL_MIN_HOPS: usize = 1;
     pub const POOL_MAX_HOPS: usize = 3;
     pub const POOL_TIMEOUT_MS: u64 = 100;
+
+    /// Returns the default Tycho Fynd endpoint URL for the given chain.
+    pub fn default_tycho_url(chain: &str) -> &str {
+        match chain.to_lowercase().as_str() {
+            "ethereum" => "tycho-fynd-ethereum.propellerheads.xyz",
+            "base" => "tycho-fynd-base.propellerheads.xyz",
+            "unichain" => "tycho-fynd-unichain.propellerheads.xyz",
+            _ => "localhost:4242",
+        }
+    }
 }
--- a/fynd-rpc/src/config.rs
+++ b/fynd-rpc/src/config.rs
@@ -152,7 +152,7 @@ pub mod defaults {
     pub const MIN_TVL: f64 = 10.0;
     pub const MIN_TOKEN_QUALITY: i32 = 100;
     pub const TRADED_N_DAYS_AGO: u64 = 3;
-    pub const TVL_BUFFER_MULTIPLIER: f64 = 1.1;
+    pub const TVL_BUFFER_RATIO: f64 = 1.1;
     pub const RECONNECT_DELAY_SECS: u64 = 5;
 
     // Gas
--- a/fynd-rpc/src/config.rs
+++ b/fynd-rpc/src/config.rs
@@ -172,12 +172,17 @@ pub mod defaults {
     pub const POOL_TIMEOUT_MS: u64 = 100;
 
     /// Returns the default Tycho Fynd endpoint URL for the given chain.
-    pub fn default_tycho_url(chain: &str) -> &str {
+    ///
+    /// Returns an error if the chain is not recognized.
+    pub fn default_tycho_url(chain: &str) -> Result<&str, String> {
         match chain.to_lowercase().as_str() {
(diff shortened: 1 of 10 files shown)


fynd-rpc/src/config.rs
```diff
@@ -169,4 +170,47 @@ pub mod defaults {
...
+                "uniswap_v4",
+                "pancakeswap_v3",
+                "aerodrome_slipstreams",
+            ],
+            "unichain" => vec![
+                "uniswap_v2",
+                "uniswap_v3",
+                "uniswap_v4",
+                "velodrome_slipstreams",
+            ],
```
louise-poole: Noooooooooooooo 😭 I've just merged a PR that defaults to all protocols marked as available by the Tycho RPC. If you query the RPC with a fynd key you'll get only protocols available to that key. We want no hardcoding here. See here: https://github.com/propeller-heads/fynd/pull/88

fynd-rpc/src/config.rs
```diff
@@ -151,7 +151,8 @@ pub mod defaults {
     // Tycho stream
     pub const MIN_TVL: f64 = 10.0;
     pub const MIN_TOKEN_QUALITY: i32 = 100;
-    pub const TVL_BUFFER_MULTIPLIER: f64 = 1.1;
+    pub const TRADED_N_DAYS_AGO: u64 = 3;
+    pub const TVL_BUFFER_RATIO: f64 = 1.1;
```
louise-poole, line 155: We need to check if this plays well with the strict parameter enforcement on the service... If we define a buffer, at what tvl does the client query the rpc? I seem to recall that we query using the upper bound of the tvl threshold range which in this case will be 11...

zizou0x: I made this change exactly for that: https://github.com/propeller-heads/fynd/pull/89/changes/e3a8ee23b3a47cee79f19f7c1b691eb98fceecdf

zizou0x: So it's the lower value that is used to query the components

fynd-rpc/src/config.rs
```diff
@@ -169,4 +170,14 @@ pub mod defaults {
...
     pub const POOL_MAX_HOPS: usize = 3;
     pub const POOL_TIMEOUT_MS: u64 = 100;
+
+    /// Returns the default Tycho Fynd endpoint URL for the given chain.
+    pub fn default_tycho_url(chain: &str) -> &str {
+        match chain.to_lowercase().as_str() {
+            "ethereum" => "tycho-fynd-ethereum.propellerheads.xyz",
+            "base" => "tycho-fynd-base.propellerheads.xyz",
+            "unichain" => "tycho-fynd-unichain.propellerheads.xyz",
+            _ => "localhost:4242",
```
zizou0x: Does it make sense to keep this?

louise-poole: Maybe default to ethereum?

fynd-rpc/src/config.rs
louise-poole: I seeee - okay then we need to update some docs. I remember the config doc specifically wording how the threshold is used and switching how the buffer is applied changes that.

fynd-rpc/src/config.rs
zizou0x: The chain argument is already defaulting to Ethereum. I wouldn't double default. I think I'll just make it return a result if it's unknown

fynd-rpc/src/config.rs
zizou0x: Updated gitbook in https://github.com/propeller-heads/fynd/pull/89/changes/45af0edd1292a8ec8992e8e1c00914f9bc957947

PR #90, opened 2026-03-12
feat(client): expose Permit2 transfer support + permit2 example

kayibal:
## Summary

- Tutorial example rewritten to use \`FyndClient\` end-to-end
- \`FyndClient\` now supports Permit2 (AllowanceTransfer) as a first-class token transfer method alongside the existing \`approve\` + \`transferFrom\` path
- \`PermitSingle::eip712_signing_hash()\` computes the Permit2 EIP-712 typed-data hash so callers can sign externally — the library never touches private keys
- New \`permit2\` example mirrors the tutorial but uses Permit2 end-to-end

## Tutorial example (\`examples/tutorial\`)

Complete rewrite from raw \`tycho-execution\` usage to \`FyndClient\`:

- Uses \`FyndClient::quote\`, \`signable_payload\`, and \`execute\` throughout
- New \`--execute\` flag: default mode is a dry-run simulation; pass \`--execute\` to submit on-chain
- ERC-20 helpers (\`balanceOf\`, \`allowance\`, storage slot detection) extracted into \`erc20.rs\` module
- Checks sell-token allowance before on-chain execution and approves the exact sell amount
- README rewritten to match the new flow
- Removed unused \`types.rs\`

## Client changes (\`clients/rust\`)

**\`router_address\` removed** from \`FyndClientBuilder\` and \`FyndClient\` — the router address now comes from \`Quote::transaction\`, which is populated by the server when \`EncodingOptions\` is set. \`TxEip1559\` fields \`to\`, \`value\`, and \`input\` are sourced from there.

**New public types** — \`PermitDetails\`, \`PermitSingle\`, \`UserTransferType\`, \`Transaction\` (all with \`Debug + Clone\`)

**\`UserTransferType\`** gains two variants:
- \`TransferFromPermit2\` — use Permit2 AllowanceTransfer
- \`None\` — funds already in the router, skip transfer entirely

**\`EncodingOptions\`** gains two builder methods:
- \`with_permit2(permit, signature)\` — sets transfer type, permit struct, and pre-computed 65-byte signature; returns \`Result\` and validates signature length
- \`with_no_transfer()\` — sets \`UserTransferType::None\`

**\`PermitSingle::eip712_signing_hash(chain_id, permit2_address)\`** — computes the Permit2 EIP-712 signing hash. Callers pass the result to their signer and supply the signature to \`with_permit2()\`.

**\`Quote::transaction()\`** — returns the server-encoded \`Transaction\` (router address, native value, calldata). Present only when \`EncodingOptions\` was set in the request.

**\`mapping.rs\`** — \`From<QuoteOptions>\` converted to \`TryFrom\` to propagate address-validation errors from permit fields.

## Permit2 example (\`examples/permit2\`)

Uses \`PermitSingle::eip712_signing_hash()\` + \`EncodingOptions::with_permit2()\`. Key differences from the tutorial:

- \`--router\` arg required (the router is the Permit2 \`spender\` and must be known before the quote)
- On-chain: reads current nonce via \`eth_call\` on Permit2, derives expiration/deadline from latest block timestamp, checks ERC-20 allowance to Permit2 (not the router)
- Dry-run: ephemeral key has nonce 0 in Permit2 naturally — no Permit2 state override needed; only ERC-20 balance and allowance-to-Permit2 are injected via storage overrides


examples/permit2/erc20.rs
```diff
@@ -0,0 +1,121 @@
+//! ERC-20 helpers for the permit2 example: on-chain reads and storage slot detection.
```
dianacarvalho1: do we need to have this file repeated? can we move it somewhere where it can be reused?

examples/tutorial_with_permit2/erc20.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+/// Maximum storage slot position to probe when detecting balance/allowance slots.
+const MAX_PROBE_SLOT: u64 = 20;
+
+/// Find the storage slot position used by an ERC-20 token's `balances` mapping.
+///
+/// Probes slot positions 0..=[`MAX_PROBE_SLOT`] by applying a state override and calling
+/// `balanceOf`. Returns the first position whose override is reflected in the return value.
+pub async fn find_balance_slot(
```
dianacarvalho1, line 29: again I'm confused with our motivation to rewrite things that already exist in our codebase.. we already have this in tycho-ethereum. I know you don't want to bloat the dependencies here. but then I don't understand our motivation for making libraries in the first place if we will be rewriting a lot of things again in new repos

examples/tutorial/main.rs
```diff
@@ -1,1395 +1,276 @@
...
-    let solver_url = cli.solver_url.clone();
-    let tycho_url =
-        env::var("TYCHO_URL").unwrap_or_else(|_| "tycho-beta.propellerheads.xyz".to_string());
-    let tycho_api_key = env::var("TYCHO_API_KEY").ok();
-    let rpc_url = env::var("RPC_URL").ok();
-    let private_key = env::var("PRIVATE_KEY").ok();
-
-    // Create tycho-client RPC client
-    let rpc_options = HttpRPCClientOptions::new().with_auth_key(tycho_api_key.clone());
+    let rpc_url = env::var("RPC_URL").map_err(|_| "RPC_URL environment variable is required")?;
```
dianacarvalho1: let's use a public RPC by default please

examples/tutorial/main.rs
```diff
@@ -1,1395 +1,276 @@
...
-                    swap.component_id
-                );
-            }
+    println!("Status:       {:?}", quote.status());
+    println!("Amount in:    {}", quote.amount_in());
+    println!("Amount out:   {}", quote.amount_out());
+    println!("Gas estimate: {}", quote.gas_estimate());
+    if let Some(impact) = quote.price_impact_bps() {
+        println!("Price impact: {:.2}%", impact as f64 / 100.0);
+    }
```
dianacarvalho1: price impact is not even implemented yet. can we remove this? It might give the impression that it is implemented

examples/tutorial/main.rs
```diff
@@ -1,1395 +1,276 @@
...
+    println!("Gas cost (wei): {}", settled.gas_cost());
 
     Ok(())
 }
 
-/// Run eth_simulate without a wallet provider (for --sender without private key).
-#[allow(clippy::too_many_arguments)]
-async fn run_eth_simulation_no_wallet(
+/// Read the current `allowance(owner, spender)` from an ERC-20 token via `eth_call`.
+async fn read_erc20_allowance(
```
dianacarvalho1: this should be in the erc20 file

examples/tutorial_with_permit2/main.rs
```diff
@@ -0,0 +1,356 @@
+//! Permit2 Example: Quote and Execute a Swap using Permit2 Token Authorization
```
dianacarvalho1, line 1: let's called it `tutorial_with_permit2`? just `permit2` makes it sound it doesn't do the whole thing end to end

examples/tutorial/main.rs
```diff
@@ -1,1395 +1,276 @@
...
-//! 1. Call an already-running tycho-router solver for a swap quote
-//! 2. Display the route and pricing
-//! 3. Encode the swap via tycho-execution
-//! 4. Optionally simulate (eth_simulate or Tenderly) or execute on-chain
-
-mod types;
+//! 1. Build a FyndClient and check solver health
+//! 2. Request a swap quote with server-side calldata encoding
+//! 3. Display the route and pricing
+//! 4. Execute the swap — dry-run (default) or on-chain (--execute)
```
dianacarvalho1: using regular ERC20 approvals

dianacarvalho1: can we also make this docstring more coherent with the one in permit2 example?

examples/tutorial_with_permit2/permit2.rs
```diff
@@ -0,0 +1,44 @@
...
+    network::Ethereum,
+    primitives::{address, Address, Bytes as AlloyBytes, TxKind},
+    providers::{Provider, RootProvider},
+    rpc::types::TransactionRequest,
+    sol,
+    sol_types::SolCall,
+};
+
+/// Canonical Permit2 deployment address (same on all EVM chains).
+pub const PERMIT2_ADDRESS: Address = address!("000000000022D473030F116dDEE9F6B43aC78BA3");
```
dianacarvalho1, line 13: I believe all of this is already defined [here](https://github.com/propeller-heads/tycho-execution/blob/main/src/encoding/evm/approvals/permit2.rs)

examples/permit2/main.rs
```diff
@@ -0,0 +1,356 @@
...
+        Some(amount) => println!("Settled amount: {}", amount),
+        None => println!("Settled amount: (no matching Transfer log)"),
+    }
+    println!("Gas cost (wei): {}", settled.gas_cost());
+
+    Ok(())
+}
+
+/// Read the current `allowance(owner, spender)` from an ERC-20 token via `eth_call`.
+async fn read_erc20_allowance(
```
dianacarvalho1: this should be in erc20.rs

clients/rust/src/mapping.rs
```diff
@@ -43,13 +47,30 @@ fn tycho_to_bytes(addr: TychoAddress) -> bytes::Bytes {
...
 
+// ============================================================================
+// PRIMITIVE CONVERSIONS
+// ============================================================================
+
+/// Convert a [`num_bigint::BigUint`] to an [`alloy::primitives::U256`].
+///
+/// Serialises `n` as a big-endian byte string, right-pads into 32 bytes, and
+/// constructs a `U256` from the resulting fixed-width representation.
+pub(crate) fn biguint_to_u256(n: &num_bigint::BigUint) -> alloy::primitives::U256 {
```
dianacarvalho1, line 52: I think this is unnecessarily complex. just this works
````
    let bytes = n.to_bytes_be();
    U256::from_be_slice(&bytes)
````

dianacarvalho1 (commented):
Thank you @kayibal ! I have some comments..
I wonder if we really need two separate tutorials - one with regular approvals and one with permit2. Couldn't we do everything in the same tutorial and let the user decide?
Please update the gitbook docs with the new tutorial as well

examples/permit2/erc20.rs
kayibal: No unfortunately examples don't allow us to do that.

examples/tutorial_with_permit2/erc20.rs
kayibal: Yeah I think this is a a very small functionality that atm I feel it's not worth pulling in tycho-ethereum with all its other dependencies. It also serves only as an example. Honestly I find this example to complex and would prefer to actually hardcode the tokens and this way remove this whole directory. I decided against this since most likely BD wants to keep this example.

I think a better way here would be to implement a Fynd CLI binary. Here I would be fine to depend on tycho-ethereum for these things.

examples/tutorial_with_permit2/permit2.rs
kayibal: `tycho-execution` is not a dependency atm, I think we may want to re-export these constants on the client though.

kayibal:
@dianacarvalho1 

> I wonder if we really need two separate tutorials - one with regular approvals and one with permit2. Couldn't we do everything in the same tutorial and let the user decide?

In my opinion examples should be small pedagogic code snippets, ideally as short as possible, highlighting mainly how to use the library. Parsing user input and providing actual funtionality there is almost too much already, it distracts the reader from the main purpose. I know we violated this multiple times in other projects but trying to do better this time.

I think the Permit2 usage using Rust will be relevant to a much smaller user base (since Rust is more used server side -> bots, services, etc and therefore the approvals flow is more likely to be used).

> Please update the gitbook docs with the new tutorial as well

👍

examples/tutorial_with_permit2/permit2.rs
dianacarvalho1: no no, I meant the whole functionality is already in tycho-execution

kayibal: Both the address and the functionality are very unlikely to change so I am just going to keep this here for now.

dianacarvalho1 (approved):
Thank you @kayibal ! 🙏🏼

PR #93, opened 2026-03-12
feat: Comparison tool

tamaralipows:
Forgot to squash commits [ca3ac58](https://github.com/propeller-heads/fynd/pull/93/commits/ca3ac580127dbf9467c10063ef0dfffcb0eae29a) and [eabd07c](https://github.com/propeller-heads/fynd/pull/93/commits/eabd07cd9d9d75bb30e3a9e7f0461d84b65b7b4e) before merging. Sorry!


dianacarvalho1 (commented):
Thank you @tamaralipows ! I didn't give a thorough review though but overall looks good. 
1. Can we have this in the benchmark directory though? 
2. Do we really need the comparison_results.json to be commited?

kayibal:
@tamaralipows 

I didn't look at this correctly but let's please not have this in examples. I think this should be most likely a crate in the workspace. E.g. `tools/fynd-compare` or a integration test (although the CLI suggest it should be a separate tool).

examples/compare/main.rs
```diff
@@ -0,0 +1,504 @@
...
+    let round_trip_ms = start.elapsed().as_millis() as u64;
+
+    match result {
+        Ok(resp) => match resp.json::<Value>().await {
+            Ok(json) => (json, round_trip_ms),
+            Err(e) => (serde_json::json!({"_error": format!("parse error: {e}")}), round_trip_ms),
+        },
+        Err(e) => (serde_json::json!({"_error": format!("{e}")}), round_trip_ms),
+    }
+}
```
kayibal: Can we use the FyndClient here please instead of working with raw JSON values? Will probably make the whole file a lot shorter.

clients/rust/src/client.rs
```diff
@@ -532,7 +532,7 @@ where
         token_out: Bytes,
         receiver: Bytes,
     ) -> Result<Quote, FyndError> {
-        let url = format!("{}/v1/quote", self.base_url);
```
tamaralipows: I had to change this to quote to run the benchmarking. Should I keep this as quote and update the failing tests? I assumed this would be caught in the tutorial refactor PR anyway though

tamaralipows: updating this in my docs PR

tools/benchmark/compare.rs
```diff
@@ -0,0 +1,408 @@
...
+
+    /// Random seed for reproducibility
+    #[arg(long, default_value_t = 42)]
+    seed: u64,
+}
+
+#[derive(Debug, Serialize)]
+struct Metrics {
+    status: String,
+    amount_out: String,
```
tvinagre, line 57: Net gas or not? Can't we have both?

tools/benchmark/README.md
```diff
@@ -0,0 +1,174 @@
...
+**Prerequisites:** Both tools require a running solver. See the [Quickstart](../../docs/get-started/quickstart/README.md) for setup instructions.
+
+---
+
+## Benchmark
+
+Measures Fynd's performance with various parallelization strategies.
+
+```bash
+cargo run --example benchmark --release -- [OPTIONS]
```
tvinagre, line 14: outdated command I guess

tools/benchmark/README.md
```diff
@@ -0,0 +1,174 @@
...
+Console output shows real-time progress, summary statistics, and ASCII histograms of timing distributions. Results can optionally be exported to JSON.
+
+---
+
+## Compare
+
+Sends identical quote requests to two running Fynd instances and compares output quality (amount out, gas, routes).
+
+```bash
+cargo run --example compare --release -- [OPTIONS]
```
tvinagre, line 64: Same

tools/benchmark/requests.rs
```diff
@@ -0,0 +1,244 @@
...
+    Token { symbol: "UNI", address: "0x1f9840a85d5aF5bf1D1762F925BDADdC4201F984", decimals: 18 },
+    Token { symbol: "AAVE", address: "0x7Fc66500c84A76Ad7e9c93437bFc5Ac33E2DDaE9", decimals: 18 },
+    Token { symbol: "GNO", address: "0x6810e776880C02933D47DB1b9fc05908e5386b96", decimals: 18 },
+];
+
+// Pairs and sell amounts (in human units, each >= ~$1M)
+// Approximate prices: ETH~$2K, BTC~$65K, LINK~$15, UNI~$7,
+// AAVE~$200, GNO~$300, BNB~$600, PEPE~$0.000008, SHIB~$0.00001
+const PAIRS: &[Pair] = &[
+    Pair { token_in: "WETH", token_out: "USDC", amounts: &[500.0, 1000.0, 2500.0, 5000.0] },
```
tvinagre: Can't we use the `requests_set.json` here?
Or, move this to a JSON file (sorry, I'm a hater of configuration on the code)

tvinagre (approved):

tvinagre:
:) 
 🔈 🔈

tools/benchmark/README.md
tamaralipows: You mean the "example" part? I guess I can put it under [[bin]] in the Cargo.toml instead? We only have the following options
```
[[bin]], [[example]], [[test]], or [[bench]]
```
and bench is referring to "cargo bench / libtest benchmarks" so not exactly what we're doing

tamaralipows: nevermind...
```
  They can't be changed to [[bin]] without moving fynd-client, serde, alloy, fastrand, and several other crates from [dev-dependencies] into [dependencies] — which would bloat the main fynd binary's dependency
  tree. That's why they're registered as [[example]] even though they live in tools/.

  The commands cargo run --example benchmark and cargo run --example compare are correct. The --example flag is a Cargo artifact — there's no [[tool]] target type, and [[bin]] would require restructuring
  dependencies. You could reply to tvinagre explaining this, or alternatively these could be extracted into their own workspace crate (tools/benchmark/Cargo.toml) to get a proper cargo run -p fynd-benchmark
  invocation — but that's a bigger refactor.
```

tools/benchmark/compare.rs
tamaralipows: gonna specify this that it doesnt include gas but also add the gas computation separately

clients/rust/src/client.rs
tamaralipows: cherry-picked it onto this one so we can merge it to main without rushing the docs PR

PR #95, opened 2026-03-13
feat: improve error handling in computation manager

Troshchk:
Before this change, when derived data computation failed for a block, workers waiting for that data had no way to know: they'd hold incoming orders until their timeout expired (several seconds), then return a generic "timeout" error.                                                                                                                                                                                                                                                                   
After this change, a failed computation immediately notifies waiting workers, which return a COMPUTATION_FAILED error to the client. This lets clients distinguish "system is slow" from "computation has failed".                                                                                                                                                                                                                                                            
Partial failure pools failures are tracked until success. Pools whose spot price or depth calculation failed are included in routing with zero weight, keeping them in the graph but deprioritized

Code changes:
- Add ComputationFailed { computation_id, block } variant to DerivedDataEvent and emit it from all Err branches in ComputationManager::compute_all                                                                                                                       
 - Replace the single last_block: Option<u64> in DerivedData with a per-field ComputedValue<T> wrapper; add per-computation block getters; last_block() returns the max across all fields                                                                               
 - Add failed_for_block set to ReadinessTracker with is_blocked_for_current_block(): tracks require_fresh failures for the current block, cleared on NewBlock                                                                                                             
 - Add SolveError::ComputationFailed and use it in wait_until_ready for an immediate early exit instead of waiting out the timeout

### DIFF
--- a/fynd-core/src/algorithm/most_liquid.rs
+++ b/fynd-core/src/algorithm/most_liquid.rs
@@ -94,27 +94,27 @@ impl crate::graph::EdgeWeightFromSimAndDerived for DepthAndPrice {
     ) -> Option<Self> {
         let key = (component_id.clone(), token_in.address.clone(), token_out.address.clone());
 
-        // Use pre-computed spot price; fall back to zero-weight on failure.
+        // Use pre-computed spot price; skip edge if unavailable.
         let spot_price = match derived
             .spot_prices()
             .and_then(|p| p.get(&key).copied())
         {
             Some(p) => p,
             None => {
-                trace!(component_id = %component_id, "spot price failed, using zero weight");
-                return Some(Self { spot_price: 0.0, depth: 0.0 });
+                trace!(component_id = %component_id, "spot price not found, skipping edge");
+                return None;
             }
         };
 
-        // Look up pre-computed depth; fall back to zero-weight on failure.
+        // Look up pre-computed depth; skip edge if unavailable.
         let depth = match derived
             .pool_depths()
             .and_then(|d| d.get(&key))
         {
             Some(d) => d.to_f64().unwrap_or(0.0),
             None => {
-                trace!(component_id = %component_id, "pool depth failed, using zero weight");
-                return Some(Self { spot_price: 0.0, depth: 0.0 });
+                trace!(component_id = %component_id, "pool depth not found, skipping edge");
+                return None;
             }
         };
 
@@ -765,7 +765,7 @@ mod tests {
     }
 
     #[test]
-    fn test_from_sim_and_derived_failed_spot_price_returns_zero_weight() {
+    fn test_from_sim_and_derived_failed_spot_price_returns_none() {
         let key = pair_key("pool1", 0x01, 0x02);
         let key_str = pair_key_str("pool1", 0x01, 0x02);
         let tok_in = token(0x01, "A");
@@ -790,13 +790,11 @@ mod tests {
                 &sim, &key.0, &tok_in, &tok_out, &derived,
             );
 
-        let val = result.unwrap();
-        assert_eq!(val.spot_price, 0.0);
-        assert_eq!(val.depth, 0.0);
+        assert!(result.is_none());
     }
 
     #[test]
-    fn test_from_sim_and_derived_failed_pool_depth_returns_zero_weight() {
+    fn test_from_sim_and_derived_failed_pool_depth_returns_none() {
         let key = pair_key("pool1", 0x01, 0x02);
         let key_str = pair_key_str("pool1", 0x01, 0x02);
         let tok_in = token(0x01, "A");
@@ -824,13 +822,11 @@ mod tests {
                 &sim, &key.0, &tok_in, &tok_out, &derived,
             );
 
-        let val = result.unwrap();
-        assert_eq!(val.spot_price, 0.0);
-        assert_eq!(val.depth, 0.0);
+        assert!(result.is_none());
     }
 
     #[test]
-    fn test_from_sim_and_derived_both_failed_returns_zero_weight() {
+    fn test_from_sim_and_derived_both_failed_returns_none() {
         let key = pair_key("pool1", 0x01, 0x02);
         let key_str = pair_key_str("pool1", 0x01, 0x02);
         let tok_in = token(0x01, "A");
@@ -862,24 +858,7 @@ mod tests {
                 &sim, &key.0, &tok_in, &tok_out, &derived,
             );
 
-        let val = result.unwrap();
-        assert_eq!(val.spot_price, 0.0);
-        assert_eq!(val.depth, 0.0);
-    }
-
-    #[test]
-    fn test_try_score_path_with_zero_weight_edge_returns_zero() {
-        let (a, b, _, _) = addrs();
-        let mut m = linear_graph();
-
-        m.set_edge_weight(&"ab".to_string(), &a, &b, DepthAndPrice::new(0.0, 0.0), false)
-            .unwrap();
-
-        let graph = m.graph();
-        let paths = MostLiquidAlgorithm::find_paths(graph, &a, &b, 1, 1).unwrap();
-        assert_eq!(paths.len(), 1);
-        let score = MostLiquidAlgorithm::try_score_path(&paths[0]);
-        assert_eq!(score, Some(0.0));
+        assert!(result.is_none());
     }
 
     // ==================== find_paths Tests ====================
--- a/fynd-core/src/algorithm/most_liquid.rs
+++ b/fynd-core/src/algorithm/most_liquid.rs
@@ -94,39 +94,27 @@ impl crate::graph::EdgeWeightFromSimAndDerived for DepthAndPrice {
     ) -> Option<Self> {
         let key = (component_id.clone(), token_in.address.clone(), token_out.address.clone());
 
-        // Use pre-computed spot price; fall back to zero-weight for explicit failures.
+        // Use pre-computed spot price; fall back to zero-weight on failure.
         let spot_price = match derived
             .spot_prices()
             .and_then(|p| p.get(&key).copied())
         {
             Some(p) => p,
             None => {
-                if derived
-                    .spot_price_failure(&key)
-                    .is_some()
-                {
-                    trace!(component_id = %component_id, "spot price failed, using zero weight");
-                    return Some(Self { spot_price: 0.0, depth: 0.0 });
-                }
-                return None; // never computed
+                trace!(component_id = %component_id, "spot price failed, using zero weight");
+                return Some(Self { spot_price: 0.0, depth: 0.0 });
             }
         };
 
-        // Look up pre-computed depth; fall back to zero-weight for explicit failures.
+        // Look up pre-computed depth; fall back to zero-weight on failure.
         let depth = match derived
             .pool_depths()
             .and_then(|d| d.get(&key))
         {
             Some(d) => d.to_f64()?,
             None => {
-                if derived
-                    .pool_depth_failure(&key)
-                    .is_some()
-                {
-                    trace!(component_id = %component_id, "pool depth failed, using zero weight");
-                    return Some(Self { spot_price: 0.0, depth: 0.0 });
-                }
-                return None; // never computed
+                trace!(component_id = %component_id, "pool depth failed, using zero weight");
+                return Some(Self { spot_price: 0.0, depth: 0.0 });
             }
         };
 
@@ -837,25 +825,6 @@ mod tests {
         assert_eq!(val.depth, 0.0);
     }
 
-    #[test]
-    fn test_from_sim_and_derived_never_computed_returns_none() {
-        let key = pair_key("pool1", 0x01, 0x02);
-        let tok_in = token(0x01, "A");
-        let tok_out = token(0x02, "B");
-
-        let mut derived = DerivedData::new();
-        // both maps are empty (no successes, no failures)
-        derived.set_spot_prices(Default::default(), vec![], 10, true);
-        derived.set_pool_depths(Default::default(), vec![], 10, true);
-
-        let sim = make_mock_sim();
-        let result =
-            <DepthAndPrice as crate::graph::EdgeWeightFromSimAndDerived>::from_sim_and_derived(
-                &sim, &key.0, &tok_in, &tok_out, &derived,
-            );
-
-        assert!(result.is_none());
-    }
 
     #[test]
     fn test_from_sim_and_derived_both_failed_returns_zero_weight() {
--- a/fynd-core/src/algorithm/most_liquid.rs
+++ b/fynd-core/src/algorithm/most_liquid.rs
@@ -111,7 +111,7 @@ impl crate::graph::EdgeWeightFromSimAndDerived for DepthAndPrice {
             .pool_depths()
             .and_then(|d| d.get(&key))
         {
-            Some(d) => d.to_f64()?,
+            Some(d) => d.to_f64().unwrap_or(0.0),
             None => {
                 trace!(component_id = %component_id, "pool depth failed, using zero weight");
                 return Some(Self { spot_price: 0.0, depth: 0.0 });
@@ -777,7 +777,7 @@ mod tests {
             Default::default(),
             vec![FailedItem {
                 key: key_str,
-                error: FailedItemError::SpotPriceComputationFailed("sim error".into()),
+                error: FailedItemError::SimulationFailed("sim error".into()),
             }],
             10,
             true,
@@ -812,7 +812,7 @@ mod tests {
             Default::default(),
             vec![FailedItem {
                 key: key_str,
-                error: FailedItemError::PoolDepthComputationFailed("depth error".into()),
+                error: FailedItemError::SimulationFailed("depth error".into()),
             }],
             10,
             true,
@@ -841,7 +841,7 @@ mod tests {
             Default::default(),
             vec![FailedItem {
                 key: key_str.clone(),
-                error: FailedItemError::SpotPriceComputationFailed("spot error".into()),
+                error: FailedItemError::SimulationFailed("spot error".into()),
             }],
             10,
             true,
@@ -850,7 +850,7 @@ mod tests {
             Default::default(),
             vec![FailedItem {
                 key: key_str,
-                error: FailedItemError::PoolDepthComputationFailed("depth error".into()),
+                error: FailedItemError::SimulationFailed("depth error".into()),
             }],
             10,
             true,
--- a/fynd-core/src/algorithm/most_liquid.rs
+++ b/fynd-core/src/algorithm/most_liquid.rs
@@ -661,7 +661,7 @@ mod tests {
         }
 
         let mut derived_data = DerivedData::new();
-        derived_data.set_token_prices(token_prices, 1);
+        derived_data.set_token_prices(token_prices, vec![], 1, true);
         Arc::new(RwLock::new(derived_data))
     }
     // ==================== try_score_path Tests ====================
--- a/fynd-core/src/algorithm/most_liquid.rs
+++ b/fynd-core/src/algorithm/most_liquid.rs
@@ -777,7 +777,7 @@ mod tests {
             Default::default(),
             vec![FailedItem {
                 key: key_str,
-                error: FailedItemError::SpotPriceComputation("sim error".into()),
+                error: FailedItemError::SpotPriceComputationFailed("sim error".into()),
             }],
             10,
             true,
@@ -812,7 +812,7 @@ mod tests {
             Default::default(),
             vec![FailedItem {
                 key: key_str,
-                error: FailedItemError::PoolDepthComputation("depth error".into()),
+                error: FailedItemError::PoolDepthComputationFailed("depth error".into()),
             }],
             10,
             true,
@@ -841,7 +841,7 @@ mod tests {
             Default::default(),
             vec![FailedItem {
                 key: key_str.clone(),
-                error: FailedItemError::SpotPriceComputation("spot error".into()),
+                error: FailedItemError::SpotPriceComputationFailed("spot error".into()),
             }],
             10,
             true,
@@ -850,7 +850,7 @@ mod tests {
             Default::default(),
             vec![FailedItem {
                 key: key_str,
-                error: FailedItemError::PoolDepthComputation("depth error".into()),
+                error: FailedItemError::PoolDepthComputationFailed("depth error".into()),
             }],
             10,
             true,
--- a/fynd-core/src/algorithm/most_liquid.rs
+++ b/fynd-core/src/algorithm/most_liquid.rs
@@ -650,7 +650,7 @@ mod tests {
             fixtures::{addrs, diamond_graph, linear_graph, parallel_graph},
             market_read, order, setup_market, token, MockProtocolSim, ONE_ETH,
         },
-        derived::{types::TokenGasPrices, DerivedData},
+        derived::{computation::FailedItem, types::TokenGasPrices, DerivedData},
         graph::GraphManager,
         types::OrderSide,
     };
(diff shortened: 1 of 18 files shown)


fynd-core/src/derived/manager.rs
```diff
@@ -477,6 +513,94 @@ mod tests {
...
+        // Note: no update_gas_price() — gas price is intentionally absent
+        market.update_last_updated(BlockInfo::new(10, "0xhash".into(), 0));
+        market.upsert_components(std::iter::once(pool));
+        market.update_states([("pool".to_string(), Box::new(MockProtocolSim::new(2000.0)) as _)]);
+        market.upsert_tokens([eth, usdc]);
+        Arc::new(RwLock::new(market))
+    }
+
+    #[tokio::test]
+    async fn spot_price_failure_broadcasts_computation_failed() {
```
dianacarvalho1: test_spot_price...

fynd-core/src/derived/tracker.rs
```diff
@@ -399,6 +438,89 @@ mod tests {
         assert!(tracker.is_ready()); // Both satisfied again
     }
 
+    // ==================== ComputationFailed / is_blocked_for_current_block Tests
+    // ====================
+
+    #[test]
+    fn is_blocked_when_require_fresh_computation_fails_for_current_block() {
```
dianacarvalho1: test_is_blocked_...

dianacarvalho1 (commented):
Thank you @Troshchk ! Tbh it's quite hard for me to give a meaningful review here because I'm not yet so familiar with these gritty details. Can you explain what this changes in a general Fynd run please? Not the code that was changed but how this changes functionality

fynd-core/src/derived/manager.rs
```diff
@@ -280,6 +280,12 @@ impl ComputationManager {
             }
             Err(e) => {
                 warn!(error = ?e, elapsed_ms = spot_elapsed.as_millis(), "spot price computation failed");
+                let _ = self
+                    .event_tx
+                    .send(DerivedDataEvent::ComputationFailed {
+                        computation_id: SpotPriceComputation::ID,
+                        block,
+                    });
```
louise-poole, line 320: Things are a bit more complex than just this no? Like the spot price computation only returns an error if ALL spot prices fail to compute. Should we be aware of partial failures maybe? Not sure how to approach it...

fynd-core/src/derived/store.rs
```diff
@@ -34,9 +39,20 @@ impl DerivedData {
         Arc::new(RwLock::new(Self::new()))
     }
 
-    /// Returns the block number at which data was last computed.
+    /// Returns the block number at which any data was last computed.
+    ///
+    /// Returns the maximum block across all computed values, so it returns `Some` as long as
+    /// any computation has run. Used by health checks.
     pub fn last_block(&self) -> Option<u64> {
```
louise-poole: I wonder if this is not more informative to get the oldest block we're bound by? Like if I have a restriction on my algorithm that I need current data only, if I only see the max block then I'd assume all derived data is in the max block and is recent. But this might not be the case.

fynd-core/src/derived/tracker.rs
```diff
@@ -399,6 +438,89 @@ mod tests {
         assert!(tracker.is_ready()); // Both satisfied again
     }
 
+    // ==================== ComputationFailed / is_blocked_for_current_block Tests
+    // ====================
```
louise-poole: This comment wrapped to 2 lines looks a bit weird. We can probably remove it.

fynd-core/src/derived/tracker.rs
```diff
@@ -399,6 +438,89 @@ mod tests {
...
+
+    #[test]
+    fn test_is_blocked_when_require_fresh_computation_fails_for_current_block() {
```
louise-poole: I find super long and verbose test names not so nice, maybe just this is a bit nicer:
```suggestion
    fn test_require_fresh_computation_fails_for_current_block() {
```

fynd-core/src/derived/tracker.rs
```diff
@@ -399,6 +438,89 @@ mod tests {
...
+        tracker.handle_event(&DerivedDataEvent::ComputationFailed {
+            computation_id: "spot_prices",
+            block: 100,
+        });
+
+        assert!(tracker.is_blocked_for_current_block());
+    }
+
+    #[test]
+    fn is_not_blocked_for_allow_stale_computation_failure() {
```
louise-poole: All tests should start with 'test_'. Also I feel like a lot of these names can easily be shortened without losing what is being tested. Can you revisit all the tests?
```suggestion
    fn test_allow_stale_computation_failure() {
```

fynd-core/src/derived/store.rs
```diff
@@ -105,13 +145,21 @@ impl DerivedData {
...
+        self.spot_prices
+            .as_ref()
+            .map(|v| &v.data)
+    }
+
+    /// Returns the block at which spot prices were last computed.
+    pub fn spot_prices_block(&self) -> Option<u64> {
+        self.spot_prices
+            .as_ref()
+            .map(|v| v.block)
```
tamaralipows, line 245: Quick question for my own understanding - why do we prefer to add an extra method for this instead of just returning all the spot prices and using the block from that return?

tvinagre (commented):
I believe this task to be a little more complex - I think @louise-poole exemplified this well. 
To do a proper error handling, first, derived computation error handling might need to be improved to distinguish between partial failures and total failures - and include information on which tokens / pools failed and Error reasons. 

With this info, we can treat failures individually and handle the errors better on the computation manager.
 
This is essential, for example, for `ComputationRequirements`, so we can start differentiating between a stale price on a token due to failure on computing the data for this token (only), or stale price because the price didn't actually move.

fynd-core/src/derived/store.rs
Troshchk: I’m wondering if this function might not be needed anymore. Each computation now records the block it was last computed for.

Since we now have dedicated getters (`token_prices_block`, `token_prices_deps_block`, `pool_depths_block`, `spot_prices_block`), it might make more sense to rely on those to verify that the specific data is current instead of using a single shared `last_block()` value. What do you think?

Currently `last_block()` is only used by the health tracker, which just checks `is_some()` to confirm that derived data has been computed at least once.

louise-poole: Yeah - let's refactor this then to better reflect where it's used: it should check all derived data has been computed at least once.

fynd-core/src/derived/store.rs
Troshchk: I think this approach is cleaner for derived_data_ready(), since it avoids unnecessary tuple destructuring. Happy to adjust if you see it differently

fynd-core/src/derived/store.rs
```diff
@@ -21,10 +28,33 @@ struct ComputedValue<T> {
 #[derive(Debug, Default)]
 pub struct DerivedData {
     token_prices: Option<ComputedValue<TokenGasPrices>>,
+    /// Failure map for token_prices: key → (block_when_failed, error_reason).
+    /// Persists across incremental runs — entries survive until the key is re-attempted
+    /// and succeeds. Full recomputes replace the map entirely.
+    token_prices_failed: Option<HashMap<TokenGasPriceKey, (u64, String)>>,
```
louise-poole: Do these need to be options? They can just be empty hashmaps when they have no failures? Avoids us having to unwrap it whenever we read/update it.

fynd-core/src/algorithm/most_liquid.rs
```diff
@@ -95,16 +95,41 @@ impl crate::graph::EdgeWeightFromSimAndDerived for DepthAndPrice {
...
-        // Look up pre-computed depth
-        let depth = derived
+        // Look up pre-computed depth; fall back to zero-weight for explicit failures.
+        let depth = match derived
             .pool_depths()
-            .and_then(|depths| depths.get(&key))?
-            .to_f64()?;
+            .and_then(|d| d.get(&key))
+        {
+            Some(d) => d.to_f64()?,
```
louise-poole: What happens to failures here? They should probably also result in us defaulting to 0 weights.

fynd-core/src/algorithm/most_liquid.rs
```diff
@@ -762,6 +787,156 @@ mod tests {
...
+        let key = pair_key("pool1", 0x01, 0x02);
+        let key_str = pair_key_str("pool1", 0x01, 0x02);
+        let tok_in = token(0x01, "A");
+        let tok_out = token(0x02, "B");
+
+        let mut derived = DerivedData::new();
+        // spot price fails, pool depth not computed
+        derived.set_spot_prices(
+            Default::default(),
+            vec![crate::derived::computation::FailedItem {
```
louise-poole: You import this repeatedly in the test mod. Why not import it once at the start of the mod?

fynd-core/src/algorithm/most_liquid.rs
Troshchk: Both cases(a failed computation and an absent pool) fall into the same None arm and return 0 weights. Or do you mean something else?

tvinagre: What's the motivation to default to 0.0 instead of returning an error?

fynd-core/src/derived/computations/pool_depth.rs
```diff
@@ -683,4 +719,62 @@ mod tests {
             );
         }
     }
+
+    #[tokio::test]
+    async fn test_compute_partial_failure_missing_spot_price() {
```
tvinagre, line 731: Should we add more test cases or they cover all the failure scenarios?

fynd-core/src/derived/computation.rs
```diff
@@ -99,6 +99,41 @@ impl ComputationRequirements {
...
 
+/// A single item that failed during a computation.
+#[derive(Debug, Clone)]
+pub struct FailedItem {
+    /// Human-readable key for the failed item.
+    /// - spot_prices/pool_depths: "component_id/token_in/token_out"
+    /// - token_prices: "token_address"
+    pub key: String,
+    /// Display string of the error.
+    pub error: String,
```
tvinagre: Ideally, errors here should be custom enums for each error type

fynd-core/src/derived/manager.rs
```diff
@@ -264,22 +264,45 @@ impl ComputationManager {
...
+            Ok(output) => {
+                let count = output.data.len();
+                if output.has_failures() {
+                    warn!(
+                        count,
+                        failed = output.failed_items.len(),
+                        "spot prices partial failures"
+                    );
+                    for item in &output.failed_items {
+                        debug!(key = %item.key, error = %item.error, "spot price failed item");
```
tvinagre, line 291: Let's keep an eye if this doesn't become too repetitive and verbose. If it does, let's downgrade to `trace`?

fynd-core/src/derived/store.rs
```diff
@@ -1,25 +1,60 @@
...
+}
+
 /// Typed storage for derived data computations.
 ///
 /// Provides typed access to previously computed derived data.
 /// Each field is `Option` to indicate whether the computation has run.
 #[derive(Debug, Default)]
 pub struct DerivedData {
-    token_prices: Option<TokenGasPrices>,
+    token_prices: Option<ComputedValue<TokenGasPrices>>,
```
tvinagre, line 30: So now, we don't allow for failures, right?

Let's say, on
> block 1, I calculate the price for USDC. Save it on token_prices
> block 2, I fail to cal the price for USDC - remove it from token_prices and add them to token_prices_failed

Is that correct?
Another option (way more complex) would to set, individually, the number of the block where it was last updated - and if there's a failure on the later block, it'd get added to token_prices_failed and the algorithm could choose whether to use outdated data or not.
But I think it then becomes too complex, so it's just and idea for the future.

fynd-core/src/derived/store.rs
```diff
@@ -1,25 +1,60 @@
...
+    /// Failure map for token_prices: key → (block_when_failed, error_reason).
+    /// Persists across incremental runs — entries survive until the key is re-attempted
```
tvinagre: Also, these comments are quite repetitive - can you compress them pls?

fynd-core/src/derived/store.rs
```diff
@@ -45,18 +83,63 @@ impl DerivedData {
...
+        if is_full_recompute {
+            self.token_prices_failed = new_failures;
+        } else {
+            self.token_prices_failed
+                .retain(|k, _| !prices.contains_key(k));
+            self.token_prices_failed
+                .extend(new_failures);
+        }
+
+        self.token_prices = Some(ComputedValue { data: prices, block });
```
tvinagre, line 122: if u haven't - it's worth double-checking if `prices` always is made of:
last prices (b0) + updated prices (b1) - failed prices (b1)

tvinagre (commented):
Looks good! Just some small comments and sanity checks (since this is quite an important feature)
Amazing work @Troshchk !

fynd-core/src/algorithm/most_liquid.rs
Troshchk: To remove a component from the graph if the calculation is a strict requirement.

Now, if `from_sim_and_derived` would return None, the edge weight will not be updated in `update_edge_weights_with_derived` and stale data from the previous calculation would be used.
By setting the weight to 0 when a calculation fails we ensure such pools are excluded from graph.

fynd-core/src/derived/store.rs
Troshchk: That's correct 👍 
I will make a ticket for it

Troshchk: Noted [here](https://propeller-heads.atlassian.net/wiki/spaces/GEN/pages/3469148161/Fynd+Release+2)

fynd-core/src/derived/store.rs
Troshchk: You mean in the test or in the code?

fynd-core/src/derived/manager.rs
Troshchk: Ok, since it is only for the failed items, it shouldn't be, but will keep an eye

fynd-core/src/algorithm/most_liquid.rs
louise-poole: I'm talking about the conversion error here: `d.to_f64()?` This is in the Some arm, not the None arm...

fynd-core/src/algorithm/most_liquid.rs
```diff
@@ -830,7 +837,10 @@ mod tests {
         // pool depth fails
         derived.set_pool_depths(
             Default::default(),
-            vec![FailedItem { key: key_str, error: "depth error".to_string() }],
+            vec![FailedItem {
+                key: key_str,
+                error: FailedItemError::PoolDepthComputation("depth error".into()),
```
louise-poole: The error string isn't adding any info though? We already know it's a depth error from the error enum... can we add any info here or remove the need for a string?

fynd-core/src/derived/computation.rs
```diff
@@ -99,15 +99,43 @@ impl ComputationRequirements {
...
+    MissingSpotPrice,
+
+    #[error("extreme decimal mismatch ({from}\u{2192}{to})")]
+    ExtremeDecimalMismatch { from: u32, to: u32 },
+
+    #[error("spot price too small: {0}")]
+    SpotPriceTooSmall(f64),
+
+    #[error("spot price computation failed: {0}")]
+    SpotPriceComputation(String),
```
louise-poole: This doesn't sound like an error though... Maybe call it SpotPriceFailed or something similar?

fynd-core/src/derived/computation.rs
```diff
@@ -99,15 +99,43 @@ impl ComputationRequirements {
...
+
+    #[error("pool depth computation failed: {0}")]
+    PoolDepthComputation(String),
```
louise-poole: Honestly I think this might be too generic a name. Like if I look up a depth and get an error with it that sas PoolDepthComputation - it's not really telling me much. These errors should rather be more informative on what failed? Was it simulation? Was it decoding? Etc. The ExtremeDecimalMismatch and ExtremeDecimalMismatch are good!

fynd-core/src/algorithm/most_liquid.rs
Troshchk: `d.to_f64()?`  always returns Some, in case of overflow it returns Some(f64::INFINITY). 
Would you expect something else to be returned for a big value of `d`?

fynd-core/src/algorithm/most_liquid.rs
Troshchk: The error indeed does not contain any info in the test, but it does add more info for the failures in the code, please see [here](https://github.com/propeller-heads/fynd/pull/95/changes#diff-633096af02f5414e9f45f6ed44114cb07075c6657ad57b2bd882a9b9f53db52eR317)

fynd-core/src/derived/computation.rs
Troshchk: It contains more info in the string, see [here for pool depth](https://github.com/propeller-heads/fynd/pull/95/changes#diff-633096af02f5414e9f45f6ed44114cb07075c6657ad57b2bd882a9b9f53db52eR317-R318) and [here for spot price](https://github.com/propeller-heads/fynd/pull/95/changes#diff-43aea46edffad938d4ddc2b2efe063fc40c3582ea9ff86762ca5b6fb8be955feR151), do you think it is still worth making it more explicit?

louise-poole: Yes - I would say it's worth it. It's generally good rust practice to have well defined Error enums that aren't unnecessarily coupled with where they are reported. For example, the spot price errors are simply SimulationErrors, which is an error type that might one day appear in a different computation that is added (say we add one to compute Fees). If we simply have `SimulationError` here, then the new computation can return the same error type (and potentially be handled in the same way down the line). As it is now, I'll need to add a new error type every time I add a new Computation to the derived computation module.

fynd-core/src/algorithm/most_liquid.rs
louise-poole: Hmmm... I see. The `?` implies we are handling a failure case that you state will never occur. I think this would mislead future coders that read this code. I would rather we explicitly handle this (even if it will never trigger) with an `.unwrap_or(0.0)` or something?

Troshchk: Understood and fixed 👍

fynd-core/src/derived/computation.rs
Troshchk: Makes sense, thank you for the explanation! Changed the error type

fynd-core/src/derived/store.rs
Troshchk: @tvinagre could you explain what you mentioned here, please?

fynd-core/src/worker_pool/worker.rs
```diff
@@ -323,6 +324,13 @@ where
                     )));
                 }
                 _ = notified => {
+                    // Check if any require_fresh computation permanently failed this block
```
tvinagre, line 338: There's an edge case here that is not covered - I've asked Claude to illustrate it:

The `is_blocked_for_current_block()` check only executes after a notification fires:

```rust
_ = notified => {
    if self.readiness_tracker.is_blocked_for_current_block() {
        return Err(SolveError::ComputationFailed(...));
    }
    continue;
}
```

But if the `ComputationFailed` event was already processed by the `run()` loop before the solve task arrived (which is the common case because biased select! prioritizes derived events over solve tasks), the tracker is already in a blocked state but no new notification will ever come. 

The flow:
1. Derived event branch processes ComputationFailed → readiness_tracker.handle_event() → ready_notify.notify_waiters()
2. No one is listening to notified yet (no solve task in progress)
3. Solve task arrives → wait_until_ready() starts
4. Checks is_ready() → false
5. Creates notified future and enters select!
6. Waits forever for a notification that already happened → times out with SolveError::NotReady instead of ComputationFailed

The fix is simple I think: 

Add the blocked check at the top of each loop iteration (before select!), not just after being notified. This covers both the pre-processed case and the race where the notification arrives between loop iterations.

The test `wait_until_ready_returns_immediately_on_blocked_state` passes because it manually spawns a tokio::task to fire notify_waiters() after 20ms — this wouldn't happen in production where the notifier is in the same select! loop.

Please, also add a test for that.

Finally, this case affects 2 scenarios:
1. if we have `required_fresh` computations (which we don't rn, so this issue is minor)
2.if it fails on startup, which can happen quite often.

It's not super critical though, so I'm fine if you want to address in another PR (but pls address and add the test), but it's quite simple so maybe worth doing this right away

fynd-rpc/src/api/error.rs
tvinagre, line 43: Error handling here is missing. ComputationFailed would go on the fallback, and return a 422 that indicates a client error - means "your request was well-formed but semantically invalid.
ComputationFailed needs to be handled here and return a server error, probably `SERVICE_UNAVAILABLE`

fynd-core/src/derived/manager.rs
```diff
@@ -264,22 +264,45 @@ impl ComputationManager {
...
                     .event_tx
                     .send(DerivedDataEvent::ComputationComplete {
                         computation_id: SpotPriceComputation::ID,
                         block,
+                        failed_items: output.failed_items,
                     });
             }
             Err(e) => {
                 warn!(error = ?e, elapsed_ms = spot_elapsed.as_millis(), "spot price computation failed");
+                let _ = self
```
tvinagre, line 315: Shouldn't we broadcast the other errors as well before returning?

```
let _ = self.event_tx.send(DerivedDataEvent::ComputationFailed {
    computation_id: TokenGasPriceComputation::ID,
    block,
});
let _ = self.event_tx.send(DerivedDataEvent::ComputationFailed {
    computation_id: PoolDepthComputation::ID,
    block,
});
return;
```

Minor issue for the current use-case, but solvers should be aware that computation has also failed for the subsequent comps bc of the `spot_prices` failure

fynd-core/src/derived/store.rs
tvinagre: In the code, it was just a warning for you to check if `token_gas_prices` (and the other computations) are doing the following:
`result = existing_prices - failed_recomputed_tokens + succeeded_recomputed_token` for every block.

I've checked and they do, so nothing to worry about :)

fynd-core/src/algorithm/most_liquid.rs
```diff
@@ -95,16 +95,29 @@ impl crate::graph::EdgeWeightFromSimAndDerived for DepthAndPrice {
...
+        // Use pre-computed spot price; fall back to zero-weight on failure.
+        let spot_price = match derived
             .spot_prices()
-            .and_then(|prices| prices.get(&key).copied())?;
+            .and_then(|p| p.get(&key).copied())
+        {
+            Some(p) => p,
+            None => {
+                trace!(component_id = %component_id, "spot price failed, using zero weight");
+                return Some(Self { spot_price: 0.0, depth: 0.0 });
```
tvinagre: I'm still not fully convinced of this behavior bc previously, a missing spot price returned `None` → the pool had no edge in the graph. Now it returns Some(DepthAndPrice { 0.0, 0.0 }) → the pool is in the graph with zero weight.

If `ALL` pools for a token pair fail (say DAI -> USDC), the algorithm would return a path with score 0.0 instead of "no route found". This path would be sent to simulation with unknown pricing (and most likely will fail on simulation, leading to useless computation). 

The same will happen with the bottom case. I don't think that's dangerous, but is this the behavior we want? What's the benefit of setting to 0?

tvinagre (approved):
Approved with some comments :)

fynd-core/src/derived/store.rs
Troshchk: Thank you for checking it!

fynd-core/src/algorithm/most_liquid.rs
Troshchk: I didn't think about the edge case, that all pools would fail. I think in this case it is indeed better to set the failed one to None, changing it.

fynd-core/src/worker_pool/worker.rs
Troshchk: Thank you for the explanation, it is very clear. I have added this check

tvinagre (approved):

PR #98, opened 2026-03-16
feat: handle gas price RPC errors instead of panicking

tanayjain14:
## Summary
- Replace `.unwrap()` on `get_latest_fee_price()` with error handling that
  warns and continues instead of crashing the solver
- On RPC failure (timeout, rate limit), the gas fetcher logs a warning,
  increments `gas_price_fetch_failures_total` metric, acks the oneshot to
  prevent TychoFeed from hanging, and retries on the next block signal
- Solver continues with the last known gas price (or `None` if first fetch
  hasn't succeeded yet)

## Context
The quickstart defaults to `llamarpc`, which rate-limits
under aggregate RPC load. Gas price fetch timeouts cause a panic in
`gas.rs:39`, killing the gas worker task, which shuts down the entire solver.

## Follow-up recommendations for separate PR
1. Add `gas_price_age_ms` to `GET /v1/health` response
2. Return 503 if gas price is stale >120s

### DIFF
--- a/fynd-core/src/feed/gas.rs
+++ b/fynd-core/src/feed/gas.rs
@@ -1,5 +1,5 @@
 use chrono::Utc;
-use metrics::gauge;
+use metrics::{counter, gauge};
 use tokio::sync::{mpsc, oneshot};
 use tracing::warn;
 use tycho_simulation::{tycho_core::traits::FeePriceGetter, tycho_ethereum::gas::BlockGasPrice};
@@ -32,11 +32,17 @@ impl<C: FeePriceGetter<FeePrice = BlockGasPrice>> GasPriceFetcher<C> {
                 .await
                 .ok_or(DataFeedError::GasPriceFetcherError("Trigger channel closed".to_string()))?;
 
-            let fee_price = self
-                .client
-                .get_latest_fee_price()
-                .await
-                .unwrap();
+            let fee_price = match self.client.get_latest_fee_price().await {
+                Ok(price) => price,
+                Err(e) => {
+                    counter!("gas_price_fetch_failures_total").increment(1);
+                    warn!(error = ?e, "Failed to fetch gas price, skipping update");
+                    if update_tx.send(()).is_err() {
+                        warn!("Failed to send update notification");
+                    }
+                    continue;
+                }
+            };
 
             {
                 let mut lock = self.shared_market_data.write().await;
--- a/fynd-core/src/feed/gas.rs
+++ b/fynd-core/src/feed/gas.rs
@@ -36,7 +36,7 @@ impl<C: FeePriceGetter<FeePrice = BlockGasPrice>> GasPriceFetcher<C> {
                 Ok(price) => price,
                 Err(e) => {
                     counter!("gas_price_fetch_failures_total").increment(1);
-                    warn!(error = ?e, "Failed to fetch gas price, skipping update");
+                    warn!(error = ?e, "Failed to fetch gas price, skipping update. Configure --gas-price-stale-threshold-secs to surface this in health checks");
                     if update_tx.send(()).is_err() {
                         warn!("Failed to send update notification");
                     }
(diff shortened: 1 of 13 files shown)


tanayjain14:
I have two regression tests for the error path that I can add if you'd like:

1. **`fetch_error_does_not_crash_and_acks_oneshot`** — Simulates a single RPC failure. Verifies the fetcher doesn't panic, the oneshot ack is still sent (preventing TychoFeed hangs), gas price stays `None`, and the next successful fetch updates it normally.
2. **`persistent_failure_keeps_loop_alive`** — Simulates 3 consecutive RPC failures. Verifies the loop survives all of them and recovers on the 4th call.

Both use a mock `FeePriceGetter` — no network needed. Happy to include them in a follow-up commit.

tanayjain14:
## Follow-up commits pushed (3 new commits on this branch)

Per feedback, the follow-up items from the original PR description are now implemented on this same branch:

### 1. `a427147` — Regression tests + warn message hint
* Two tests for `GasPriceFetcher` error resilience:   
  * `fetch_error_does_not_crash_and_acks_oneshot` — 1 failure then recovery
  * `persistent_failure_keeps_loop_alive` — 3 consecutive failures then recovery
* Updated the fetch failure warning to hint at `--gas-price-stale-threshold-secs`

### 2. `999b7f4` — `gas_price_age_ms` in health response
* New `gas_price_age_ms: Option<u64>` field on `GET /v1/health`
* Measures time since the block that produced the last gas price (block freshness, not fetch recency)
* Omitted from JSON when gas price hasn't been fetched yet (`serde(skip_serializing_if)`)
* Updated both Rust and TypeScript SDK clients + regenerated OpenAPI spec and TS autogen types
* Backwards-compatible: `#[serde(default)]` on deserialization, omitted when `None`

### 3. `b3f274a` — Optional 503 on stale gas price
* New `--gas-price-stale-threshold-secs` CLI flag (disabled by default)
* When set, health returns 503 if `gas_price_age_ms > threshold`
* **Startup grace period**: missing gas price during the first `threshold` seconds is not considered stale (avoids false 503s before the first fetch completes)
* After the grace period, missing gas price = stale (RPC genuinely broken)

fynd-core/src/feed/gas.rs
```diff
@@ -32,11 +32,17 @@ impl<C: FeePriceGetter<FeePrice = BlockGasPrice>> GasPriceFetcher<C> {
...
-                .unwrap();
+            let fee_price = match self.client.get_latest_fee_price().await {
+                Ok(price) => price,
+                Err(e) => {
+                    counter!("gas_price_fetch_failures_total").increment(1);
+                    warn!(error = ?e, "Failed to fetch gas price, skipping update. Configure --gas-price-stale-threshold-secs to surface this in health checks");
+                    if update_tx.send(()).is_err() {
+                        warn!("Failed to send update notification");
+                    }
+                    continue;
```
kayibal, line 43: It looks like the main run loop handles retries here? I think it would be much cleaner if this was handled within the `client.get_latest_fee_price`

tanayjain14: `EthereumRpcClient::get_gas_price()` already retries internally via `retry_policy.retry_request()` (3 retries, exponential backoff 100ms→5s). By the time `Err` reaches this match, the client has exhausted all retries.

The loop isn't retrying - it's gracefully skipping the update and waiting for the next block signal. The oneshot ack on L40-42 must stay in the loop since it's coordination between `TychoFeed` and `GasPriceFetcher` (prevents `TychoFeed` from hanging).

So only the `counter!` + `warn!` (L38-39) could move into the client. Are you suggesting we wrap the client to relocate those 2 lines, or did you have a different restructuring in mind?

kayibal: OK so what happens if the solvers have no gas price? I agree panicking is pretty bad... Maybe we need to adjust the retry policy in that case?

kayibal (approved):

PR #100, opened 2026-03-16
feat: update defaults to match fynd-basic plan restrictions

zizou0x:
### Summary
- fix: connect min_token_quality CLI arg to builder — the --min-token-quality flag was accepted but silently ignored
- feat: add traded_n_days_ago parameter — new --traded-n-days-ago flag (default: 3) passed through CLI → builder → feed → load_all_tokens
- fix: apply TVL buffer as lower bound for hysteresis — the buffer ratio now defines the remove threshold (min_tvl / ratio), so min_tvl is correctly sent as tvl_gt to the indexer
- refactor: rename tvl_buffer_multiplier to tvl_buffer_ratio — better reflects its role as a hysteresis ratio
- feat: add chain-specific defaults for tycho_url — --tycho-url is now optional; defaults resolve per chain to the fynd indexer endpoints (e.g. tycho-fynd-ethereum.propellerheads.xyz)
- With these changes, a user with a fynd-basic plan key can run fynd serve --tycho-api-key <key> with no extra args and get a working solver.


docs/get-started/quickstart/executing-the-solutions.md
```diff
@@ -126,7 +120,7 @@ cargo run --example tutorial -- \
 | `--sell-token`    | USDC address          | Token address to sell                                                |
 | `--buy-token`     | WETH address          | Token address to buy                                                 |
 | `--sell-amount`   | 10.0                  | Amount to sell (human readable)                                      |
-| `--chain`         | ethereum              | Blockchain (currently only ethereum is available)                    |
+| `--chain`         | ethereum              | Blockchain (ethereum, base, unichain)                                |
```
louise-poole, line 125: I don't think we should have this hardcoded in the docs here. We'll likely forget to update it when we release new chains...

docs/README.md
```diff
@@ -46,9 +46,8 @@ Fynd puts you in control:
 ### Supported Chains
 
 * Ethereum Mainnet
-* **Coming soon:**
-  * Base&#x20;
-  * Unichain
+* Base
+* Unichain
```
louise-poole, line 51: Same comment here: I think we should probably remove this from here.

fynd-rpc/src/builder.rs
```diff
@@ -109,9 +111,15 @@ impl FyndBuilder {
...
-    /// Sets the TVL buffer multiplier (default: 1.1).
-    pub fn tvl_buffer_multiplier(mut self, multiplier: f64) -> Self {
-        self.tvl_buffer_multiplier = multiplier;
+    /// Sets the traded_n_days_ago filter (default: 3).
+    pub fn traded_n_days_ago(mut self, days: u64) -> Self {
+        self.traded_n_days_ago = days;
+        self
+    }
+
+    /// Sets the TVL buffer ratio (default: 1.1).
```
louise-poole: Since this is a user-facing fn I think it could have as much info as the `tvl_buffer_ratio` property has:
```suggestion
    /// Sets the ratio used to define the lower bound of the TVL filter for hysteresis (default: 1.1).
```

fynd-rpc/src/builder.rs
```diff
@@ -109,9 +111,15 @@ impl FyndBuilder {
         self
     }
 
-    /// Sets the TVL buffer multiplier (default: 1.1).
-    pub fn tvl_buffer_multiplier(mut self, multiplier: f64) -> Self {
-        self.tvl_buffer_multiplier = multiplier;
+    /// Sets the traded_n_days_ago filter (default: 3).
```
louise-poole: We should clarify this filter applies to tokens and not pools

louise-poole (approved):
A few minor comments, otherwise looks good!

docs/get-started/quickstart/executing-the-solutions.md
zizou0x: I tend to agree but wouldn't include it here as some people might disagree (this is just an in-place update of what was already here).

I would prefer to have this in the last docs polishing

docs/README.md
zizou0x: Same comment as above, I would prefer to keep docs changes just as an update and refactor them in the later PR

PR #102, opened 2026-03-16
feat(benchmark): parallel requests, net-of-gas, timing & route depth comparison

Haikane:
## Summary

Improves the compare subcommand based on learnings from the BF vs ML benchmarking (PR #43):

- **Parallel requests**: Send quotes to both solvers via `tokio::join!` so they solve against the same market state (previously sequential, causing state drift)
- **Server-side net-of-gas comparison**: Uses the solver's `amount_out_net_gas` field directly instead of client-side gas price approximation. This works for all token pairs, not just WETH-paired trades. The field was already in the API response but was dropped by the Rust client SDK during DTO-to-Quote mapping; this PR fixes the SDK to expose it.
- **Solve time comparison**: p50/p95/p99 percentiles with relative speed summary
- **Route depth comparison**: Average and max swaps per solver
- **Win-rate framing**: Head-to-head summary shows win/loss/tie counts with percentages instead of just avg bps diff
- **Default -n raised to 500**: 100 requests yielded too few contested trades for statistical significance
- **Real aggregator trades**: Embedded 50-trade sample from Dune Analytics for zero-config use; `download-trades` subcommand fetches the full 10k dataset from GitHub Releases

### Changes

**`fynd-client` (Rust SDK)**
- Added `amount_out_net_gas` field to `Quote` type with public getter
- Fixed DTO-to-Quote mapping to preserve `amount_out_net_gas` instead of dropping it

**`fynd-benchmark`**
- Removed `--rpc-url` flag and client-side gas price estimation logic
- Compare tool now uses server-computed `amount_out_net_gas` for net-of-gas diff
- Default request source changed from random pair generation to embedded aggregator trades
- Added `download-trades` subcommand to fetch full 10k dataset from GitHub Releases
- Deleted `trades_1k_requests.json` (replaced by embedded sample + downloadable full set)
- Updated README and docs

## Test plan

- [x] `cargo build -p fynd-benchmark` passes
- [x] `cargo test -p fynd-benchmark` passes (63 tests)
- [x] `cargo clippy -p fynd-benchmark -- -D warnings` clean
- [x] `cargo test -p fynd-client` passes (74 tests)
- [x] `cargo clippy -p fynd-client -- -D warnings` clean
- [x] GitHub Release `benchmark-data-v1` created with `aggregator_trades_10k.json`
- [ ] Manual test with two live solvers


fynd-core/src/feed/protocol_registry.rs
```diff
@@ -130,7 +130,7 @@ pub(crate) fn register_exchanges(
                 continue;
             }
             _ => {
-                return Err(DataFeedError::Config(format!("Unknown protocol: {}", protocol)));
+                warn!("Skipping unknown protocol: {}", protocol);
```
dianacarvalho1, line 133: why should we skip instead of erroring?

clients/rust/src/mapping.rs
```diff
@@ -431,7 +432,7 @@ mod tests {
     // -----------------------------------------------------------------------
 
     #[test]
-    fn quote_try_from_dto_drops_amount_out_net_gas_and_defaults_to_fynd() {
+    fn quote_try_from_dto_preserves_amount_out_net_gas_and_defaults_to_fynd() {
```
dianacarvalho1: this test name is unnecessarily complex

dianacarvalho1 (approved):
Thank you @Haikane 🙏🏼 Gave a very loose review! After it is rebased onto main and CI passes, this looks good to me!

fynd-core/src/feed/protocol_registry.rs
Haikane: This change is from main (came in via the merge commit), not from this PR. The reasoning: an unknown protocol in the config shouldn't crash the whole solver, it should just skip it and keep running with the protocols it does support. But happy to revert if you'd prefer it to error.

clients/rust/src/mapping.rs
Haikane: Good call, simplified to `quote_from_dto` in 15d6367.

PR #103, opened 2026-03-16
refactor: replace DFS market price with Bellman-Ford SPFA

Haikane:
## Summary

- Replace DFS-based `TokenGasPriceComputation` with Bellman-Ford SPFA one-to-all pricing
- New `bellman_ford_pricing.rs` module: single SPFA pass from WETH prices all tokens using actual `get_amount_out()` simulation during relaxation, replacing DFS path enumeration + spot-price heuristic scoring
- Remove `SpotPriceComputation` dependency from token gas pricing (still used by pool depths and MostLiquid)

## Benchmark results

| Config | Bootstrap | Incr median | Tokens |
|---|---|---|---|
| Eth all protocols TVL 50 | 104ms (was 2,599ms, **25x**) | 101ms (was 1,352ms, **13x**) | 803 (was 817) |
| Eth native TVL 50 | 3ms (was 20ms, **6.7x**) | 7ms (was 9.5ms, **1.4x**) | 726 (was 730) |
| Eth native TVL 10 | 11ms (was 51ms, **4.6x**) | 7ms (was 21.5ms, **3x**) | 1,974 (was 1,976) |
| Base all TVL 50 | 4ms (was 57ms, **14x**) | 3ms (was 25ms, **8x**) | 205 (was 209) |

Token coverage is 0.1-1.9% lower due to BF's forbid-revisits pruning marginal paths through low-liquidity pools.

## Test plan

- [x] 238/240 lib tests pass (2 pre-existing CLI test failures on base branch)
- [x] `cargo clippy --lib -- -D warnings`: zero warnings
- [x] Benchmarked on 4 configurations against DFS baseline
- [ ] Review the 0.1-1.9% token coverage gap on real data


fynd-core/src/algorithm/bellman_ford_pricing.rs
```diff
@@ -0,0 +1,595 @@
+//! Bellman-Ford SPFA for one-to-all token pricing.
+//!
+//! Runs a single flat-array SPFA from a source token (e.g. WETH) with a probe amount,
+//! propagating to all reachable tokens within `max_hops`. Uses forbid-revisits to prevent
+//! paths through arbitrage loops that would distort prices.
+//!
+//! This is the pricing counterpart to the routing BF in `bellman_ford.rs`:
```
dianacarvalho1: bellman_ford.rs doesn't exist in this branch

fynd-core/src/algorithm/bellman_ford_pricing.rs
```diff
@@ -0,0 +1,595 @@
...
+    source: NodeIndex,
+    token_map: HashMap<NodeIndex, Token>,
+}
+
+impl SpfaAllResult {
+    /// Returns the best forward amount reachable at `node`.
+    #[allow(dead_code)]
+    pub fn amount_at(&self, node: NodeIndex) -> &BigUint {
+        &self.distance[node.index()]
+    }
```
dianacarvalho1: let's not introduce dead code if we don't need it 🧹

fynd-core/src/algorithm/bellman_ford_pricing.rs
```diff
@@ -0,0 +1,595 @@
...
+        let is_vm = component_state
+            .as_any()
+            .downcast_ref::<EVMPoolState<PreCachedDB>>()
+            .is_some();
+
+        let state_override = if is_vm {
+            vm_state_override.as_ref()
+        } else {
+            native_state_overrides.get(component_id)
+        };
```
dianacarvalho1: why is this needed?

dianacarvalho1 (commented):
Ah this PR should be bundled with https://github.com/propeller-heads/fynd/pull/43 so we don't have repeated code? 
What is your merge plan here? I would suggest cleaning up https://github.com/propeller-heads/fynd/pull/43 in a way that the necessary methods for this PR are available and can just be called here instead of repeating the same code twice

fynd-core/src/algorithm/bellman_ford_pricing.rs
Haikane: Fixed. Rebased on main (which now includes PR #43), and changed this to a proper module link: `[`bellman_ford`](super::bellman_ford)`.

fynd-core/src/algorithm/bellman_ford_pricing.rs
Haikane: Removed from production code. Moved behind `#[cfg(test)]` since it's only used in test assertions.

fynd-core/src/algorithm/bellman_ford_pricing.rs
Haikane: Good catch, it wasn't needed. Forbid-revisits guarantees each pool appears at most once in a path, so state overrides can never trigger. Removed the entire VM/native state-override tracking, the EVMPoolState downcast, and the related imports.

Haikane:
Addressed all review comments. Rebased on main (which now includes PR #43) and deduplicated the shared code:

- Created `bf_helpers.rs` with shared free functions (`path_has_conflict`, `reconstruct_path`, `extract_subgraph_edges`) used by both routing and pricing BF
- Updated `bellman_ford.rs` to call `bf_helpers::` instead of associated methods
- Removed dead code and unnecessary state-override logic from pricing module

285 lib tests pass, zero clippy warnings.

dianacarvalho1 (approved):
Thank you @Haikane

tvinagre:
Let's not merge this now as it interferes with https://github.com/propeller-heads/fynd/pull/95/changes

PR #108, opened 2026-03-18
fix: recalculate token price when any competing path pool changes

Troshchk:
TokenPriceEntry::path_components previously only tracked pools from the currently-selected best path. If a competing (non-best) path's pool changed and became better, the system wouldn't know to recompute the token's price until a restart.

In token_gas_price.rs, after finding the best price for each token, collect component IDs from all candidate paths (not just the winner) and merge them into path_components. This ensures recalculation fires whenever any pool that could affect which path is best changes.

**[Benchmarks]**
This change introduces a ~**10 ms overhead (~24%) on the background `token_gas_price` calculation.** 
I ran a simple comparison by executing both branches and measuring elapsed_ms for token price recomputation across 260 block updates.
main: 43.6 ms mean 
price_recalc: 54.0 ms mean 

I also benchmarked **route quality** with both solvers running:
Route match rate: 98/100
price_recalc found 2 better routes (up to +0.02 bps output)
Gas estimates matched in 98/100 cases; in 2 cases price_recalc was higher (max +75%)

No influence on the **solve_time** according to the benchmarks.

### DIFF
--- a/fynd-core/src/derived/computations/token_gas_price.rs
+++ b/fynd-core/src/derived/computations/token_gas_price.rs
@@ -441,15 +441,7 @@ impl TokenGasPriceComputation {
             }
         }
 
-        // Extend each token's path_components with all candidate path components so
-        // incremental recomputation fires when any competing path's pool changes.
-        for (token, (_, _, components)) in best_prices.iter_mut() {
-            if let Some(all_comps) = all_candidate_components.get(token) {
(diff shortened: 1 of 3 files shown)


dianacarvalho1 (approved):
Thank you 🙏🏼

zizou0x (approved):

tvinagre (approved):

PR #109, opened 2026-03-19
feat: harden fynd-core public API

kayibal:
## Summary

- Add `#[non_exhaustive]` to 6 public enums and 11 struct variants to allow adding variants/fields without breaking downstream matches
- Make fields private on 8 structs (`PoolConfig`, `SolverParts`, `OrderQuote`, `ComputationManagerConfig`, `AlgorithmConfig`, `UnsupportedChainError`, `UnknownAlgorithmError`, `ComputationRequirements`) and add public getters/constructors
- Reduce visibility to `pub(crate)` for 5 internal items (`GasPriceFetcher`, `TychoFeed`, `TychoFeedConfig`, `DataFeedError`, `NATIVE_TOKEN`)
- Add `#[must_use]` to `Quote`, `OrderQuote`, `Route`, `SolverBuilder`, `WorkerPoolBuilder`
- Update downstream match arms in fynd-rpc and fynd-rpc-types with wildcard fallbacks

## Test plan

- [x] `cargo clippy --workspace -- -D warnings` passes
- [x] `cargo test --workspace` passes (415 tests, 0 failures)
- [ ] Verify fynd-rpc-types `core` feature still compiles: `cargo check -p fynd-rpc-types --features core`
- [ ] Verify experimental feature compiles: `cargo check -p fynd-rpc --features experimental`

### DIFF
--- a/fynd-core/src/solver.rs
+++ b/fynd-core/src/solver.rs
@@ -88,25 +88,111 @@ fn default_algo_timeout_ms() -> u64 {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct PoolConfig {
     /// Algorithm name for this pool (e.g., `"most_liquid"`).
-    pub algorithm: String,
+    algorithm: String,
     /// Number of worker threads for this pool.
     #[serde(default = "num_cpus::get")]
-    pub num_workers: usize,
+    num_workers: usize,
     /// Task queue capacity for this pool.
     #[serde(default = "default_task_queue_capacity")]
-    pub task_queue_capacity: usize,
+    task_queue_capacity: usize,
     /// Minimum hops to search (must be >= 1).
     #[serde(default = "default_min_hops")]
-    pub min_hops: usize,
+    min_hops: usize,
     /// Maximum hops to search.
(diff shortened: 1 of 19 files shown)


fynd-core/src/solver.rs
```diff
@@ -171,7 +258,7 @@ struct CustomPoolEntry {
 }
 
 /// Builder for assembling the full solver pipeline.
-///
+#[must_use = "a builder does nothing until .build() is called"]
```
dianacarvalho1, line 261: weird placing between the docstring

dianacarvalho1: this doesn't force the user to actually use `build` right? it will trigger just if nothing is done with `FyndBuilder`, but is the user does something else (other than build) then it this won't be raised right?

dianacarvalho1 (approved):
Thank you @kayibal ! tiny comment only

fynd-core/src/solver.rs
kayibal: It ensures sure the builder is consumed and not just left hanging in a variable. It will only warn the user if it detects this case.

PR #110, opened 2026-03-19
refactor: make AlgorithmConfig.max_routes private and add getter

louise-poole:


Troshchk (approved):
Thank you @louise-poole!

PR #113, opened 2026-03-19
feat: harden fynd-rpc-types

kayibal:
## Summary

- **Remove `SingleOrderQuote`** — confirmed zero active usage across all consumers; was re-exported by `fynd-rpc` but never constructed or referenced downstream
- **`#[non_exhaustive]`** on `UserTransferType` and `QuoteStatus` — downstream `match` arms no longer break when new variants are added; wildcard arm added to `From<dto::QuoteStatus>` in `fynd-client`
- **`#[must_use]`** on `Quote`, `OrderQuote`, `Route` — silently discarding a computed quote is a bug
- **Private fields + constructors/getters on all 15 structs** — struct literal construction in downstream crates no longer breaks when fields are added; public API is now methods, not fields
  - Request types (`QuoteRequest`, `QuoteOptions`, `EncodingOptions`, `PermitSingle`, `PermitDetails`, `Order`): `new()` + builder `with_*()` setters
  - Response types (`HealthStatus`, `ErrorResponse`, `BlockInfo`, `Quote`, `Route`, `Swap`, `Transaction`, `OrderQuote`): `new()` constructors where constructed externally + ref getters on all
  - `conversions` module (internal, `feature = "core"`) is unchanged — child modules retain access to private fields
- **Remove dead `fynd-rpc-types` dependency** from root `fynd` crate — zero references in `src/main.rs` or `src/cli.rs`

## Test plan

- [ ] `cargo +nightly fmt -- --check` passes
- [ ] `cargo +nightly clippy --locked --all --all-features --all-targets -- -D warnings` passes
- [ ] `cargo nextest run --workspace --locked --all-targets --all-features --bin fynd` passes
- [ ] Existing mapping tests in `fynd-client` updated to use serde deserialization for DTO test fixtures


Cargo.toml
```diff
@@ -8,7 +8,6 @@ edition = "2021"
 
 [dependencies]
 fynd-rpc = { workspace = true }
-fynd-rpc-types = { workspace = true }
```
dianacarvalho1, line 11: why?

dianacarvalho1 (approved):
👍🏼

Cargo.toml
kayibal: The direct dependency isn't actually required, because `fynd-rpc` re-exports the types.

PR #116, opened 2026-03-19
feat: PriceProvider skeleton + helpers

tamaralipows:
- PriceProviderRegistry
- PriceProvider trait
- PriceProviderError + ExternalPrice
- PriceGuardConfig
- common helper functions

### DIFF
--- a/fynd-core/src/price_guard/common.rs
+++ b/fynd-core/src/price_guard/common.rs
@@ -1,0 +1,140 @@
+//! Helper functions for PriceGuard calculations.
+
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+use num_bigint::BigUint;
+use tycho_simulation::tycho_common::models::Address;
+
+use super::provider::PriceProviderError;
+use crate::feed::market_data::SharedMarketDataRef;
+
+/// Maximum age of price data before it is considered stale.
+pub const STALENESS_THRESHOLD: Duration = Duration::from_secs(30);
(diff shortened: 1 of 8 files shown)


tamaralipows:
So does this mean the PR title shouldn't be `feat` ?

kayibal:
It means it doesn't have to be, but it is allowed to be e.g. if you add a new feature it should be feat

fynd-core/src/price_guard/common.rs
```diff
@@ -0,0 +1,140 @@
...
+pub fn normalize_symbol(symbol: &str) -> &str {
+    match symbol.to_uppercase().as_str() {
+        "WETH" => "ETH",
+        "WBTC" => "BTC",
+        "WBNB" => "BNB",
+        "WMATIC" => "MATIC",
+        "WAVAX" => "AVAX",
+        _ => symbol,
+    }
+}
```
kayibal: This seems to be brittle.. is this the best we can do? Maybe this should be happening within the individual provider implementations and our API keeps using the token addresses.

Troshchk (approved):
Thank you @tamaralipows! Looks good to me 👍

fynd-core/src/price_guard/utils.rs
dianacarvalho1, line 1: utils?

dianacarvalho1 (approved):
Thank you! 🙏🏼

PR #117, opened 2026-03-20
feat: publish @fynd/client to npm on release

kayibal:
## Summary

- Adds a `publish-npm` job to `release.yaml` that builds and publishes `@fynd/client` to npm on each GitHub release. Uses `--dry-run` until `NPM_TOKEN` is configured and the repo is public (same pattern as the Rust crates job).
- Extends `release.config.js` to bump `clients/typescript/client/package.json` version alongside Rust crates during semantic-release, and includes it in the release commit.
- Bundles `@fynd/autogen` into `@fynd/client`: moves `schema.d.ts` and inlines `autogen/src/index.ts` as `client/src/autogen.ts`, removing the workspace package so only `@fynd/client` is published.
- Updates the OpenAPI drift check in CI to reference the new schema path (`client/src/schema.d.ts`).
- Fixes two type errors that were masked by the previous separate-compilation approach: conditional spread for nullable `gasPriceAgeMs`, and an `unknown` cast for the health endpoint error response.

## Before going public

Remove `--dry-run` from both publish steps and configure the secrets:
- `NPM_TOKEN` — npm access token with publish rights to `@fynd/client`
- `CRATESIO_REGISTRY_TOKEN` — already noted in `release.yaml`


dianacarvalho1 (approved):
LFG 🚀

tvinagre (approved):
great work @kayibal

zizou0x (approved):
GG kayibal

PR #119, opened 2026-03-20
docs: Set up extensive claude docs structure

louise-poole:


.claude/knowledge/rust.md
```diff
@@ -0,0 +1,62 @@
+# Coding in Rust
+
+## Commands
+
+During development, use simpler commands to help you iterate:
+
+```bash
+cargo check --all-features
+cargo test
+cargo clippy --all-features
```
kayibal, line 10: why not use the usual commands from check.sh?

I actually added running check.sh as a precommit hook on my git. and it's been great experience so far. You could potential reduce this whole section to just point to `check.sh` and mention that it is ran anyway as precommit hook.

.claude/knowledge/rust.md
```diff
@@ -0,0 +1,62 @@
...
+
+### General Rust guidelines
+
+- Prefer `for` loops with mutable accumulators over long iterator chains when clearer
+- Use `let...else` for early returns; keep happy path unindented
+- No wildcard matches — explicit destructuring catches field changes at compile time
+- Newtypes over primitives where the domain warrants it
+- Enums for state machines, not boolean flags
+- Write efficient code by default — correct algorithm, appropriate data structures, no unnecessary allocations. Profile
+  before micro-optimizing.
```
kayibal, line 25: I have this in my global Claude.md already, would use up unnecessary context....

.claude/knowledge/typescript.md
```diff
@@ -0,0 +1,59 @@
...
+
+## OpenAPI Codegen
+
+The TypeScript types are generated from `clients/openapi.json`:
+
+1. Update the OpenAPI spec: `./scripts/update-openapi.sh`
+2. This regenerates `clients/openapi.json` from Rust types and `autogen/src/schema.d.ts` via
+   `openapi-typescript`
+
+CI checks for drift in both the OpenAPI spec and the generated TypeScript types.
```
kayibal: I think this needs to be put into the rust module that defines the openapi output: `fynd-rpc-types`

.claude/knowledge/version_control.md
kayibal, line 1: Pretty sure I have a lot of this already in my global Claude.md from tob.

.claude/skills/plan/SKILL.md
```diff
@@ -0,0 +1,145 @@
...
+
+After collecting all answers, summarize the proposal back to the user in a compact format:
+
+```
+**Goal:** ...
+**Must have:** ...
+**Nice to have:** ...
+**Consider:** ...
+**Avoid:** ...
+```
```
kayibal, line 43: Honestly for an AI I think the "Nice To Have" section kind of leaves the door open for ti being vague...

.claude/skills/plan/SKILL.md
```diff
@@ -0,0 +1,145 @@
...
+
+Ask the user to confirm the proposal is correct before proceeding.
+
+## Phase 2: Research and Assumptions
+
+Read the codebase documentation to understand the problem space:
+
+1. Read `.claude/CODEBASE.md` and any linked docs relevant to the goal.
+2. Use Glob/Grep/Read to examine specific files as needed.
```
kayibal: I think using the `rust_analyzer_lsp` mcp plugin preserves context a lot better compared to glob/grep/read...

.claude/skills/plan/SKILL.md
```diff
@@ -0,0 +1,145 @@
...
+3. Identify the modules, traits, and types involved.
+
+Then list your **assumptions** — things you believe to be true based on your research that affect the solution design.
+Present these to the user as a numbered list and ask them to confirm, correct, or add to them **as plain text** (free
+text response needed — do not use `AskUserQuestion`).
```
kayibal: I feel this is a bit restrictive, afaik recently I have always been able to answer Claude questions with a custom field. I really like the interview style multiple choice options it gives me. 75% of the time I am ok with one of the choices. Sometimes I write my own thing and it usually gives me the option to do so.

.claude/skills/plan/SKILL.md
```diff
@@ -0,0 +1,145 @@
...
+---
+
+## Planning guidelines
+
+- **No speculative features** — Don't add features, flags, or configuration unless actively needed.
+- **No premature abstraction** — Don't create utilities until you've written the same code three times.
+- **Clarity over cleverness** — Prefer explicit, readable code over dense one-liners.
+- **Justify new dependencies** — Each dependency is attack surface and maintenance burden.
+- **Finish the job** — Handle edge cases you can see. Clean up what you touched. Flag broken adjacent code. But don't
+  invent new scope.
```
kayibal, line 136: Again taken 1-1 from tob which a lot of us have globally. Would drop these.

.claude/skills/run-ci/SKILL.md
```diff
@@ -0,0 +1,92 @@
+---
+allowed-tools: Bash(cargo:*), Bash(pnpm:*), Bash(npx:*), Bash(jq:*), Bash(diff:*), Bash(git diff:*), Bash(git branch:*), Bash(git status:*), Read
+description: "Run the full CI pipeline locally to catch failures before pushing. Use this skill before creating a PR, before pushing commits, or whenever you want to verify that CI will pass. Also use it when the user says 'run ci', 'check ci', 'run tests', 'lint', or 'will ci pass'."
+user-invocable: true
+---
+
+# Run CI Locally
+
+Run the same checks that GitHub Actions CI runs, locally, to catch failures before they hit the
```
kayibal, line 9: I think only the checks that previosu edits could have affected should run. E.g. no need to always run the typescript tests.

fynd-core/CLAUDE.md
```diff
@@ -0,0 +1,76 @@
...
+    fn update_edge_weights_with_derived(&mut self, market: &SharedMarketData, derived: &DerivedData) -> usize;
+}
+```
+
+## Two Builder Levels
+
+- **`FyndBuilder`** (`solver.rs`): Assembles feed + gas + computations + pools + encoder + router.
+  Returns a `Solver` that can `quote()` directly. For standalone (non-HTTP) use.
+- **`FyndRPCBuilder`** (`fynd-rpc/builder.rs`): Wraps `FyndBuilder` and adds HTTP server. For
+  server deployments.
```
kayibal: This shouldn't be mentioned here.

kayibal (approved):

.claude/knowledge/rust.md
louise-poole: These are intentionally kept in the project-level knowledge files so they work for all contributors, not just those with matching global CLAUDE.md settings. The context cost is small and worth it to ensure consistency imo.

.claude/knowledge/version_control.md
louise-poole: Same comment as above - I think it's valuable to have this on a project level too and doesn't add much overhead to be honest.

.claude/skills/plan/SKILL.md
louise-poole: In the few iterations I did using this skill, most of the 'nice to haves' listed were decent considerations, though not many I actually added to the plan. Though you're right - I think the 'Consider' part is enough for this.

.claude/skills/plan/SKILL.md
louise-poole: Not everyone would have that plugin (though I would encourage it). So I'll add an instruction to give it preference if it's available but fallback to these as defaults.

.claude/knowledge/rust.md
kayibal: Based on the last company survey, everyone should have this in their global Claude.md. The problem is if you use claude to set it up, you never look at this file.

PR #121, opened 2026-03-21
feat: PriceGuard validation

tamaralipows:
1. Do you think it's okay that the `get_expected_out` is sync (see latest commit)? I realized it just reads from cache so this should be fine anyway? 
2. See comment below.

### DIFF
--- a/fynd-core/src/price_guard/config.rs
+++ b/fynd-core/src/price_guard/config.rs
@@ -23,6 +23,12 @@ pub struct PriceGuardConfig {
     /// `true`: let solutions pass through when no provider can be reached.
     allow_on_provider_error: bool,
 
+    /// Controls behavior when every provider returns `PriceNotFound` for the
+    /// requested token pair.
+    /// `false` (default): reject solutions for unknown token pairs.
+    /// `true`: let solutions pass through when no provider has a price.
+    allow_on_token_price_not_found: bool,
+
     /// Whether the price guard is enabled.
     /// Default: `true`.
     enabled: bool,
@@ -34,6 +40,7 @@ impl Default for PriceGuardConfig {
             lower_tolerance_bps: 300,
             upper_tolerance_bps: 10_000,
             allow_on_provider_error: false,
+            allow_on_token_price_not_found: false,
             enabled: true,
         }
     }
@@ -52,6 +59,10 @@ impl PriceGuardConfig {
         self.allow_on_provider_error
     }
 
+    pub fn allow_on_token_price_not_found(&self) -> bool {
+        self.allow_on_token_price_not_found
+    }
+
     pub fn enabled(&self) -> bool {
         self.enabled
     }
@@ -68,6 +79,11 @@ impl PriceGuardConfig {
 
     pub fn with_allow_on_provider_error(mut self, allow: bool) -> Self {
         self.allow_on_provider_error = allow;
+        self
+    }
+
+    pub fn with_allow_on_token_price_not_found(mut self, allow: bool) -> Self {
+        self.allow_on_token_price_not_found = allow;
         self
     }
 
--- a/fynd-core/src/price_guard/config.rs
+++ b/fynd-core/src/price_guard/config.rs
@@ -1,8 +1,10 @@
+use serde::{Deserialize, Serialize};
+
 /// Configuration for the PriceGuard external price validation.
 ///
 /// Controls tolerance thresholds, fail-open behavior, and whether validation
 /// is enabled at all. All fields have sensible defaults via [`Default`].
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct PriceGuardConfig {
     /// Maximum allowed deviation when `amount_out < expected`, in basis points.
     /// Solutions where the user gets less than expected by more than this are rejected.
@@ -15,7 +17,8 @@ pub struct PriceGuardConfig {
     /// Default: 10_000 (100%).
     upper_tolerance_bps: u32,
 
-    /// Controls behavior when all providers error (network issues, API down).
+    /// Controls behavior when all providers error for reasons unrelated to pricing
+    /// (network issues, API down).
     /// `false` (default): reject solutions when no provider can return a price.
     /// `true`: let solutions pass through when no provider can be reached.
     allow_on_provider_error: bool,
--- a/fynd-core/src/price_guard/config.rs
+++ b/fynd-core/src/price_guard/config.rs
@@ -41,7 +41,7 @@ impl Default for PriceGuardConfig {
             upper_tolerance_bps: 10_000,
             allow_on_provider_error: false,
             allow_on_token_price_not_found: false,
-            enabled: true,
+            enabled: false,
         }
     }
 }
--- a/fynd-core/src/price_guard/guard.rs
+++ b/fynd-core/src/price_guard/guard.rs
@@ -77,6 +77,7 @@ impl PriceGuard {
     /// Checks that a successful quote has a route with input/output tokens.
     /// Returns the token pair if valid, `None` otherwise.
     fn check_structure(&self, quote: &OrderQuote) -> Option<(Address, Address)> {
+        //invalid route would be rejected earlier; this prevents using expect
         let Some(route) = quote.route() else {
             warn!(order_id = quote.order_id(), "successful quote has no route");
             return None;
--- a/fynd-core/src/price_guard/guard.rs
+++ b/fynd-core/src/price_guard/guard.rs
@@ -40,12 +40,12 @@ impl PriceGuard {
     ///
     /// For each successful quote with a route:
     /// 1. Checks that the quote has a well-formed route
-    /// 2. Queries all registered providers concurrently
+    /// 2. Queries all registered providers
     /// 3. Passes if at least one provider validates within BPS tolerance
     ///
     /// Failures are always per-quote: the quote's status is set to
     /// `PriceCheckFailed` and processing continues. Never aborts the batch.
-    pub async fn validate(
+    pub fn validate(
         &self,
         mut quotes: Vec<OrderQuote>,
         config: &PriceGuardConfig,
@@ -66,10 +66,7 @@ impl PriceGuard {
                 quote.set_status(QuoteStatus::PriceCheckFailed);
                 continue;
             };
-            if !self
-                .check_price(quote, &token_in, &token_out, config)
-                .await
-            {
+            if !self.check_price(quote, &token_in, &token_out, config) {
                 quote.set_status(QuoteStatus::PriceCheckFailed);
             }
         }
@@ -94,7 +91,7 @@ impl PriceGuard {
     }
 
     /// Queries all providers and returns `true` if at least one validates.
-    async fn check_price(
+    fn check_price(
         &self,
         quote: &OrderQuote,
         token_in: &Address,
@@ -103,8 +100,7 @@ impl PriceGuard {
     ) -> bool {
         let results = self
             .registry
-            .get_all_expected_out(&token_in, &token_out, quote.amount_in())
-            .await;
+            .get_all_expected_out(token_in, token_out, quote.amount_in());
 
         let mut any_validated = false;
         let mut all_errors = true;
@@ -184,7 +180,6 @@ impl PriceGuard {
 mod tests {
     use std::str::FromStr;
 
-    use async_trait::async_trait;
     use num_bigint::BigUint;
     use rstest::rstest;
     use tokio::task::JoinHandle;
@@ -211,13 +206,12 @@ mod tests {
         source: String,
     }
 
-    #[async_trait]
     impl PriceProvider for MockProvider {
         fn start(&mut self, _market_data: SharedMarketDataRef) -> JoinHandle<()> {
             tokio::spawn(std::future::ready(()))
         }
 
-        async fn get_expected_out(
+        fn get_expected_out(
             &self,
             _token_in: &Address,
             _token_out: &Address,
@@ -229,13 +223,12 @@ mod tests {
 
     struct FailingProvider;
 
-    #[async_trait]
     impl PriceProvider for FailingProvider {
         fn start(&mut self, _market_data: SharedMarketDataRef) -> JoinHandle<()> {
             tokio::spawn(std::future::ready(()))
         }
 
-        async fn get_expected_out(
+        fn get_expected_out(
             &self,
             _token_in: &Address,
             _token_out: &Address,
@@ -316,8 +309,8 @@ mod tests {
     #[case::within_upper(1000, 1500, 300, 10_000, true)]
     #[case::at_upper_boundary(1000, 2000, 300, 10_000, true)]
     #[case::beyond_upper(1000, 2500, 300, 10_000, false)]
-    #[tokio::test]
-    async fn test_deviation_bounds(
+    #[test]
+    fn test_deviation_bounds(
         #[case] provider_amount: u64,
         #[case] fynd_amount: u64,
         #[case] lower_bps: u32,
@@ -329,9 +322,7 @@ mod tests {
             .with_upper_tolerance_bps(upper_bps);
         let guard = price_guard(vec![mock_provider(provider_amount)]);
 
-        let result = guard
-            .validate(vec![make_quote(fynd_amount)], &config)
-            .await;
+        let result = guard.validate(vec![make_quote(fynd_amount)], &config);
 
         let expected_status =
             if should_pass { QuoteStatus::Success } else { QuoteStatus::PriceCheckFailed };
@@ -341,37 +332,33 @@ mod tests {
     #[rstest]
     #[case::all_error_allow(true, true)]
     #[case::all_error_deny(false, false)]
-    #[tokio::test]
-    async fn test_all_providers_error(#[case] allow_on_error: bool, #[case] should_pass: bool) {
+    #[test]
+    fn test_all_providers_error(#[case] allow_on_error: bool, #[case] should_pass: bool) {
         let config = PriceGuardConfig::default().with_allow_on_provider_error(allow_on_error);
         let guard = price_guard(vec![Box::new(FailingProvider), Box::new(FailingProvider)]);
 
-        let result = guard
-            .validate(vec![make_quote(500)], &config)
-            .await;
+        let result = guard.validate(vec![make_quote(500)], &config);
 
         let want = if should_pass { QuoteStatus::Success } else { QuoteStatus::PriceCheckFailed };
         assert_eq!(result[0].status(), want);
     }
 
-    #[tokio::test]
-    async fn test_disabled_guard() {
+    #[test]
+    fn test_disabled_guard() {
         let config = PriceGuardConfig::default().with_enabled(false);
 
         // Guard is disabled via config. Expected amount out of the provider is irrelevant,
         // because the provider is never called.
         let guard = price_guard(vec![mock_provider(10_000)]);
 
-        let result = guard
-            .validate(vec![make_quote(50)], &config)
-            .await;
+        let result = guard.validate(vec![make_quote(50)], &config);
 
         assert_eq!(result.len(), 1);
         assert_eq!(result[0].status(), QuoteStatus::Success);
     }
 
-    #[tokio::test]
-    async fn test_one_pass_one_fail() {
+    #[test]
+    fn test_one_pass_one_fail() {
         // Test that the quote status is success even with one failing provider,
         // as long as the second provider passes.
         let config = PriceGuardConfig::default().with_lower_tolerance_bps(300);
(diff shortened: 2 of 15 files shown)


fynd-core/src/types/quote.rs
```diff
@@ -85,6 +85,13 @@ pub struct QuoteOptions {
     /// Options for encoding the solution into an on-chain transaction.
     /// If `None`, the solution is returned without an encoded transaction.
     encoding_options: Option<EncodingOptions>,
+    /// TODO: Should we move to `Order` for per-order price guard control if needed?
+    /// One might want to turn off the price guard for a few orders of their list, but not all,
+    /// depending on tokens. At the same time - I kept it in here to be coherent with where we
+    /// are passing the encoding options. Please let me know if you disagree.
```
tamaralipows: What do you think?

fynd-core/src/price_guard/guard.rs
```diff
@@ -0,0 +1,411 @@
...
+        mut quotes: Vec<OrderQuote>,
+        config: &PriceGuardConfig,
+    ) -> Vec<OrderQuote> {
+        if !config.enabled() {
+            return quotes;
+        }
+
+        if self.registry.is_empty() {
+            warn!("price guard is enabled but no providers are registered");
+        }
```
louise-poole, line 70: Surely this is an error then? And we should error here? Or move this check to `new` and error there?

fynd-core/src/price_guard/guard.rs
```diff
@@ -0,0 +1,411 @@
...
+            return None;
+        };
+        match (route.input_token(), route.output_token()) {
+            (Some(token_in), Some(token_out)) => Some((token_in, token_out)),
+            _ => {
+                warn!(order_id = quote.order_id(), "successful quote has empty route");
+                None
+            }
+        }
+    }
```
louise-poole, line 102: Very strange to me that this might occur. Surely an invalid quote should be rejected at an earlier level and not reach our price validator? Or is an empty route how we return a failed solution? I'm afraid I don't fully understand the case here.

fynd-core/src/price_guard/guard.rs
```diff
@@ -0,0 +1,411 @@
...
+                    debug!(error = %e, "price provider error");
+                }
+            }
+        }
+
+        if all_errors {
+            config.allow_on_provider_error()
+        } else {
+            any_validated
+        }
```
louise-poole: Personal preference: (I'm kinda not usually a fan of a status like 'all errors' being true when nothing has errored yet). 
```suggestion
        let mut has_price = false;

        for result in &results {
            match result {
                Ok(price) => {
                    if self.price_within_tolerance(quote, price, config) {
                        return true;
                    }
                    has_price = true;
                }
                Err(e) => debug!(error = %e, "price provider error"),
            }
        }

        if has_price { false } else { config.allow_on_provider_error() }
```

fynd-core/src/price_guard/guard.rs
```diff
@@ -0,0 +1,411 @@
...
+        quote: &OrderQuote,
+        provider_price: &ExternalPrice,
+        config: &PriceGuardConfig,
+    ) -> bool {
+        if provider_price
+            .expected_amount_out()
+            .is_zero()
+        {
+            return true;
+        }
```
louise-poole, line 158: Why does a provider price of 0 return a valid 'is in tolerance' result?

fynd-core/src/price_guard/guard.rs
```diff
@@ -0,0 +1,411 @@
...
+
+        // Fynd price is higher than the price provider
+        if fynd_amount_out >= provider_amount_out {
+            diff = fynd_amount_out - provider_amount_out;
+            tolerance = config.upper_tolerance_bps();
+        // Fynd price is lower than the price provider
+        } else {
+            diff = provider_amount_out - fynd_amount_out;
+            tolerance = config.lower_tolerance_bps();
+        }
```
louise-poole: Why not the usual let if structure? It's more rusty
```suggestion
        let (diff, tolerance) = if fynd_amount_out >= provider_amount_out {
            (fynd_amount_out - provider_amount_out, config.upper_tolerance_bps())
        } else {
            (provider_amount_out - fynd_amount_out, config.lower_tolerance_bps())
        }
```

fynd-core/src/types/quote.rs
louise-poole: I think let's stay in line with the EncodingOptions no? They can submit 2 order batches if they want them to have different treatment no? 
I'm not fully convinced on this decision though - would like others to weigh in too.

louise-poole (commented):
Nice! A couple suggestions, otherwise looks solid 👍

fynd-core/src/price_guard/guard.rs
tamaralipows: This is all to prevent using an `expect` when getting the routes and tokens :/

fynd-core/src/price_guard/guard.rs
```diff
@@ -0,0 +1,411 @@
...
+        if self.registry.is_empty() {
+            warn!("price guard is enabled but no providers are registered");
+        }
+
+        for quote in &mut quotes {
+            if quote.status() != QuoteStatus::Success {
+                continue;
+            }
+            let Some((token_in, token_out)) = self.check_structure(quote) else {
+                quote.set_status(QuoteStatus::PriceCheckFailed);
```
Troshchk: Do you think this status is accurate here? May be `NoRouteFound` would be more accurate?

Troshchk: I saw in a later comment that this is meant to avoid using expect, imo it would still be worth having a more relevant status here

fynd-core/src/price_guard/guard.rs
```diff
@@ -0,0 +1,411 @@
...
+    fn check_price(
+        &self,
+        quote: &OrderQuote,
+        token_in: &Address,
+        token_out: &Address,
+        config: &PriceGuardConfig,
+    ) -> bool {
+        let results = self
+            .registry
+            .get_all_expected_out(token_in, token_out, quote.amount_in());
```
Troshchk, line 114: Tbh, I think it would be cleaner to exit early before constructing the full results, especially if `get_all_expected_out` is sync now. Would it be possible to call `price_within_tolerance` directly on each `p.get_expected_out(token_in, token_out, amount_in)` instead of first collecting everything via `get_all_expected_out` and iterating over it afterward?

Troshchk (commented):
Thank you @tamaralipows!
Just a couple of small things 💫

fynd-core/src/types/quote.rs
tamaralipows: I agree with this. I'd like to prioritize the coherence too.

fynd-core/src/price_guard/guard.rs
tamaralipows: In which case would we like to exit early? You mean when one provider already returns an affirmative result?

tamaralipows: Is the refactor worth it to avoid a simple cache pull and multiplication? 
I checked that making it iterate lazily instead of performing collect would mean introducing lifetimes (gross). Making the function itself stop early would mean the registry would now have to have knowledge about tolerances (which isn't its job). Do you have some other suggestion I'm missing?

Troshchk: Maybe something like this?

It removes a vec allocation, which is negligible with few providers, but scales better as more are added. Also avoids doing work we don’t need.

Lmk what you think, happy to keep it as is if you don’t see it as an improvement.
```       
let mut all_errors = true;

for provider in self.registry.providers.iter() {
    match provider.get_expected_out(
        token_in,
        token_out,
        quote.amount_in(),
    ) {
        Ok(price) => {
            all_errors = false;

            if self.price_within_tolerance(quote, &price, config) {
                return true;
            }
        }
        Err(e) => {
            debug!(error = %e, "price provider error");
        }
    }
}

if all_errors {
    config.allow_on_provider_error()
} else {
    false
}
      
```

fynd-core/src/types/quote.rs
Troshchk: I will keep it as it is now for coherence. Removing the TODO note.

fynd-core/src/price_guard/guard.rs
Troshchk: Since we won't have many providers, let's keep it as it is.

louise-poole (approved):
Looks good to me! For future reference try avoid generic commit messages like 'address PR comments' etc... we like to keep the commit history clean and easy to navigate and a too generic message gives us little info on what was done.

fynd-core/src/price_guard/guard.rs
```diff
@@ -134,27 +127,14 @@ impl PriceGuard {
         provider_price: &ExternalPrice,
         config: &PriceGuardConfig,
     ) -> bool {
-        if provider_price
-            .expected_amount_out()
-            .is_zero()
-        {
-            return true;
-        }
```
louise-poole: This early return is actually needed to avoid division by 0 issues later. But it needs to return false and not true.

fynd-core/src/types/quote.rs
```diff
@@ -85,6 +85,9 @@ pub struct QuoteOptions {
     /// Options for encoding the solution into an on-chain transaction.
     /// If `None`, the solution is returned without an encoded transaction.
     encoding_options: Option<EncodingOptions>,
+    /// Per-request price guard overrides. If `None`, uses server defaults.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    price_guard: Option<PriceGuardConfig>,
```
louise-poole, line 90: This needs to be added to `fynd_rpc_types::QuoteOptions` too.

tamaralipows:
@louise-poole @Troshchk FYI when I write "PR comments" in the commit, I always expand on them in the message (click ...) to exactly which comments were fixed, and I try to only use this for minor comments like refactors or moving unnecessary blocks, which I think would spam commit history if I put them all in separate commits. Bigger/important fixes are a separate commit

fynd-core/src/price_guard/guard.rs
Troshchk: Fixed

fynd-core/src/types/quote.rs
Troshchk: Added

fynd-core/src/price_guard/guard.rs
Troshchk: Added returning an error

tamaralipows (commented):
Thanks @Troshchk - looks good for now! (can't approve since I opened the PR) We can make future improvements in the next PRs as they come up. Feel free to approve/merge this!

fynd-core/src/price_guard/guard.rs
```diff
@@ -0,0 +1,418 @@
...
+};
+use crate::types::{OrderQuote, QuoteStatus};
+
+/// Errors returned by [`PriceGuard::validate`].
+#[derive(Error, Debug)]
+pub enum PriceGuardError {
+    /// Price guard is enabled but no providers are registered.
+    #[error("price guard is enabled but no providers are registered")]
+    NoProviders,
+}
```
tamaralipows, line 23: I'm curious if we should just use the existing `PriceProviderError` since this is just one value in the enum. No strong feelings though

dianacarvalho1: I think yes. Let's not bloat the code with enums with only 1 variant!

Troshchk: Tbh, I’d prefer to keep it as an explicit PriceGuardError -- I think it makes it clearer where the issue originates

fynd-core/src/price_guard/guard.rs
```diff
@@ -0,0 +1,418 @@
...
+        if self.registry.is_empty() {
+            return Err(PriceGuardError::NoProviders);
+        }
+
+        for quote in &mut quotes {
+            if quote.status() != QuoteStatus::Success {
+                continue;
+            }
+            let Some((token_in, token_out)) = self.check_structure(quote) else {
+                // This should not happen.
```
dianacarvalho1, line 77: great comment xD

fynd-core/src/price_guard/guard.rs
```diff
@@ -0,0 +1,418 @@
...
+                quote.set_status(QuoteStatus::PriceCheckFailed);
+            }
+        }
+
+        Ok(quotes)
+    }
+
+    /// Checks that a successful quote has a route with input/output tokens.
+    /// Returns the token pair if valid, `None` otherwise.
+    fn check_structure(&self, quote: &OrderQuote) -> Option<(Address, Address)> {
```
dianacarvalho1: I think check_structure is too obscure.. maybe `check_route`?

fynd-core/src/price_guard/guard.rs
```diff
@@ -0,0 +1,418 @@
...
+        quote: &OrderQuote,
+        token_in: &Address,
+        token_out: &Address,
+        config: &PriceGuardConfig,
+    ) -> bool {
+        let results = self
+            .registry
+            .get_all_expected_out(token_in, token_out, quote.amount_in());
+
+        let mut has_price = false;
```
dianacarvalho1: `has_price` is more like `validated_price` no? it means two things: that there is a price from a provider and that the fynd price is within the tolerance

fynd-core/src/price_guard/guard.rs
```diff
@@ -0,0 +1,418 @@
...
+                    if self.price_within_tolerance(quote, price, config) {
+                        return true;
+                    }
+                    has_price = true;
+                }
+                Err(e) => debug!(error = %e, "price provider error"),
+            }
+        }
+        if has_price {
+            false
```
dianacarvalho1: what if we just didn't get any provider prices because we are looking at long tail tokens? 🤔 I think this logic needs to be more nuanced? not sure tbh.
We can have two scenarios:
1. We don't get prices because the providers are down or wtv
2. We don't get prices because the providers just don't have the tokens we are looking for. In this case I think we have no choice but just ignore the price guard 😕

dianacarvalho1 (commented):
Thank you @tamaralipows and @Troshchk ! I'm sorry for the late review 😓 
I have a few comments but it's ok if they are addressed in a separate PR. Just remember that now Fynd has been launched, so we shouldn't merge to main if the current PR changes how things work significantly

fynd-core/src/price_guard/guard.rs
Troshchk: if the price is validated, it will return `true` early:
``` 
 if self.price_within_tolerance(quote, price, config) {
     return true;
}
```
` has_price` only holds info that the price exists to distinguish between all providers failing and no providers validating the price

fynd-core/src/price_guard/guard.rs
Troshchk: since it checks and returns token pair, may be `validated_token_pair`?

fynd-core/src/price_guard/guard.rs
dianacarvalho1: doesn't the `PriceProviderError` has a string where we can write the reason more explicitly? I don't know honestly

fynd-core/src/price_guard/guard.rs
Troshchk: if all providers are down, than `has_price` will be `false` and the result of `check_price` is derived from `config.allow_on_provider_error()`. This is true for both cases you mention

fynd-core/src/price_guard/guard.rs
Troshchk: It does have a string based variant, but using `PriceProviderError` here feels semantically incorrect, since the error originates from the `PriceGuard`, not the `PriceProvider`

fynd-core/src/price_guard/guard.rs
dianacarvalho1: ah then it's more like a `has_price_but_out_of_tolerance`??

fynd-core/src/price_guard/guard.rs
dianacarvalho1: do you see how that is an issue?

Troshchk: I think this is a valid setup. `allow_on_provider_error` is configurable, so it gives flexibility to decide how to behave when all providers are down and the price can’t be validated.

Just to clarify my previous comment -- when I said “this is true for both cases you mention,” I was referring to the execution flow, not the value of `allow_on_provider_error`. By default, it’s set to false.

fynd-core/src/price_guard/guard.rs
Troshchk: yes, exactly
`price_out_of_tolerance` to make it shorter?

fynd-core/src/price_guard/guard.rs
dianacarvalho1: Imagine we have a user that wants to use the price guard but that it trades a lot of end tail tokens (that the price providers don't have prices for). How should their setup be?? Do we expect them to have two Fynd servers running? One for regular tokens and one for long tail tokens? How are they supposed to even know the difference?

Troshchk: From the price guard’s perspective, both cases: providers being unavailable and no price data existing for a long tail pair mean the same thing: the quote could not be externally validated. In both cases, the relevant user decision is whether to allow that quote through or reject it, and that is exactly what `allow_on_provider_error` controls.

I don’t think we should expect users to run two separate Fynd servers, since `allow_on_provider_error` is set per batch.

dianacarvalho1: aaahh "is set per batch" you mean that it is set per request. Ok I thought this was defined at startup only once. I misunderstood that

fynd-core/src/price_guard/config.rs
```diff
@@ -23,6 +23,12 @@ pub struct PriceGuardConfig {
     /// `true`: let solutions pass through when no provider can be reached.
     allow_on_provider_error: bool,
 
+    /// Controls behavior when every provider returns `PriceNotFound` for the
+    /// requested token pair.
+    /// `false` (default): reject solutions for unknown token pairs.
+    /// `true`: let solutions pass through when no provider has a price.
+    allow_on_token_price_not_found: bool,
```
tamaralipows, line 30: I would call this `allow_on_token_not_found` to indicate the token doesn't exist at all for that provider (and it's shorter)

Troshchk: I chose this name to make it clear we’re handling `PriceNotFound`, not `TokenNotFound`. I think it helps make the failure case more explicit, what do you think?

fynd-core/src/price_guard/guard.rs
```diff
@@ -120,11 +124,19 @@ impl PriceGuard {
                     }
                     has_price = true;
                 }
-                Err(e) => debug!(error = %e, "price provider error"),
+                Err(e) => {
+                    if !matches!(e, PriceProviderError::PriceNotFound { .. }) {
+                        no_price_for_token_found = false;
```
tamaralipows: I think this logic is wrong. We don't want to error if one provider doesn't return a price. This should only be if all providers are unable to find the token

fynd-core/src/price_guard/config.rs
dianacarvalho1: but I think `TokenNotFound` is only used when we are looking for tokens in Tycho data, not on external providers tokens (see [here](https://github.com/propeller-heads/fynd/blob/main/fynd-core/src/price_guard/utils.rs#L23)). So I would agree with @tamaralipows original comment and name it `allow_on_token_not_found`

fynd-core/src/price_guard/guard.rs
```diff
@@ -429,4 +461,53 @@ mod tests {
         assert_eq!(result[0].status(), QuoteStatus::Success);
         assert_eq!(result[1].status(), QuoteStatus::PriceCheckFailed);
     }
+
+    #[rstest]
+    #[case::allow(true, QuoteStatus::Success)]
+    #[case::deny(false, QuoteStatus::PriceCheckFailed)]
+    #[test]
+    fn test_all_price_not_found(#[case] allow: bool, #[case] result_status: QuoteStatus) {
```
dianacarvalho1, line 479: we are missing a test where one provider has a price and another doesn't have it. I think this test shouldn't work now because of the issue @tamaralipows mentioned [here](https://github.com/propeller-heads/fynd/pull/121/changes/c7ef87e3d78da142d3591522774dee0b3aafcc6f#r3010946774)

dianacarvalho1 (commented):
Thank you @Troshchk ! I have some comments

fynd-core/src/price_guard/guard.rs
Troshchk: `no_price_for_token_found` remains true only if all price providers return `PriceProviderError::PriceNotFound`.

If at least one provider returns a different result, `no_price_for_token_found` becomes false, and the final outcome is determined by `config.allow_on_provider_error()`.

Do you think this should work differently?

dianacarvalho1: > no_price_for_token_found remains true only if all price providers return PriceProviderError::PriceNotFound.

I don't think this is true.. 
1. here we are looping through the results
2. if one of them is PriceNotFound then we are changing `no_price_for_token_found` to false

dianacarvalho1: aaaaaaahhhh nevermind! we are checking for inequality 🥴  `!matches!(e, PriceProviderError::PriceNotFound { .. })`.  So if the error IS NOT PriceNotFound then we set no_price_for_token_found to false? 🤔 

Can we make this more obvious please? I think there are unnecessary negations here

Troshchk: okok, will do and will also add a test you mention below

tamaralipows: Ahh yeah. Maybe if we avoid naming vars starting with `no_...` this would be clearer

fynd-core/src/price_guard/guard.rs
Troshchk: Added

fynd-core/src/price_guard/guard.rs
```diff
@@ -113,32 +113,35 @@ impl PriceGuard {
...
                 }
                 Err(e) => {
-                    if !matches!(e, PriceProviderError::PriceNotFound { .. }) {
-                        no_price_for_token_found = false;
+                    if let PriceProviderError::Unavailable(_) |
+                    PriceProviderError::TokenNotFound { .. } |
+                    PriceProviderError::StaleData { .. } = e
+                    {
+                        has_provider_error = true;
                     }
```
dianacarvalho1: wouldn't it be easier to check for the PriceNotFoundError? 😅 instead of enumerating all the other possible errors here? if we extend the variants then we need to keep this up to date as well

dianacarvalho1 (commented):
Thank you! Tiny comment only!!

dianacarvalho1 (approved):

PR #123, opened 2026-03-23
feat(rpc): always return JSON error bodies

kayibal:
## Summary

- Register `JsonConfig` and `QueryConfig` error handlers so every extraction failure (malformed JSON body, wrong `Content-Type`, empty body, invalid query param) returns a typed `{"error": "...", "code": "BAD_REQUEST"}` response instead of actix-web's plain-text default
- Add a `default_service` so unknown routes return `{"error": "not found", "code": "NOT_FOUND"}` JSON 404 instead of actix-web's plain-text default
- Fix `SolveError::MarketDataStale` returning 422 instead of 503 — stale data means the server is temporarily unavailable, consistent with `QueueFull` and `StaleData`
- Fix solver `Timeout` returning 504 Gateway Timeout — 504 implies an upstream proxy timed out; the solver timeout is an internal condition, 503 is the correct code and consistent with other transient failures
- Add constructors for `#[non_exhaustive]` `SolveError` struct variants (`NoRouteFound`, `InsufficientLiquidity`, `Timeout`, `MarketDataStale`) so they are constructible from outside `fynd-core`
- Map all known server error codes in the Rust client (`ALGORITHM_ERROR`, `INTERNAL_ERROR`, `FAILED_ENCODING` → `ServerError`; `NOT_FOUND` → `NotFound`); `Unknown` is now truly reserved for unseen codes
- Regenerate OpenAPI spec and TypeScript schema

## Test plan

- [ ] All `ApiError` variants produce the correct JSON body and HTTP status code (`api::error` tests)
- [ ] Malformed JSON body, empty body, wrong `Content-Type`, bad field type → 400 JSON (`api::handlers` tests)
- [ ] Invalid query param (`?limit=not-a-number`) → 400 JSON
- [ ] Unknown route → 404 JSON
- [ ] All known server error codes map to typed `ErrorCode` variants in the Rust client

### DIFF
--- a/clients/openapi.json
+++ b/clients/openapi.json
@@ -7,7 +7,7 @@
       "name": "MIT",
       "identifier": "MIT"
     },
-    "version": "0.22.0"
+    "version": "0.24.2"
   },
   "paths": {
     "/v1/health": {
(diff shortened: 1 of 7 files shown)


dianacarvalho1 (approved):

louise-poole (approved):
Nice - I like it 👌

PR #124, opened 2026-03-23
feat(rpc): add GET /v1/info endpoint

kayibal:
## Summary

- Adds `GET /v1/info` endpoint returning the chain ID, Tycho Router address, and canonical Permit2 address
- Clients no longer need to hardcode addresses or do a dummy quote to discover the router
- Extracts `router_address` from the encoder at build time and threads it through `Solver`/`SolverParts` into `AppState`
- `InstanceInfo` DTO added to `fynd-rpc-types`; OpenAPI spec and TypeScript autogen schema regenerated

## Test plan

- `cargo nextest run --workspace` — 429 passed, 3 skipped
- `cargo clippy --all-targets --all-features -- -D warnings` — clean
- Handler tests verify: 200 response, required fields present, correct Ethereum router address

### DIFF
--- a/fynd-core/src/encoding/encoder.rs
+++ b/fynd-core/src/encoding/encoder.rs
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
+
 use alloy::{
     primitives::{aliases::U48, Address, Keccak256, U160, U256},
     sol_types::SolValue,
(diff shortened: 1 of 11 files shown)


fynd-core/src/encoding/encoder.rs
```diff
@@ -18,15 +20,31 @@ use tycho_simulation::tycho_common::{models::Chain, Bytes};
...
+///
+/// `tycho_execution::encoding::evm::constants` is a private module, so we embed a copy here.
+/// When upgrading tycho-execution, verify this stays in sync with
+/// `tycho-execution/config/router_addresses.json`.
+const ROUTER_ADDRESSES_JSON: &str = r#"{
+  "ethereum": "0xfD0b31d2E955fA55e3fa641Fe90e08b677188d35",
+  "base": "0xea3207778e39EB02D72C9D3c4Eac7E224ac5d369",
+  "unichain": "0xFfA5ec2e444e4285108e4a17b82dA495c178427B"
+}"#;
+
```
kayibal: We should make those public on tycho-execution and then import here instead.

dianacarvalho1: yikes yes please let's not do this. We just need to add a getter to `TychoRouterEncoder` to get it

dianacarvalho1 (commented):
Thank you @kayibal ! I agree that we should expose this in the TychoRouter (tycho-execution) before merging this PR

dianacarvalho1 (approved):
Thank you 🙏🏼  Aaaahhh damn it. you are going to have issues merging this before my PR that bring the new tycho-execution to Fynd https://github.com/propeller-heads/fynd/pull/135 🫠

PR #125, opened 2026-03-23
feat(rust-client): add approval flow (info, approval, execute_swap, execute_approval)

kayibal:
## Summary

Adds an approval flow to `FyndClient` alongside a rename of the swap execution API for clarity.

### New methods

- **`info()`** — fetches and caches `GET /v1/info` (router address, permit2 address, chain ID). Cached via `OnceLock` after the first successful call.
- **`approval(params: &ApprovalParams, hints: &SigningHints)`** — builds an unsigned EIP-1559 `approve(router, amount)` transaction. On-chain allowance check is opt-in via `ApprovalParams::check_allowance` (off by default for latency-sensitive callers).
- **`execute_approval(signed: SignedApproval)`** — broadcasts a signed approval and polls for inclusion, returning `TxReceipt { tx_hash, gas_cost }`.

### Renamed for clarity

| Old | New |
|---|---|
| `signable_payload()` | `swap_payload()` |
| `SignablePayload` | `SwapPayload` |
| `SignedOrder` | `SignedSwap` |
| `execute()` | `execute_swap()` |

### New public types

`ApprovalParams`, `ApprovalPayload`, `SignedApproval`, `MinedTx`, `TxReceipt`, `InstanceInfo`

`execute_approval()` reuses the `send_raw()` private helper that was extracted from `execute_swap()`.

## Test plan

- `cargo nextest run -p fynd-client` — 75 tests pass
- `cargo clippy --all-targets --all-features -- -D warnings` — clean
- Tests cover: `info()` caching, calldata selector (`0x095ea7b3`), `check_allowance` both branches, `execute_approval` broadcast + poll

### DIFF
--- a/.github/workflows/examples.yaml
+++ b/.github/workflows/examples.yaml
@@ -19,6 +19,7 @@ jobs:
     env:
       TYCHO_API_KEY: ${{ secrets.TYCHO_API_KEY }}
       FORK_RPC_URL: ${{ secrets.ETH_RPC_URL }}
+      RUST_LOG: info
     steps:
       - name: Checkout
         uses: actions/checkout@v4
--- a/.github/workflows/examples.yaml
+++ b/.github/workflows/examples.yaml
@@ -18,7 +18,7 @@ jobs:
     timeout-minutes: 60
     env:
       TYCHO_API_KEY: ${{ secrets.TYCHO_API_KEY }}
-      TYCHO_URL: ${{ secrets.TYCHO_URL }}
+      TYCHO_URL: tycho-beta.propellerheads.xyz
       FORK_RPC_URL: ${{ secrets.ETH_RPC_URL }}
       RUST_LOG: info
     steps:
--- a/.github/workflows/examples.yaml
+++ b/.github/workflows/examples.yaml
@@ -37,5 +37,8 @@ jobs:
         with:
           cache-on-failure: true
 
+      - name: Build fynd (release)
+        run: cargo build --release --locked --quiet
+
       - name: Run all examples
         run: ./scripts/run-all-examples.sh
--- a/.github/workflows/examples.yaml
+++ b/.github/workflows/examples.yaml
@@ -18,7 +18,7 @@ jobs:
     timeout-minutes: 60
     env:
       TYCHO_API_KEY: ${{ secrets.TYCHO_API_KEY }}
-      FORK_RPC_URL: ${{ secrets.FORK_RPC_URL }}
+      FORK_RPC_URL: ${{ secrets.ETH_RPC_URL }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
--- a/.github/workflows/examples.yaml
+++ b/.github/workflows/examples.yaml
@@ -18,6 +18,7 @@ jobs:
     timeout-minutes: 60
     env:
       TYCHO_API_KEY: ${{ secrets.TYCHO_API_KEY }}
+      TYCHO_URL: ${{ secrets.TYCHO_URL }}
       FORK_RPC_URL: ${{ secrets.ETH_RPC_URL }}
       RUST_LOG: info
     steps:
--- a/.github/workflows/examples.yaml
+++ b/.github/workflows/examples.yaml
@@ -1,0 +1,40 @@
+name: Examples
+
+on:
+  pull_request:
+  workflow_call:
+
+permissions:
+  contents: read
+
+env:
+  CARGO_TERM_COLOR: always
+  CARGO_INCREMENTAL: 0
+
+jobs:
+  run_examples:
+    name: Run Rust client examples
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    env:
+      TYCHO_API_KEY: ${{ secrets.TYCHO_API_KEY }}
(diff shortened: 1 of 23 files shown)


clients/rust/src/client.rs
```diff
@@ -709,6 +742,181 @@ where
...
+    }
+
+    /// Build an unsigned EIP-1559 `approve(router, amount)` transaction for the given token.
+    ///
+    /// 1. Calls [`info()`](Self::info) to resolve the router address (used as spender).
+    /// 2. Resolves nonce and EIP-1559 fees via `hints` (same semantics as
+    ///    [`swap_payload`](Self::swap_payload)).
+    /// 3. Encodes the `approve(router, amount)` calldata using the ERC-20 ABI.
+    /// 4. Optionally checks the current allowance if `params.check_allowance` is `true` and sets
+    ///    [`ApprovalPayload::is_needed`] accordingly.
```
kayibal: If it is determined, that the approval is not needed, the function below should simply return None. So the interface needs to be changed to require the approval.

clients/rust/src/client.rs
```diff
@@ -251,6 +251,36 @@ pub struct ExecutionOptions {
     pub storage_overrides: Option<StorageOverrides>,
 }
 
+// ============================================================================
+// APPROVAL PARAMS
+// ============================================================================
+
+/// Parameters for [`FyndClient::approval`].
+pub struct ApprovalParams {
```
kayibal, line 270: This struct is missing the spender. It should also support approvals against the Permit2 contract. By default though, the spender should be the router contract.

Note that ideally we add the user transfer type here and then the corresponding approval payload is built by the client. Do not reference the router or permit to by address directly.

clients/rust/src/client.rs
```diff
@@ -709,6 +742,181 @@ where
...
+            max_fee_per_gas,
+            max_priority_fee_per_gas,
+            gas_limit,
+            to: alloy::primitives::TxKind::Call(token_addr),
+            value: alloy::primitives::U256::ZERO,
+            input: AlloyBytes::from(calldata),
+            access_list: alloy::eips::eip2930::AccessList::default(),
+        };
+
+        // Optionally check current allowance.
```
kayibal: This should happen as soon as possible to avoid other unnecessary RPC calls such as nonce and gas price calls.

clients/rust/src/mapping.rs
```diff
@@ -288,6 +288,26 @@ impl From<dto::BlockInfo> for BlockInfo {
...
+        )));
+    }
+    if permit2.len() != 20 {
+        return Err(FyndError::Protocol(format!(
+            "permit2_address must be 20 bytes, got {}",
+            permit2.len()
+        )));
+    }
+    Ok(crate::types::InstanceInfo::new(router, permit2, dto.chain_id()))
+}
```
zizou0x: Can't this be a TryFrom?

clients/rust/src/client.rs
```diff
@@ -709,6 +761,187 @@ where
...
+    ///
+    /// The result is fetched at most once per [`FyndClient`] instance; subsequent calls return the
+    /// cached value without making a network request.
+    pub async fn info(&self) -> Result<&InstanceInfo, FyndError> {
+        if let Some(cached) = self.info_cache.get() {
+            return Ok(cached);
+        }
+        let fetched = self.fetch_info().await?;
+        let _ = self.info_cache.set(fetched);
+        Ok(self.info_cache.get().expect("just set"))
```
zizou0x: Can't we return the `fetched` directly? 

Looks very safe but this is still a panic in a lib (I think its very fine though, just not sure why we decide to read the cache instead of returning the value)

clients/rust/src/client.rs
```diff
@@ -709,6 +761,187 @@ where
...
+            None => self
+                .provider
+                .get_transaction_count(sender)
+                .await
+                .map_err(FyndError::Provider)?,
+        };
+
+        // Resolve EIP-1559 fees.
+        let (max_fee_per_gas, max_priority_fee_per_gas) =
+            match (hints.max_fee_per_gas(), hints.max_priority_fee_per_gas()) {
```
zizou0x, line 936: Seems weird to me that these two are settable individually on `SigningHints` but if only one of them is set we ignore it.

Shouldn't we still use it if only one of them is set (fetch the other one from rpc)?

.github/workflows/examples.yaml
```diff
@@ -0,0 +1,40 @@
...
+  CARGO_TERM_COLOR: always
+  CARGO_INCREMENTAL: 0
+
+jobs:
+  run_examples:
+    name: Run Rust client examples
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    env:
+      TYCHO_API_KEY: ${{ secrets.TYCHO_API_KEY }}
```
zizou0x, line 20: Depending on the key type we will be limited to 2 connections, should we make it `scale` (10 connections)?

zizou0x (approved):
LGTM, just some small comments

.github/workflows/examples.yaml
kayibal: It's my key, I think it already is

PR #127, opened 2026-03-23
feat(ENG-5578): add integration tests with recorded market replay

tvinagre:
## Summary

- Add `record-market` CLI tool that captures Tycho protocol stream updates and generates golden quote outputs by replaying through the full Fynd pipeline
- Add integration test suite (`fynd-core --test integration`) that replays recorded fixtures and verifies solution availability, quality baselines, derived data, and timing
- Worker pools and algorithm config loaded from `worker_pools.toml` to match production behavior
- Golden file stores exact derived data metrics (spot prices, pool depths, token prices counts) for deterministic replay assertions
- Shared golden types (`GoldenFile`, `TestScenario`, etc.) in `fynd_core::recording::golden`
- `parse_chain` moved to `fynd_core::types::constants` for reuse across crates
- Gas price captured from RPC as raw wei string (no lossy float conversion)
- VM-backed protocol states (UniswapV4) filtered during recording serialization
- Git LFS configured for compressed fixture files

## Test plan

- [x] `cargo +nightly fmt -- --check` passes
- [x] `cargo +nightly clippy --locked --all --all-features --all-targets -- -D warnings` passes
- [x] `cargo nextest run --workspace --locked --all-targets --all-features --bin fynd` — 416 passed, 0 failed
- [x] All 8 integration tests pass (41/44 scenarios successful, 3 BNB pairs expected NoRouteFound)
- [ ] CI runs on GitHub Actions

Generated with [Claude Code](https://claude.com/claude-code)


.github/workflows/nightly-canary.yaml
louise-poole, line 1: Cooool! And then if this fails? Can we make it alert us on slack? I'm scared it might silently fail and never be noticed/fixed.

fixtures/integration/README.md
```diff
@@ -0,0 +1,92 @@
...
+by the fynd-core integration tests.
+
+## Files
+
+- `market_recording.json.zst` — Zstd-compressed JSON recording of Tycho
+  protocol stream updates. Captured from live Tycho mainnet using the
+  `record-market` tool.
+- `golden_outputs.json` — Expected quote outputs for a set of canonical
+  trading pairs. Generated by replaying the recording through the full Fynd
+  pipeline (the same `record-market` tool).
```
louise-poole: Soooo - Fynd itself is determining what the expected outputs are? Then it'll always pass no? Or is the goal of these integration tests to fail a PR that breaks current performance?

What about if there is an improvement though? These should be "minimal expected quote outputs"?

fixtures/integration/README.md
louise-poole, line 1: I think this readme is misplaced. It should be split into 2 and be in fynd-core/src/recording and fynd-core/test/integration.

fixtures/integration/README.md
```diff
@@ -0,0 +1,92 @@
...
+```bash
+cargo test -p fynd-core --test integration
+```
+
+The tests replay the recorded market state through the full Fynd pipeline
+(TychoFeed → ComputationManager → WorkerPools → Router) and verify:
+
+- **Solution availability**: All pairs that succeeded in the golden baseline
+  also succeed during replay.
+- **Quality baseline**: Output amounts stay within 1% of the golden baseline.
```
louise-poole: Surely we don't want to restrict positive improvement?

fynd-core/src/feed/mod.rs
```diff
@@ -12,7 +13,7 @@ pub mod tycho_feed;
 
 /// Configuration for the TychoFeed.
 #[derive(Debug, Clone)]
-pub(crate) struct TychoFeedConfig {
+pub struct TychoFeedConfig {
```
louise-poole: Is it needed outside of the crate? I'l scared of @zizou0x shouting at us for making something `pub` (loooool)

fynd-core/src/recording/golden.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// Number of unique pools with at least one pool depth.
+    pub pool_depth_pools: usize,
+    /// Number of tokens with gas price conversions.
+    pub token_prices: usize,
+}
+
+/// Standard path for golden outputs in the fixtures directory.
+pub fn golden_file_path() -> PathBuf {
+    std::path::Path::new(env!("CARGO_MANIFEST_DIR"))
+        .join("../fixtures/integration/golden_outputs.json")
```
louise-poole: Should the fixtures not maybe live closer to the tests? What's convention here, I'm not too sure. Seems odd to have the test in fynd-core/tests/integration but have the fixtures up where they are.

fynd-core/src/recording/golden.rs
```diff
@@ -0,0 +1,155 @@
...
+
+/// A test scenario: a single token swap to quote.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TestScenario {
+    pub token_in: Address,
+    pub token_out: Address,
+    pub amount: BigUint,
+    pub side: OrderSide,
+    pub name: String,
+}
```
louise-poole: Why are some structs defined here and others in types.rs? If there's a clear reason then AG - at first glance it feels inconsistent.

fynd-core/src/types/constants.rs
```diff
@@ -58,3 +58,20 @@ pub fn native_token(chain: &Chain) -> Result<Address, UnsupportedChainError> {
...
+///
+/// Returns an error if the chain name is not recognized.
+pub fn parse_chain(chain: &str) -> Result<Chain, ParseChainError> {
+    let candidate = format!("\"{}\"", chain.to_ascii_lowercase());
+    serde_json::from_str::<Chain>(&candidate).map_err(|_| ParseChainError(chain.to_string()))
+}
+
+/// Error returned when a chain name string cannot be parsed.
+#[derive(Debug, Clone, thiserror::Error)]
+#[error("unsupported chain '{0}'. Try values like 'ethereum', 'base', 'unichain'")]
```
louise-poole: Would prefer we not list the chains here as this easily becomes outdated

fynd-core/tests/integration/derived_data_tests.rs
```diff
@@ -0,0 +1,75 @@
...
+    let derived = harness.derived_data.read().await;
+
+    assert!(derived.spot_prices().is_some(), "spot_prices should be computed");
+    assert!(derived.pool_depths().is_some(), "pool_depths should be computed");
+    assert!(derived.token_prices().is_some(), "token_prices should be computed");
+}
+
+/// Derived data metrics should exactly match the golden baseline.
+/// Since replay is deterministic (same recording → same derived data),
+/// any deviation indicates a real bug, not expected variance.
```
louise-poole: Really? I would say a positive deviation is allowed. What about if there's a tycho-simulation bug that we fix. Suddenly we get more successful spot prices compared to the latest golden baseline - this is a good thing but would fail the check?

fynd-core/tests/integration/solution_tests.rs
```diff
@@ -0,0 +1,178 @@
...
+                }
+            }
+        }
+    }
+
+    assert!(
+        regressions.is_empty(),
+        "quality regressions (>1% degradation):\n{}",
+        regressions.join("\n")
+    );
```
louise-poole, line 144: Oh nice! So it was just the doc that was misleading... for solution quality it is a gte check with a 1% degradation allowance. Cool cool.

tools/record-market/src/expected.rs
louise-poole, line 1: Maybe all these integration test stuff should be feature gated? Or collected into its own mod or something? I feel like a lot of the integration test specific stuff leaked into fynd-core etc and it should be more contained.

tools/record-market/src/recorder.rs
```diff
@@ -0,0 +1,209 @@
...
+        // EIP-1559: use max_fee_per_gas as the upper bound
+        other => {
+            tracing::warn!(?other, "non-legacy gas price, falling back to 10 gwei");
+            num_bigint::BigUint::from(10_000_000_000u64)
+        }
+    };
+    Ok(gas_price_wei.to_string())
+}
+
+async fn fetch_protocol_systems(
```
louise-poole: We should probably make a tycho RPC client fn to fetch all protocol systems like we have for tokens... I'll add a cooldown task.

.github/workflows/nightly-canary.yaml
tvinagre: Yes - good point. Will do that as the next step!

fixtures/integration/README.md
tvinagre: >Or is the goal of these integration tests to fail a PR that breaks current performance?
Yes, that's the primary goal - is to be able to catch errors introduced in future PRs.

If it is an improvement, the CI will flag and the user should regenerate the new specs, once validated that the output is actually good and not slop / overfit

fixtures/integration/README.md
tvinagre: Over-optimistic improvement can be bad - if they're good, we need to regenerate the assets and they will pass

fynd-core/src/feed/mod.rs
tvinagre: 🤣 - yes I launched a subagent only for that

fynd-core/tests/integration/derived_data_tests.rs
tvinagre: > Since replay is deterministic
I can rephrase this but this stands for "same recording, same code" should return the same result

> this is a good thing but would fail the check?
Yeah the flow would be:
1. dev would have a baseline to compare and will be very transparent about the improvements (wouldn't have to benchmark or do any manual testing).
2. If they look good, dev will have to update the baseline - new numbers will be easily spottable on the PR review for the reviewer.

fynd-core/src/feed/mod.rs
kayibal: I made this private recently, why does it have to be public to add a test?

tools/record-market/src/expected.rs
kayibal: Can you run the audit skill on these changes, por favor?

tools/record-market/src/recorder.rs
kayibal: I think tycho client already has it...

EDIT: I guess you meant he cycling through pages logic?

tools/fynd-swap-cli/src/main.rs
```diff
@@ -31,8 +31,9 @@ use fynd_client::{
     QuoteParams, SignedOrder, SigningHints, StorageOverrides,
 };
 use fynd_rpc::{
-    builder::{parse_chain, FyndRPCBuilder},
+    builder::FyndRPCBuilder,
     config::WorkerPoolsConfig,
+    parse_chain,
```
kayibal: Why did we move this?

tools/record-market/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+version.workspace = true
+edition = "2021"
+description = "Capture Tycho market state for integration testing"
+
+[dependencies]
+fynd-core.workspace = true
+tycho-simulation.workspace = true
+tycho-execution.workspace = true
+tokio = { workspace = true, features = ["full"] }
+clap = { version = "4", features = ["derive"] }
```
kayibal: surprised we can't take it from the workspace

tools/record-market/src/recorder.rs
```diff
@@ -0,0 +1,209 @@
...
+
+    Ok(MarketRecording {
+        metadata: RecordingMetadata {
+            chain: "ethereum".to_string(),
+            recorded_at_unix_s: std::time::SystemTime::now()
+                .duration_since(std::time::UNIX_EPOCH)
+                .expect("time went backwards")
+                .as_secs(),
+            fynd_version: env!("CARGO_PKG_VERSION").to_string(),
+            recording_duration_s: actual_duration,
```
kayibal: why the `_s` is this supposed to mean seconds??

tools/record-market/src/recorder.rs
```diff
@@ -0,0 +1,209 @@
...
+            num_updates: updates.len(),
```
kayibal: kind of pointless to have this attribute...unless you plean reading only the metadata.

tools/record-market/src/recorder.rs
```diff
@@ -0,0 +1,209 @@
...
+        .await
+        .map_err(|e| anyhow::anyhow!("failed to fetch gas price: {e}"))?;
+    let gas_price_wei = match block_gas_price.pricing {
+        GasPrice::Legacy { gas_price } => gas_price,
+        // EIP-1559: use max_fee_per_gas as the upper bound
+        other => {
+            tracing::warn!(?other, "non-legacy gas price, falling back to 10 gwei");
+            num_bigint::BigUint::from(10_000_000_000u64)
+        }
+    };
```
kayibal: Looks hacky... comment is not true?

fynd-core/src/recording/mod.rs
kayibal, line 1: This whole module, I feel should not be here, instead it can live directly in the tool? If exposed as a lib, I think it should be in tycho-simulation

fixtures/integration/README.md
kayibal: I guess some of this information should also be put into the knowledge documents for LLMs?

fynd-core/src/recording/golden.rs
```diff
@@ -0,0 +1,155 @@
...
+            self.amount.clone(),
+            self.side,
+            Address::zero(20), // dummy sender for testing
+        )
+    }
+}
+
+/// Expected output for a scenario.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct GoldenOutput {
```
kayibal: `ExpectedOutput` would be a way better name for this struct and you can probably extend this naming pattern up through to the module name.

fynd-core/src/recording/golden.rs
```diff
@@ -0,0 +1,155 @@
...
+    let content = std::fs::read_to_string(&path).expect("failed to read golden_outputs.json");
+    Some(serde_json::from_str(&content).expect("failed to parse golden_outputs.json"))
+}
+
+/// Load test scenarios from pairs.json (the canonical source of trading pairs).
+pub fn load_test_scenarios() -> Vec<TestScenario> {
+    let content = include_str!("../../../tools/benchmark/src/pairs.json");
+    let raw: serde_json::Value = serde_json::from_str(content).expect("failed to parse pairs.json");
+
+    let tokens: std::collections::HashMap<String, (Address, u32)> = raw["tokens"]
```
kayibal: unhandled panics here all over the place. Since this is a public library function, that is unfortunately not acceptable. But I think we should consider moving as much as possible from here to the tool binary.

fynd-core/src/recording/types.rs
```diff
@@ -0,0 +1,104 @@
...
+    tycho_client::feed::SynchronizerState,
+    tycho_core::simulation::protocol_sim::ProtocolSim,
+};
+
+/// A serializable mirror of [`Update`].
+///
+/// `Update` itself is `#[derive(Debug, Clone)]` only. All its fields
+/// implement `Serialize`/`Deserialize` individually (`Box<dyn ProtocolSim>`
+/// via `#[typetag::serde]`), so this wrapper just adds the derives.
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
kayibal: It would be a lot better to fix this quickly in tycho-simulation

kayibal (commented):
Partial review for now. Mostly concerned about the newly exposed api surface from this PR. I think a lot of this API would be better suited to be placed within tycho-simulation if exposed publically.

fixtures/integration/README.md
tamaralipows: Couldn't we use the price guard providers ad-hoc here instead of having a dedicated golden outputs json file? This way we aren't comparing fynd against fynd? Or is that too much in the CI?

fynd-core/tests/integration/derived_data_tests.rs
tamaralipows: Positive deviation can also indicate a nasty bug promising us 20 BTC for 1 ETH like the curve situation no?

fynd-core/tests/integration/solution_tests.rs
```diff
@@ -0,0 +1,178 @@
...
+                "unknown token should not produce a successful quote"
+            );
+        }
+        Err(_) => { /* Expected */ }
+    }
+}
+
+/// Quality: each pair's amount_out_net_gas should be within 1% of golden baseline.
+#[tokio::test]
+async fn test_quality_within_golden_baseline() {
```
tamaralipows: So it seems this is the most important test that we should be running in case of any code change? It took me a while to find this as an outsider - should we add this to some readme? It would be a lot easier to develop this way than to run every cargo test every time? Not sure where the best place to put this info is.

tools/record-market/src/golden.rs
```diff
@@ -0,0 +1,289 @@
...
+    #[serde(default)]
+    max_routes: Option<usize>,
+}
+
+fn default_num_workers() -> usize {
+    num_cpus::get()
+}
+fn default_min_hops() -> usize {
+    1
+}
```
tamaralipows: Would it not be better to increase this for integration testing purposes? I feel a number of bugs happened thx to sequential swaps for example on consecutive vm protocols?

tamaralipows (commented):
Thanks - some questions mostly for my understanding

.github/workflows/nightly-canary.yaml
dianacarvalho1: What is the motivation to run this periodically? 🤔 To make sure that Tycho's output is not weird? If yes, then we should support this for all chains we support 😕

fixtures/integration/README.md
```diff
@@ -0,0 +1,92 @@
...
+## Limitations
+
+- **Ethereum only**: The recording tool and tests are currently hardcoded to
+  `Chain::Ethereum`. Multi-chain support would require parameterizing the
+  chain throughout the pipeline.
+- **VM-backed protocols filtered**: Protocol states that depend on EVM engine
+  state (e.g., UniswapV4) cannot be serialized and are dropped during
+  recording. Pools for these protocols are registered as components but have
+  no simulation state, so they cannot compute spot prices or be used for
+  routing.
```
dianacarvalho1: oohhh 😞 how difficult would it be to serialise the vm protocols? (I know that it isn't in scope of this PR!)

.github/workflows/nightly-canary.yaml
dianacarvalho1: and this can give a lot of warnings if the price for tokens are just "normally" moving.. 😕

dianacarvalho1 (commented):
Thank you @tvinagre ! I gave a quick review - there are already a lot of good suggestions (that I agree 👍🏼)!

fixtures/integration/README.md
tvinagre: Too much in the CI IMO, and the price providers are not reliable in this case, as they provide offchain data, which is not quite comparable to onchain data.

The goal of the golden outputs is to have a frozen scenario to be able to compare the impact of the changes

tools/record-market/src/recorder.rs
tvinagre: Yeah - the goal was to be able to support reading the metadata without deserializing the whole updates array

.github/workflows/nightly-canary.yaml
tvinagre: @dianacarvalho1 - I believe it's a misunderstanding. This would fail only if Tycho changes its output format OR it is down.
We use the recording tool to regenerate the recordings, targets and run Fynd with the recorded data. This would also error if we are having non-deterministic values (which is most likely a bug since the data is the same).

fixtures/integration/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+This directory contains recorded market state and golden output baselines used
+by the fynd-core integration tests.
+
+## Files
+
+- `market_recording.json.zst` — Zstd-compressed JSON recording of Tycho
+  protocol stream updates. Captured from live Tycho mainnet using the
+  `record-market` tool.
+- `golden_outputs.json` — Expected quote outputs for a set of canonical
```
tvinagre: wrong, new file is `expected_outputs.json`

fynd-core/src/solver.rs
```diff
@@ -708,6 +711,167 @@ impl Solver {
...
+    /// VM-backed protocol states that couldn't be serialized will be absent from
+    /// the recording. Pools without states will be registered as components but
+    /// won't contribute to routing.
+    pub async fn from_recording(
+        chain: Chain,
+        updates: Vec<tycho_simulation::protocol::models::Update>,
+        pools: std::collections::HashMap<String, PoolConfig>,
+        gas_price_wei: Option<num_bigint::BigUint>,
+    ) -> Result<Self, SolverBuildError> {
+        use tycho_simulation::tycho_ethereum::gas::{BlockGasPrice, GasPrice};
```
tvinagre: These should be top-level imports since this is not feature-gated nor test-only imports.

fynd-core/src/solver.rs
```diff
@@ -708,6 +711,167 @@ impl Solver {
...
+
+        // Inject gas price (recorded value or default 10 gwei)
+        let gas_price =
+            gas_price_wei.unwrap_or_else(|| num_bigint::BigUint::from(10_000_000_000u64));
+        let block_number = market_data
+            .read()
+            .await
+            .last_updated()
+            .map(|b| b.number())
+            .unwrap_or(0);
```
tvinagre: Should at least add a warning that if failed to extract block number. Same for gas price

fynd-rpc/src/builder.rs
tvinagre, line 381: Double check if this move + the reexport were necessary

Cargo.toml
```diff
@@ -123,10 +125,14 @@ metrics = "0.24"
 metrics-exporter-prometheus = "0.16"
 
 # Tycho
-tycho-simulation = ">=0.254.0"
+# TODO: revert to crates.io version once tycho-simulation PR #568 merges
+tycho-simulation = { path = "/Users/thales/Environments/data_revenue/propeller_heads/tycho-simulation" }
```
tvinagre: Important! Revert once PR is merged

.github/workflows/nightly-canary.yaml
tvinagre: Added a Slack notification on failure: posts to the `SLACK_WEBHOOK_URL` repo secret with a link to the failed run, and degrades to a workflow warning if the secret isn't configured.

fixtures/integration/README.md
tvinagre: Split into `fynd-core/tests/integration/README.md` (tests) and `tools/record-market/README.md` (tool); `fixtures/integration/` is gone. Also added the tool, the `fynd-test-fixtures` crate, and the integration suite to the LLM knowledge docs (`.claude/CODEBASE.md`, `tools/CLAUDE.md`, `fynd-core/CLAUDE.md`).

fynd-core/src/feed/mod.rs
tvinagre: Reverted — `TychoFeedConfig` is `pub(crate)` again. Replay now goes through the `test-utils`-feature-gated `Solver::from_recording` instead of exposing feed internals.

fynd-core/src/recording/golden.rs
tvinagre: Moved to `fynd-core/tests/fixtures/` next to the tests; loaded via `CARGO_MANIFEST_DIR/tests/fixtures`.

fynd-core/src/recording/golden.rs
tvinagre: Consolidated into the `fynd-test-fixtures` crate with one module per concern (`recording.rs`, `expected.rs`, `scenarios.rs`); `golden.rs` and `types.rs` are deleted.

fynd-core/src/types/constants.rs
tvinagre: Done — the error is now `unsupported chain '{0}'` without enumerating chains.

fynd-core/tests/integration/derived_data_tests.rs
tvinagre: Doc comment rephrased: replay is deterministic (same recording + same code = same result), so exact equality is intentional — any intentional change regenerates the baseline via the ignored `regenerate_expected_outputs` test.

tools/record-market/src/expected.rs
tvinagre: Done on both counts. `Solver::from_recording` and the `Replay` error variant are behind `cfg(feature = "test-utils")`, and the recording types moved out of fynd-core into the unpublished `fynd-test-fixtures` crate. Also ran the security review skill on the branch diff — no findings: fixture deserialization is data-only and test-gated, no secrets are logged, and the workflows only pass quoted env secrets on trusted triggers.

tools/record-market/src/recorder.rs
tvinagre: Follow-up landed: the recorder now uses `fynd_rpc::protocols::fetch_protocol_systems` (added on main) instead of the hand-rolled pagination.

tools/fynd-swap-cli/src/main.rs
tvinagre: `parse_chain` moved to `fynd_core::types` so record-market and the integration harness can use it without a fynd-rpc dependency; fynd-rpc re-exports it for tools that depend only on fynd-rpc (benchmark). fynd-swap-cli no longer imports it at all.

tools/record-market/Cargo.toml
tvinagre: Fixed — `clap` and `tracing-subscriber` both come from the workspace now.

tools/record-market/src/recorder.rs
tvinagre: Renamed to `recording_duration_secs` / `recorded_at_secs`.

tools/record-market/src/recorder.rs
tvinagre: Removed `num_updates` from the metadata.

tools/record-market/src/recorder.rs
tvinagre: Replaced with `block_gas_price.effective_gas_price()`, which handles EIP-1559 properly — the 10 gwei fallback and the misleading comment are gone from the recorder.

fynd-core/src/recording/mod.rs
tvinagre: `fynd-core/src/recording/` is deleted. The types live in the unpublished `fynd-test-fixtures` crate and the tool, and `Update` serialization was fixed upstream in tycho-simulation so the wrapper is unnecessary.

fynd-core/src/recording/golden.rs
tvinagre: Renamed throughout: `ExpectedOutput`, `expected_outputs.json`, `expected.rs` — no "golden" references remain.

fynd-core/src/recording/golden.rs
tvinagre: Replaced with Result-returning `load_test_scenarios` / `load_expected_file`; the only remaining unwraps in the crate are inside `#[cfg(test)]` modules.

fynd-core/src/recording/types.rs
tvinagre: Fixed upstream (tycho-simulation #568) — the wrapper is gone and `Update` is serialized directly. After merging main the workspace requires tycho-simulation >=0.304.

fynd-core/tests/integration/solution_tests.rs
tvinagre: Documented in `fynd-core/tests/integration/README.md`: how to run the suite, a per-test table, and a debugging-failures section including when to regenerate baselines.

tools/record-market/src/golden.rs
tvinagre: The custom default is gone — expected outputs are now generated with the production `worker_pools.toml` (its SHA-256 is stored in the recording metadata to detect drift), so hop limits match production and multi-hop routes are exercised (the expected outputs include 3-swap routes).

fixtures/integration/README.md
tvinagre: Old README deleted; the tests README references `expected_outputs.json` correctly.

fynd-core/src/solver.rs
tvinagre: Moved to top-level imports (the gas types are feature-gated at top level).

fynd-core/src/solver.rs
tvinagre: Added `tracing::warn!` for both fallbacks (missing gas price → 10 gwei, missing block number → 0).

fynd-rpc/src/builder.rs
tvinagre: Checked — both are needed: `src/main.rs` and fynd-benchmark consume `parse_chain` via the fynd-rpc re-export, while fynd-core and record-market use it directly. The old `fynd_rpc::builder::parse_chain` path is the (declared) breaking change.

Cargo.toml
tvinagre: Done — back on crates.io; after merging main the workspace requires tycho-simulation >=0.304 (lock at 0.304.2).

fixtures/integration/README.md
tvinagre: Hmm Idk to be honest - I know some people tried but I have very little context

fynd-core/tests/integration/harness.rs
```diff
@@ -0,0 +1,53 @@
...
+        let pools = load_pools();
+
+        let solver = Solver::from_recording(Chain::Ethereum, recording.updates, pools, gas_price)
+            .await
+            .expect("failed to build solver from recording");
+
+        solver
+            .wait_until_ready(Duration::from_secs(120))
+            .await
+            .expect("solver not ready after 120s");
```
louise-poole, line 39: There are a lot of groups multiple tests that call this and with a potential 120s wait each... we risk making our tests excessively slow. Could we maybe reuse the same test harness in all tests/one per test group? Initialise it only once?

tools/record-market/src/recorder.rs
```diff
@@ -0,0 +1,158 @@
...
+    pub min_tvl: f64,
+    pub min_token_quality: i32,
+    pub traded_n_days_ago: u64,
+    pub rpc_url: Option<String>,
+}
+
+/// Connect to Tycho, capture raw Update messages for the configured
+/// duration, and return a MarketRecording.
+pub async fn record_market(opts: &RecordingOptions) -> anyhow::Result<MarketRecording> {
+    let chain = Chain::Ethereum;
```
louise-poole: Could we not make this chain agnostic? 👀

louise-poole (commented):
Only saw 2 smallish things - otherwise looks good to me!

fynd-core/tests/integration/README.md
```diff
@@ -0,0 +1,48 @@
...
+state through `Solver::from_recording()` and verify quote results against an expected
+baseline.
+
+## Running
+
+```bash
+cargo nextest run -p fynd-core --test integration
+```
+
+## Test Descriptions
```
tamaralipows, line 16: Perfect - thank you! Nothing else to say here other than Louise's comments

fynd-core/tests/integration/harness.rs
tvinagre: Good instinct, but the 120s here is a poll *deadline*, not a wait — `wait_until_ready` returns as soon as derived data is ready, which is ~15-20s in practice (you can see it in the per-test durations). Reuse isn't really available to us either: nextest runs each test in its own process, so a shared/static harness can never be hit twice, and the Solver's background tasks bind to each test's tokio runtime so sharing would be unsound even under plain `cargo test`. The one thing that would cut builds is collapsing the tests into fewer `#[tokio::test]` fns, which we'd rather not do (loses per-test granularity).

The saving grace is that the builds run in parallel: on the last CI run the whole Integration Tests job took 128s vs 216s for Compile & Test, so the suite isn't on the critical path. Happy to revisit (e.g. one harness per logical group via a shared binary) if it grows enough to become the bottleneck.

tools/record-market/src/recorder.rs
tvinagre: Done — `record-market` now takes a `--chain` flag (default `ethereum`). This reverses the earlier review ask to hardcode Ethereum, but the plumbing already supported it: `Solver::from_recording` was chain-parametric and `RecordingMetadata.chain` was already a string. The chain is written into the recording metadata and both replay paths (expected-output generation and the test harness) read it from there, so fixtures are self-describing and replay can't disagree with the recording.

Trading-pair scenarios are chain-specific, so they moved from a compile-time include of the benchmark's `pairs.json` to per-chain files at `fynd-core/tests/fixtures/pairs/<chain>.json` (seeded with the existing Ethereum set). Recording a new chain just needs a pairs file for it.

louise-poole (approved):
BLC!

PR #128, opened 2026-03-23
feat: remove blacklist from fynd

Troshchk:
- Move blocklist filtering upstream. Previously, blocklisted components were received from the Tycho stream and filtered out manually in TychoFeed::handle_tycho_message. Now the blocklist is passed directly to ComponentFilter when subscribing, so blocklisted components are never sent to us.                                                                                                         
- Rename blacklist → blocklist across config files, CLI flags, builder methods, and types.                                                                                                                                                             
- Remove blocklist unit tests from TychoFeed since the filtering no longer happens at that layer.                                                                                         
                                                                                                                                                                                                                                  
The blocklist filter itself lives in tycho-indexer, but blocklist.toml stays in fynd because it defines which pools fynd user considers problematic.


docs/guides/server-configuration.md
```diff
@@ -86,7 +86,7 @@ Run `cargo run --release -- serve --help` for the full list.
...
 | `-w, --worker-pools-config`        | `WORKER_POOLS_CONFIG` | `worker_pools.toml`        | Worker pools config file path                                                                                                                                                                                  |
-| `--blacklist-config`               | `BLACKLIST_CONFIG`    | `blacklist.toml`           | Blacklist config file path                                                                                                                                                                                     |
+| `--blocklist-config` | `BLOCKLIST_CONFIG` | `blocklist.toml` | Path to blocklist TOML config file. Components listed here are excluded from the Tycho stream.                                                                                                                                                                                     |
```
louise-poole: Could you keep the easy-to-read formatting that was here before?

docs/guides/server-configuration.md
```diff
@@ -142,12 +142,18 @@ To use a custom config file:
 cargo run --release -- serve -w my_worker_pools.toml
 ```
 
-## Blacklist (`blacklist.toml`)
+## Blocklist config
 
-Exclude specific components from routing, useful for components with known simulation issues (e.g., [rebasing tokens on UniswapV3 pools](https://docs.uniswap.org/concepts/protocol/integration-issues)):
+By default, Fynd loads `blocklist.toml` from the working directory. The shipped default excludes components with known simulation issues (e.g., [rebasing tokens on UniswapV3 pools](https://docs.uniswap.org/concepts/protocol/integration-issues)). Override with `--blocklist-config`:
```
louise-poole: Not sure I like us saying 'shipped' here. Let's just leave the word out entirely.

louise-poole (commented):
This likely can be simplified more if the tycho-simulation is updated to take a config file directly. So far looks good though.

docs/guides/server-configuration.md
louise-poole: Maybe the default blocklist file should move to the tycho-simulation repo? As the reason it's blocklisted is because of simulation issues and not Fynd-specific issues.

dianacarvalho1: Do we want to always blocklist the pools defined here? even for solver users (not Fynd)? Or are these issues only for Fynd? I think I'm missing a bit of context here

Troshchk: I have asked Thales and he mentioned, that the most common use cases are:
- this pool is faulty, so we want to to quickly remove this pool from the calculations
- user wants to avoid going through this pool - maybe they give fees to someone they don’t like, or this pool has a security flaw or doesn’t comply to any user’s internal compliance

fynd-core/src/feed/tycho_feed.rs
```diff
@@ -133,6 +130,11 @@ impl TychoFeed {
                     ComponentFilter::with_tvl_range(
                         self.config.min_tvl / self.config.tvl_buffer_ratio,
                         self.config.min_tvl,
+                    )
+                    .blocklist(
```
dianacarvalho1, line 134: isn't it `blocklist_components(..)`? [here](https://github.com/propeller-heads/tycho-simulation/pull/559/changes#diff-c785f19d1ae91486de58c60daaa738f1df83069e5f678de6120ccaaf2abb8c47R371)

fynd-core/src/feed/tycho_feed.rs
```diff
@@ -133,6 +130,11 @@ impl TychoFeed {
...
+                        self.config
+                            .blocklisted_components
+                            .clone(),
```
dianacarvalho1, line 137: and shouldn't it take a file path? 😵‍💫 I'm quite confused tbh

fynd-core/src/solver.rs
```diff
@@ -367,9 +364,9 @@ impl FyndBuilder {
         self
     }
 
-    /// Replaces the set of component addresses that are excluded from routing.
-    pub fn blacklisted_components(mut self, components: HashSet<String>) -> Self {
-        self.blacklisted_components = components;
+    /// Sets component IDs to exclude from the Tycho stream.
```
dianacarvalho1, line 383: why are we keeping this here? Can you please explain the motivation in the PR description?

dianacarvalho1: reviewing this is quite confusing.. we accept a file in one place and read it and then in another place we accept already a hashset? I'm really not following 😕 Can you explain please?

dianacarvalho1 (commented):
I'm quite confused 😬  Did you also have a look at this PR https://github.com/propeller-heads/fynd/pull/166 ?

fynd-core/src/feed/tycho_feed.rs
Troshchk: It is `blocklist` from `ComponentFilter` from [here](https://github.com/propeller-heads/tycho-indexer/pull/886/changes#diff-3655b79ddd4dfce0d05744283742e8776fd23c58cc6af1340d4c66094ffa255aR100)

Troshchk:
> I'm quite confused 😬 Did you also have a look at this PR #166 ?

I did see it and as far as I understand it was an adhoc change, that would be replaced with a change in this PR

fynd-core/src/feed/tycho_feed.rs
Troshchk: This add components to blocklist in the `ComponentFilter` and it takes an iterator of the component ids.
If it makes it easier to follow, I can add components to blocklist on the `ProtocolStreamBuilder` level, it is basically the same.

fynd-core/src/solver.rs
Troshchk: `blocklisted_components()` on FyndBuilder is the builder setter that lets callers configure which components to exclude. In `build()`, the value is passed to `TychoFeedConfig` ([here](https://github.com/propeller-heads/fynd/pull/128/changes#diff-b6bb00073fa62eb1e1888795fff3d9f237119987d020bc0a823f0b519e3e6b4eR473)), which forwards it to `TychoFeed` and that's where the actual filtering happens.

dianacarvalho1: aaaahhh ok then AG! I was just very confused when reviewing 😵‍💫

dianacarvalho1 (approved):
Looks good to me! let's just not merge this before we merge all the other related PRs and change the cargo.toml here!

Troshchk:
Since we now provide a blocklist file, do we still need `builtin_default()`?
Also, should tycho-simulation follow the same behavior for blocklist?
@kayibal tagging you here, since you added it

PR #130, opened 2026-03-24
refactor: reduce SharedMarketData read lock hold times

louise-poole:
The tokio::RwLock protecting SharedMarketData is write-preferring — queued writers block new readers. Several read paths held the lock during expensive computation (EVM simulation, route scoring), causing request drops during block updates.                                                                                                                                                                                        
                  
## Changes:                                                                                                                                                                                                            
-  find_best_route — Merged two separate lock acquisitions (gas price check + extract_subset) into a single brief lock. Route scoring now runs entirely on a local snapshot with no lock held.                       
- TokenGasPriceComputation — Both full and incremental paths now use snapshot-then-simulate: brief lock to call extract_subset, then simulate on the local copy. Store reads are no longer nested inside the market
  lock.                                                                                                                                                                                                               
- PoolDepthComputation — Unnested store reads from the market lock. Store data is read first, then a brief market lock acquires the simulation snapshot.
- SpotPriceComputation — Reordered store reads to happen before the market lock. Kept the lock held during computation since spot_price() is a HashMap lookup (microseconds per pool).                              
                                                                                                                                                                                                                      
## Benchmark
  Before → After (under load across 8 blocks):                                                                                                                                                                        
                                                                                                                     
<img width="500" height="135" alt="Screenshot 2026-03-24 at 12 38 07" src="https://github.com/user-attachments/assets/e6a6c5c2-e20d-41f6-a8e7-da9215062d68" />

### DIFF
--- a/fynd-core/src/derived/computations/spot_price.rs
+++ b/fynd-core/src/derived/computations/spot_price.rs
@@ -2,6 +2,11 @@
 //!
 //! Computes spot prices for all pools in both directions using `ProtocolSim::spot_price()`.
 //! Spot prices are the instantaneous exchange rates without slippage.
+//!
+//! `ProtocolSim::spot_price()` is cheap for all pool types: VM pools return a pre-computed
+//! HashMap lookup, native pools do simple arithmetic. This means the read lock hold time is
+//! negligible (microseconds per pool), so we hold it for the full loop rather than paying the
+//! cost of cloning simulation states via `extract_subset()`.
 
 use async_trait::async_trait;
 use itertools::Itertools;
@@ -46,9 +51,7 @@ impl DerivedComputation for SpotPriceComputation {
         store: &SharedDerivedDataRef,
         changed: &ChangedComponents,
     ) -> Result<Self::Output, ComputationError> {
-        let market = market.read().await;
-
-        // Start with existing prices (or empty for full recompute)
+        // Start with existing prices (or empty for full recompute).
         let mut spot_prices = if changed.is_full_recompute {
             SpotPrices::new()
         } else {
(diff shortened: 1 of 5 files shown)


fynd-core/src/derived/computations/token_gas_price.rs
```diff
@@ -396,39 +405,41 @@ impl TokenGasPriceComputation {
...
+
+            let component_ids: HashSet<ComponentId> = market_guard
+                .component_topology()
+                .keys()
+                .cloned()
+                .collect();
+            let snapshot = market_guard.extract_subset(&component_ids);
+
+            (snapshot, gas_price, block)
+        };
```
kayibal: We have this code block repeated twice. 

I am not sure what ids are returned by `component_topology()` is this small enough to justify the clone instead?

kayibal (approved):

fynd-core/src/derived/computations/token_gas_price.rs
louise-poole: `component_topology` returns all currently tracked components... So we're cloning the full market here. I'll improve this - nice catch!

fynd-core/src/derived/computations/spot_price.rs
```diff
@@ -58,43 +61,45 @@ impl DerivedComputation for SpotPriceComputation {
                 .spot_prices()
                 .cloned()
                 .unwrap_or_default();
-            // Remove spot prices for removed components
+            // Remove spot prices for removed components.
```
tvinagre, line 64: 🙏🏽 🙏🏽 🙏🏽

tvinagre (approved):
pure engineering! <3

tvinagre:
and grammar correctness!

PR #131, opened 2026-03-24
docs: Benchmark Bellman-Ford and most_liquid scaling across 8 protocols

kayibal:
## Summary

- Wires the missing `scale` subcommand into `fynd-benchmark` and adds `fynd-rpc` as a dependency
- Adds `OUTPUT_FILE` and `RPC_URL` env var support to `scripts/bench-remote.sh`
- Adds pool config TOMLs for all benchmark variants under `tools/benchmark/`
- Runs CPU scaling benchmarks for `bellman_ford` (2-hop, 3-hop) and `most_liquid` (2-hop, 3-hop) across 8 protocols on AWS `c7a.8xlarge`
- Documents results in `docs/benchmark-results.md` with setup tables, per-algorithm analysis, and a cross-algorithm comparison
- Adds shutdown tracing to `FyndRPC::run` and `scale.rs` to aid future debugging
- Adds `docs/investigation-ml-fluid-v1.md` — a guide for investigating the non-monotonic `most_liquid` 3-hop throughput caused by `fluid_v1`

## Key findings

- `bellman_ford` is **2.5–4× faster** than `most_liquid` at 3 hops across the same 8-protocol set
- `most_liquid` 2-hop scales cleanly and crosses 1000 req/s at 3 workers
- `most_liquid` 3-hop peaks at ~384 req/s (28 workers) with `fluid_v1` included; without it, scaling is monotonic up to ~462 req/s at 28 workers
- `fluid_v1` causes non-monotonic throughput in `most_liquid` 3-hop — root cause is under investigation

### DIFF
--- a/.claude/plans/investigation-ml-fluid-v1.md
+++ b/.claude/plans/investigation-ml-fluid-v1.md
@@ -39,7 +39,24 @@ Clean near-monotonic scaling. Peaks at 28 workers (462 req/s). P99 stays tightly
 |      28 |             400.13 |             94 |         238 |
 |      32 |             274.16 |            123 |         337 |
 
-Non-monotonic. Throughput dips at 16 and 32 workers with no consistent trend.
+Non-monotonic (partial run, 12–32 only). Throughput dips at 16 and 32 workers with no consistent trend.
+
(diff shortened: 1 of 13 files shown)


kayibal:
Reopening with correct branch name: ah/bellman-ford-benchmark

PR #135, opened 2026-03-25
feat: support Router v3 (tycho-execution=0.167.0)

dianacarvalho1:
Update encoder to the tycho-execution v3 interface. The new router requires ClientFeeParams in all calldata. Fees are optional per-request params signed offline by the fee receiver — Fynd never handles private keys.

fynd-core:
- Add ClientFeeParams struct to types/quote.rs
- Update encoder to build v3 calldata with client fee tuple (zero
  defaults when absent)
- Support all 9 router methods (single/sequential/split ×
  normal/Permit2/Vault). Split methods were missing
- Add integration tests with real TychoRouterEncoder

fynd-rpc-types:
- Add ClientFeeParams DTO with ToSchema and Into<fynd_core> conversion
- Rename UserTransferType::None → UseVaultsFunds

Rust client:
- Add ClientFeeParams with eip712_signing_hash()
- Add with_client_fee() builder on EncodingOptions
- Rename with_no_transfer() → with_vault_funds()
- Add swap_client_fee.rs example

TypeScript client:
- Add ClientFeeParams interface to types, client-fee.ts with
  clientFeeSigningHash() and withClientFee()
- Rename 'none' → 'use_vaults_funds', withNoTransfer → withVaultFunds

fynd-swap-cli:
- Add use-vaults-funds transfer type

Docs:
- New docs/guides/client-fees.md guide
- Update architecture, api-specs, swap-cli, and CLAUDE.md files

Misc:
- Rename tycho-solver to fynd everywhere (docker-compose, etc)


clients/rust/src/types.rs
```diff
@@ -97,6 +99,91 @@ impl PermitSingle {
...
+    /// `router_address` is the 20-byte address of the TychoRouter contract.
+    pub fn eip712_signing_hash(
+        bps: u16,
+        receiver: &Bytes,
+        max_contribution: &BigUint,
+        deadline: &BigUint,
+        chain_id: u64,
+        router_address: &Bytes,
+    ) -> Result<[u8; 32], crate::error::FyndError> {
+        use alloy::{
```
Troshchk: why do you import it here?

clients/rust/src/types.rs
```diff
@@ -97,6 +99,91 @@ impl PermitSingle {
...
+
+        let domain_type_hash = keccak256(
+            b"EIP712Domain(string name,string version,\
+uint256 chainId,address verifyingContract)",
+        );
+        let domain_separator = keccak256(
+            (
+                domain_type_hash,
+                keccak256(b"TychoRouter"),
+                keccak256(b"1"),
```
Troshchk, line 164: Would this version ever change?

clients/typescript/autogen/src/schema.d.ts
```diff
@@ -81,8 +81,43 @@ export interface components {
...
+             * @example 100
+             */
+            bps: number;
+            /**
+             * @description Unix timestamp after which the signature is invalid.
+             * @example 1893456000
+             */
+            deadline: string;
+            /**
+             * @description Maximum subsidy from the client's vault balance.
```
Troshchk, line 104: Should this be also mentioned above in the ClientFeeParams description?

docs/guides/client-fees.md
```diff
@@ -0,0 +1,96 @@
...
+
+## EIP-712 signing
+
+The fee receiver must sign a typed data hash with the following structure:
+
+| Field                   | Type      | Description                       |
+|-------------------------|-----------|-----------------------------------|
+| `clientFeeBps`          | `uint16`  | Fee in basis points (0–10,000)    |
+| `clientFeeReceiver`     | `address` | Address receiving the fee         |
+| `maxClientContribution` | `uint256` | Maximum subsidy from client vault |
```
Troshchk, line 40: Same as previous comment -- should there be any mention of what is parameter means in"how it works?"

Troshchk (commented):
Thank you @dianacarvalho1! 
I have a couple of smol questions/comments, but it looks good to me!

One thing I have a general comment about is `maxClientContribution`. I only saw the  description (“Maximum subsidy from client vault”), but didn’t notice it being mentioned in the high-level overview.

Maybe I missed it, but if not, it could be helpful to introduce it a bit earlier, rather than only in the field description.

clients/rust/examples/swap_client_fee.rs
```diff
@@ -0,0 +1,140 @@
...
+    let signer = PrivateKeySigner::random();
+    let sender = signer.address();
+
+    let client = FyndClientBuilder::new(FYND_URL, RPC_URL)
+        .with_sender(sender)
+        .build()
+        .await?;
+
+    // The fee receiver signs the fee params. In production this would be a
+    // separate key controlled by the integrator; here we reuse the sender key.
```
kayibal: Hmmm I wonder if we shouldn't just use a separate key here...

clients/rust/examples/swap_client_fee.rs
```diff
@@ -0,0 +1,140 @@
...
+    // [doc:start client-fee-rust]
+    // Compute the EIP-712 signing hash for the client fee.
+    let hash = ClientFeeParams::eip712_signing_hash(
+        FEE_BPS,
+        &fee_receiver_bytes,
+        &max_contribution,
+        &deadline,
+        1, // chainId = Ethereum mainnet
+        &router_address,
+    )?;
```
kayibal: This is a bit weird, no? Couldn't we make a struct here first and then call a method on that struct to get the hash? This way you save yourself basically constructing this twice...

clients/rust/examples/swap_client_fee.rs
```diff
@@ -0,0 +1,140 @@
...
+    let sig = signer
+        .sign_hash(&alloy::primitives::B256::from(hash))
+        .await?;
+    let signature = Bytes::copy_from_slice(&sig.as_bytes()[..]);
+
+    // Build encoding options with the client fee attached.
+    let fee =
+        ClientFeeParams::new(FEE_BPS, fee_receiver_bytes, max_contribution, deadline, signature);
+    let encoding_options = EncodingOptions::new(SLIPPAGE).with_client_fee(fee);
+    // [doc:end client-fee-rust]
```
kayibal, line 95: I'd recommend to go fully through with this and actually build a transaction that would execute if sender had WETH and then execute it.

fynd-rpc-types/src/lib.rs
```diff
@@ -132,8 +132,76 @@ pub enum UserTransferType {
...
+    )]
+    receiver: Bytes,
+    /// Maximum subsidy from the client's vault balance.
+    #[serde_as(as = "DisplayFromStr")]
+    #[cfg_attr(feature = "openapi", schema(value_type = String, example = "0"))]
+    max_contribution: BigUint,
+    /// Unix timestamp after which the signature is invalid.
+    #[serde_as(as = "DisplayFromStr")]
+    #[cfg_attr(feature = "openapi", schema(value_type = String, example = "1893456000"))]
+    deadline: BigUint,
```
kayibal: Does this need to be BigUint, could be u64 instead?

docs/guides/client-fees.md
```diff
@@ -0,0 +1,96 @@
+---
+icon: coins
+---
+
+# Client Fees
+
+The Tycho Router supports optional client fees — a percentage of the swap output charged on behalf of an integrator.
```
kayibal:
```suggestion
Fynd supports optional client fees — a percentage of the swap output charged on behalf of an integrator.
```

docs/guides/client-fees.md
```diff
@@ -0,0 +1,96 @@
...
+Fees are configured per-request via `ClientFeeParams` in the encoding options.
+
+## How it works
+
+1. The integrator chooses a fee in basis points (e.g. `50` = 0.5%) and a receiver address.
+2. The fee receiver signs an EIP-712 `ClientFee` message authorizing the fee parameters.
+3. The signed params are attached to the quote request via `EncodingOptions.clientFeeParams`.
+4. The router verifies the signature on-chain and deducts the fee from the swap output.
```
kayibal:
```suggestion
4. The router contract verifies the signature on-chain and deducts the fee from the swap output.
```

clients/rust/src/types.rs
dianacarvalho1: claude is gonna claude. so sorry! fixing it

clients/rust/src/types.rs
dianacarvalho1: only if we change in the TychoRouter contract too (meaning if we deploy a new version and want to up this)

clients/typescript/autogen/src/schema.d.ts
dianacarvalho1: conceptually we are bundling this in the fee scope as well (same in the actual router contract) 🤔 I think explaining it in the docs is good enough

docs/guides/client-fees.md
dianacarvalho1: adding it here!

kayibal (approved):

PR #136, opened 2026-03-25
feat: add Hyperliquid oracle price provider

louise-poole:
## Summary                                                
                                                                                                                                                                                                                      
  - Implement `HyperliquidProvider` — polls `POST /info` (`metaAndAssetCtxs`), caches USD oracle prices (weighted median across 8 CEXs) per asset symbol
  - Add shared `utils.rs` helpers: `normalize_symbol`, `check_staleness`, `expected_out_from_price`, `resolve_token`
  - Hardcode stablecoins (USDC, USDT, DAI, FRAX) at $1.00 since they aren't listed as perps                                                                                                                           
  - Cross-pair pricing via `price_in / price_out`
  - Polling interval is configurable (defaults to 3s)


fynd-core/src/price_guard/hyperliquid.rs
```diff
@@ -0,0 +1,494 @@
...
+
+    /// Overrides the Hyperliquid API endpoint URL
+    pub fn new_with_api_url(poll_interval: Duration, api_url: impl Into<String>) -> Self {
+        Self {
+            api_url: api_url.into(),
+            price_cache: Arc::new(RwLock::new(HashMap::new())),
+            token_cache: Arc::new(RwLock::new(HashMap::new())),
+            poll_interval,
+        }
+    }
```
tamaralipows: I don't imagine this will ever be necessary

fynd-core/src/price_guard/hyperliquid.rs
```diff
@@ -0,0 +1,494 @@
...
+use tycho_simulation::tycho_common::models::Address;
+
+use super::{
+    provider::{ExternalPrice, PriceProvider, PriceProviderError},
+    utils::{check_staleness, expected_out_from_price, normalize_symbol},
+};
+use crate::feed::market_data::SharedMarketDataRef;
+
+const DEFAULT_API_URL: &str = "https://api.hyperliquid.xyz/info";
+const API_URL_ENV_VAR: &str = "HYPERLIQUID_API_URL";
```
tamaralipows: Whoa I got confused seeing the env var _name_ assigned to a constant like this lol. I think it's fine to hardcode this in the init or perhaps not allow env vars at all here? We should make sure we keep these up to date in the code itself maybe

fynd-core/src/price_guard/hyperliquid.rs
```diff
@@ -0,0 +1,494 @@
...
+const DEFAULT_POLL_INTERVAL: Duration = Duration::from_secs(3);
+
+/// Stablecoins pegged to USD that aren't listed as Hyperliquid perps.
+const USD_STABLECOINS: &[&str] = &["USDC", "USDT", "DAI", "FRAX"];
```
tamaralipows: Not sure how we should keep this maintained - but there are a few other ones like RLUSD, USDE, USDS that I worry we will miss. Maybe it's fine if we just add a warning later in the docs, or make this configurable? Will think a bit more about it later

https://www.kraken.com/en-ca/categories/stablecoins

tamaralipows (commented):
Thank you! Only minor things that our team can take over! Commenting them so we don't forget

fynd-core/src/price_guard/hyperliquid.rs
```diff
@@ -0,0 +1,494 @@
...
+}
+
+/// Shared price cache. Key is the Hyperliquid asset name (e.g. "ETH").
+type PriceCache = Arc<RwLock<HashMap<String, OraclePrice>>>;
+
+/// Cached token metadata resolved from on-chain addresses.
+type TokenCache = Arc<RwLock<HashMap<Address, TokenInfo>>>;
+
+#[derive(Debug, Clone)]
+struct TokenInfo {
```
dianacarvalho1: why not just use the Token struct we already have? It's not that much bigger

fynd-core/src/price_guard/hyperliquid.rs
```diff
@@ -0,0 +1,494 @@
...
+            .await?;
+
+        let (meta, asset_ctxs): (Meta, Vec<AssetCtx>) = resp.json().await?;
+
+        let now_ms = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_millis() as u64;
+
+        // Build the next snapshot off-lock, then swap it in atomically.
```
dianacarvalho1: swap it?

fynd-core/src/price_guard/hyperliquid.rs
```diff
@@ -0,0 +1,494 @@
...
+        }
+
+        let mut cache = self
+            .price_cache
+            .write()
+            .map_err(|e| format!("price cache poisoned: {e}"))?;
+
+        *cache = new_cache;
+
+        Ok(count)
```
dianacarvalho1, line 281: this count is not used for anything right?

fynd-core/src/price_guard/utils.rs
```diff
@@ -11,6 +11,20 @@ use crate::feed::market_data::SharedMarketDataRef;
 /// Maximum age of price data before it is considered stale.
 pub const STALENESS_THRESHOLD: Duration = Duration::from_secs(30);
 
+/// Maps wrapped on-chain token symbols to their offchain exchange equivalents.
+///
+/// On-chain tokens like WETH, WBTC are listed as ETH, BTC on offchain exchanges.
+pub fn normalize_symbol(symbol: &str) -> String {
```
dianacarvalho1: 😬 we are going to have probleeeems

dianacarvalho1 (commented):
Thank you @louise-poole ! 🙏🏼 Only have small comments

fynd-core/src/price_guard/utils.rs
tamaralipows: So, I've looked into perhaps just removing the W. Doesn't seem smart though.

> The main risk is wstETH (wrapped staked ETH). It's a top-10 DeFi token, and stripping "W" gives "STETH". If Hyperliquid lists a STETH perp, we'd silently match the wrong price — wstETH trades at ~1.17x stETH  
  due to accumulated staking yield. Today's hardcoded version safely falls through to "WSTETH", which fails with PriceNotFound (loud and correct).  
  
  Considering Hyperliquid only has 100s of token pairs, not often changing, I think we could perhaps store a json list of mapping for each provider? And just auto claude-extract these mappings instead of doing it programmatically

tamaralipows: Actually just claude-analyzed all existing hyperliquid pairs, and the normalization list is quite small (and includes a concept about k-tokens that we missed). I think it can live in the file of the individual providers and moved away from the utils.

fynd-core/src/price_guard/hyperliquid.rs
tamaralipows: logging

fynd-core/src/price_guard/guard.rs
```diff
@@ -260,10 +260,7 @@ mod tests {
             _token_out: &Address,
             _amount_in: &BigUint,
         ) -> Result<ExternalPrice, PriceProviderError> {
-            Err(PriceProviderError::PriceNotFound {
-                token_in: "0xdead".to_string(),
-                token_out: "0xbeef".to_string(),
-            })
+            Err(PriceProviderError::TokenNotFound { address: "0xdead".to_string() })
```
Troshchk: Is the `struct PriceNotFoundProvider` needed if the `PriceNotFound` variant is gone?

fynd-core/src/price_guard/hyperliquid.rs
```diff
@@ -0,0 +1,542 @@
...
+    ) -> Result<ExternalPrice, PriceProviderError> {
+        let (sym_in, dec_in, scale_in) = self.resolve_token(token_in)?;
+        let (sym_out, dec_out, scale_out) = self.resolve_token(token_out)?;
+
+        let cache = self
+            .price_cache
+            .read()
+            .map_err(|e| PriceProviderError::Unavailable(format!("price cache poisoned: {e}")))?;
+
+        let price_in = cache
```
Troshchk, line 148: Here, I would expect `PriceNotFound`, so it is clear that this case means the provider does not have a price for the token, rather than `TokenNotFound`(as in `resolve_token`), which means the token itself could not be resolved. What do you think?

fynd-core/src/price_guard/hyperliquid.rs
```diff
@@ -0,0 +1,542 @@
...
+                    );
+                    count += 1;
+                }
+            }
+        }
+
+        for stable in USD_STABLECOINS.iter() {
+            new_cache
+                .entry(stable.clone())
+                .or_insert(OraclePrice { usd_price: 1.0, timestamp_ms: now_ms });
```
Troshchk, line 271: Is that always true? I think stablecoins can fluctuate slightly -- do we care about those fluctuations?

Troshchk (commented):
Thank you @louise-poole and @tamaralipows!
Looks very nice, I just have a couple of quetions

fynd-core/src/price_guard/guard.rs
dianacarvalho1: afaik there are more variants..

fynd-core/src/price_guard/hyperliquid.rs
```diff
@@ -0,0 +1,542 @@
...
+
+/// Maps on-chain token symbols to their Hyperliquid asset names.
+///
+/// Returns `(hyperliquid_symbol, price_scale)` where `price_scale` adjusts the oracle price
+/// to a per-token basis. For most tokens this is `1.0`. Curated from the Hyperliquid perp
+/// universe across three categories:
+///
+/// 1. **Wrapped native tokens** — on-chain "W"-prefixed wrappers of chain gas tokens. Tokens like
+///    WLD/WIF whose names happen to start with W are NOT included.
+/// 2. **k-prefix tokens** — Hyperliquid quotes these per 1,000 tokens to avoid tiny decimals.
```
dianacarvalho1, line 46: what does this mean?

fynd-core/src/price_guard/hyperliquid.rs
```diff
@@ -0,0 +1,542 @@
...
+///    `price_scale` is `0.001` so callers get the per-token price.
+fn normalize_symbol(symbol: &str) -> (String, f64) {
```
dianacarvalho1, line 48: should we try to keep this list up to date regularly? 😬

dianacarvalho1: how did you come up with it to begin with?

fynd-core/src/price_guard/hyperliquid.rs
```diff
@@ -0,0 +1,542 @@
...
+const DEFAULT_POLL_INTERVAL: Duration = Duration::from_secs(3);
+
+/// USD-pegged stablecoins. Loaded from `stable_usd.json` — shared across providers.
+static USD_STABLECOINS: LazyLock<HashSet<String>> = LazyLock::new(|| {
+    serde_json::from_str(include_str!("stable_usd.json")).expect("stable_usd.json is valid")
+});
+
+/// Cached oracle price entry (USD-denominated).
+#[derive(Debug, Clone)]
+struct OraclePrice {
```
dianacarvalho1, line 71: why Oracle? I thought we were calling them Price Providers

fynd-core/src/price_guard/hyperliquid.rs
```diff
@@ -0,0 +1,542 @@
...
+
+/// Shared price cache. Key is the Hyperliquid asset name (e.g. "ETH").
+type PriceCache = Arc<RwLock<HashMap<String, OraclePrice>>>;
+
+/// Cached token metadata resolved from on-chain addresses.
+type TokenCache = Arc<RwLock<HashMap<Address, Token>>>;
+
+/// Hyperliquid oracle price provider.
+///
+/// All oracle prices are in USD, so pricing any pair is `price_in / price_out`.
```
dianacarvalho1: I don't understand this sentence.. can you explain?

fynd-core/src/price_guard/hyperliquid.rs
dianacarvalho1: tbh I'm quite confused about the difference between TokenNotFound and PriceNotFound. Conceptually aren't they equivalent?? That the provider doesn't support this token -> so there is no price for it. I'm guessing that is why @tamaralipows removed the `PriceNotFound` variant

fynd-core/src/price_guard/hyperliquid.rs
dianacarvalho1: aaaahh I understand now. Maybe we can write:
````
/// All oracle prices are in USD, so pricing any Fynd pair is `oracle_price_in / oracle_price_out`.
````

File: fynd-core/src/price_guard/hyperliquid.rs [hunk 15]
```diff
@@ -0,0 +1,542 @@
...
+    market_data: SharedMarketDataRef,
+    client: Client,
+    poll_interval: Duration,
+    api_url: String,
+}
+
+impl HyperliquidWorker {
+    async fn run(&self) {
+        loop {
+            self.refresh_token_cache().await;
```
dianacarvalho1, line 189: do we really need to update the token cache every 3 seconds? I'm not sure how expensive this is tbh

fynd-core/src/price_guard/hyperliquid.rs
```diff
@@ -0,0 +1,542 @@
...
+    }
+}
+
+const API_URL: &str = "https://api.hyperliquid.xyz/info";
+const DEFAULT_POLL_INTERVAL: Duration = Duration::from_secs(3);
+
+/// USD-pegged stablecoins. Loaded from `stable_usd.json` — shared across providers.
+static USD_STABLECOINS: LazyLock<HashSet<String>> = LazyLock::new(|| {
+    serde_json::from_str(include_str!("stable_usd.json")).expect("stable_usd.json is valid")
+});
```
dianacarvalho1: can we have the consts at the top of the file please?

fynd-core/src/price_guard/hyperliquid.rs
dianacarvalho1: not really

fynd-core/src/price_guard/hyperliquid.rs
```diff
@@ -0,0 +1,542 @@
...
+                    new_cache.insert(
+                        asset.name.clone(),
+                        OraclePrice { usd_price, timestamp_ms: now_ms },
+                    );
+                    count += 1;
+                }
+            }
+        }
+
+        for stable in USD_STABLECOINS.iter() {
```
dianacarvalho1, line 268: why is this needed? 🤔 we are never trying to hop through any other usd stable coin to begin with..

fynd-core/src/price_guard/hyperliquid.rs
```diff
@@ -0,0 +1,542 @@
...
+                .expect("valid address");
+            let aave = make_token(aave_addr, "AAVE", 18);
+            cache.insert(aave.address.clone(), aave);
+        }
+
+        provider
+    }
+
+    #[test]
+    fn price_from_usd_oracle() {
```
dianacarvalho1: all these tests are not starting with `test` 😬

dianacarvalho1 (commented):
Thank you @tamaralipows ! 🙏🏼 I still have some comments

fynd-core/src/price_guard/hyperliquid.rs
dianacarvalho1: aaahhh got it! without this loop, any pair where one side is a stablecoin (USDC, USDT, DAI, etc.) would error out

fynd-core/src/price_guard/hyperliquid.rs
dianacarvalho1: I accept oracle. resolving the comment

fynd-core/src/price_guard/hyperliquid.rs
dianacarvalho1: From Claudio:

The "k" stands for kilo (1,000). Hyperliquid lists certain very low-priced meme coins as "kPEPE", "kSHIB", "kBONK", etc., where the oracle price is quoted per 1,000
  tokens rather than per 1 token.

  Why? Tokens like PEPE trade at ~$0.00000336 per token. That's awkward for order books and prone to rounding. By quoting per 1,000 tokens, the price becomes ~$0.00336
  — much more practical. This is Hyperliquid's version of Binance's "1000PEPE" or Bybit's "1000PEPE" contracts.

  Concrete example:

  ```
┌───────┬───────────────────────────────┬────────────────────────┬────────┐
  │ Token │ Hyperliquid oracle (per 1000) │ Actual per-token price │ Factor │
  ├───────┼───────────────────────────────┼────────────────────────┼────────┤
  │ kPEPE │ $0.003362                     │ $0.00000336            │ 1,000x │
  ├───────┼───────────────────────────────┼────────────────────────┼────────┤
  │ kSHIB │ $0.005934                     │ $0.00000592            │ 1,000x │
  ├───────┼───────────────────────────────┼────────────────────────┼────────┤
  │ kBONK │ $0.005817                     │ $0.00000581            │ 1,000x │
  └───────┴───────────────────────────────┴────────────────────────┴────────┘
```

fynd-core/src/price_guard/hyperliquid.rs
dianacarvalho1: The tokens are never deleted from the shared data, they can only be added or updated (very unlikely this happens tbh). Adding here a small optimization: check if the size of the tokens in shared data is higher than the current one we have - this way we avoid unnecessary clones

fynd-core/src/price_guard/hyperliquid.rs
Troshchk: I think these errors mean different things:
TokenNotFound = we can’t resolve the token at all (like a bad address)
PriceNotFound = token is valid, but this provider doesn’t have a price for it

If `PriceNotFound` is removed, I think `allow_on_token_price_not_found` becomes a bit dangerous or at least less precise, because then we may also allow cases where the token input itself is wrong.

That’s why I think keeping them separate is useful: same outcome, but different reason

What do you think?

dianacarvalho1: but then `TokenNotFound` is only about Fynd: that we can't find the token in Fynd's token registry.. not about the price providers right?

dianacarvalho1 (approved):

fynd-core/src/price_guard/hyperliquid.rs
Troshchk: There are two `TokenNotFound`'s:
PriceProviderError::TokenNotFound = Token address not found in the market data registry
GraphError::TokenNotFound = Token not found in graph

I want only talking about the PriceProviderError

Troshchk: Ah, you mean while resolving we also work with cache basically

Troshchk (approved):

fynd-core/src/price_guard/hyperliquid.rs
tamaralipows: I asked Claude to poll all token pairs, and give me the ones that need normalization

fynd-core/src/price_guard/hyperliquid.rs
tamaralipows: I think it's fine to do this wayyy less often! Good point!

PR #138, opened 2026-03-25
feat: add Binance WebSocket price provider

louise-poole:
## Summary
                                                                                                                                                                                                                      
  - Add `BinanceWsProvider` that subscribes to Binance `bookTicker` WebSocket streams for real-time bid/ask prices                                                                                                    
  - Pair discovery cross-references `SharedMarketData` tokens with Binance `/api/v3/exchangeInfo`, re-syncs every 60s
  - Price resolution: direct pair (bid) → reverse pair (1/ask) → intermediate routing through USDT/USDC/ETH/BTC, using ask prices on the buy-side leg
  - Reconnects with exponential backoff (5s → 60s max)                                            
  - Register as second default provider alongside Hyperliquid

### DIFF
--- a/fynd-core/src/price_guard/binance_ws.rs
+++ b/fynd-core/src/price_guard/binance_ws.rs
@@ -7,7 +7,7 @@
 
 use std::{
     collections::{HashMap, HashSet},
-    sync::{Arc, RwLock},
+    sync::{Arc, LazyLock, RwLock},
     time::{Duration, SystemTime, UNIX_EPOCH},
 };
 
@@ -47,6 +47,11 @@ const DEFAULT_EXCHANGE_INFO_URL: &str = "https://api.binance.com/api/v3/exchange
 
 /// Quote assets used for intermediate price routing.
 const INTERMEDIATE_ASSETS: &[&str] = &["USDT", "USDC", "ETH", "BTC"];
+
+/// USD-pegged stablecoins. Loaded from `stable_usd.json` — shared across providers.
+static USD_STABLECOINS: LazyLock<HashSet<String>> = LazyLock::new(|| {
+    serde_json::from_str(include_str!("stable_usd.json")).expect("stable_usd.json is valid")
+});
 
 /// How often to check for new tokens and subscribe to additional streams.
 const RESYNC_INTERVAL: Duration = Duration::from_secs(60);
@@ -387,6 +392,11 @@ impl BinanceWsWorker {
             return;
         };
         cache.insert(symbol.clone(), TickerData { bid, ask, timestamp_ms });
+
+        // When a ticker quotes against USDT or USDC, inject synthetic entries for
+        // all USD stablecoins so unlisted stablecoins (DAI, GHO, …) get priced
+        // as if they were USDT/USDC.
+        inject_stablecoin_tickers(&mut cache, symbol, bid, ask, timestamp_ms);
     }
 
     /// Fetches valid TRADING symbols from Binance exchange info.
@@ -534,6 +544,44 @@ fn discover_pairs(
     }
 
     pairs.into_iter().collect()
+}
+
+/// For a ticker like `ETHUSDT`, injects synthetic entries `ETHDAI`, `ETHGHO`, etc. for every
+/// USD stablecoin in `stable_usd.json` that doesn't already have a real Binance pair cached.
+/// Similarly handles `USDTETH`-style pairs (quote asset is a stablecoin).
+fn inject_stablecoin_tickers(
+    cache: &mut HashMap<String, TickerData>,
+    symbol: &str,
+    bid: f64,
+    ask: f64,
+    timestamp_ms: u64,
+) {
+    for quote in ["USDT", "USDC"] {
+        if let Some(base) = symbol.strip_suffix(quote) {
+            for stable in USD_STABLECOINS.iter() {
+                if stable == quote {
+                    continue;
+                }
+                let synthetic = format!("{base}{stable}");
+                cache
+                    .entry(synthetic)
+                    .or_insert(TickerData { bid, ask, timestamp_ms });
+            }
+            return;
+        }
+        if let Some(base) = symbol.strip_prefix(quote) {
+            for stable in USD_STABLECOINS.iter() {
+                if stable == quote {
+                    continue;
+                }
+                let synthetic = format!("{stable}{base}");
+                cache
+                    .entry(synthetic)
+                    .or_insert(TickerData { bid, ask, timestamp_ms });
+            }
+            return;
+        }
+    }
 }
 
 fn now_ms() -> u64 {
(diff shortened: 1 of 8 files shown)


fynd-core/src/price_guard/binance_ws.rs
```diff
@@ -0,0 +1,942 @@
...
+                    .as_ref()
+                    .map(|f| format!("code={}, reason={}", f.code, f.reason))
+                    .unwrap_or_else(|| "no reason".to_string());
+                debug!(reason, "Binance WebSocket close frame received");
+            }
+            _ => {}
+        }
+    }
+
+    fn handle_ticker(&self, text: &str) {
```
tamaralipows: Vague function name with no explanation what it does, could use a docstring or comment at least

fynd-core/src/price_guard/binance_ws.rs
```diff
@@ -0,0 +1,942 @@
...
+                    if let Err(e) = self.resync_subscriptions(&mut write).await {
+                        warn!(error = %e, "failed to resync Binance subscriptions");
+                    }
+                }
+            }
+        }
+    }
+
+    /// Handles a single WebSocket message.
+    fn handle_message(&self, msg: &Message) {
```
tamaralipows: Vague doc here

fynd-core/src/price_guard/binance_ws.rs
```diff
@@ -0,0 +1,942 @@
...
+                    self.handle_message(&msg);
+                }
+                _ = resync_interval.tick() => {
+                    self.refresh_token_cache().await;
+                    if let Err(e) = self.resync_subscriptions(&mut write).await {
+                        warn!(error = %e, "failed to resync Binance subscriptions");
+                    }
+                }
+            }
+        }
```
tamaralipows, line 365: So I understand this gathers two tasks: one to handle messages, and one to resync the token pairs at the hardcoded interval. Maybe with some better naming or comments we could make this clearer

tamaralipows (commented):
Thank you for taking this @louise-poole ! Just a few comments to make this easier if we ever need to debug in the future - again we can take it from here though 🙏

fynd-core/src/price_guard/binance_ws.rs
```diff
@@ -0,0 +1,942 @@
...
+}
+
+/// Shared ticker cache. Key is the uppercase Binance symbol (e.g. "ETHUSDT").
+type TickerCache = Arc<RwLock<HashMap<String, TickerData>>>;
+
+/// Shared token metadata cache.
+type TokenCache = Arc<RwLock<HashMap<Address, TokenInfo>>>;
+
+#[derive(Debug, Clone)]
+struct TokenInfo {
```
dianacarvalho1: same comment as in the Hyperliquid PR. Why not just use the regular `Token` struct?

fynd-core/src/price_guard/binance_ws.rs
```diff
@@ -0,0 +1,942 @@
...
+                    timestamp_ms: ticker.timestamp_ms,
+                });
+            }
+        }
+
+        None
+    }
+
+    /// Resolves a price between two symbols, trying direct/reverse first,
+    /// then routing through intermediate assets.
```
dianacarvalho1, line 201: is there a reason why we shouldn't route through intermediate assets in Hyperliquid too?

dianacarvalho1 (commented):
Thank you @louise-poole ! Only have smol comments

fynd-core/src/price_guard/binance_ws.rs
tamaralipows: IIUC, we don't need to. Hyperliquid's oracle prices are all USD-denominated. So pricing any pair is just `usd_price_in` / `usd_price_out`. USD is the intermediate for every single pair.  (Note each asset returns a single oraclePx value - not a pair price)                                                                                                  
                                                                                                                                                                                                                   
Binance is different: it has specific trading pairs (ETHUSDT, LINKBTC, etc.), so if there's no direct LINK/ETH pair, you have to explicitly route through USDT or BTC.   

Let me know if I missed something though

tamaralipows:
I rebased but it should still be fine to see the latest commits since they're signed with my name instead of Louise's

fynd-core/src/price_guard/binance_ws.rs
```diff
@@ -0,0 +1,1144 @@
...
+    /// updates, close frames are logged, and everything else is ignored.
+    fn handle_message(&self, msg: &Message) {
+        match msg {
+            Message::Text(text) => self.update_price_cache(text),
+            Message::Close(frame) => {
+                let reason = frame
+                    .as_ref()
+                    .map(|f| format!("code={}, reason={}", f.code, f.reason))
+                    .unwrap_or_else(|| "no reason".to_string());
+                debug!(reason, "Binance WebSocket close frame received");
```
Troshchk, line 378: Should we try to reconnect here immediately?

Troshchk (dismissed):
Thank you!
Looks good, have only one question

fynd-core/src/price_guard/binance_ws.rs
```diff
@@ -780,19 +847,19 @@ mod tests {
...
     }
 
-    // -- handle_message / handle_ticker tests (M4) --
+    // -- handle_message / update_price_cache tests (M4) --
 
     #[test]
-    fn handle_ticker_writes_to_cache() {
-        let ticker_cache: TickerCache = Arc::new(std::sync::RwLock::new(HashMap::new()));
-        let worker = make_worker(&ticker_cache);
+    fn update_price_cache_writes_to_cache() {
```
dianacarvalho1: these tests don't start with test either :(

dianacarvalho1 (commented):
Thank you @tamaralipows ! I see that the USDC like tokens are way more relevant here than in Hyperliquid!

fynd-core/src/price_guard/binance_ws.rs
dianacarvalho1: I think it is not necessary since reconnection is already handled at a higher level!

Troshchk: Could you explain the reconnecting flow, please?

As far as I understand, if the close frame is received, we will attempt to reconnect  in the `connect_and_stream` after waiting the backoff.
Does it mean, that for some time(until the end of the backoff) we will not have a websocket connection?
Is it expected?

Troshchk (commented):
Thank you @dianacarvalho1!
Looks good, have one question about the websocket reconnects

fynd-core/src/price_guard/binance_ws.rs
dianacarvalho1: yes I also think that is how it is working right now. And I don't think it's such a big issue to wait until the end of the backoff. We would still have the cached prices for that time and we don't expect the prices to move soooooooo fast

Troshchk: okok, makes sense, approving!

Troshchk (approved):

PR #147, opened 2026-03-26
feat: publish fynd as library and add Docker build pipeline

kayibal:
## Summary

- Add `src/lib.rs` that re-exports `fynd_core` and `fynd_rpc`, making the `fynd` crate usable as a library dependency for anyone building a custom Fynd-based CLI
- Add `description` and `license` to the root `Cargo.toml` so the crate is publishable to crates.io
- Add `fynd` as the final publish step in `release.yaml` (after `fynd-rpc`, which it depends on)
- Add `.github/workflows/docker.yaml`: builds the Docker image on every PR, builds and pushes to `ghcr.io/propeller-heads/fynd` on release with semver and `latest` tags

## Test plan

- [x] CI passes on this PR (Docker build step runs without push)
- [ ] On next release, `cargo publish --package fynd` succeeds and the Docker image is pushed to `ghcr.io/propeller-heads/fynd`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,23 +12,36 @@ RUN apt-get update && apt-get install -y \
 
 WORKDIR /app
 
-# Dependency caching layer: copy manifests and build deps first
+# Dependency caching layer: copy all workspace manifests and build deps first
 COPY Cargo.toml Cargo.lock ./
 COPY fynd-core/Cargo.toml fynd-core/
 COPY fynd-rpc/Cargo.toml fynd-rpc/
-RUN mkdir -p src fynd-core/src fynd-rpc/src && \
+COPY fynd-rpc-types/Cargo.toml fynd-rpc-types/
+COPY clients/rust/Cargo.toml clients/rust/
+COPY tools/benchmark/Cargo.toml tools/benchmark/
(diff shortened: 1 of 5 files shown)


tvinagre (approved):

Dockerfile
```diff
@@ -12,23 +12,40 @@ RUN apt-get update && apt-get install -y \
...
 
 # Copy real source and rebuild
 COPY src/ src/
 COPY fynd-core/src/ fynd-core/src/
 COPY fynd-rpc/src/ fynd-rpc/src/
-RUN touch src/main.rs src/lib.rs fynd-core/src/lib.rs fynd-rpc/src/lib.rs && cargo build --release
+COPY fynd-rpc-types/src/ fynd-rpc-types/src/
+RUN mkdir -p clients/rust/src tools/benchmark/src tools/fynd-swap-cli/src && \
+    echo "" > clients/rust/src/lib.rs && \
+    echo "fn main() {}" > tools/benchmark/src/main.rs && \
```
tvinagre, line 44: is this correct? 
I understand the previous step but shouldn't we have the real code here now? Or we're excluding the non-essential packages (tools / cli) on the dockerfile?

kayibal: Was thinking to only serve the fynd binary

PR #158, opened 2026-03-26
fix: polish crate docs, add READMEs, enforce missing_docs

kayibal:
All four Rust crates now have complete, navigable docs on docs.rs. Previously, `missing_docs` was not enforced, intra-doc links were broken, and the example code in `fynd-client` used a verbose address instantiation pattern and had a broken doctest.

**What changed:**
- Enforce `#![deny(missing_docs)]` across all crates; fill every gap
- Fix broken intra-doc links in `fynd-core` and `fynd-rpc`
- Expand the `encoding` module doc with links to the tycho-execution and Fynd encoding guide
- Add direct link to the API reference (`docs.fynd.xyz/reference/api`) in `fynd-rpc`
- Fix `fynd-client` doctest: succinct address instantiation, correct `Result` return context, crate-relative example path
- Unpin `fynd-client` README version (`"0.35"` → `"0"`)
- Add a Rust API Reference section to the GitBook sidebar linking to docs.rs for all four crates

Users integrating the Rust crates get accurate, fully linked API docs on docs.rs and can navigate from the GitBook site directly to crate-level reference.

### DIFF
--- a/clients/rust/README.md
+++ b/clients/rust/README.md
@@ -26,7 +26,7 @@ use num_bigint::BigUint;
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Start a local Fynd instance first: https://docs.fynd.xyz/get-started/quickstart
-    let client = FyndClientBuilder::new("http://localhost:3000", "http://localhost:8545")
+    let client = FyndClientBuilder::new("http://localhost:3000", "https://reth-ethereum.ithaca.xyz/rpc")
         .build()
         .await?;
 
--- a/clients/rust/README.md
+++ b/clients/rust/README.md
@@ -6,9 +6,8 @@ For documentation, guides, and API reference visit **<https://docs.fynd.xyz/>**.
 
 ## Installation
 
-```toml
-[dependencies]
-fynd-client = "0"  # see crates.io for the latest version
+```sh
+cargo add fynd-client
 ```
 
 ## Quick start
--- a/clients/rust/README.md
+++ b/clients/rust/README.md
@@ -2,10 +2,7 @@
 
 Rust client for the [Fynd](https://fynd.xyz) DEX router.
 
-Request swap quotes, build signable transaction payloads, and broadcast signed orders through
-the Fynd RPC API — all from a single typed interface.
-
-For documentation, guides, and API reference, visit **<https://docs.fynd.xyz/>**.
+For documentation, guides, and API reference visit **<https://docs.fynd.xyz/>**.
 
 ## Installation
 
@@ -16,48 +13,5 @@ fynd-client = "0.35"
 
 ## Quick start
 
-```rust,no_run
-use fynd_client::{
-    FyndClientBuilder, Order, OrderSide, QuoteOptions, QuoteParams,
-};
-use bytes::Bytes;
-use num_bigint::BigUint;
-
-#[tokio::main]
-async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    // Start a local Fynd instance first: https://docs.fynd.xyz/get-started/quickstart
-    let client = FyndClientBuilder::new("http://localhost:3000", "https://reth-ethereum.ithaca.xyz/rpc")
-        .build()
-        .await?;
-
-    let weth = Bytes::copy_from_slice(
-        alloy::primitives::address!("C02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2").as_slice(),
-    );
-    let usdc = Bytes::copy_from_slice(
-        alloy::primitives::address!("A0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48").as_slice(),
-    );
-    let sender = Bytes::copy_from_slice(
-        alloy::primitives::address!("d8dA6BF26964aF9D7eEd9e03E53415D37aA96045").as_slice(),
-    );
-
-    let quote = client
-        .quote(QuoteParams::new(
-            Order::new(
-                weth,
-                usdc,
-                BigUint::from(1_000_000_000_000_000_000u64), // 1 WETH
-                OrderSide::Sell,
-                sender,
-                None,
-            ),
-            QuoteOptions::default(),
-        ))
-        .await?;
-
-    println!("amount out: {}", quote.amount_out());
-    Ok(())
-}
-```
-
-See the [`examples/`](examples/) directory for complete, runnable programs covering ERC-20
-approvals, Permit2 transfers, and client fees.
+See the [quickstart guide](https://docs.fynd.xyz/get-started/quickstart) and the
+[`examples/`](examples/) directory for complete, runnable programs.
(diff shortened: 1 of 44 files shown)


clients/rust/src/lib.rs
```diff
@@ -1,48 +1,75 @@
...
-//! let weth: Bytes = Bytes::copy_from_slice(&[0xC0; 20]); // placeholder
-//! let usdc: Bytes = Bytes::copy_from_slice(&[0xA0; 20]); // placeholder
-//! let sender: Bytes = Bytes::copy_from_slice(&[0xd8; 20]); // placeholder
-//!
-//! let order = Order::new(
-//!     weth,
-//!     usdc,
-//!     BigUint::from(1_000_000_000_000_000_000u64), // 1 WETH (18 decimals)
-//!     OrderSide::Sell,
-//!     sender,
```
kayibal, line 40: I am sure there is a more succinct form to instantiate an alloy::Address?

clients/rust/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+Rust client for the [Fynd](https://fynd.xyz) DEX router.
+
+For documentation, guides, and API reference visit **<https://docs.fynd.xyz/>**.
+
+## Installation
+
+```toml
+[dependencies]
+fynd-client = "0.35"
```
kayibal: Is it possible to acoid pinning a version here?

fynd-core/src/algorithm/most_liquid.rs
```diff
@@ -47,7 +47,7 @@ pub struct MostLiquidAlgorithm {
 pub struct DepthAndPrice {
     /// Spot price (token_out per token_in) for this edge direction.
     pub spot_price: f64,
-    /// Liquidity depth in USD (or native token terms).
+    /// Liquidity depth denominated in the chain's native token.
     pub depth: f64,
 }
```
kayibal, line 52: @tvinagre can you confirm the denomination here please?

fynd-core/src/encoding/mod.rs
```diff
@@ -1 +1,2 @@
+/// Route encoder: converts solver output into ABI-encoded on-chain calldata.
```
kayibal, line 1: I think it would be great if the module docs of encoder could be a bit more detailed and link to the fynd encoding docs: https://docs.fynd.xyz/guides/encoding-options and tycho-execution, which powers it: https://docs.propellerheads.xyz/tycho/for-solvers/execution

fynd-core/src/feed/gas.rs
```diff
@@ -7,6 +7,7 @@ use tycho_simulation::{tycho_core::traits::FeePriceGetter, tycho_ethereum::gas::
 use crate::feed::{market_data::SharedMarketDataRef, DataFeedError};
 
 // TODO: Refactor gas price fetching into a `DerivedComputation`.
```
kayibal, line 9: @tvinagre is this still a TODO?

fynd-rpc/src/lib.rs
```diff
@@ -1,25 +1,26 @@
...
-//! # Main Components
-//!
-//! - **builder**: `FyndRPCBuilder` for assembling and configuring the HTTP server
-//! - **api**: HTTP endpoint handlers (`/v1/quote`, `/v1/health`) and OpenAPI documentation
-//! - **config**: Configuration types for pools, algorithms, and blacklists
+//! | Endpoint | Description |
+//! |---|---|
+//! | `POST /v1/quote` | Request an optimal swap route |
+//! | `GET /v1/health` | Data freshness and solver readiness |
+//! | `GET /v1/info` | Static instance metadata (chain ID, contract addresses) |
```
kayibal, line 16: It would be nice to reference the api spec here: https://docs.fynd.xyz/reference/api

fynd-rpc/README.md
```diff
@@ -0,0 +1,21 @@
...
+| Endpoint | Description |
+|---|---|
+| `POST /v1/quote` | Request an optimal swap route |
+| `GET /v1/health` | Data freshness and solver readiness |
+| `GET /v1/info` | Static instance metadata (chain ID, contract addresses) |
+
+## Quick start
+
+See the [server configuration guide](https://docs.fynd.xyz/guides/server-configuration) and the
+[quickstart](https://docs.fynd.xyz/get-started/quickstart).
```
kayibal, line 22: Same here probably a good idea to link the api reference:

https://docs.fynd.xyz/reference/api

clients/rust/src/lib.rs
```diff
@@ -1,48 +1,72 @@
...
 //! ```rust,no_run
-//! # use fynd_client::{FyndClient, FyndClientBuilder};
+//! # use fynd_client::FyndClientBuilder;
 //! # #[tokio::main] async fn main() -> Result<(), Box<dyn std::error::Error>> {
+//! // Fynd default: http://localhost:3000  RPC default (Anvil): https://reth-ethereum.ithaca.xyz/rpc
 //! let client = FyndClientBuilder::new(
-//!     "https://rpc.fynd.exchange",
-//!     "https://mainnet.infura.io/v3/YOUR_KEY",
+//!     "http://localhost:3000",
+//!     "https://reth-ethereum.ithaca.xyz/rpc",
```
dianacarvalho1, line 24: I think in other places we are using `https://eth.llamarpc.com`. Should we be consistent?

clients/rust/README.md
```diff
@@ -0,0 +1,17 @@
...
+Rust client for the [Fynd](https://fynd.xyz) DEX router.
+
+For documentation, guides, and API reference visit **<https://docs.fynd.xyz/>**.
+
+## Installation
+
+```toml
+[dependencies]
+fynd-client = "0"  # see crates.io for the latest version
+```
```
dianacarvalho1: maybe just `cargo add fynd-client` then? 👀

fynd-core/src/encoding/mod.rs
```diff
@@ -1 +1,7 @@
...
+///
+/// Wraps [tycho-execution](https://docs.propellerheads.xyz/tycho/for-solvers/execution) to
+/// produce ABI-encoded calldata for single and sequential swaps, with and without Permit2. See
```
dianacarvalho1: single, sequential and split swaps

dianacarvalho1: I know we don't have an algo that does splits yet but the encoding already supports it

dianacarvalho1 (approved):
Thank you @kayibal ! 🙏🏼 only have smol comments

fynd-core/src/algorithm/most_liquid.rs
tvinagre: I've checked the code and I think we have a bug..  depth is in raw units of sell_token - which is fine if we're comparing same In-out token paths, but that is not the case. I'm investigating this further and will open a PR soon if the bug exists

fynd-core/src/feed/gas.rs
tvinagre: Yes

tvinagre: Unless we think it's irrelevant. 
It's a discussion if `gas` should be treated as derived data or not - and if it should block solving or not

PR #160, opened 2026-03-27
feat: harden public API across fynd-core, fynd-rpc, fynd-rpc-types, fynd-client

kayibal:
## Summary

Audited all four library crates for backwards-compatibility risks and applied
targeted hardening. Changes are grouped by type across crates.

- **`#[non_exhaustive]` on all `OrderSide` enums** — `Buy` is explicitly documented as coming; without this attribute adding it would be semver-breaking (3 crates: `fynd-core`, `fynd-rpc-types`, `fynd-client`)
- **`#[non_exhaustive]` on `NoPathReason` and `PriceProviderError`** — both are extensible error enums with no exhaustive downstream matches; `PriceProviderError` also gets variant-level attributes on its three struct variants (`fynd-core`)
- **`#[must_use]` on builder and config types** — silently discarding a consuming `with_*` return is always a bug; covers `PoolConfig`, `AlgorithmConfig`, `PriceGuardConfig`, `WorkerPoolRouterConfig`, `QuoteOptions`, `EncodingOptions` in `fynd-core` and `QuoteRequest`, `Order`, `EncodingOptions`, `QuoteOptions`, `ErrorResponse` in `fynd-rpc-types`
- **Remove dead `GAS_PRICE_DEPENDENCY_ID` constant** and narrow `feed::gas` to `pub(crate)` (`fynd-core`)
- **Remove unused crate-root re-exports** in `fynd-rpc` (`ApiError`, `AppState`, `HealthStatus`, `ExternalPrice`, `PriceProvider`, `PriceProviderError`) — none consumed by any dependent crate
- **Remove `POOL_*` constants from `fynd-rpc`'s public defaults** — only used in own tests; updated tests to import from `fynd_core::solver::defaults` directly

## Test plan

- [x] `./check.sh` — 503 tests pass, 3 skipped (pre-existing)
- [x] All 7 commits pass the pre-commit hook individually
- [x] No wildcard match arms added — all `OrderSide` matches are within their defining crates

### DIFF
--- a/fynd-core/src/algorithm/mod.rs
+++ b/fynd-core/src/algorithm/mod.rs
@@ -38,6 +38,7 @@ use crate::{
 };
 
 /// Configuration for an Algorithm instance.
+#[must_use]
 #[derive(Debug, Clone)]
 pub struct AlgorithmConfig {
     /// Minimum hops to search (must be >= 1).
--- a/fynd-core/src/algorithm/mod.rs
+++ b/fynd-core/src/algorithm/mod.rs
@@ -257,6 +257,7 @@ pub enum AlgorithmError {
 }
 
 /// Reason why no path was found between tokens.
+#[non_exhaustive]
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum NoPathReason {
     /// Source token not present in the routing graph.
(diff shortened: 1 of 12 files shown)


fynd-core/src/algorithm/mod.rs
```diff
@@ -38,6 +38,7 @@ use crate::{
 };
 
 /// Configuration for an Algorithm instance.
+#[must_use]
```
louise-poole, line 41: I quite like this 👀 protecting our users - nice!

louise-poole (approved):
I like it 👌

PR #163, opened 2026-03-30
fix: remove runtime workaround for tycho-execution encoding

dianacarvalho1:
tycho-execution no longer uses block_in_place, so callers don't need to provide a dedicated multi-thread runtime for encoding. This removes the encoding_rt field, the custom Drop impl, and the std::thread::scope dispatch from Encoder -- encode_solutions is now called directly.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>


tamaralipows (approved):
Thank you!

PR #166, opened 2026-03-30
fix(config): embed default configs in binary to fix cargo publish

kayibal:
## Summary

- Fixes `cargo publish` failure for `fynd-rpc`: `include_str!("../../worker_pools.toml")` resolved correctly in the workspace but failed during crate verification (the file is outside the package tree)
- Inlines both `worker_pools.toml` and `blacklist.toml` as string constants in `fynd-rpc/src/config.rs`
- Adds `WorkerPoolsConfig::builtin_default()` and `BlacklistConfig::builtin_default()`
- Falls back to built-in defaults in `src/main.rs` when the default path is missing, with a `warn!` log; custom paths that don't exist still fail fast
- Adds unit tests for both `builtin_default()` methods

The repo-root `.toml` files stay as the user-facing example configs; a comment in each constant notes they should be kept in sync.

Verified with `cargo publish --dry-run --package fynd-rpc`.

## Test plan

- [ ] `cargo test -p fynd-rpc` passes
- [ ] Re-run the release workflow and confirm `cargo publish --package fynd-rpc` succeeds
- [ ] Start `fynd serve` in a directory without either `.toml` file — confirm it warns and starts

### DIFF
--- a/fynd-rpc/src/config.rs
+++ b/fynd-rpc/src/config.rs
@@ -91,7 +91,43 @@ pub struct BlacklistConfig {
     components: HashSet<String>,
 }
 
+/// The default blacklist configuration embedded at compile time.
+///
+/// Keep in sync with the repo-root `blacklist.toml` (the user-facing example config).
+/// Cannot use `include_str!` here because `cargo publish` verifies the crate in isolation,
+/// and the file lives outside the `fynd-rpc` package directory.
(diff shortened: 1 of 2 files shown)


michcio1234 (approved):
blind approval

dianacarvalho1 (approved):
yikes this is not so nice 😬 but let's go for now

PR #167, opened 2026-03-30
fix(swap-cli): make dry-run work without real funds

kayibal:
## Summary

- Default sell/buy tokens swapped to WETH→USDC (1 WETH = `1e18` atomic units)
- Set gas limit in `SigningHints` to skip gas estimation, it failed due to allowance slots not yet set
- Dry-run sender set to `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045` (vitalik.eth) so `eth_call` simulations have sufficient ETH for gas; signatures are not validated in dry-run mode
- Balance/allowance slot detection probe switched from `U256::MAX` to `0xdeadbeef`: USDC V2.2 packs a blacklist flag in bit 255 and masks it in `balanceOf`, so the exact-equality check against `MAX` always failed; a low-bit sentinel survives any upper-bit masking scheme
- Remove embedded solver (`--tycho-url` and related flags) from `fynd-swap-cli` to drop `fynd-rpc`, `actix-web`, `tycho-simulation`, and `toml` from its direct dependencies to speed up build time
- Add `fynd-swap-cli` binary to the Docker image
- Update `docs/guides/swap-cli.md` to reflect current flags and defaults; add `tools/fynd-swap-cli/README.md`

### DIFF
--- a/docs/guides/swap-cli.md
+++ b/docs/guides/swap-cli.md
@@ -20,17 +20,27 @@ cargo build --release -p fynd-swap-cli
 # binary: target/release/fynd-swap-cli
 ```
 
+{% hint style="info" %}
+Alternatively, the binary is included in the official Docker image (`ghcr.io/propeller-heads/fynd`) and can be invoked with `--entrypoint fynd-swap-cli`.
+{% endhint %}
+
 ## Dry-run a swap (ERC-20)
 
-By default, `fynd-swap-cli` runs a **dry-run**: it generates an ephemeral key and injects ERC-20 storage overrides so the simulation succeeds without any real funds or wallet approvals.
+By default, `fynd-swap-cli` runs a **dry-run**: it uses a well-funded sender address and injects ERC-20 storage overrides so the simulation succeeds without any real funds or wallet approvals.
 
 ```bash
 export RPC_URL=https://your-rpc-provider.com/v1/your_key
 
+./target/release/fynd-swap-cli
+```
+
+This sells 1 WETH for USDC using the defaults. Pass explicit tokens and amounts to customise:
+
+```bash
 ./target/release/fynd-swap-cli \
-  --sell-token  0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48 \
-  --buy-token   0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2 \
-  --sell-amount 1000000000
+  --sell-token  0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2 \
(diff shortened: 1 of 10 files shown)


fynd-rpc/src/config.rs
```diff
@@ -17,27 +17,11 @@ use serde::{Deserialize, Serialize};
 /// Cannot use `include_str!` here because `cargo publish` verifies the crate in isolation,
 /// and the file lives outside the `fynd-rpc` package directory.
 const DEFAULT_WORKER_POOLS_TOML: &str = r#"
-[pools.most_liquid_2_hops_fast]
```
tvinagre, line 20: Any reason to not do 3hops? 2 is kinda limited and already super fast with ML

tools/fynd-swap-cli/src/erc20.rs
```diff
@@ -26,6 +26,13 @@ sol! {
 
 pub const MAX_PROBE_SLOT: u64 = 20;
 
+/// Sentinel written to a candidate storage slot during balance/allowance probing.
+///
+/// Deliberately avoids high bits so tokens that mask upper bits (e.g. USDC packs a
```
tvinagre, line 31: we should prob improve our token balance / allowance detection to use this (unrelated)

tvinagre: or merge them

docs/guides/swap-cli.md
```diff
@@ -20,17 +20,27 @@ cargo build --release -p fynd-swap-cli
...
+```
+
+This sells 1 WETH for USDC using the defaults. Pass explicit tokens and amounts to customise:
+
+```bash
 ./target/release/fynd-swap-cli \
-  --sell-token  0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48 \
-  --buy-token   0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2 \
-  --sell-amount 1000000000
+  --sell-token  0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2 \
```
tvinagre, line 41: we should never sell eth 😭

fynd-rpc/src/config.rs
tvinagre: (this comment applies for the configfile, not this)

tvinagre (approved):

fynd-rpc/src/config.rs
kayibal: I think the default should be as low latency and have as much converage as possible. This is the best tradeoff, but I am happy to discuss this further. The more hops the less likely you get a bad pool or bad settlement.

PR #168, opened 2026-03-30
fix(fynd-rpc-types): remove unconditional tycho-simulation dependency

kayibal:
## Summary

`fynd-rpc-types` previously pulled in `tycho-simulation` unconditionally, dragging ~575 crates
(revm, alloy, AWS SDK, ekubo, etc.) into every crate that depends on it — including `fynd-client`
and `fynd-swap-cli`.

The fix replaces the two tycho primitives used by the wire types (`Address = Bytes`, a
`bytes::Bytes` wrapper with hex serde) with a local equivalent that has identical wire format:
`"0x{lowercase hex}"` on serialize, accepts with or without `0x` prefix on deserialize.

- Add `fynd_rpc_types::Bytes(pub bytes::Bytes)` + `type Address = Bytes` with a self-contained
  serde module. No new dependencies.
- Move `tycho-simulation` to optional, only enabled by the `core` feature (used by `fynd-rpc`
  for DTO↔domain conversions).
- Add `From<TychoBytes>/From<Bytes>` bridge impls inside the `core` feature gate for zero-copy
  conversion (both types expose their inner `bytes::Bytes` as a `pub` field).
- Remove `tycho-simulation` from `fynd-client` entirely; update `mapping.rs` to use `dto::Bytes`
  directly.
- Add `.into()` at the conversion sites where local `Bytes` is passed to `fynd-core` constructors
  expecting `TychoBytes`.
- Fix a pre-existing bug: clap `env` feature was missing from the workspace dep, causing
  `fynd-swap-cli` to fail to compile in isolation due to feature unification masking it.

Wire-format safety: 25 tests added first to pin the exact JSON representation of every
`Address`/`Bytes` field across all public DTO types. All 119 tests pass.

**Result:** `fynd-client` dep tree is now free of `tycho-simulation`. Combined with the
`ah/fix-swap-cli` changes (removing the embedded solver), `fynd-swap-cli` cold-build time
drops from ~3 minutes to **21 seconds** (fresh package cache, no artifacts).

### DIFF
--- a/fynd-rpc-types/src/lib.rs
+++ b/fynd-rpc-types/src/lib.rs
@@ -1229,381 +1229,139 @@ fn generate_order_id() -> String {
 }
 
 // ============================================================================
-// SERIALIZATION TESTS
+// WIRE FORMAT TESTS
 // ============================================================================
 //
-// These tests pin the exact JSON wire format for every field that uses `Address`
-// or `Bytes`. They must pass before and after any dependency swap, so we know the
-// wire format hasn't changed.
+// These tests pin the JSON wire format for the key API types. They catch
+// field renames, enum case changes, wrong numeric types, and structural
+// changes that would silently break API clients.
 
 #[cfg(test)]
-mod serde_tests {
+mod wire_format_tests {
     use num_bigint::BigUint;
 
     use super::*;
 
-    // ── helpers ──────────────────────────────────────────────────────────────
-
-    /// Build an Address from a 20-byte array.
-    fn addr(raw: [u8; 20]) -> Address {
-        Address::from(raw)
-    }
-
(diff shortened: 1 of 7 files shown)


fynd-rpc-types/src/lib.rs
```diff
@@ -1154,8 +1619,27 @@ fn generate_order_id() -> String {
 ///   violate the orphan rule.)
 #[cfg(feature = "core")]
 mod conversions {
+    use tycho_simulation::tycho_core::Bytes as TychoBytes;
```
kayibal, line 1380: Please use `tycho_simulation::tycho_common` instead, since core is basically `deprecated`

tvinagre (approved):

PR #171, opened 2026-03-31
feat: Enable using PriceGuard in FyndBuilder

tamaralipows:
sorry for chained PRs :( 

fynd-rpc/fynd-client:
- On our hosted endpoint, PriceGuard is running with default providers, and caching provider prices, though the default config disables the checks by default. If the user wants to submit their order with PriceGuard checks, they need to enable it via the config. 

fynd-core: Users running their own fynd instance:
- Lets you define your own price provider via the FyndBuilder, starts all providers in the FyndBuilder.build method
- Via the CLI, the user can pass the fynd config args. I am assuming that the user will use the FyndBuilder directly if they want to register custom providers. This is not supported in any way via CLI.
- The meat is in the WorkerPoolRouter. So we pass the guard+config from the FyndBuilder into there.

TODO:
- Double-check reasoning above
- Double-check that nothing is missing on the client side (backwards compatibility?)
- Double-check error handling
- Include integration test? 

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/clients/rust/src/types.rs
+++ b/clients/rust/src/types.rs
@@ -269,6 +269,7 @@ pub struct EncodingOptions {
     pub(crate) permit: Option<PermitSingle>,
     pub(crate) permit2_signature: Option<Bytes>,
     pub(crate) client_fee_params: Option<ClientFeeParams>,
+    pub(crate) price_guard: Option<PriceGuardConfig>,
 }
 
 impl EncodingOptions {
@@ -283,6 +284,7 @@ impl EncodingOptions {
             permit: None,
             permit2_signature: None,
             client_fee_params: None,
+            price_guard: None,
         }
     }
 
@@ -320,6 +322,14 @@ impl EncodingOptions {
     /// Attach client fee configuration with a pre-signed EIP-712 signature.
     pub fn with_client_fee(mut self, params: ClientFeeParams) -> Self {
         self.client_fee_params = Some(params);
+        self
+    }
+
+    /// Override server-side price guard defaults for this request.
+    ///
+    /// Fields left as `None` in [`PriceGuardConfig`] inherit the server defaults.
+    pub fn with_price_guard(mut self, config: PriceGuardConfig) -> Self {
+        self.price_guard = Some(config);
         self
     }
 }
@@ -454,7 +464,6 @@ pub struct QuoteOptions {
     pub(crate) min_responses: Option<usize>,
     pub(crate) max_gas: Option<BigUint>,
     pub(crate) encoding_options: Option<EncodingOptions>,
-    pub(crate) price_guard: Option<PriceGuardConfig>,
 }
 
 impl QuoteOptions {
@@ -486,16 +495,6 @@ impl QuoteOptions {
         self
     }
 
-    /// Override server-side price guard defaults for this request.
-    ///
-    /// When the server has the price guard enabled, this allows per-request
-    /// tuning of tolerance thresholds and fail-open behavior. Fields left as
-    /// `None` in [`PriceGuardConfig`] inherit the server defaults.
-    pub fn with_price_guard(mut self, config: PriceGuardConfig) -> Self {
-        self.price_guard = Some(config);
-        self
-    }
-
     /// The configured timeout in milliseconds, or `None` if using the server default.
     pub fn timeout_ms(&self) -> Option<u64> {
         self.timeout_ms
@@ -511,10 +510,6 @@ impl QuoteOptions {
         self.max_gas.as_ref()
     }
 
-    /// Per-request price guard overrides, or `None` if using server defaults.
-    pub fn price_guard(&self) -> Option<&PriceGuardConfig> {
-        self.price_guard.as_ref()
-    }
 }
 
 /// All inputs needed to call [`FyndClient::quote`](crate::FyndClient::quote).
--- a/clients/rust/src/types.rs
+++ b/clients/rust/src/types.rs
@@ -439,6 +439,12 @@ impl Order {
     }
 }
 
+/// Per-request overrides for price guard validation.
+///
+/// All fields are optional. When `None`, the server's configured defaults are used.
+/// Re-exported from `fynd-rpc-types` for wire compatibility.
+pub use fynd_rpc_types::PriceGuardConfig;
+
 /// Optional parameters that tune solving behaviour for a [`QuoteParams`] request.
 ///
 /// Build via the builder methods; unset options use server defaults.
@@ -448,6 +454,7 @@ pub struct QuoteOptions {
     pub(crate) min_responses: Option<usize>,
     pub(crate) max_gas: Option<BigUint>,
     pub(crate) encoding_options: Option<EncodingOptions>,
+    pub(crate) price_guard: Option<PriceGuardConfig>,
 }
 
 impl QuoteOptions {
@@ -479,6 +486,16 @@ impl QuoteOptions {
         self
     }
 
+    /// Override server-side price guard defaults for this request.
+    ///
+    /// When the server has the price guard enabled, this allows per-request
+    /// tuning of tolerance thresholds and fail-open behavior. Fields left as
+    /// `None` in [`PriceGuardConfig`] inherit the server defaults.
+    pub fn with_price_guard(mut self, config: PriceGuardConfig) -> Self {
+        self.price_guard = Some(config);
+        self
+    }
+
     /// The configured timeout in milliseconds, or `None` if using the server default.
     pub fn timeout_ms(&self) -> Option<u64> {
         self.timeout_ms
@@ -492,6 +509,11 @@ impl QuoteOptions {
     /// The configured gas cap, or `None` if no cap was set.
     pub fn max_gas(&self) -> Option<&BigUint> {
         self.max_gas.as_ref()
+    }
+
+    /// Per-request price guard overrides, or `None` if using server defaults.
+    pub fn price_guard(&self) -> Option<&PriceGuardConfig> {
+        self.price_guard.as_ref()
     }
 }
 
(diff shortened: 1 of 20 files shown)


clients/rust/src/types.rs
```diff
@@ -434,6 +440,7 @@ pub struct QuoteOptions {
     pub(crate) min_responses: Option<usize>,
     pub(crate) max_gas: Option<BigUint>,
     pub(crate) encoding_options: Option<EncodingOptions>,
+    pub(crate) price_guard: Option<PriceGuardConfig>,
```
tamaralipows: :/ will this break backwards compatibility

tamaralipows: Okay I just realized this is misplaced: we need to move it into the `EncodingOptions` not here (won't have time to do this before leaving).

But same question for the `EncodingOptions` - will it break backwards compatibility there?

fynd-core/src/solver.rs
```diff
@@ -578,10 +624,24 @@ impl FyndBuilder {
...
+        // Start price providers and construct the guard.
+        // Providers are started even when `enabled: false` so caches stay warm
+        // for per-request opt-in. The `enabled` flag only controls whether
+        // validation runs at request time.
         let router_config = WorkerPoolRouterConfig::default()
             .with_timeout(self.router_timeout)
             .with_min_responses(self.router_min_responses);
-        let router = WorkerPoolRouter::new(solver_pool_handles, router_config, encoder);
+        let mut router = WorkerPoolRouter::new(solver_pool_handles, router_config, encoder);
+        if !self.price_providers.is_empty() {
```
Troshchk: They are always set above, so maybe this check is not necessary?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -132,6 +152,23 @@ impl WorkerPoolRouter {
...
 
+        // Validate against external prices if enabled.
+        // Per-request config overrides the server default.
+        if let Some(ref guard) = self.price_guard {
+            let config = request
+                .options()
+                .price_guard()
+                .cloned()
+                .unwrap_or(self.price_guard_config.clone());
+            match guard.validate(order_quotes, &config) {
```
Troshchk, line 166: Not sure this is correct -- in the list of validated quotes are also quotes where the price check failed

Troshchk: ah, ok, it is filtered out in the `encode()`

clients/rust/src/types.rs
dianacarvalho1: > for the EncodingOptions - will it break backwards compatibility there?

I don't think so because it is None and they are set with a builder like pattern

fynd-core/src/price_guard/guard.rs
```diff
@@ -125,7 +125,7 @@ impl PriceGuard {
                     price_out_of_tolerance = true;
                 }
                 Err(e) => {
-                    if let PriceProviderError::PriceNotFound { .. } = e {
+                    if let PriceProviderError::TokenNotFound { .. } = e {
```
dianacarvalho1: ah this is from a wrong merge. it should be `PriceNotFound`

fynd-core/src/price_guard/hyperliquid.rs
```diff
@@ -147,10 +147,7 @@ impl PriceProvider for HyperliquidProvider {
 
         let price_in = cache
             .get(&sym_in)
-            .ok_or(PriceProviderError::PriceNotFound {
-                token_in: sym_in.clone(),
-                token_out: sym_out.clone(),
-            })?;
+            .ok_or(PriceProviderError::TokenNotFound { address: sym_in.clone() })?;
```
dianacarvalho1: same here

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -80,6 +82,11 @@ pub struct WorkerPoolRouter {
     config: WorkerPoolRouterConfig,
     /// Encoder for encoding solutions into on-chain transactions.
     encoder: Encoder,
+    /// Validates solution outputs against external price sources.
+    price_guard: Option<PriceGuard>,
+    /// Server-side default config for the price guard.
+    /// Used when a request does not include per-request overrides.
+    price_guard_config: PriceGuardConfig,
```
dianacarvalho1: why would we want to have this? I thought we were enforcing passing this in the request 🤔

dianacarvalho1: so we can have a master config from startup??

dianacarvalho1 (commented):
Thank you! Small comments only.

I'm not sure I agree with the overall config for the price guard at the `serve` cli 😕 I need to think about this

fynd-core/src/worker_pool_router/mod.rs
Troshchk: I think [this](https://github.com/propeller-heads/fynd/pull/171/changes#diff-20a45894038b25a91d18beea517ece0569617f2a25f25aa8fb298a171e5ec469R155) is the place where the config is currently used, but I’m not sure it’s actually needed.
If no price config is provided in the QuoteRequest, the WorkerPoolRouter price_guard default acts as a fallback if price_guard is set.

If we plan to keep price_guard disabled by default, we can simplify it: when there’s no price_guard_config in the QuoteRequest, we just skip price validation entirely even if a price_guard is set on the router.

If we do not plan to keep price_guard disabled by default, the only place I’d set both is in with_price_guard( set config to default or provided)

dianacarvalho1: I agree with most of your comment, except this part

> If we do not plan to keep price_guard disabled by default, the only place I’d set both is in with_price_guard( set config to default or provided)

We need to always give the option to disable the price guard per request (because of weird tokens for example)

src/cli.rs
```diff
@@ -125,13 +125,13 @@ pub struct ServeArgs {
...
-    /// Default: true.
-    #[arg(long, default_value_t = true)]
+    /// Default: false.
+    #[arg(long, default_value_t = false)]
     pub price_guard_allow_on_provider_error: bool,
 
     /// Allow solutions through when no provider has a price for the token pair.
-    /// Default: true.
-    #[arg(long, default_value_t = true)]
+    /// Default: false.
```
dianacarvalho1, line 134: @Troshchk you changed this so that the default behaviour is the most restrict one correct? 
but if then in the request, `price_guard_allow_on_provider_error=true` then true is used right?

dianacarvalho1 (commented):
Thank you @Troshchk ! Just one small question

src/cli.rs
Troshchk: I have changed it, because if default is true, there is no way to disable it: before, you would add --price-guard-allow-on-provider-error to set it true and it would be true by default too. 

When you want to enable it, you just add: **--price-guard-allow-on-provider-error**

```
docker compose -f docker-compose.yml run --rm \
  -p 3000:3000 \
  fynd serve \
  --chain Ethereum \
  --protocols uniswap_v2,uniswap_v3 \
  --enable-price-guard \
  --price-guard-allow-on-provider-error 
  ```
  
  
  or 
  
  
  ```
  cargo run --release -- serve \
  --enable-price-guard \
  --price-guard-allow-on-provider-error \
  --price-guard-allow-on-token-not-found
  ```

Comment (dianacarvalho1) [reply src/cli.rs:134]: > because if default is true, there is no way to disable it

why not? setting it to false in the request doesn't overwrite it?

Comment (Troshchk) [reply src/cli.rs:134]: Setting it to false in the request does overwrite it.

What I meant is that if you want to run with price_guard_allow_on_provider_error being false by default, there was no way to configure that.

Comment (tamaralipows) [reply src/cli.rs:134]: Sorry I'm a bit confused here. I would say that the default should be the least restrictive (i.e. true - allow on provider error, allow on token not found)

If the user wants to overwrite this per request to disallow this, couldn't they just do something like?
```
cargo run --release -- serve \
--enable-price-guard \
--price-guard-allow-on-provider-error false \
--price-guard-allow-on-token-not-found false
```

Not sure I understand the problem

File: clients/rust/src/types.rs (hunk 1 above)
Comment (tamaralipows) [reply clients/rust/src/types.rs]: @Troshchk Could you please move the price guard into the encoding options?

**Review by tamaralipows — COMMENTED** [review]
Thanks @Troshchk - looks mostly good, but I think there is still some confusion about the defaults

File: src/cli.rs (hunk 7 above)
Comment (Troshchk) [reply src/cli.rs:134]: This will not work:
```
cargo run --release -- serve \
--enable-price-guard \
--price-guard-allow-on-provider-error false \
--price-guard-allow-on-token-not-found false
```

Boolean flags do not accept values, so they're true when present, and false when absent.
To make both true:
```
cargo run --release -- serve \
--enable-price-guard \
--price-guard-allow-on-provider-error \
--price-guard-allow-on-token-not-found
```
To make both false:
```
cargo run --release -- serve \
--enable-price-guard
```

Being true by default means, that the flags were true both when present and absent, so there was no way to disable them

Comment (tamaralipows) [reply src/cli.rs:134]: What about reversing this to call it something else then? 
`--price-guard-fail-on-provider-error` and `--price-guard-fail-on-token-not-found`
and setting it to false by default?

Comment (tamaralipows) [reply src/cli.rs:134]: the motivation is backwards compatibility here - we want it by default to be least restrictive

File: fynd-core/src/worker_pool_router/mod.rs [hunk 8]
```diff
@@ -147,9 +147,13 @@ impl WorkerPoolRouter {
...
 
         // Validate against external prices if enabled.
-        // Per-request config overrides the server default.
+        // Per-request config (inside encoding_options) overrides the server default.
         if let Some(ref guard) = self.price_guard {
-            let config = if let Some(request_config) = request.options().price_guard() {
+            let request_config = request
+                .options()
+                .encoding_options()
+                .and_then(|e| e.price_guard());
```
tamaralipows: I think this var is/was named wrong - it's actually the price guard config not the whole request config

src/cli.rs
Troshchk: Makes sense, changed!

tamaralipows (commented):
Thank you! I'd approve but I opened the PR :)

Troshchk (approved):

PR #172, opened 2026-03-31
feat(swap-cli): use client approval API, fetch router from /v1/info

kayibal:
## Summary

- **Remove `--router` flag** — the Tycho Router address is now fetched automatically from `GET /v1/info`. The Permit2 path uses this to construct the EIP-712 permit and storage overrides.

- **Automatic approvals** — both ERC-20 and Permit2 execute paths now call `client.approval()` + `client.execute_approval()` instead of printing a `cast send` command. If the allowance is already sufficient, the approval is skipped. Removes the dead `erc20::read_erc20_allowance` helper.

- **Fix Permit2 dry-run simulation** — three issues fixed:
  - Permit2 validates `ecrecover(sig) == owner`: the dry-run now uses the ephemeral signer's own address as the order sender so this check passes.
  - `eth_call` was rejecting calls from zero-ETH accounts: `maxFeePerGas=0` is set in dry-run signing hints, bypassing the balance pre-check without needing a funded address. `DRY_RUN_SENDER` (Vitalik's address) is removed.
  - `TRANSFER_FROM_FAILED` revert: the TychoRouter may call `token.transferFrom(sender, router, amount)` directly, so the dry-run now injects `allowance[sender][router]` in addition to `allowance[sender][permit2]`.

- **Fix dry-run for tokens with bit-packed storage** — storage overrides now inject `U256::MAX >> 1` instead of `U256::MAX`. USDC packs a blacklist flag into bit 255 of its balance slot, so injecting all-ones caused a "Blacklistable: account is blacklisted" revert. Clearing the top bit avoids this while still providing an effectively unlimited balance/allowance.

- **Refactor**: all per-transfer-type logic (encoding options, approvals, dry-run spender addresses) is consolidated into a single `match cli.transfer_type` block.

- **Docs**: removes `--router` from examples, describes automatic approval handling, drops the stale "Insufficient allowance / cast send" troubleshooting entry.

Tested onchain with a couple of transactions. Example:
https://etherscan.io/tx/0x128238213a9281b7789c013986c99e341bcd863e1f2131adfc274d81bf600df4

### DIFF
--- a/docs/guides/swap-cli.md
+++ b/docs/guides/swap-cli.md
@@ -76,7 +76,7 @@ The output prints the quote (amount\_in, amount\_out, gas estimate, route) follo
 
 ## Dry-run a swap (Permit2)
 
-Add `--transfer-type transfer-from-permit2`. The router address is fetched automatically from the Fynd server. The dry-run uses nonce 0 and maximum deadlines, so no chain reads are needed.
+Add `--transfer-type transfer-from-permit2`. The dry-run uses nonce 0 and maximum deadlines, so no chain reads are needed.
 
 ```bash
 fynd-swap-cli --transfer-type transfer-from-permit2
@@ -99,7 +99,7 @@ fynd-swap-cli --execute
 
 ## Execute on-chain (Permit2)
 
-Add `--execute --transfer-type transfer-from-permit2`. The router address is resolved automatically from the Fynd server. The CLI checks the Permit2 allowance and submits an approval transaction first if needed, then reads the current nonce, builds the EIP-712 permit, signs it, and submits the swap.
+Add `--execute --transfer-type transfer-from-permit2`. The CLI checks the Permit2 allowance and submits an approval transaction first if needed, then reads the current nonce, builds the EIP-712 permit, signs it, and submits the swap.
 
 ```bash
 export RPC_URL=https://your-rpc-provider.com/v1/your_key
--- a/docs/guides/swap-cli.md
+++ b/docs/guides/swap-cli.md
@@ -76,22 +76,18 @@ The output prints the quote (amount\_in, amount\_out, gas estimate, route) follo
 
 ## Dry-run a swap (Permit2)
 
-Add `--transfer-type transfer-from-permit2` and supply the TychoRouter address with `--router`. The dry-run uses nonce 0 and maximum deadlines, so you don't need any chain reads for Permit2 state.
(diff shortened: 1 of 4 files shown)


tools/fynd-swap-cli/src/main.rs
```diff
@@ -375,34 +384,35 @@ async fn main() -> anyhow::Result<()> {
...
-            }
+            ensure_approval(
+                &client,
+                &signer,
+                sell_token_bytes,
+                amount.clone(),
+                UserTransferType::TransferFrom,
+                sender,
+            )
+            .await?;
```
kayibal: It's not great that we have this match on transfer_type twice (here and on the encoding options).

docs/guides/swap-cli.md
```diff
@@ -76,22 +76,18 @@ The output prints the quote (amount\_in, amount\_out, gas estimate, route) follo
 
 ## Dry-run a swap (Permit2)
 
-Add `--transfer-type transfer-from-permit2` and supply the TychoRouter address with `--router`. The dry-run uses nonce 0 and maximum deadlines, so you don't need any chain reads for Permit2 state.
+Add `--transfer-type transfer-from-permit2`. The router address is fetched automatically from the Fynd server. The dry-run uses nonce 0 and maximum deadlines, so no chain reads are needed.
```
dianacarvalho1: > The router address is fetched automatically from the Fynd server

I think the user doesn't need to know this.. I would just remove this sentence entirely

docs/guides/swap-cli.md
```diff
@@ -103,18 +99,13 @@ fynd-swap-cli --execute
 
 ## Execute on-chain (Permit2)
 
-Add `--execute --transfer-type transfer-from-permit2 --router <addr>`. The CLI reads the current Permit2 nonce from the contract, builds the EIP-712 permit, signs it with your key, and submits the swap in one step.
-
-Your wallet must have already approved the Permit2 contract (`0x000000000022D473030F116dDEE9F6B43aC78BA3`) to spend the sell token. The CLI prints the required `cast send` command if that approval is missing.
+Add `--execute --transfer-type transfer-from-permit2`. The router address is resolved automatically from the Fynd server. The CLI checks the Permit2 allowance and submits an approval transaction first if needed, then reads the current nonce, builds the EIP-712 permit, signs it, and submits the swap.
```
dianacarvalho1: > The router address is resolved automatically from the Fynd server

same here

dianacarvalho1 (approved):
Thank you @kayibal ! Only have a small comment about the docs

PR #174, opened 2026-04-01
feat(swap-cli): poll health for 30s and show amount out net gas

tvinagre:
## Summary
- Health check now polls every 2s for up to 30s instead of failing immediately, giving the solver time to load market data
- Added `amount_out_net_gas` to the CLI quote output (the gas-adjusted amount the solver uses to rank routes)


dianacarvalho1 (approved):
Thank you @tvinagre !

PR #175, opened 2026-04-01
fix: normalize pool depth to ETH for path scoring

tvinagre:
## Summary

- Normalize pool depth from raw `token_in` units to ETH-equivalent in `DepthAndPrice::from_sim_and_derived`, using `TokenGasPrices` already available in `DerivedData`
- Fix `min(depths)` comparison across hops — previously meaningless when comparing different token denominations (e.g., WETH 18 decimals vs USDC 6 decimals)
- Guard against zero numerator/denominator in token price

## Problem

`MostLiquidAlgorithm::try_score_path` computes `score = product(spot_prices) × min(depths)` to prioritize which paths to simulate. Depth was stored in raw `token_in` units, so a WETH edge depth of `1e18` (1 WETH) appeared 10^12x larger than a USDC depth of `1e9` (1000 USDC). This caused misordered simulation priority, and with `max_routes` truncation or timeout, genuinely good paths could be permanently excluded.

## Changes

Single file change: `fynd-core/src/algorithm/most_liquid.rs`

**Production code:**
- `from_sim_and_derived`: after reading raw `BigUint` depth, converts to ETH using `depth_eth = raw_depth * denominator / numerator` from `TokenGasPrices`
- Returns `None` (skips edge) if token price is missing or has zero numerator/denominator
- Updated `depth` field doc comment and `computation_requirements()` comment

**Tests (6 new/updated):**
- 3 existing tests updated to provide token prices in `DerivedData`
- `test_from_sim_and_derived_missing_token_price_returns_none`
- `test_from_sim_and_derived_normalizes_depth_to_eth` (integer price ratio)
- `test_from_sim_and_derived_normalizes_depth_fractional_price` (fractional price ratio)

## Benchmark Results (500 requests, `fynd-benchmark compare`)

```
Coverage:
  Both found route:          371
  Only main found route:       0
  Only normalized-depth:       0
  Neither found route:       129

Head-to-head (amount out before gas):
  normalized-depth wins:  65  (13.0%)
  main wins:              29  ( 5.8%)
  Ties:                  406  (81.2%)
  Avg diff:           +0.50 bps

Head-to-head (net of gas, server-side):
  normalized-depth wins:  96  (19.2%)
  main wins:              26  ( 5.2%)
  Ties:                  378  (75.6%)
  Avg diff:           +0.67 bps

Solve Time:
  main:             avg=34ms  p50=35  p95=87   p99=103
  normalized-depth: avg=34ms  p50=34  p95=88   p99=102

Route Depth:
  main avg swaps: 1.6  (max 3)
  normalized-depth avg swaps: 1.7  (max 3)
  Identical routes: 415/500
```

**Key takeaways:**
- Zero coverage regression — no routes lost from skipping edges without token prices
- 2:1 win ratio pre-gas (65 vs 29), improving to ~4:1 net-of-gas (96 vs 26)
- Top outliers show +18-20 bps on multi-hop routes (WETH → small tokens) where the fix matters most
- Solve time unchanged — normalization adds negligible overhead
- Worst single trade: -11.9 bps; best: +19.9 bps; average consistently positive

## Test plan

- [x] All 324 fynd-core tests pass
- [x] Clippy clean (`-D warnings`)
- [x] Formatting clean
- [x] Benchmark comparison against main (500 requests)

### DIFF
--- a/fynd-core/src/algorithm/most_liquid.rs
+++ b/fynd-core/src/algorithm/most_liquid.rs
@@ -47,7 +47,7 @@ pub struct MostLiquidAlgorithm {
 pub struct DepthAndPrice {
     /// Spot price (token_out per token_in) for this edge direction.
     pub spot_price: f64,
-    /// Liquidity depth normalized to gas token (ETH) units.
+    /// Liquidity depth normalized to gas token (native token) units.
     pub depth: f64,
 }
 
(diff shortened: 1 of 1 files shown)


fynd-core/src/algorithm/most_liquid.rs
```diff
@@ -47,7 +47,7 @@ pub struct MostLiquidAlgorithm {
 pub struct DepthAndPrice {
     /// Spot price (token_out per token_in) for this edge direction.
     pub spot_price: f64,
-    /// Liquidity depth in raw units of the sell token.
+    /// Liquidity depth normalized to gas token (ETH) units.
```
louise-poole: Maybe don't mention ETH? We'll forget to update this when we expand to new chains that don't necessarily use ETH...
```suggestion
    /// Liquidity depth normalized to gas token (native token) units.
```

fynd-core/src/algorithm/most_liquid.rs
```diff
@@ -119,6 +119,40 @@ impl crate::graph::EdgeWeightFromSimAndDerived for DepthAndPrice {
             }
         };
 
+        // Normalize depth from raw token_in units to gas token (ETH) units.
```
louise-poole:
```suggestion
        // Normalize depth from raw token_in units to gas token units.
```

fynd-core/src/algorithm/most_liquid.rs
```diff
@@ -108,7 +108,7 @@ impl crate::graph::EdgeWeightFromSimAndDerived for DepthAndPrice {
         };
 
         // Look up pre-computed depth; skip edge if unavailable.
-        let depth = match derived
+        let raw_depth = match derived
             .pool_depths()
```
tvinagre, line 112: Please note that I chose to not  change how `pool_depth` was calculated. It's still denominated in `sell_token`, we just normalize here for this algorithm usage.

Fixing the calculation would change a lot, as it would create another dependency layer (`pool_depth` needs `token_prices`). IMO this is fine to be done here - we can move to the derived_data layer if preferred.

louise-poole (approved):
Two minor comments, otherwise looks good! Interesting that pre-normalising results had some better paths it found... I expected with normalising we'd find the same or better always. What `max_routes` value did you run these benchmarks with?

fynd-core/src/algorithm/most_liquid.rs
```diff
@@ -119,6 +119,40 @@ impl crate::graph::EdgeWeightFromSimAndDerived for DepthAndPrice {
...
+                let den = price
+                    .denominator
+                    .to_f64()
+                    .unwrap_or(0.0);
+                if num == 0.0 || den == 0.0 {
+                    trace!(
+                        component_id = %component_id,
+                        token_in = %token_in.address,
+                        "token price has zero numerator or denominator, skipping edge"
+                    );
```
zizou0x: This could also mean that to_f64 failed, should we have different paths to make each case clearer?

PR #176, opened 2026-04-01
fix: set min token quality filter on tycho stream

louise-poole:
Previously it was only used at start-up to filter the initial tokens retrieved.


zizou0x (approved):

dianacarvalho1 (approved):
👍🏼 
I think you are blocked by this PR though https://github.com/propeller-heads/fynd/pull/128

PR #180, opened 2026-04-02
feat: improve swap-cli UX

zizou0x:
## Summary
  
  - Return transaction hash from execute_swap in the Rust client instead of discarding it
  - Add AllowanceCheck enum to control ERC-20 approval behavior — supports skipping checks or checking against a minimum threshold, so users can approve once (e.g. unlimited for Permit2) and skip redundant approvals on subsequent swaps
  - Approve unlimited amount for Permit2 in the swap CLI to avoid repeated approval transactions
  - Improve swap CLI console output with progress messages for allowance checks, transaction submissions, token addresses, and transaction hashes
  - Update protocol list in docs (replace vm:curve with ekubo and fluid)

### DIFF
--- a/tools/fynd-swap-cli/src/main.rs
+++ b/tools/fynd-swap-cli/src/main.rs
@@ -186,21 +186,23 @@ async fn ensure_approval(
     transfer_type: UserTransferType,
     sender: Address,
 ) -> anyhow::Result<()> {
+    println!("Checking ERC-20 allowance...");
     let params = ApprovalParams::new(sell_token, amount, true).with_transfer_type(transfer_type);
     let hints = SigningHints::default().with_sender(sender);
     let Some(approval_payload) = client.approval(&params, &hints).await? else {
-        return Ok(()); // allowance already sufficient
+        println!("Allowance sufficient, no approval needed.");
+        return Ok(());
     };
-    info!("Submitting approval transaction...");
+    println!("Allowance insufficient — submitting approval transaction...");
     let sig = signer
         .sign_hash(&approval_payload.signing_hash())
         .await?;
     let signed = SignedApproval::assemble(approval_payload, sig);
     let receipt = client.execute_approval(signed).await?;
-    tokio::time::timeout(Duration::from_secs(120), receipt)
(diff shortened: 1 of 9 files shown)


tools/fynd-swap-cli/src/main.rs
```diff
@@ -350,11 +352,14 @@ async fn main() -> anyhow::Result<()> {
...
                 .map_err(|_| anyhow::anyhow!("invalid router address from /v1/info"))?;
             if cli.execute {
+                // Check against the swap amount but approve max — subsequent swaps won't
+                // need re-approval even after Permit2 deducts from the ERC-20 allowance.
                 ensure_approval(
                     &client,
                     &signer,
                     sell_token_bytes.clone(),
-                    amount.clone(),
+                    max_uint160(),
```
Troshchk: Is it safe to approve max here?

Troshchk (commented):
Thank you!
Just have one question

tvinagre (approved):
LGTM 
I agree with the comment on max approval, we can maybe 3x the swap amount instead

tools/fynd-swap-cli/src/main.rs
tvinagre: Is this approving Max to Permit2 or the router?
If permit2, then it's fine for me. If it is the router, I'd stick to `amount`

Troshchk: Why do we need to approve more than the swap amount?

zizou0x: Happy to discuss but IMO it's safe to approve permit2 yes

zizou0x: When the router transfer_from it consumes the allowance. So swapping becomes 2 transaction every time (approve then swap) instead of just one transaction

zizou0x: Update: we decided to not approve more than needed even if it's against Permit2 for security reasons. See https://github.com/propeller-heads/fynd/pull/180/changes/03bad3a9092a81df5920e3e0201b25c61637f3db

PR #183, opened 2026-04-03
refactor: change server config to be binary-oriented

zizou0x:
- Updated README and server configuration documentation to reflect the new command usage without 'cargo run --release --'.
- Changed the ENTRYPOINT in Dockerfile to make it more flexible (like the binary).


kayibal (approved):
Looks better, but we need to adapt the installtion / setup instructions to actually put the `fynd` binary in your path. Mainly the compile from source one.

Other than this looks good.

PR #184, opened 2026-04-06
feat(fynd-client): Clone derives + batch_quote method

carloszanella:
## Summary

- Adds `Clone` to all public structs in `fynd-client` that were missing it (`Order`, `QuoteOptions`, `QuoteParams`, `InstanceInfo`, `HealthStatus`, `SigningHints`, `ExecutionOptions`, `ApprovalParams`, `SettledOrder`, `MinedTx`)
- Adds `FyndClient::batch_quote(BatchQuoteParams) -> Result<Vec<Quote>, FyndError>` — submits multiple orders in a single round-trip to the server, with the same retry logic as `quote()`
- Adds `BatchQuoteParams` to the public API

## Test plan

- [ ] `cargo test -p fynd-client --all-features` passes (96 tests)
- [ ] New mapping tests cover: empty-orders error, per-order `token_out`/`receiver` extraction, count-mismatch error, `solve_time_ms` propagation


clients/rust/src/mapping.rs
```diff
@@ -60,6 +60,58 @@ pub(crate) fn quote_params_to_dto(params: QuoteParams) -> Result<dto::QuoteReque
...
+    Ok((request, order_meta))
+}
+
+/// Maps a batch DTO response to client [`Quote`]s using per-order metadata.
+///
+/// `order_meta` must be index-aligned with the request orders produced by
+/// [`batch_quote_params_to_dto`]. Returns an error if the response order count differs from
+/// the request order count.
+pub(crate) fn dto_to_quotes(
+    ds: dto::Quote,
```
carloszanella: what is "ds"? This method seems too similar to dto_to_batch_quote and the names are terribly confusing between the two. pls suggest some refactoring

carloszanella:
> what is "ds"? This method seems too similar to dto_to_batch_quote and the names are terribly confusing between the two. pls suggest some refactoring

`ds` is a leftover abbreviation for "deserialized" (dto). Agreed on the naming — and the two functions are genuinely redundant.

Suggested refactor:
- Delete `dto_to_batch_quote` and the `BatchQuote` wrapper type entirely (both are `pub(crate)` so no public API break).
- Rename `dto_to_quotes` → `map_quote_response(response: dto::Quote, order_meta: Vec<(Bytes, Bytes)>)` — one function handles both the single and multi-order case.
- `request_quote` calls it with a 1-element meta vec and takes the first element.
- `request_batch_quote` calls it with the full meta vec.

Implementing this now.

carloszanella:
dumb robot didn't sign the comment

clients/rust/src/mapping.rs
```diff
@@ -283,17 +256,33 @@ impl From<dto::Transaction> for Transaction {
...
+    order_meta: Vec<(bytes::Bytes, bytes::Bytes)>,
+) -> Result<Vec<Quote>, FyndError> {
+    let solve_time_ms = response.solve_time_ms();
+    let order_quotes = response.into_orders();
+    if order_quotes.len() != order_meta.len() {
+        return Err(FyndError::Protocol(format!(
+            "server returned {} quotes but {} were requested",
+            order_quotes.len(),
+            order_meta.len()
+        )));
```
carloszanella, line 277: pls make sure error is the correct behaviour for Fynd on this case @fynd-devs

clients/rust/src/client.rs
```diff
@@ -614,6 +617,56 @@ where
         Ok(quote)
     }
 
+    /// Request quotes for multiple swap orders in a single round-trip.
+    ///
+    /// All orders share the same [`QuoteOptions`]. The returned vec is index-aligned with the
+    /// input: `quotes[i]` corresponds to `params.orders[i]`.
+    ///
+    /// Retries automatically on transient failures according to the client's [`RetryConfig`].
```
kayibal, line 620: Should we mention here that it retries the entire request? How does failure work of a single order work in the batch response btw? I am not sure... In general I think it would be nice to document this here.

clients/rust/src/mapping.rs
```diff
@@ -60,6 +60,58 @@ pub(crate) fn quote_params_to_dto(params: QuoteParams) -> Result<dto::QuoteReque
     Ok(dto::QuoteRequest::new(vec![order]).with_options(options))
 }
 
+/// Converts [`BatchQuoteParams`] into a DTO request plus per-order `(token_out, receiver)` metadata.
+///
+/// The metadata vec is index-aligned with the DTO orders and must be threaded through to
+/// [`dto_to_quotes`] so each response quote gets the correct token/receiver fields.
+pub(crate) fn batch_quote_params_to_dto(
+    params: BatchQuoteParams,
+) -> Result<(dto::QuoteRequest, Vec<(bytes::Bytes, bytes::Bytes)>), FyndError> {
```
kayibal, line 70: I think I'd appreciate a private type for the per-order metadata.. It wasn't very obivous to me that this is basically a pass-through for the client structs.

kayibal: I see this was later improved, I am ok with that solution.

kayibal (approved):

PR #187, opened 2026-04-14
docs: add price guard documentation

Troshchk:
Document the price guard feature in encoding-options guide (overview, config fields, tolerance, fallback behavior, usage example) and add CLI flags to the server-configuration flag reference table.


docs/guides/encoding-options.md
```diff
@@ -47,6 +48,97 @@ Draws tokens from the sender's vault balance in the TychoRouter contract (ERC-69
 permit needed — tokens must have been deposited into the vault beforehand. See the
 [vault mechanism](https://docs.propellerheads.xyz/tycho/for-solvers/execution) in the Tycho docs.
 
+## Price guard
+
+The price guard validates each successful quote's `amount_out` against external price providers
+(Hyperliquid and Binance by default). It runs after solving and before the response is returned.
```
dianacarvalho1: > It runs after solving and before the response is returned.

I think this is unnecessary

docs/guides/encoding-options.md
```diff
@@ -47,6 +48,97 @@ Draws tokens from the sender's vault balance in the TychoRouter contract (ERC-69
...
+Providers are queried concurrently, and a quote passes if **at least one** provider returns a price
+within tolerance. When validation fails, the quote's status is set to `price_check_failed`; other
+orders in the same batch are unaffected.
+
+The guard is disabled by default. Operators enable it server-side with `--enable-price-guard`
+(see [server configuration](server-configuration.md)); tolerances and fallback behavior can be
+tuned via the matching `--price-guard-*` flags. When enabled without further overrides, the server
+uses the defaults shown in the table below.
```
dianacarvalho1: I find it weird that inside of the page for "encoding options" we are talking about server side things..

docs/guides/encoding-options.md
```diff
@@ -47,6 +48,97 @@ Draws tokens from the sender's vault balance in the TychoRouter contract (ERC-69
...
+
+Clients can override the configuration per request through `encoding_options.price_guard`. When
+present, the request config **replaces** the server config entirely — any field omitted from the
+request falls back to the default in the table below, not to the server's configured value. Send
+every field whose server default you do not want.
```
dianacarvalho1: Send? or set?

docs/guides/encoding-options.md
```diff
@@ -47,6 +48,97 @@ Draws tokens from the sender's vault balance in the TychoRouter contract (ERC-69
...
+| `fail_on_provider_error`         | `boolean` | `false` | See [fallback behavior](#fallback-behavior).                                                                                  |
+| `fail_on_token_price_not_found`  | `boolean` | `false` | See [fallback behavior](#fallback-behavior).                                                                                  |
+
+### Tolerance
+
+The quote's `amount_out` is compared to each provider's expected amount out and the check
+short-circuits as soon as one provider validates within tolerance — the remaining providers are
+not consulted. If `amount_out` is below the expected amount, the deviation is checked against
+`lower_tolerance_bps`; otherwise against `upper_tolerance_bps`. The two bounds are separate so
+operators can be strict on under-delivery while leaving headroom for over-delivery that may
```
dianacarvalho1: users instead of operators

docs/guides/encoding-options.md
```diff
@@ -47,6 +48,97 @@ Draws tokens from the sender's vault balance in the TychoRouter contract (ERC-69
...
+indicate a pricing error in Fynd rather than a genuine arbitrage. The default `upper_tolerance_bps`
```
dianacarvalho1: genuine good price instead of arbitrage

docs/guides/encoding-options.md
```diff
@@ -47,6 +48,97 @@ Draws tokens from the sender's vault balance in the TychoRouter contract (ERC-69
...
+to catch suspicious outputs.
+
+### Fallback behavior
+
+The fallback flags apply only when every provider response is an error or "not found" — in other
+words, no provider returned a price at all. If any provider returns an in-tolerance price, the
+quote passes regardless of these flags; if any provider returns an out-of-tolerance price and
+none passes, the quote is rejected regardless of these flags.
+
+- `fail_on_provider_error`: all providers failed with an infrastructure error (network issue, API
```
dianacarvalho1: this sentence is strange..

dianacarvalho1 (commented):
Thank you @Troshchk ! Overall this look good! I do think we are putting too much context inside of "encoding options". I would rather make a completely independent guide for "Price Guard/Validation" where we explain all the details. and then in "Encoding options" and "Server Configuration" I would just mention the flags/options and point to the more complete guide for more. 
Also we missing information on adding your own price provider if the user wants

docs/guides/price-guard.md
```diff
@@ -0,0 +1,135 @@
...
+status is set to `price_check_failed`; other orders in the same batch are unaffected.
+
+Each provider runs a background worker that continuously populates an in-memory cache. On the
+solve path, `get_expected_out()` reads from cache — no blocking API calls during solving.
+
+## Enabling the price guard
+
+The guard is disabled by default. Users enable it server-side with `--enable-price-guard`
+(see [server configuration](server-configuration.md)); tolerances and fallback behavior can be
+tuned via the matching `--price-guard-*` flags. When enabled without further overrides, the server
```
tamaralipows: Should we say exactly what the flags are called server-side?

docs/guides/price-guard.md
```diff
@@ -0,0 +1,135 @@
...
+uses the defaults shown in the table below.
+
+Clients can override the configuration per request through `encoding_options.price_guard`
+(see [encoding options](encoding-options.md)). When present, the request config **replaces** the
+server config entirely — any field omitted from the request falls back to the default in the table
+below, not to the server's configured value. Set every field whose server default you do not want.
```
tamaralipows: Confused, they are not falling back to the server's configured value, but we need to set every field whose server default we do not want? 

I actually didn't think about this - I would think the server's configured value should be the default, otherwise what purpose does it serve?

docs/guides/price-guard.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+- `amount_out < expected` → reject if deviation exceeds `lower_tolerance_bps`
+- `amount_out >= expected` → reject if deviation exceeds `upper_tolerance_bps`
+
+By default the lower bound is much stricter (`300` bps = 3%) than the upper bound (`10000` bps =
+100%). The lower bound catches under-delivery — the user getting less than expected. The upper
+bound catches cases where Fynd's price is significantly better than the external price, which may
+indicate a pricing bug that would fail to land on-chain. The default of `10000` allows `amount_out`
+up to twice the expected amount; lower it to catch suspicious outputs.
```
tamaralipows: I think this can be a lot more concise. Did you try running the editor claude skill on this PR?

docs/guides/price-guard.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Fallback behavior
+
+The fallback flags apply only when every provider response is an error or "not found" — in other
+words, no provider returned a price at all. If any provider returns an in-tolerance price, the
+quote passes regardless of these flags; if any provider returns a price but none are within
```
tamaralipows: I think this means to say "if all provider return prices"?

docs/guides/price-guard.md
```diff
@@ -0,0 +1,135 @@
...
+tolerance, the quote is rejected regardless of these flags.
+
+- **`fail_on_provider_error`** — controls behavior when all providers fail with an infrastructure
+  error (network issue, API down, rate-limited). `false` (default) lets the quote pass; `true`
+  rejects it.
+- **`fail_on_token_price_not_found`** — controls behavior when every provider was reached but no
+  provider lists the token (it is not traded on that venue). `false` (default) lets the quote
```
tamaralipows: not all providers are trading venues, some aggregate onchain dexs, so maybe remove the parenthesis

docs/guides/price-guard.md
```diff
@@ -0,0 +1,135 @@
...
+  pass; `true` rejects it.
+
+When responses are mixed — some `price_not_found`, some infrastructure errors — the guard treats
+the pair as potentially supported but unreachable and falls back to `fail_on_provider_error`.
```
tamaralipows: I think this line can be a bit clearer

docs/guides/price-guard.md
```diff
@@ -0,0 +1,135 @@
...
+    .build()
+    .await?;
+```
+
+Providers follow a worker+cache pattern: `start()` spawns a background task that populates an
+in-memory cache, and `get_expected_out()` reads from that cache synchronously without blocking or
+making network calls.
+
+If no providers are registered before `build()`, the two built-in providers (Hyperliquid, Binance)
+are added automatically. Calling `register_price_provider()` before `build()` skips the defaults —
```
tamaralipows: They can also add the defaults to the ones they register right?

tamaralipows (commented):
Thank you! Just a few comments!

docs/guides/price-guard.md
Troshchk: hmm, actually I think the default ones are hardcoded, so I don't think there is a possibility to add a new registered one to the default ones.

docs/guides/price-guard.md
Troshchk: If at least one provider returns price and it is in tolerance, than the quote price check is successful. I think it is "If any" rather than "if all", or do you mean something else?

docs/guides/price-guard.md
Troshchk: Currently, the server’s configured values are only used when the request does not include a price_guard configuration. If a price_guard is provided in the request, it fully overrides the server configuration. Any fields set in the request are applied on top of the PriceGuardConfig defaults, not on top of the server’s configured values.

Should we document this behavior as is for now, and add a todo to revisit both the implementation and the documentation to fallback to the server defaults?

docs/guides/price-guard.md
Troshchk: I have added more info, lmk what you think

docs/guides/price-guard.md
tamaralipows: No no I meant something like 
```
.add_default_providers()
.register_provider(...)
```
I don't remember what we did but this would be a nice thing to allow and mention in docs

docs/guides/price-guard.md
tamaralipows: > Any fields set in the request are applied on top of the PriceGuardConfig defaults, not on top of the server’s configured values.

I don't fully understand what this means - can you elaborate please? 

> any field omitted from the request falls back to the default in the table
below, not to the server's configured value. Set every field whose server default you do not want.

I'm confused whether it falls back to the PriceGuardConfig default or the server default because these sentences imply both? I think it should always fall back to the server default, perhaps this is incorrect? Let me know what you think, also @dianacarvalho1

docs/guides/price-guard.md
tamaralipows: I'm referring to this sorry:

> if any provider returns a price but none are within tolerance, the quote is rejected regardless of these flags.

docs/guides/price-guard.md
```diff
@@ -0,0 +1,134 @@
...
+# Price Guard
+
+The price guard validates each successful quote's `amount_out` against external price providers
+(Hyperliquid and Binance by default), catching mispriced quotes before they are encoded into
+transactions and submitted on-chain. Providers are queried concurrently, and a quote passes if
+**at least one** provider returns a price within tolerance. When validation fails, the quote's
+status is set to `price_check_failed`; other orders in the same batch are unaffected.
+
+Each provider runs a background worker that continuously populates an in-memory cache. On the
+solve path, `get_expected_out()` reads from cache — no blocking API calls during solving.
```
dianacarvalho1: here I think we are missing a paragraph on why is the price guard important and why should anyone use it

docs/guides/price-guard.md
dianacarvalho1: yes I'm also confused about this honestly 😅

docs/guides/price-guard.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Enabling the price guard
+
+The guard is disabled by default. Users enable it server-side with `--enable-price-guard`;
```
dianacarvalho1: I find these two paragraphs a bit confusing.. first we talk about the server config but then we are already introducing the table below and mixing things..
Could we re organize this a bit maybe?
I would suggest moving the table first as the price guard config attributes. and only then introduce that this can be set at the server or encoding side

docs/guides/price-guard.md
```diff
@@ -0,0 +1,134 @@
...
+- **`fail_on_token_price_not_found`** — applies when every provider was reached but none list the
+  token. `false` (default) lets the quote pass; `true` rejects it.
+
+When responses mix `price_not_found` with infrastructure errors, the token might be listed on one
+of the unreachable providers, so the guard applies `fail_on_provider_error` rather than
+`fail_on_token_price_not_found`.
+
+## Custom providers
+
+The `PriceProvider` trait, `ExternalPrice`, and `PriceProviderError` are public and re-exported
```
dianacarvalho1: " re-exported from `fynd-core`" why is this relevant?

docs/guides/price-guard.md
```diff
@@ -0,0 +1,134 @@
...
+from `fynd-core`. Implement the trait to add your own price source and register it via
```
dianacarvalho1: price provider

docs/guides/price-guard.md
```diff
@@ -0,0 +1,134 @@
...
+
+Providers follow a worker+cache pattern: `start()` spawns a background task that populates an
+in-memory cache, and `get_expected_out()` reads from that cache synchronously without blocking or
+making network calls.
+
+If no providers are registered before `build()`, the two built-in providers (Hyperliquid, Binance)
+are added automatically. Calling `register_price_provider()` before `build()` skips the defaults —
+register only what you need.
+
+## Example
```
dianacarvalho1: Example Quote with Price Guard protection

dianacarvalho1 (commented):
Thank you @Troshchk ! I have a couple more suggestions!

docs/guides/price-guard.md
Troshchk: Removed

docs/guides/price-guard.md
Troshchk: If at least one registered provider returns a price, the solution will always resolve to either pass or fail, regardless of the `fail_on_price_not_found` and `fail_on_provider_error flags`. These flags are only used as a fallback when no providers return a price, that's why it is "if any". Do you think it should be phrased differently?

docs/guides/price-guard.md
dianacarvalho1: > Currently, the server’s configured values are only used when the request does not include a price_guard configuration. If a price_guard is provided in the request, it fully overrides the server configuration. Any fields set in the request are applied on top of the PriceGuardConfig defaults, not on top of the server’s configured values.

I think this is a very weird behaviour.. but aren't the defaults the same everywhere anyway?

dianacarvalho1 (approved):
Thank you! Looks good to me now

PR #189, opened 2026-04-15
feat: Cap tycho-execution version

dianacarvalho1:
The latest versions of tycho-execution are not compatible with the currently deployed router. This cap will be removed once we redeploy the router at the end of the audit


tamaralipows (approved):
Good catch! Thank you!

PR #191, opened 2026-04-17
feat: keep ranked quote candidates for price guard fallback

Troshchk:
This PR improves quote selection by introducing a fallback mechanism during PriceGuard validation. Instead of validating only the top-ranked quote per order, we keep all quotes, rank them by amount_out_net_gas, and return the first one that passes the PriceGuard check. If PriceGuard is disabled, the top-ranked quote is returned as before.

### DIFF
--- a/fynd-core/src/price_guard/guard.rs
+++ b/fynd-core/src/price_guard/guard.rs
@@ -90,8 +90,7 @@ impl PriceGuard {
         let mut last = None;
         for candidate in candidates {
             if candidate.status() != QuoteStatus::Success {
-                last = Some(candidate);
-                continue;
+                return Ok(candidate);
             }
             let Some((token_in, token_out)) = self.validated_token_pair(&candidate) else {
                 last = Some(candidate);
@@ -104,9 +103,7 @@ impl PriceGuard {
         }
 
         if let Some(mut order_quote) = last {
-            if order_quote.status() == QuoteStatus::Success {
-                order_quote.set_status(QuoteStatus::PriceCheckFailed);
-            }
+            order_quote.set_status(QuoteStatus::PriceCheckFailed);
             Ok(order_quote)
         } else {
             // should never happen since the solver should always return at least one candidate per
--- a/fynd-core/src/price_guard/guard.rs
+++ b/fynd-core/src/price_guard/guard.rs
@@ -20,6 +20,9 @@ pub enum PriceGuardError {
     /// Price guard is enabled but no providers are registered.
     #[error("price guard is enabled but no providers are registered")]
     NoProviders,
+    /// Received an empty candidates list.
+    #[error("received empty candidates list")]
+    EmptyQuoteCandidates,
 }
 
 /// Validates solver outputs against external price sources.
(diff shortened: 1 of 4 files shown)


fynd-core/src/price_guard/guard.rs
```diff
@@ -48,43 +51,68 @@ impl PriceGuard {
...
 
-        Ok(quotes)
+        if let Some(mut order_quote) = last {
+            if order_quote.status() == QuoteStatus::Success {
+                order_quote.set_status(QuoteStatus::PriceCheckFailed);
+            }
+            Ok(order_quote)
+        } else {
+            // should never happen since the solver should always return at least one candidate per
+            // order
```
dianacarvalho1: it might not find any candidate for a certain order but I guess the list would wouldn't even be present here

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -345,27 +348,30 @@ impl WorkerPoolRouter {
...
+            return valid_quotes
+                .into_iter()
+                .map(|(_, q)| q.clone())
+                .collect();
         }
 
         // No valid quote found - return a NoRouteFound response
         // Try to get any response to extract block info, or create a placeholder
-        if let Some((_, any_q)) = responses.quotes.first() {
+        let placeholder = if let Some((_, any_q)) = responses.quotes.first() {
```
dianacarvalho1: placeholder?

fynd-core/src/price_guard/guard.rs
```diff
@@ -48,43 +51,68 @@ impl PriceGuard {
...
+    fn select_first_valid(
+        &self,
+        candidates: Vec<OrderQuote>,
+        config: &PriceGuardConfig,
+    ) -> Result<OrderQuote, PriceGuardError> {
+        let mut last = None;
+        for candidate in candidates {
+            if candidate.status() != QuoteStatus::Success {
+                last = Some(candidate);
                 continue;
```
dianacarvalho1: we already filter the `QuoteStatus` before reaching here no? in the rank method. So this logic is unnecessary at this point I would say

dianacarvalho1 (commented):
Thank you @Troshchk ! 🙏🏼 I like this! only have minor suggestions

fynd-core/src/price_guard/guard.rs
```diff
@@ -48,43 +51,68 @@ impl PriceGuard {
...
-                quote.set_status(QuoteStatus::PriceCheckFailed);
+            if self.check_price(&candidate, &token_in, &token_out, config) {
+                return Ok(candidate);
             }
+            last = Some(candidate);
         }
 
-        Ok(quotes)
+        if let Some(mut order_quote) = last {
+            if order_quote.status() == QuoteStatus::Success {
```
tamaralipows: Should we even be running the price check if the quote status isn't `Success`?

fynd-core/src/price_guard/guard.rs
```diff
@@ -48,43 +51,68 @@ impl PriceGuard {
...
-    /// 1. Checks that the quote has a well-formed route
-    /// 2. Queries all registered providers
-    /// 3. Passes if at least one provider validates within BPS tolerance
-    ///
-    /// Failures are always per-quote: the quote's status is set to
-    /// `PriceCheckFailed` and processing continues. Never aborts the batch.
+    /// Each inner `Vec<OrderQuote>` contains ranked candidates for a single order
+    /// (sorted by `amount_out_net_gas` descending). For each order, returns the
+    /// first candidate that passes price validation. If none pass, returns the
+    /// last candidate with status `PriceCheckFailed`.
```
tamaralipows, line 59: Should we maybe still return the candidate with the highest amount out then? Could be a sign of a symbol mismatch, no? In that case we would still want to show them the best route?

tamaralipows (commented):
Looking good! Just two things to discuss

fynd-core/src/price_guard/guard.rs
Troshchk: It will be here, but not empty, the `rank_quotes()` always return a list with at least one quote

fynd-core/src/worker_pool_router/mod.rs
Troshchk: Do you think it is a bad name?

fynd-core/src/price_guard/guard.rs
Troshchk: If there were no successful quotes, we return the so-called placeholder from `rank_quotes()`. Its status is not `QuoteStatus::Success`, so this is why this check is still needed

fynd-core/src/price_guard/guard.rs
Troshchk: True!
The case when the quote is not success is only when there were no successful quotes returned and `rank_quotes()` returns one quote with a status different to Success.
Changed

fynd-core/src/price_guard/guard.rs
Troshchk: Yes, makes sense, changed to return the best one👍

fynd-core/src/price_guard/guard.rs
Troshchk: I have removed that check by exiting early if the first quote is not sucessful

fynd-core/src/worker_pool_router/mod.rs
dianacarvalho1: yes 😅

Troshchk: fallback is better?

dianacarvalho1 (approved):
obridanke

PR #192, opened 2026-04-20
feat: Simplify Price Guard Config

tamaralipows:
Server controls on/off, clients set all params per-request. Request cannot enable if guard is disabled on the server side (gives more power to server-runner to completely disable reliance on external providers).

This was done because the previous behaviour (configurable on both the client and the request side) was becoming quite confusing. Server-side set parameters wre obscure and hidden from the request-maker, and defaults became meaningless.

### DIFF
--- a/docs/guides/price-guard.md
+++ b/docs/guides/price-guard.md
@@ -91,6 +91,19 @@ When responses mix `price_not_found` with infrastructure errors, the token might
 of the unreachable providers, so the guard applies `fail_on_provider_error` rather than
 `fail_on_token_price_not_found`.
 
+## Symbol collisions and long-tail tokens
+
+Price providers (Binance, Hyperliquid) identify tokens by their trading symbol — e.g. "ETH",
+"LINK", "PEPE". On-chain, symbols are not unique: any token can declare itself "PEPE", and
+multiple unrelated tokens on the same chain may share a symbol. The guard resolves tokens by
+matching the on-chain symbol from `SharedMarketData` to a provider's symbol, so a long-tail token
+whose symbol collides with a well-known token will be priced as if it were that token.
+
+In practice this means the guard works reliably for major tokens listed on CEXs, but may produce
+false rejections (or false passes) for obscure tokens that happen to share a symbol with a listed
+asset. Clients trading long-tail tokens should consider leaving `enabled: false` for those
+requests.
+
 ## Custom providers
 
 The `PriceProvider` trait, `ExternalPrice`, and `PriceProviderError` are public. Implement the
--- a/docs/guides/price-guard.md
+++ b/docs/guides/price-guard.md
@@ -15,31 +15,46 @@ to `price_check_failed`; other orders in the same batch are unaffected.
 
 ## Configuration
 
+The server controls only whether the guard is on or off. All other parameters — tolerance
+thresholds and fallback behavior — are set per-request by the client. Omitted fields fall back to
+struct defaults.
+
+### Server-side
+
+The guard is disabled by default. Enable it with `--enable-price-guard`:
+
+```bash
+fynd --enable-price-guard
+```
+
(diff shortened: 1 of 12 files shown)


fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -147,20 +147,21 @@ impl WorkerPoolRouter {
             .collect();
 
         // Validate against external prices if enabled.
-        // Per-request config (inside encoding_options) overrides the server default.
+        // Per-request overrides are merged onto the server config (not replaced).
         if let Some(ref guard) = self.price_guard {
-            let price_guard_config = request
+            let Some(server_config) = &self.price_guard_config else {
```
Troshchk: I am not sure about this check tbh -- I think the `self.price_guard_config` is only set is the PG is enabled in the server config with `with_price_guard()`. I think, that if the PG is disabled by default and it is overriden per request, the `self.price_guard_config` would not exist.
Should we fallback on`PriceGuardConfig` defaults in this case?

Troshchk (commented):
Thank you @tamaralipows! 💫 
I have one thing to discuss about the fallback, otherwise the breaking change for the PGOverrides looks ok to me

fynd-core/src/worker_pool_router/mod.rs
dianacarvalho1: > I think the self.price_guard_config is only set is the PG is enabled in the server config with with_price_guard()

I think this is already checked one line above
> if let Some(ref guard) = self.price_guard

dianacarvalho1 (approved):
Mercii! This looks good to me 🙏🏼

fynd-core/src/worker_pool_router/mod.rs
Troshchk: It means, that if the server price guard is not set(`self.price_guard` is None), than the overrides will be ignored?

dianacarvalho1: then it means that the price is not even enabled to begin with no? and no price providers should be running in the background - or did I misunderstand this?

Troshchk: Currently, FyndBuilder always starts default providers, but PriceGuard (and its config) is only added if enabled at startup. This means if a user tries to enable PG per request, the overrides are ignored.

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -146,35 +146,31 @@ impl WorkerPoolRouter {
...
-                return Err(SolveError::Internal("no price guard config available".to_string()));
-            };
+        // Validate against external prices when the client explicitly enables it.
+        let price_guard_config = request
+            .options()
+            .encoding_options()
+            .map(|e| e.price_guard())
+            .filter(|c| c.enabled());
+
+        if price_guard_config.is_some() && self.price_guard.is_none() {
```
Troshchk: cannot this be one of  the arms of the following `match (&self.price_guard, price_guard_config) {` ?

Troshchk (commented):
Thank you @tamaralipows!
It looks good to me, have just one small q

docs/guides/price-guard.md
```diff
@@ -76,6 +91,19 @@ When responses mix `price_not_found` with infrastructure errors, the token might
 of the unreachable providers, so the guard applies `fail_on_provider_error` rather than
 `fail_on_token_price_not_found`.
 
+## Symbol collisions and long-tail tokens
+
+Price providers (Binance, Hyperliquid) identify tokens by their trading symbol — e.g. "ETH",
```
dianacarvalho1, line 96: 👍🏼

dianacarvalho1 (approved):
yeeeess this is much better!! 🧹  Thank yooouu! 🙏🏼

PR #195, opened 2026-04-27
fix: make graph construction and SPFA relaxation deterministic

tvinagre:
## Summary

- Sort tokens and components before graph construction in `initialize_graph` and `add_components`, giving deterministic `NodeIndex` assignment and edge insertion order across process runs
- Sort the SPFA `active_nodes` set after collecting from `HashSet`, ensuring deterministic relaxation order between rounds
- Sort token lists within each component before generating edge pairs (defense-in-depth)

## Problem

The solver produced different routing results across process runs. Rust's `HashMap`/`HashSet` use random SipHash seeds per process, and graph construction iterated these collections to assign node indices and insert edges. Since petgraph's `graph.edges(node)` returns edges in insertion order, different seeds produced different BFS traversal and SPFA relaxation orders.

In the Bellman-Ford SPFA loop, within-round amount propagation means the order active nodes are processed affects which intermediate amounts get used — producing different final routes depending on iteration order.

## Test plan

- [ ] Existing unit tests pass (363 tests, 0 failures)
- [ ] Verify determinism: run integration quality test multiple times — same result each run

### DIFF
--- a/fynd-core/src/algorithm/bellman_ford.rs
+++ b/fynd-core/src/algorithm/bellman_ford.rs
@@ -531,6 +531,8 @@ impl Algorithm for BellmanFordAlgorithm {
             }
 
             active_nodes = next_active.into_iter().collect();
+            // Deterministic order: HashSet iteration is random per process.
+            active_nodes.sort_unstable();
         }
 
         // Check if destination was reached
--- a/fynd-core/src/algorithm/bellman_ford.rs
+++ b/fynd-core/src/algorithm/bellman_ford.rs
@@ -12,6 +12,16 @@
 //! - **SPFA (Shortest Path Faster Algorithm) queuing**: Only re-relaxes edges from nodes whose
 //!   amount improved
 //! - **Forbid revisits**: Skips edges that would revisit a token or pool already in the path
+//!
+//! # Known limitation: SPFA order-dependence
+//!
+//! SPFA reads and writes the same `amount[]` array within a round (unlike
+//! textbook Bellman-Ford which snapshots between rounds). Processing node B
+//! before node C can update intermediate amounts that C then builds on,
+//! producing different routes depending on iteration order. Active nodes are
+//! sorted by `NodeIndex` for determinism, but the chosen ordering is not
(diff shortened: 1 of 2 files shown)


fynd-core/src/graph/petgraph.rs
```diff
@@ -388,23 +392,33 @@ impl<D: Clone + Send + Sync> GraphManager<StableDiGraph<D>> for PetgraphStableDi
         self.edge_map.clear();
         self.node_map.clear();
 
-        let unique_tokens: HashSet<Address> = component_topology
+        // Sort tokens for deterministic NodeIndex assignment across processes.
```
zizou0x: It's assuming the same order of events (abstracting away the fact that tycho dataset changes over time)

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -531,6 +531,8 @@ impl Algorithm for BellmanFordAlgorithm {
             }
 
             active_nodes = next_active.into_iter().collect();
+            // Deterministic order: HashSet iteration is random per process.
```
zizou0x, line 544: Let's be more specific regarding one thing: this ordering could mean that we don't end up with the best solution (if the round gets exhausted before we finished processing all the active nodes) but given that this order is currently random, statistically ordering it doesn't change anything regarding the chances to end up with the best solution

zizou0x (approved):
LGTM

fynd-core/src/graph/petgraph.rs
tvinagre: Yes, this is only valid if the input is the same, will update the comment

fynd-core/src/algorithm/bellman_ford.rs
tvinagre: Yes, this is an known limitation of the algo. The algorithm's output quality depends on which lucky ordering the HashSet happens to give.
I'll check if this is explicited on the docs / algo code and, if not, will add a note there. I'll also add a note here to make sure this limitation is clear to the reader.

PR #196, opened 2026-05-02
feat: add --metrics-port flag to serve command

PoloX2021:
I'd like to be able to customize the metrics port. This can be very useful to run 2 fynd servers on different chains without opening them with a container.


kayibal (approved):
Thank you! LGTM

PR #197, opened 2026-05-07
feat: Improve gas estimations + add new audit gas estimations tool

dianacarvalho1:
This PR does a few things:
- Adds a new tool to audit gas estimations in Fynd. Originally @tvinagre started on this during cooldown. I took over and cleaned it up to this state. In a nutshell this tool:
  - Gets N samples from the `aggregator_trades_10k.json`
  - Asks Fynd for quotes
  - Runs simulations on them all
  - Compares the estimated gas from Tycho vs the node simulation gas usage
  - Outputs a report
  - Right now I added a `single_swaps.toml` worker config so we can easily get the results only for single swaps. The results for 500 samples are in the README
-  Updates Fynd to use the latest tycho-execution [feat: Upgrade to new tycho-execution that handles gas estimations](https://github.com/propeller-heads/fynd/pull/197/commits/89217832ecd4fe30b765d763a6ffef3fae18ebfa)
   -  Adds tokens to `Route`. This was a necessary change because tycho-execution `Swap` now takes `Token` instead of `Bytes`. I thought about adding `Token` to Fynd's `Swap` but this would be a really big breaking change, it was much more contained to just add it to `Route`. [feat: Add tokens attribute to Route](https://github.com/propeller-heads/fynd/pull/197/commits/53fdc9d4e7eb1e523d0e01846923cb75fd2e4a35)
   - Sets the tycho-execution `estimated_gas` in `OrderQuote`
- Small refactor: Extract erc20-overrides crate to dedupe slot detection (it was copied over from fynd-swap-cli)

TODO next (in other tasks):
- [ ] After the improvements to the individual protocols are stabilised + router overhead is added, re run the analysis and see how the results look like
- [ ] Run the analysis for sequential swaps


Cargo.toml
```diff
@@ -129,8 +131,8 @@ metrics = "0.24"
 metrics-exporter-prometheus = "0.16"
 
 # Tycho
-tycho-simulation = ">=0.254.0"
-tycho-execution = "<=0.175.0"
+tycho-simulation = { git = "https://github.com/propeller-heads/tycho-indexer", tag = "0.286.0" }
+tycho-execution = { git = "https://github.com/propeller-heads/tycho-indexer", tag = "0.286.0" }
```
dianacarvalho1: this is temporary until we push to crates again

tools/fynd-gas-audit/README.md
```diff
@@ -0,0 +1,197 @@
...
+Absolute ETH numbers scale linearly with gas price — the **relative** figures
+are the ones that matter.
+
+**Sample yield (500 trades attempted):**
+
+| outcome | count |
+|---|---|
+| success (quote + simulation) | 155 |
+| Fynd returned no route | 312 |
+| simulation reverted | 33 |
```
dianacarvalho1, line 73: some are because we can't find the token balance slots, others are real reverts (I didn't investigate more though since this tycho-execution doesn't even match fully with the deployed contracts)

fynd-core/src/types/quote.rs
```diff
@@ -929,12 +931,26 @@ impl BlockInfo {
 pub struct Route {
     /// Ordered sequence of swaps to execute.
     swaps: Vec<Swap>,
+    /// Full `Token` objects keyed by address, populated by the algorithm that
+    /// built the route. Skipped during (de)serialization — only the in-process
+    /// encoding path consumes it; wire consumers receive an empty map and
+    /// should not rely on it.
```
tamaralipows: So this means we get this information from Tycho - but the client will receive an empty hashmap here? There's no way to remove this totally for the client? I guess not, without overcomplicating things right (I'm lost in all the structs)

fynd-core/src/types/quote.rs
```diff
@@ -946,6 +962,12 @@ impl Route {
     pub fn into_swaps(self) -> Vec<Swap> {
         self.swaps
     }
+
+    /// Returns the token map attached to this route. Empty unless populated via
+    /// [`Route::with_tokens`].
+    pub fn tokens(&self) -> &HashMap<Bytes, Token> {
+        &self.tokens
+    }
```
tamaralipows, line 1031: Do we need this to be pub if the client will never access this?

fynd-core/src/worker_pool/worker.rs
```diff
@@ -196,6 +196,8 @@ where
                 let gas_price = result.gas_price().clone();
                 let route = result.into_route();
 
+                // This is a first naive approach to getting the total gas of this quote
+                // A finer estimation is done during encoding
```
tamaralipows, line 221: Do we need to keep this at all? Why?

tools/erc20-overrides/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "erc20-overrides"
+version.workspace = true
+edition = "2021"
+description = "Brute-force ERC-20 balance/allowance storage-slot detection for eth_call state overrides"
+license = "MIT"
```
tamaralipows, line 6: Is this the right license?

tools/fynd-gas-audit/src/main.rs
```diff
@@ -0,0 +1,238 @@
...
+    #[arg(long, default_value_t = 100)]
+    n: usize,
+
+    /// Max trades kept per (token_in, token_out) pair.
+    #[arg(long, default_value_t = 5)]
+    max_per_pair: usize,
+
+    /// Deterministic seed for the stratified shuffle.
+    #[arg(long, default_value_t = 42)]
+    seed: u64,
```
tamaralipows, line 55: I think I commented this in the doc but I forge the solution. This should ideally be different every time right? Else we might overfit to a specific sample?

tools/fynd-gas-audit/src/main.rs
```diff
@@ -0,0 +1,238 @@
...
+                    }
+                }
+            }
+        };
+
+        let error_gas = match (&gas_estimate, actual_gas) {
+            (Some(est), Some(actual)) => match est.to_i128() {
+                Some(e) => Some(e - i128::from(actual)),
+                None => {
+                    warn!("gas_estimate overflows i128: {est}");
```
tamaralipows, line 180: That's an expensive boi. Something's very wrong in this case no?

tools/fynd-gas-audit/README.md
```diff
@@ -0,0 +1,197 @@
...
+drifts from the actual gas a transaction would consume.
+
+---
+
+## Why this audit exists
+
+Fynd ranks route candidates by `amount_out_net_gas = amount_out − gas_cost`. If
+`gas_estimate` is biased, that ranking is biased: a systematically-low
+`gas_estimate` makes every route look cheaper than it is, which flatters
+routes in a way that doesn't reflect execution economics. This tool quantifies
```
tamaralipows: > flatters routes in a way that doesn't reflect execution economics

Very poetic. Can we simplify this haha

tools/fynd-gas-audit/README.md
tamaralipows: Should we open a task to investigate once we do the switch?

tools/fynd-gas-audit/src/types.rs
```diff
@@ -0,0 +1,71 @@
...
+///
+/// `BigUint` fields use `DisplayFromStr` so they serialise as decimal strings
+/// in CSV (the default serde representation is a sequence of u32 digits, which
+/// the csv crate can't put in a column).
+#[serde_as]
+#[derive(Debug, Clone, Serialize)]
+pub struct AuditRow {
+    pub token_in: String,
+    pub token_out: String,
+    pub amount_in: String,
```
tamaralipows, line 43: for my info: It's fine to keep pub attrs if it's just in tools? I guess it's more for ourself so its ok?

tools/fynd-gas-audit/README.md
```diff
@@ -0,0 +1,197 @@
...
+
+- **Portfolio bias:** Fynd's `amount_out_net_gas` is consistently
+  over-estimated by roughly 38% of the gas cost. At the 6.09 gwei this run
+  saw, the per-trade loss is ~$1.20 on the mean (the -0.000532 ETH headline
+  at ETH ≈ $2,285); at 20 gwei it scales to ~$4/trade.
+- **Per-trade ranking:** a systematic bias doesn't by itself flip the winner,
+  but it narrows margins. Higher-liquidity multi-hop routes that cost more gas
+  lose ground to lower-liquidity single-hop routes whenever the gas cost is
+  the tie-breaker. So the bias is more likely to matter when top-2 routes
+  have similar `amount_out`.
```
tamaralipows, line 149: This readme in general is quite wordy. Do we want to keep it like that? Isn't it meant for human eyes?

tools/fynd-gas-audit/src/report.rs
```diff
@@ -0,0 +1,448 @@
...
+        .collect();
+    let errors: Vec<f64> = successes
+        .iter()
+        .filter_map(|r| error_eth(r))
+        .collect();
+    if errors.is_empty() {
+        return summary;
+    }
+
+    let abs_errors: Vec<f64> = errors.iter().map(|x| x.abs()).collect();
```
tamaralipows, line 62: Do we need to keep the abs errors :/

tools/fynd-gas-audit/src/report.rs
```diff
@@ -0,0 +1,448 @@
...
+        "| Trades over-charged | {} / {} |",
+        summary.over_charged, summary.succeeded
+    )?;
+    writeln!(
+        &mut out,
+        "| Trades under-charged | {} / {} |",
+        summary.under_charged, summary.succeeded
+    )?;
+    writeln!(&mut out, "| Sum of signed error (ETH) | {:+.6} |", summary.sum_signed_eth)?;
+    writeln!(&mut out, "| Sum of absolute error (ETH) | {:.6} |", summary.sum_abs_eth)?;
```
tamaralipows: Why is this in ETH and not wei?

clients/rust/src/client.rs
```diff
@@ -231,6 +244,13 @@ fn storage_overrides_to_alloy(so: &StorageOverrides) -> Result<StateOverride, Fy
             .collect::<Result<alloy::primitives::map::B256HashMap<B256>, FyndError>>()?;
         result.insert(addr, AccountOverride { state_diff: Some(state_diff), ..Default::default() });
     }
+    for (addr_bytes, wei) in &so.balances {
+        let addr = mapping::bytes_to_alloy_address(addr_bytes)?;
+        let entry = result
+            .entry(addr)
+            .or_insert_with(AccountOverride::default);
+        entry.balance = Some(mapping::biguint_to_u256(wei));
+    }
```
tamaralipows, line 253: Not sure I understand what this change is for? ELI5 pls?

clients/rust/src/client.rs
```diff
@@ -212,12 +214,23 @@ impl StorageOverrides {
             .insert(slot, value);
     }
 
-    /// Merge all slot overrides from `other` into `self`.
+    /// Override the native (ETH) balance of an account for dry-run simulation.
+    ///
+    /// Useful when simulating transactions from a synthetic sender that has no real ETH —
+    /// many nodes (e.g. reth) reject `eth_estimateGas` if the sender cannot afford
+    /// `gas_limit * max_fee_per_gas`.
```
tamaralipows, line 221: Oh that's important. We already do that for our integration test I guess right? Was this broken before in fynd? Or were we assuming a synthetic sender was never used?

tamaralipows (commented):
Thanks @dianacarvalho1 ! A bunch of questions here. Please forgive me if I'm asking something you've already answered in some commit message or description, it was a lot to go through :)

tools/fynd-gas-audit/README.md
dianacarvalho1: just added it [here](https://propeller-heads.atlassian.net/browse/ENG-5928)

fynd-core/src/types/quote.rs
dianacarvalho1: no, actually this comment is misleading.. the dto and client types don't even have this attribute. I will remove it.

fynd-core/src/worker_pool/worker.rs
dianacarvalho1: to not do this, I would need to change the OrderQuote interface to make `gas_estimate` optional 😕 do you think it's worth it?

tools/erc20-overrides/Cargo.toml
dianacarvalho1: yes

tools/fynd-gas-audit/src/main.rs
dianacarvalho1: yes, it should be different every time. I will add a note to the docs to mention this. I think this is here just for reproducibility

tools/fynd-gas-audit/src/main.rs
dianacarvalho1: yes xD

tools/fynd-gas-audit/src/types.rs
dianacarvalho1: yes I think so.. we are not making big commitments in these tools! they are mostly for us/curious users

tools/fynd-gas-audit/README.md
dianacarvalho1: yes, it's for human eyes but it was made by claude obviously 😂  Do you think it's worth simplifying? Tbh this is just a tool for us, I think we don't need to make it extra perfect and squeaky clean.

clients/rust/src/client.rs
dianacarvalho1: > We already do that for our integration test I guess right? 

yes, in the integration test we gives moneys to the sender. 

> Was this broken before in fynd?

I think it is for the simulation case, that's why I had to add it here. But the user is responsible for setting this themselves.

clients/rust/src/client.rs
dianacarvalho1: this is to add the balance overwrites to the `StateOverride`
what is ELI5?

tools/fynd-gas-audit/src/report.rs
dianacarvalho1: changing it

tools/fynd-gas-audit/src/report.rs
dianacarvalho1: no

fynd-core/src/types/quote.rs
dianacarvalho1: made it pub(crate)

fynd-core/src/worker_pool/worker.rs
tamaralipows: couldn't we just set it to 0? No strong opinion though I'm fine with keeping it

clients/rust/src/client.rs
tamaralipows: explain like I'm 5 lol haha. Got it though, thanks

tamaralipows (approved):
Thank you!

fynd-core/src/worker_pool/worker.rs
dianacarvalho1: ah yes, we could set it to 0.. I just left the old behaviour as a default

louise-poole: We need an initial estimate here because it's used when ranking and comparing solutions. Can you maybe mention that in the comment? We should defs not set this to 0 though!

fynd-core/src/types/quote.rs
```diff
@@ -929,12 +931,25 @@ impl BlockInfo {
...
+        Self { swaps, tokens: HashMap::new() }
+    }
+
+    /// Attaches a token map (address → full `Token`) to the route. The encoder
+    /// uses this to convert each swap's `token_in`/`token_out` addresses back
+    /// into the full `Token` objects required by `tycho-execution`.
+    pub(crate) fn with_tokens(mut self, tokens: HashMap<Bytes, Token>) -> Self {
+        self.tokens = tokens;
+        self
     }
```
louise-poole: This makes it too easy for someone implementing an algorithm to overlook adding tokens to their route. I know you have a warning at execution time if this has happened, but a compilation warning would be much better. What's the argument against just updating `new` to take a tokens param too? If we expect someone to add an algorithm that won't use tycho-execution then maybe we can have a special `new_without_tokens`? I'd rather the non-default behaviour (not using our tycho-execution) use a non-default constructor.

tools/fynd-gas-audit/single_hops.toml
louise-poole, line 1: We use the terminology of 'hops' when in the solver... could we name this file `single_hops`?

fynd-core/src/types/quote.rs
louise-poole: Side note: we're still pre-V1 so breaking interfaces isn't the end of the world.

fynd-core/src/worker_pool/worker.rs
dianacarvalho1: ah damn.. but that raises another question.. we are only encoding (and estimating gas) for best solution already 🫠 so when comparing solutions, the gas estimations might be really off 😬

fynd-core/src/types/quote.rs
dianacarvalho1: it was only not to break the interface. but you are right. will break it then 🪓

fynd-core/src/worker_pool/worker.rs
dianacarvalho1: will address this in a separate PR to make review nicer

PR #198, opened 2026-05-13
feat: add connector token restriction for routing algorithms

kayibal:
## Summary

Adds optional per-pool connector token allowlists to reduce on-chain reversion risk from illiquid intermediate hops, and a CLI subcommand to derive good connector tokens from live market data.

**Connector token restriction**

`AlgorithmConfig` gains an optional `connector_tokens` allowlist. When set, only those tokens may appear as intermediates in a route — the source and destination are always allowed regardless. Absent means no restriction, preserving existing behaviour.

- BellmanFord: filter applied inline in the SPFA edge-relaxation loop, before simulation
- MostLiquid: filter applied in BFS path enumeration; destination node always passes
- `PoolConfig` gains a TOML-deserializable `connector_tokens` field (list of `0x`-prefixed addresses)
- `FyndBuilder::add_pool` now returns `Result` instead of panicking on malformed addresses
- `FyndRPCBuilder::new` now returns `Result`; uses `try_fold` for pool construction

**`fynd derive-connector-tokens` subcommand**

Connects to Tycho, waits for the initial market snapshot, scores every token by pool count, and outputs a ranked list as TOML, JSON, or plain text. Useful for generating a chain-specific allowlist to paste into `worker_pools.toml`.

```
fynd derive-connector-tokens --chain Ethereum --top-n 10 --output toml
```

**Shared CLI helpers**

`resolve_tycho_url`, `resolve_rpc_url`, and `resolve_protocols` extracted to `main.rs`. Both `serve` and `derive-connector-tokens` share the same `--protocols` semantics: empty or `all_onchain` fetches all protocols from Tycho RPC.

**`worker_pools.toml`**

Connector token examples are commented out (default remains no restriction). Includes derived top-10 lists for Ethereum mainnet and Base as a starting point.


fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -1296,6 +1311,105 @@ mod tests {
...
+            AlgorithmConfig::new(1, max_hops, Duration::from_millis(timeout_ms), None)
+                .unwrap()
+                .with_connector_tokens(connector_tokens),
+        )
+        .unwrap()
+    }
+
+    #[tokio::test]
+    async fn test_connector_tokens_blocks_disallowed_intermediate() {
+        // Diamond: A->B->D (2x * 2x = 4x) vs A->C->D (3x * 3x = 9x).
```
dianacarvalho1: what is the `(2x * 2x = 4x)` and `(3x * 3x = 9x)` ?

dianacarvalho1: maybe it would be more explicit to actually draw the diamond?

````
     A
   /   \
  B     C
   \   /
     D
````

File: fynd-core/src/algorithm/bellman_ford.rs [hunk 2]
```diff
@@ -1296,6 +1311,105 @@ mod tests {
...
+        // C is in the allowlist; B is not.
+        // Expect A->C->D to be found; B-path is pruned.
+        let token_a = token(0x01, "A");
```
dianacarvalho1, line 1338: there is a fixture that does already this `diamond_graph`. It is being used in the most_liquid tests already

src/commands/derive_connector_tokens.rs
```diff
@@ -0,0 +1,220 @@
...
+}
+
+struct TokenScore {
+    symbol: String,
+    pool_count: usize,
+}
+
+async fn score_tokens(
+    market_data: &SharedMarketDataRef,
+    _chain: Chain,
```
dianacarvalho1: unnecessary _chain arg

src/commands/derive_connector_tokens.rs
```diff
@@ -0,0 +1,220 @@
...
+            (addr, TokenScore { symbol, pool_count: count })
+        })
+        .collect()
+}
+
+fn print_toml(candidates: &[&(Address, TokenScore)], chain: &str) {
+    use chrono::Utc;
+    let date = Utc::now().format("%Y-%m-%d");
+    println!("# Derived connector tokens for {chain} ({date})");
+    println!("# Score = pool_count. Top {} of {} tokens.", candidates.len(), candidates.len());
```
dianacarvalho1: this print is a bit weird.. it uses the `candidates.len()` twice. So it appears like "# Score = pool_count. Top 10 of 10 tokens"

dianacarvalho1 (approved):
Thank you @kayibal !! Overall this looks good to me 🙏🏼 only have small comments
It's actually interesting to see the tokens ranked like this 🤔

fynd-core/src/algorithm/bellman_ford.rs
kayibal: Yeah unfortunately that fixture is not compatible with BF.

PR #199, opened 2026-05-15
feat(fynd-core): labeled simulation-state overlays

kayibal:
## Summary

- Adds `StateLabel` newtype and labeled-state overlay support to `SharedMarketData`. External crates can register named overlays (`register_labeled_state`, `remove_labeled_state`, `clear_labeled_states`) via `SharedMarketDataRef` and solve against them by setting `QuoteOptions::with_state_label(label)`.
- `SharedMarketDataRef` converted from type alias to struct carrying an optional `StateLabel`. `with_label()` returns a cheap Arc-clone targeting an overlay; `MarketDataReadGuard::get_simulation_state` and `extract_subset_with_overlay` check the overlay first and fall back to base silently — both built-in algorithms are transparent to the change.
- `MarketEvent` and `subscribe_market_events()` on `Solver` are now public so external feed implementations (mempool, flashblocks, historical window) can react to block updates and rotate their labeled overlays without any changes to `fynd-core`.

## Test Plan

- [ ] `cargo nextest run --workspace --locked --all-features` — 624 tests pass
- [ ] `./check.sh` (nightly fmt + clippy -D warnings + docs + nextest) passes
- [ ] Labeled overlay lookup: pool in overlay returns overlay state; pool absent from overlay falls back to base
- [ ] Label threading: `QuoteOptions::with_state_label(l)` → label on all `SolveTask`s → `OrderQuote.state_label` reflects requested label
- [ ] No-label path: `OrderQuote.state_label` equals current block hash when no label requested
- [ ] `Solver::subscribe_market_events()` receiver fires on each `MarketUpdated` event

### DIFF
--- a/fynd-core/src/worker_pool/task_queue.rs
+++ b/fynd-core/src/worker_pool/task_queue.rs
@@ -7,7 +7,9 @@
 use tokio::sync::oneshot;
 use uuid::Uuid;
 
-use crate::{types::internal::SolveTask, Order, SingleOrderQuote, SolveError};
+use crate::{
+    feed::market_data::StateLabel, types::internal::SolveTask, Order, SingleOrderQuote, SolveError,
+};
 
 /// Configuration for the task queue.
 #[derive(Debug, Clone)]
@@ -51,6 +53,26 @@ impl TaskQueueHandle {
             .map_err(|_| SolveError::QueueFull)?;
 
         // Wait for response
+        response_rx
+            .await
+            .map_err(|_| SolveError::Internal("worker dropped response channel".to_string()))?
+    }
+
+    /// Enqueues a solve request targeting the given labeled market state.
+    ///
+    /// Pass `None` to use the base Tycho state.
+    pub async fn enqueue_with_label(
+        &self,
+        order: Order,
+        state_label: Option<StateLabel>,
+    ) -> Result<SingleOrderQuote, SolveError> {
+        let (response_tx, response_rx) = oneshot::channel();
+        let task_id = Uuid::new_v4();
+        let task = SolveTask::new_with_label(task_id, order, state_label, response_tx);
+        self.sender
+            .send(task)
+            .await
+            .map_err(|_| SolveError::QueueFull)?;
         response_rx
             .await
             .map_err(|_| SolveError::Internal("worker dropped response channel".to_string()))?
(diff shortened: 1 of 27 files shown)


fynd-core/src/worker_pool/task_queue.rs
```diff
@@ -56,6 +58,26 @@ impl TaskQueueHandle {
...
     }
 
+    /// Enqueues a solve request targeting the given labeled market state.
+    ///
+    /// Pass `None` to use the base Tycho state.
+    pub async fn enqueue_with_label(
+        &self,
+        order: Order,
+        state_label: Option<StateLabel>,
+    ) -> Result<SingleOrderQuote, SolveError> {
```
kayibal, line 68: Alternatively we could refactor enqueue to take a params struct

fynd-core/src/worker_pool/worker.rs
```diff
@@ -128,7 +128,11 @@ where
     }
 
     /// Returns a quote for an order.
-    pub async fn quote(&mut self, order: &Order) -> Result<SingleOrderQuote, SolveError> {
+    pub async fn quote(
+        &mut self,
+        order: &Order,
+        state_label: Option<StateLabel>,
+    ) -> Result<SingleOrderQuote, SolveError> {
```
kayibal, line 135: Same here consider a params struct instead

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -195,6 +199,7 @@ impl WorkerPoolRouter {
     async fn solve_order(
         &self,
         order: Order,
+        state_label: Option<StateLabel>,
```
kayibal, line 264: consider introducing a params struct, we have already a lot here

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -424,9 +437,11 @@ impl WorkerPoolRouter {
...
                 Bytes::default(),
                 Bytes::default(),
-                // Task 7 will thread the requested label here; for now use "none" to indicate
-                // no market state was available.
-                StateLabel::new("none"),
+                // Use the requested label when available; fall back to "none" only when no label
+                // was requested and no market state was available.
+                state_label
+                    .cloned()
+                    .unwrap_or_else(|| StateLabel::new("none")),
```
kayibal: Is this necessary if all quotes should have a label assigned to them? I agree if the orders have different labels ranking them does not make sense and we should error instead.

kayibal:
Closing in favour of a cleaner implementation. The design was revised after review: overlays are now stored on `SharedMarketDataRef` under a separate lock, decoupled from the base-state `RwLock`. Continued in #TBD.

PR #200, opened 2026-05-19
feat(fynd-core): labeled simulation-state overlays

kayibal:
## Summary

- Solve requests can now target a named state overlay via `QuoteOptions::with_state_label`. Pools absent from the overlay fall through to the base Tycho state — two O(1) lookups, no allocation.
- Overlays are stored on `SharedMarketDataRef` under a dedicated `overlays` lock, separate from the base-state `data` lock. Overlay inserts and solver reads are fully concurrent.
- `MarketDataReadGuard` pre-fetches the overlay snapshot under a brief overlay read lock, releases it, then `get_simulation_state` checks overlay first — algorithm call sites are unchanged.
- Overlays carry `valid_until: u64`. `apply_block_update` atomically evicts stale overlays then applies the block, making the correct ordering structural.
- `Solver::subscribe_market_events` lets external state managers (mempool, flashblock feeds) react to block transitions without any logic living in `fynd-core`.
- `SolveParams` flows through the full task pipeline (`QuoteOptions` → router → `TaskQueueHandle::enqueue` → `SolveTask` → `SolverWorker::quote`), keeping the public request surface and internal solve parameters distinct. `OrderQuote::solved_against` records the state actually used, not the label requested.

Replaces #199. The initial design (also in #199) stored overlays inside `SharedMarketData` under the same `RwLock` as pool state, so every overlay insert competed with solver reads of the base state. After reviewing that design we separated the two locks: base state and overlay registry are independent `Arc<RwLock<_>>` fields on `SharedMarketDataRef`, so inserts never block solvers.

### DIFF
--- a/fynd-core/src/algorithm/bellman_ford.rs
+++ b/fynd-core/src/algorithm/bellman_ford.rs
@@ -379,7 +379,10 @@ impl Algorithm for BellmanFordAlgorithm {
 
         // Acquire read lock only for market data extraction, then release.
         let (token_map, market_subset) = {
-            let market = market.read(label.as_ref()).await;
+            let market = market
+                .read_opt(label.as_ref())
+                .await
+                .map_err(|e| AlgorithmError::Other(e.to_string()))?;
 
             let token_map: HashMap<NodeIndex, Token> = token_nodes
                 .iter()
--- a/fynd-core/src/algorithm/bellman_ford.rs
+++ b/fynd-core/src/algorithm/bellman_ford.rs
@@ -44,7 +44,7 @@ use crate::{
         types::{SpotPrices, TokenGasPrices},
         SharedDerivedDataRef,
     },
-    feed::market_data::SharedMarketDataRef,
+    feed::market_data::MarketData,
     graph::{petgraph::StableDiGraph, PetgraphStableDiGraphManager},
     types::{ComponentId, Order, Route, RouteResult, Swap},
 };
@@ -338,7 +338,7 @@ impl Algorithm for BellmanFordAlgorithm {
     async fn find_best_route(
         &self,
         graph: &Self::GraphType,
-        market: SharedMarketDataRef,
+        market: MarketData,
         derived: Option<SharedDerivedDataRef>,
         order: &Order,
     ) -> Result<RouteResult, AlgorithmError> {
@@ -666,7 +666,10 @@ impl Algorithm for BellmanFordAlgorithm {
 
 #[cfg(test)]
 mod tests {
+    use std::sync::Arc;
+
     use num_bigint::BigInt;
+    use tokio::sync::RwLock;
     use tycho_simulation::{
         tycho_common::{models::Address, simulation::protocol_sim::ProtocolSim},
         tycho_ethereum::gas::{BlockGasPrice, GasPrice},
@@ -676,7 +679,7 @@ mod tests {
     use crate::{
         algorithm::test_utils::{component, order, token, MockProtocolSim},
         derived::{types::TokenGasPrices, DerivedData},
-        feed::market_data::{SharedMarketData, SharedMarketDataRef},
+        feed::market_data::{MarketData, MarketState},
         graph::GraphManager,
         types::quote::OrderSide,
     };
@@ -686,8 +689,8 @@ mod tests {
     /// Sets up market and graph with `()` edge weights for BellmanFord tests.
     fn setup_market_bf(
         pools: Vec<(&str, &Token, &Token, MockProtocolSim)>,
-    ) -> (SharedMarketDataRef, PetgraphStableDiGraphManager<()>) {
-        let mut market = SharedMarketData::new();
+    ) -> (MarketData, PetgraphStableDiGraphManager<()>) {
+        let mut market = MarketState::new();
 
         market.update_gas_price(BlockGasPrice {
             block_number: 1,
@@ -708,10 +711,7 @@ mod tests {
         let mut graph_manager = PetgraphStableDiGraphManager::default();
         graph_manager.initialize_graph(&market.component_topology());
 
-        (
-            SharedMarketDataRef::new(std::sync::Arc::new(tokio::sync::RwLock::new(market))),
-            graph_manager,
-        )
+        (MarketData::new(Arc::new(RwLock::new(market))), graph_manager)
     }
 
     fn setup_derived_with_token_prices(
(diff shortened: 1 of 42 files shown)


fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -729,7 +729,7 @@ mod tests {
 
         let mut derived_data = DerivedData::new();
         derived_data.set_token_prices(token_prices, vec![], 1, true);
-        Arc::new(RwLock::new(derived_data))
+        std::sync::Arc::new(tokio::sync::RwLock::new(derived_data))
```
zizou0x: What's the motivation? Keeping explicit which implementation we use?

fynd-core/src/feed/market_data.rs
```diff
@@ -24,8 +32,192 @@ use tycho_simulation::{
...
+    ///
+    /// All clones share the same overlay registry, so a state registered via any handle
+    /// is visible to all handles with the same label.
+    pub fn with_label(&self, label: StateLabel) -> Self {
+        Self {
+            data: Arc::clone(&self.data),
+            overlays: Arc::clone(&self.overlays),
+            label: Some(label),
+        }
+    }
```
zizou0x: I find this weird, why would we need to store the label inside the `SharedMarketDataRef`?

It should contain data for all overlays anyway, not a specific one

fynd-core/src/feed/market_data.rs
```diff
@@ -24,8 +32,192 @@ use tycho_simulation::{
...
+
+    /// Acquires a read guard that exposes both the base data and the active overlay (if any).
+    ///
+    /// The overlay lock is held only briefly to clone the overlay `Arc` pointer; it is released
+    /// before the guard is returned, so solving never holds two locks simultaneously.
+    pub async fn read(&self) -> MarketDataReadGuard<'_> {
```
zizou0x: I guess this is the motivation, to keep the interface similar to `RwLock`?

Maybe it could be another method that takes the label as arg, we keep this one simple to return the baseline state guard but without overlay

fynd-core/src/feed/market_data.rs
```diff
@@ -24,8 +32,192 @@ use tycho_simulation::{
...
+
+    /// Returns the labels of all registered overlays.
+    pub async fn labeled_state_ids(&self) -> Vec<StateLabel> {
+        self.overlays
+            .read()
+            .await
+            .keys()
+            .cloned()
+            .collect()
+    }
```
zizou0x, line 200: Do we expect this to be used?

fynd-core/src/feed/market_data.rs
```diff
@@ -24,8 +32,192 @@ use tycho_simulation::{
...
+    overlay: Option<OverlayStates>,
+}
+
+impl Deref for MarketDataReadGuard<'_> {
+    type Target = SharedMarketData;
+
+    fn deref(&self) -> &SharedMarketData {
+        &self.guard
+    }
+}
```
zizou0x: Unfortunately this is considered anti-pattern by the Rust community [see here](https://rust-unofficial.github.io/patterns/anti_patterns/deref.html)

I think we need to explicitly re-expose the methods here

fynd-core/src/price_guard/binance_ws.rs
```diff
@@ -758,7 +757,7 @@ mod tests {
 
     /// Creates a `BinanceWsWorker` that writes to the given ticker cache.
     fn make_worker(price_cache: &PriceCache) -> BinanceWsWorker {
-        let market_data = Arc::new(RwLock::new(SharedMarketData::new()));
+        let market_data = crate::feed::market_data::SharedMarketDataRef::new_shared();
```
zizou0x: Can't we import it to avoid having the full path duplicated?

fynd-core/src/feed/market_data.rs
zizou0x: I think the solution I suggested above is actually not that great because we need to carry the label around. I'd still investigate if we can get some separation between `SharedMarketDataRef` and the current label. Maybe a wrapper would be a nice solution.

fynd-core/src/feed/market_data.rs
kayibal: Are we trying to achieve polymorphism here or is this actually just a extended reference to shared market data?

fynd-core/src/feed/market_data.rs
kayibal: No this is meant to be used if you want to solve over all current open states.

fynd-core/src/feed/market_data.rs
kayibal: It's not just about the interface but also the semantics of data access a lock gives you.

fynd-core/src/feed/market_data.rs
zizou0x: From the std::Deref [docs](https://doc.rust-lang.org/std/ops/trait.Deref.html#when-to-implement-deref-or-derefmut):

> In general, deref traits should not be implemented if:
> 
> 1. the deref implementations could fail unexpectedly; or
> 2. the type has methods that are likely to collide with methods on the target type; or
> 3. committing to deref coercion as part of the public API is not desirable.

I think point 2. and 3. apply to this case so I wouldn't use it

fynd-core/src/feed/market_data.rs
```diff
@@ -53,88 +52,77 @@ pub struct OverlayEntry {
...
+    pub async fn read(&self, label: Option<&StateLabel>) -> MarketDataView<'_> {
         let guard = self.data.read().await;
-        let overlay = if let Some(ref label) = self.label {
+        let overlay = if let Some(label) = label {
             self.overlays
                 .read()
                 .await
                 .get(label)
-                .map(|e| Arc::clone(&e.states))
+                .map(|e| (label.clone(), Arc::clone(&e.states)))
```
zizou0x: Should we add a warn if the overlay isn't found?

fynd-core/src/feed/market_data.rs
```diff
@@ -24,15 +31,262 @@ use tycho_simulation::{
...
+
+/// The shared overlay registry: maps each label to its snapshot.
+type OverlayRegistry = Arc<RwLock<HashMap<StateLabel, OverlayEntry>>>;
+
+/// The main entry point for accessing market data.
+///
+/// Cloning is cheap — all clones share the same underlying data and overlay registry.
+/// Pass an optional label to `read` to scope the view to a specific overlay.
+#[derive(Clone)]
+pub struct MarketData {
```
zizou0x, line 68: Could you get your Claude to update the docs and the claude related files too please? I still see some “SharedMarketData” mentions in these files

zizou0x (approved):
LGTM! Thanks

PR #201, opened 2026-05-20
feat(gas-estimation): Use estimate_gas_usage to refine gas estimations

dianacarvalho1:
This is only done in the WorkerPoolRouter for the winning solution of each worker! Inside each worker, the ranking is done using the original naive gas estimation (just a sum of each swap simulated gas).

One thing to note: This implementation makes us create the Solution (encoding) objects twice:
1. Once for the gas refinement for all solutions from each worker
2. Once for the overall winner solution when we encode (existing code)  

Other approaches were tried to clean this up and convert only once but that lead to an even weirder design and having to pass the Solution objects in multiple places. Converting to Solution is not expensive at all, so I decided to use the simplest solution.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -155,9 +155,10 @@ Orchestrates quote requests across multiple worker pools:
 
 1. Fans out each order to all pools in parallel
 2. Manages per-request timeouts with optional early return
-3. Selects the best solution by `amount_out_net_gas`
-4. Optionally encodes winning solutions into on-chain transactions (when `EncodingOptions` are provided)
-5. Reports failures with error types and metrics
+3. Refines gas estimates using `estimate_gas_usage` (tycho-execution) before cross-pool ranking — algorithms use a fast naive estimate internally; this step applies a more accurate one that accounts for token transfers and protocol overhead
+4. Selects the best solution by refined `amount_out_net_gas`
+5. Optionally encodes winning solutions into on-chain transactions (when `EncodingOptions` are provided)
+6. Reports failures with error types and metrics
 
 ***
 
@@ -200,8 +201,8 @@ Pluggable interface for route-finding algorithms:
 
 **Built-in algorithms:**
 
-* `MostLiquidAlgorithm` -- BFS path enumeration, depth-weighted scoring, ProtocolSim simulation, gas-adjusted ranking.
-* `BellmanFordAlgorithm` -- Bellman-Ford relaxation with gas-aware edge weights, configurable via `AlgorithmConfig.gas_aware`.
+* `MostLiquidAlgorithm` -- BFS path enumeration, depth-weighted scoring, ProtocolSim simulation, gas-adjusted ranking (naive `route.total_gas()` within the algorithm; more accurate estimate applied by the router before cross-pool ranking).
+* `BellmanFordAlgorithm` -- Bellman-Ford relaxation with gas-aware edge weights, configurable via `AlgorithmConfig.gas_aware` (same two-phase gas estimation as above).
 
 ***
 
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -201,8 +201,8 @@ Pluggable interface for route-finding algorithms:
 
 **Built-in algorithms:**
 
-* `MostLiquidAlgorithm` -- BFS path enumeration, depth-weighted scoring, ProtocolSim simulation, gas-adjusted ranking (naive `route.total_gas()` within the algorithm; more accurate estimate applied by the router before cross-pool ranking).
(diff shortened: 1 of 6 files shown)


docs/ARCHITECTURE.md
```diff
@@ -200,8 +201,8 @@ Pluggable interface for route-finding algorithms:
 
 **Built-in algorithms:**
 
-* `MostLiquidAlgorithm` -- BFS path enumeration, depth-weighted scoring, ProtocolSim simulation, gas-adjusted ranking.
-* `BellmanFordAlgorithm` -- Bellman-Ford relaxation with gas-aware edge weights, configurable via `AlgorithmConfig.gas_aware`.
+* `MostLiquidAlgorithm` -- BFS path enumeration, depth-weighted scoring, ProtocolSim simulation, gas-adjusted ranking (naive `route.total_gas()` within the algorithm; more accurate estimate applied by the router before cross-pool ranking).
+* `BellmanFordAlgorithm` -- Bellman-Ford relaxation with gas-aware edge weights, configurable via `AlgorithmConfig.gas_aware` (same two-phase gas estimation as above).
```
louise-poole: Don't think it's necessary to add the gas comments here

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -437,6 +452,49 @@ impl WorkerPoolRouter {
...
+        1 => Strategy::Single,
+        _ if route
+            .swaps()
+            .iter()
+            .any(|s| *s.split() > 0.0) =>
+        {
+            Strategy::Split
+        }
+        _ => Strategy::Sequential,
+    }
```
louise-poole: I find it weird to have a match statement with only one real branch and the rest be wildcard branches `_`...

```suggestion
    let swaps = route.swaps();

    if swaps.len() == 1 {
        Strategy::Single
    } else if swaps.iter().any(|s| *s.split() > 0.0) {
        Strategy::Split
    } else {
        Strategy::Sequential
    }
```

louise-poole (approved):
Some minor comments, otherwise looks good!

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -11,8 +11,13 @@
 //!    customized, but initially it's set to relay to all solvers.
 //! 2. **Timeout**: Cancel if solver response takes too long
 //! 3. **Collection**: Wait for N responses OR timeout per order
-//! 4. **Selection**: Choose best quote (max `amount_out_net_gas`)
-//! 5. **Encoding**: If [`EncodingOptions`](crate::EncodingOptions) are provided in the request,
+//! 4. **Gas refinement**: Before cross-pool ranking, replace each candidate's naive
+//!    `route.total_gas()` estimate (used internally by algorithms for intra-pool ranking) with the
+//!    more accurate `estimate_gas_usage` from tycho-execution, which accounts for token transfer
+//!    costs and router overhead. The `amount_out_net_gas` values are rescaled proportionally so the
```
tamaralipows, line 17: What do you mean by "rescaled proportionally"?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -437,6 +452,47 @@ impl WorkerPoolRouter {
...
+    for responses in order_responses {
+        for (_, quote) in &mut responses.quotes {
+            if quote.status() != QuoteStatus::Success {
+                continue;
+            }
+            let solution = Solution::try_from(&*quote)?
+                .with_user_transfer_type(encoding_options.transfer_type().clone());
+            let refined_gas = estimate_gas_usage(&solution, derive_strategy(quote));
+            let naive_gas = quote.gas_estimate().clone();
+            if naive_gas > BigUint::ZERO {
```
tamaralipows, line 468: What does it mean for the `naive_gas` to be zero again? And why would we like it to stay zero in that case?  Can we add a comment?

It also looks like we can move the `refined_gas` into the if block?

tamaralipows (approved):
Thank you! Looks good - just small things

fynd-core/src/worker_pool_router/mod.rs
dianacarvalho1: to deduct the refined gas cost from the amountOut I would need to know about the token prices 😬 which I do not at this point.. so I'm using the initial gas estimation and amountOut diff as a proportion to infer this value!

fynd-core/src/worker_pool_router/mod.rs
dianacarvalho1: it would mean that the protocol gas was empty I think. This is a safe guard not to do crazy math there. because if this is 0 then we can't do the rescaling (by inferring the token price from this)

PR #204, opened 2026-05-22
fix(encoding): Use proper eth marker

tamaralipows:
- Easily done thanks to ROUTER_ETH_ADDRESS export from tycho-execution
- Needed to update test_encoder_new_fails_on_unsupported_chain since Arbitrum is now actually supported
- Update failing tests after gas updates (perhaps failing in base branch)

### DIFF
--- a/fynd-core/src/encoding/encoder.rs
+++ b/fynd-core/src/encoding/encoder.rs
@@ -220,18 +220,12 @@ impl Encoder {
         let min_amount_out = biguint_to_u256(fee_breakdown.min_amount_received());
         let native_address = &self.chain.native_token().address;
         let router_eth = Address::from_slice(ROUTER_ETH_ADDRESS.as_ref());
-        let raw_token_in = bytes_to_address(solution.token_in())?;
-        let raw_token_out = bytes_to_address(solution.token_out())?;
-        let token_in = if raw_token_in.as_slice() == native_address.as_ref() {
-            router_eth
-        } else {
-            raw_token_in
+        let to_router_address = |raw: Address| {
+            if raw.as_slice() == native_address.as_ref() { router_eth } else { raw }
         };
-        let token_out = if raw_token_out.as_slice() == native_address.as_ref() {
-            router_eth
-        } else {
-            raw_token_out
-        };
+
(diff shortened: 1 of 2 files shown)


fynd-core/src/encoding/encoder.rs
```diff
@@ -217,8 +218,20 @@ impl Encoder {
...
+        let token_in = if raw_token_in.as_slice() == native_address.as_ref() {
+            router_eth
+        } else {
+            raw_token_in
+        };
+        let token_out = if raw_token_out.as_slice() == native_address.as_ref() {
+            router_eth
+        } else {
+            raw_token_out
+        };
```
louise-poole, line 229: Just a rusty suggestion (up to you if you like it or not):
```suggestion
        let native_address = &self.chain.native_token().address;
        let router_eth = Address::from_slice(ROUTER_ETH_ADDRESS.as_ref());
        let to_router_address = |raw: Address| {
            if raw.as_slice() == native_address.as_ref() { router_eth } else { raw }
        };

        let token_in  = to_router_address(bytes_to_address(solution.token_in())?);
        let token_out = to_router_address(bytes_to_address(solution.token_out())?);
```

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -646,7 +646,7 @@ mod tests {
         assert_eq!(quote.orders().len(), 1);
         assert_eq!(quote.orders()[0].status(), QuoteStatus::Success);
         // amount_out_net_gas is refined using estimate_gas_usage before ranking
-        assert_eq!(*quote.orders()[0].amount_out_net_gas(), BigUint::from(837u64));
+        assert_eq!(*quote.orders()[0].amount_out_net_gas(), BigUint::from(873u64));
```
louise-poole, line 667: How did changing the eth marker change this? 
EDIT - ah, did updating tycho-execution cause this? That would make sense.

louise-poole (approved):
Thanks! 🙏

fynd-core/src/worker_pool_router/mod.rs
tamaralipows: Yeaaah I think that's what caused this - it's totally unrelated to the marker

PR #206, opened 2026-05-22
test: add ConstantProductSim xy=k test primitive

louise-poole:
The existing MockProtocolSim uses a flat multiplier in its swap math. This is linear and represents a pool with no price impact, therefore splitting a trade never helps. This is not useful for testing split routes.

Therefore we add a ConstantProductSim to test_utils.rs. It implements ProtocolSim with xy=k price math using BigUint arithmetic. For simplicity, assume no fees.

### DIFF
--- a/fynd-core/src/algorithm/test_utils.rs
+++ b/fynd-core/src/algorithm/test_utils.rs
@@ -4,6 +4,8 @@ use std::collections::HashMap;
 
 use chrono::NaiveDateTime;
 use num_bigint::BigUint;
+use num_rational::BigRational;
+use num_traits::{ToPrimitive, Zero};
 use tycho_simulation::{
     tycho_core::{
         dto::ProtocolStateDelta,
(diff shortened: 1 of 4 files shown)


fynd-core/src/algorithm/test_utils.rs
```diff
@@ -249,6 +251,128 @@ impl ProtocolSim for MockProtocolSim {
...
+    /// Reserve of token 0 (lower-address token)
+    pub reserve_0: BigUint,
+    /// Reserve of token 1 (higher-address token)
+    pub reserve_1: BigUint,
+    /// Gas units per swap
+    pub gas: u64,
+}
+
+impl ConstantProductSim {
+    fn reserves_for<'a>(&'a self, token_a: &Token, token_b: &Token) -> (&'a BigUint, &'a BigUint) {
```
tamaralipows: Do we really need to specify lifetimes here? Doesn't rust handle this automatically?

tamaralipows (approved):
Thank you! Just a smol question

PR #207, opened 2026-05-22
feat(fynd-core): add build_with_pending for embedded pending-block simulation

kayibal:
## Why

`FyndBuilder` wraps `ProtocolStreamBuilder` internally but always called `build()`, which discards the `PendingBlockProcessor`. There was no way for embedding callers to obtain one — the interface to configure and receive a `PendingBlockProcessor` was entirely missing from Fynd's public API.

Without this, block builders and other embedders who need to simulate candidate transaction bundles against live pool state had to open a second Tycho WebSocket connection themselves, defeating the point of using Fynd as an embedded library.

## Changes

**Core feature**

- Upgrades tycho-simulation 0.258.0 → 0.300.5 (required for the `evm::pending` module). Updates `fynd-rpc/src/protocols.rs` for the new `ProtocolSystemsParams` API in tycho-client 0.300.5.
- Adds `FyndBuilder::build_with_pending() -> Result<(Solver, PendingBlockProcessor), SolverBuildError>`: async variant of `build()` that calls `ProtocolStreamBuilder::build_with_pending()` internally and returns the processor alongside the solver.
- Adds `FyndBuilder::with_pending_indexer(extractor, indexer)` so callers can attach a `Box<dyn TxDeltaIndexer>` before calling `build_with_pending()`, mirroring `ProtocolStreamBuilder::with_pending_indexer`.
- Re-exports `PendingBlockProcessor`, `PendingError`, `PendingUpdate`, and `TxDeltaIndexer` from the crate root with contextualizing doc comments.

**Fixup commits**

- Removes an RFQ guard that incorrectly rejected `build_with_pending` when RFQ protocols were configured. RFQ streams are now handled alongside the EVM pending stream, identical to `run()`.
- Changes the pending oneshot channel type to `Result<PendingBlockProcessor, String>` so feed setup failures (e.g. token loading errors) surface the root cause as `SolverBuildError::FeedSetup` instead of the opaque `PendingChannelClosed`.
- Fixes the `SolverBuildError` enum-level doc to mention `build_with_pending` as a source of `PendingChannelClosed`.
- Extracts the ~150 lines of shared setup code from `build()` and `build_with_pending()` into a private `assemble_components() -> BuiltComponents` helper so future changes to the builder only need to be made in one place.

## Usage

```rust
let (solver, pending) = FyndBuilder::new(chain, tycho_url, rpc_url, protocols, min_tvl)
    .tycho_api_key(key)
    .algorithm("most_liquid")
    .build_with_pending()
    .await?;

// Per block iteration:
let update = pending.generate_pending_update(txs, header, label)?;
solver.register_labeled_state(label, update);
let quote = solver.quote(request).await?;
```

### DIFF
--- a/fynd-core/src/feed/tycho_feed.rs
+++ b/fynd-core/src/feed/tycho_feed.rs
@@ -11,7 +11,7 @@ use tokio::{sync::broadcast, task::JoinHandle};
 use tokio_stream::StreamExt;
 use tracing::{debug, info, instrument, span, trace, Instrument, Level};
 use tycho_simulation::{
-    evm::stream::ProtocolStreamBuilder,
+    evm::{pending::PendingBlockProcessor, stream::ProtocolStreamBuilder},
     protocol::models::Update,
     rfq::stream::RFQStreamBuilder,
     tycho_client::feed::{component_tracker::ComponentFilter, SynchronizerState},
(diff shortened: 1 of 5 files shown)


fynd-core/src/feed/tycho_feed.rs
```diff
@@ -255,6 +255,196 @@ impl TychoFeed {
...
+        };
+
+        loop {
+            tokio::select! {
+                msg = protocol_stream.next() => {
+                    match msg {
+                        Some(msg) => {
+                            trace!("Received message from protocol stream: {:?}", msg);
+                            let msg = msg.map_err(|e| DataFeedError::StreamError(e.to_string()))?;
+                            self.refresh_gas_price().await?;
```
zizou0x: Would there be a way to differenciate between pending tx message and normal flow to avoid having to do things like this everytime?

zizou0x (approved):
Looks good, just one small comment

fynd-core/src/feed/tycho_feed.rs
kayibal: Louise already changed this to be only triggered by a timer, which is more correct since gas price doesn't only change on block updates.

PR #208, opened 2026-05-22
feat(encoding): allow client to patch calldata signature

tamaralipows:
The updated EIP-712 ClientFee type hash includes swap-specific fields (amountIn, tokenIn, tokenOut, minAmountOut, receiver, swaps) that are only known after the server encodes the transaction. This creates a chicken-and-egg problem: the signature covers the swaps, but the calldata includes the signature.

The encoder now returns the byte offset of the client fee signature within the calldata (via Transaction::signature_offset). Clients request a quote with a placeholder signature, sign using swaps_hash from the response, and patch the 65 bytes at the returned offset. Single request, no server-side state.

This signature is returned via `tycho-execution` so that any potential future changes in the router contract are responsibility of the `tycho-execution` crate and does not need to touch fynd.

Necessary change in tycho-execution: https://github.com/propeller-heads/tycho-indexer/pull/1046

### DIFF
--- a/clients/rust/src/types.rs
+++ b/clients/rust/src/types.rs
@@ -380,7 +380,8 @@ impl EncodingOptions {
 pub struct Transaction {
     to: Bytes,
     value: BigUint,
-    data: Vec<u8>,
+    pub(crate) data: Vec<u8>,
+    pub(crate) client_fee_signature_offset: Option<usize>,
 }
 
 impl Transaction {
@@ -390,7 +391,7 @@ impl Transaction {
     /// - `value`: native token value to send with the transaction.
     /// - `data`: ABI-encoded calldata.
     pub fn new(to: Bytes, value: BigUint, data: Vec<u8>) -> Self {
-        Self { to, value, data }
+        Self { to, value, data, client_fee_signature_offset: None }
     }
 
     /// Router contract address (20 raw bytes).
@@ -406,6 +407,11 @@ impl Transaction {
     /// ABI-encoded calldata.
     pub fn data(&self) -> &[u8] {
         &self.data
+    }
+
+    /// Byte offset of the client fee signature within `data`.
+    pub fn client_fee_signature_offset(&self) -> Option<usize> {
+        self.client_fee_signature_offset
     }
 }
 
@@ -758,8 +764,6 @@ pub struct FeeBreakdown {
     min_amount_received: BigUint,
     /// keccak256 of the ABI-encoded swap bytes. Use this for EIP-712 signing.
     swaps_hash: Option<[u8; 32]>,
-    /// Byte offset of the client fee signature within the TychoRouter calldata
-    signature_offset: Option<usize>,
 }
 
 impl FeeBreakdown {
@@ -769,16 +773,8 @@ impl FeeBreakdown {
         max_slippage: BigUint,
         min_amount_received: BigUint,
         swaps_hash: Option<[u8; 32]>,
-        signature_offset: Option<usize>,
     ) -> Self {
-        Self {
-            router_fee,
-            client_fee,
-            max_slippage,
-            min_amount_received,
-            swaps_hash,
-            signature_offset,
-        }
+        Self { router_fee, client_fee, max_slippage, min_amount_received, swaps_hash }
     }
 
     /// Router protocol fee (fee on output + router's share of client fee).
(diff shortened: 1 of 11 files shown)


fynd-core/src/encoding/encoder.rs
```diff
@@ -367,6 +386,30 @@ impl Encoder {
         call_data
     }
 
+    /// Finds the byte offset of `sig_bytes` within `calldata`.
+    ///
+    /// Returns `None` if the signature is not found or appears more than once
+    /// (ambiguous match).
+    fn find_signature_offset(calldata: &[u8], sig_bytes: &[u8]) -> Option<usize> {
```
tamaralipows: This is a bit hacky. I'm open to changing this to be more elegant. Wanted to make sure you were okay with option 3 before refining it

clients/rust/src/types.rs
```diff
@@ -767,8 +769,16 @@ impl FeeBreakdown {
...
     ) -> Self {
-        Self { router_fee, client_fee, max_slippage, min_amount_received, swaps_hash }
+        Self {
+            router_fee,
+            client_fee,
+            max_slippage,
+            min_amount_received,
+            swaps_hash,
+            signature_offset,
+        }
```
tamaralipows: I assumed this was okay to change because the new method is only `pub(crate)` but I'd like to revisit this on Monday too

fynd-core/src/encoding/encoder.rs
kayibal: Would be cleaner if it just came directly from tycho-execution if this is possible?

fynd-core/src/types/quote.rs
```diff
@@ -254,6 +254,10 @@ pub struct FeeBreakdown {
     /// Clients use this to compute the 10-field EIP-712 signing hash for the client fee.
     #[serde(skip)]
     swaps_hash: Option<[u8; 32]>,
+    /// Byte offset of the client fee signature within `Transaction.data`.
+    /// Clients use this to patch the 65-byte EIP-712 signature into the calldata.
+    #[serde(skip)]
+    signature_offset: Option<usize>,
```
kayibal: I like that this is part of the rpc response. I am not sure this is the correct place, though to be honest. It seems a bit out of place.

Possible other places:

- OrderQuote
- Transaction

fynd-core/src/encoding/encoder.rs
tamaralipows: Moved this here: https://github.com/propeller-heads/tycho-indexer/pull/1046

clients/rust/src/types.rs
```diff
@@ -927,6 +945,37 @@ impl Quote {
...
+    /// 2. Read [`FeeBreakdown::swaps_hash`] and [`FeeBreakdown::signature_offset`] from the
+    ///    response.
+    /// 3. Sign the 10-field EIP-712 hash using [`ClientFeeParams::eip712_signing_hash`].
+    /// 4. Call this method to patch the signature into the calldata.
+    /// 5. Execute the transaction.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the quote has no transaction, no fee breakdown, or no
+    /// `signature_offset`.
```
louise-poole: Should it panic though? Would prefer we return an error.

docs/guides/client-fees.md
```diff
@@ -118,23 +118,58 @@ const opts = withClientFee(encodingOptions(0.005), {...feeParams, signature});
 
 {% tab title="Rust" %}
 ```rust
-    // Build the fee params (without signature).
+    // Step 1: request a quote with unsigned client fee params.
```
louise-poole: 'with' feels a bit ambiguous... like is it referring to the quote or the request action? Will the quote be 'with' the unsigned fees no matter what the user gives, or is this an instruction on what to use to request the quote?
```suggestion
    // Step 1: request a quote using unsigned client fee params.
```

fynd-core/src/types/quote.rs
louise-poole: I mean this is only relevant if the user is applying fees so it makes sense if it is bundled with the fee info like this no? Plus this FeeBreakdown is only populated if the user is using our encoding/executor option which this signature offset is for so fits well.

If we were to move it I like OrderQuote more than Transaction. I don't feel too strongly about this though (moving it vs not moving).

louise-poole (commented):
Thanks for tackling this! Left some small comments, otherwise looks great!

clients/rust/src/types.rs
tamaralipows: Oh no missed this! Thank you!

louise-poole (approved):
LGTM!

fynd-core/src/types/quote.rs
tamaralipows: I chose yesterday to move this into Transaction after thinking more because:

It is not referring to any fee amounts or anything related to the order. It is very specifically talking about the calldata and how to index the calldata, which I feel is very specific to the Transaction object.

I'm open to moving this (again) but in the end I think a good argument can be made for putting it into each of those structs. I'm going to merge this for now because it feels quite urgent but feel free to keep talking about this on slack and I can open a PR to move this later!

dianacarvalho1 (commented):
uuff thank you for finding and fixing this @tamaralipows 🙏🏼

PR #209, opened 2026-05-22
feat(split-routing): add split-route primitive types

tamaralipows:
[Task](https://propeller-heads.atlassian.net/browse/ENG-5843)

### DIFF
--- a/fynd-core/src/algorithm/split_primitives.rs
+++ b/fynd-core/src/algorithm/split_primitives.rs
@@ -30,8 +30,7 @@ pub(crate) struct PathAllocation {
     pub(crate) amount_in: BigUint,
     pub(crate) amount_out: BigUint,
     /// Product of marginal prices along all hops at the time this allocation was
-    /// last simulated. Used to compute per-path price impact:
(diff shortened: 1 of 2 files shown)


fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -0,0 +1,142 @@
...
+/// currently allocated to it. All fractions across allocations sum to 1.0.
+pub(crate) struct PathAllocation {
+    pub(crate) hops: Vec<HopDescriptor>,
+    /// Fraction of total input on this path (0 < f <= 1).
+    pub(crate) flow_fraction: f64,
+    pub(crate) amount_in: BigUint,
+    pub(crate) amount_out: BigUint,
+    /// Product of marginal prices along all hops at the time this allocation was
+    /// last simulated. Used to compute per-path price impact:
+    /// `PI = 1 - amount_out / (amount_in * marginal_price_product)`.
```
louise-poole: Not sure the equation should be here? It should live where we actually apply the calculation no?

louise-poole (approved):
Looks good!

PR #210, opened 2026-05-26
feat(split-routing): add math utilities for split-route optimisation

tamaralipows:
Implements golden_section_search, split_amount, normalize_fractions, and fractions_to_amounts in split_primitives.rs.

Note: The task description mentions "Panics if the slice is empty or all zeros.", but I don't think this was on purpose - we wouldn't like our code to panic. normalize_fractions returns Result<(), SplitMathError> instead.

[Task](https://propeller-heads.atlassian.net/browse/ENG-5844)

### DIFF
--- a/fynd-core/src/algorithm/split_primitives.rs
+++ b/fynd-core/src/algorithm/split_primitives.rs
@@ -143,8 +143,8 @@ impl ProtocolSim for ZeroGasSim {
 
 /// Find the `x` in `[lo, hi]` that maximises `f(x)` using golden-section search.
 ///
-/// Assumes `f` is roughly unimodal. `max_evals` controls the number of function
-/// evaluations (higher = more precise but slower).
+/// Assumes `f` is roughly unimodal (has one maximum). `max_evals` controls the
+/// number of function evaluations (higher = more precise but slower).
 pub(crate) fn golden_section_search(
     f: impl Fn(f64) -> f64,
     mut lo: f64,
@@ -187,21 +187,24 @@ pub(crate) fn golden_section_search(
 ///
 /// Both values always sum exactly to `total` — no tokens lost to rounding.
 pub(crate) fn split_amount(total: &BigUint, fraction: f64) -> (BigUint, BigUint) {
+    let clamped = fraction.clamp(0.0, 1.0);
     // Scale fraction to fixed-point with 18 decimal digits of precision.
     let scale: u64 = 1_000_000_000_000_000_000;
-    let numerator = (fraction * scale as f64) as u64;
+    let numerator = (clamped * scale as f64) as u64;
     let part = (total * BigUint::from(numerator)) / BigUint::from(scale);
     let remainder = total - &part;
     (part, remainder)
 }
 
 /// Errors from split-routing math utilities.
-#[derive(Debug, Clone, thiserror::Error)]
+#[derive(Debug, Clone, PartialEq, thiserror::Error)]
 pub(crate) enum SplitMathError {
     #[error("fractions slice must not be empty")]
     EmptyFractions,
     #[error("all fractions are zero, cannot normalize")]
     AllZeroFractions,
+    #[error("fractions must not be negative")]
+    NegativeFraction,
 }
 
 /// Normalize a slice of fractions so they sum to 1.0.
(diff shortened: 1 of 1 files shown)


fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -140,3 +141,170 @@ impl ProtocolSim for ZeroGasSim {
             .unwrap_or(false)
     }
 }
+
+/// Find the `x` in `[lo, hi]` that maximises `f(x)` using golden-section search.
+///
+/// Assumes `f` is roughly unimodal. `max_evals` controls the number of function
```
louise-poole: Can we expand a little on what unimodal means so future readers have an easier time parsing the code. Means we have a single max point no?

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -140,3 +141,170 @@ impl ProtocolSim for ZeroGasSim {
...
+
+/// Convert fractions (summing to 1.0) into `BigUint` amounts summing exactly to `total`.
+///
+/// The last element absorbs any rounding remainder so the sum is exact.
+pub(crate) fn fractions_to_amounts(total: &BigUint, fractions: &[f64]) -> Vec<BigUint> {
+    let n = fractions.len();
+    let mut amounts = Vec::with_capacity(n);
+    let mut running_sum = BigUint::zero();
+
+    for &frac in &fractions[..n - 1] {
```
louise-poole, line 253: I *think* this will panic if fractions is empty. Can you double check?

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -140,3 +141,170 @@ impl ProtocolSim for ZeroGasSim {
...
+
+/// Split `total` into `(part, remainder)` where `part ≈ total * fraction`.
+///
+/// Both values always sum exactly to `total` — no tokens lost to rounding.
+pub(crate) fn split_amount(total: &BigUint, fraction: f64) -> (BigUint, BigUint) {
+    // Scale fraction to fixed-point with 18 decimal digits of precision.
+    let scale: u64 = 1_000_000_000_000_000_000;
+    let numerator = (fraction * scale as f64) as u64;
+    let part = (total * BigUint::from(numerator)) / BigUint::from(scale);
+    let remainder = total - &part;
```
louise-poole, line 196: I also think this might panic here if part > total (so fraction > 1).

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -140,3 +141,170 @@ impl ProtocolSim for ZeroGasSim {
...
+    AllZeroFractions,
+}
+
+/// Normalize a slice of fractions so they sum to 1.0.
+///
+/// # Errors
+///
+/// Returns [`SplitMathError::EmptyFractions`] if the slice is empty, or
+/// [`SplitMathError::AllZeroFractions`] if every element is zero.
+pub(crate) fn normalize_fractions(fractions: &mut [f64]) -> Result<(), SplitMathError> {
```
louise-poole, line 217: With the previous comments in mind, maybe we should also check to make sure negative fractions aren't used? Or fractions greater than 1.

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -140,3 +141,170 @@ impl ProtocolSim for ZeroGasSim {
...
+        assert!((sum - 1.0).abs() < f64::EPSILON);
+    }
+
+    #[rstest]
+    #[case::empty(&[], SplitMathError::EmptyFractions)]
+    #[case::all_zeros(&[0.0, 0.0, 0.0], SplitMathError::AllZeroFractions)]
+    fn test_normalize_fractions_invalid(#[case] input: &[f64], #[case] expected: SplitMathError) {
+        let mut fractions = input.to_vec();
+        let err = normalize_fractions(&mut fractions).unwrap_err();
+        assert_eq!(err.to_string(), expected.to_string());
```
louise-poole: Could we not compare the error type directly instead of via the to_string? Might need to derive `PartialEq` for the error types to do it, but I would expect we then can just have:

```suggestion
        assert_eq!(err, expected);
```

louise-poole (commented):
Looks good 👍  a couple concerns on possible error handling/panics, otherwise the logic looks good!

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -188,21 +188,24 @@ pub(crate) fn golden_section_search(
 ///
 /// Both values always sum exactly to `total` — no tokens lost to rounding.
 pub(crate) fn split_amount(total: &BigUint, fraction: f64) -> (BigUint, BigUint) {
+    let clamped = fraction.clamp(0.0, 1.0);
```
louise-poole, line 191: Can you add this to the docstring so it's not hidden/unexpected behaviour?

louise-poole (approved):
Nice! Thanks! 👍

PR #211, opened 2026-05-26
test: setup market unweighted test util

louise-poole:
Add a market setup test util that can be used for algorithms that do not need weighted edges.


tamaralipows (approved):
Thank you!

dianacarvalho1 (approved):
Thank you @louise-poole ! Just for my understanding: which kind of algorithm doesn't need a weighted graph? BF right? this change is just to clean things up a bit (unsure about the motivation that led to this)

louise-poole:
> which kind of algorithm doesn't need a weighted graph? BF right? this change is just to clean things up a bit (unsure about the motivation that led to this)

Yes, BF does not need a weighted graph and by extension neither will our split algorithm we implement.

PR #212, opened 2026-05-26
test(split-routing): add optimal_two_pool_output analytical bounds helper

louise-poole:
Analytically finds the optimal split between two Uniswap v2-like pools (ignoring fees for simplicity).

### DIFF
--- a/fynd-core/src/algorithm/split_test_harness.rs
+++ b/fynd-core/src/algorithm/split_test_harness.rs
@@ -1,10 +1,10 @@
 //! Test helpers for split-routing algorithm scenarios.
 
 /// Returns `(fraction_for_pool_1, total_output)` — the theoretically optimal output when
-/// splitting `trade_amount` between two constant-product pools.
+/// splitting `trade_amount` between two constant-product pools with no fees.
 ///
-/// Negative allocations are clamped to `0` — a clamped value means the full trade routes through
-/// the other pool.
+/// Finds the split where both pools offer the same marginal rate on the last unit traded.
+/// Negative allocations are clamped to `0`.
 #[allow(dead_code)]
 pub(crate) fn optimal_two_pool_output(
     reserve_in_1: f64,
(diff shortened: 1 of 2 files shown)


fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -0,0 +1,74 @@
...
+    trade_amount: f64,
+) -> (f64, f64) {
+    // The optimal split equates marginal prices across pools. Let `d = √(k1/k2)` where
+    // `k_i = reserve_in_i · reserve_out_i`. Solving the marginal-price-equality condition with
+    // `a_1 + a_2 = trade_amount` gives:
+    //
+    // ```text
+    // d   = sqrt((reserve_in_1 * reserve_out_1) / (reserve_in_2 * reserve_out_2))
+    // a_2 = (trade_amount + reserve_in_1 - d * reserve_in_2) / (d + 1)
+    // a_1 = trade_amount - a_2
```
tamaralipows: I don't think this derivation is necessary here - it's almost identical to what's in the code?

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -0,0 +1,74 @@
...
+    fn test_optimal_two_pool_output_asymmetric() {
+        // Pool 1: reserve_in=100, reserve_out=400  →  k1=40_000, K1=200
+        // Pool 2: reserve_in=100, reserve_out=100  →  k2=10_000, K2=100
+        // d = sqrt(40_000/10_000) = 2
+        // a_2 = (400 + 100 - 2·100) / (2+1) = 300/3 = 100
+        // a_1 = 300  →  fraction = 300/400 = 0.75
+        //
+        // Verify optimality: marginal prices after split must be equal.
+        // Pool 1: k1 / (100+300)² = 40_000/160_000 = 0.25
+        // Pool 2: k2 / (100+100)² = 10_000/40_000  = 0.25  ✓
```
tamaralipows: We are not actually adding this assertion - should we add it?

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -0,0 +1,74 @@
...
+        let (fraction, total_out) = optimal_two_pool_output(100.0, 400.0, 100.0, 100.0, 400.0);
+        assert!(
+            (fraction - 0.75).abs() < 1e-9,
+            "asymmetric pools: expected fraction 0.75, got {fraction}"
+        );
+        // out_1 = 300·400/(100+300) = 300, out_2 = 100·100/(100+100) = 50
```
tamaralipows: Hard to tell right away what these numbers represent, maybe one more line above that shows the variable names?

tamaralipows (commented):
Thanks! Logic looks sound - just minor things

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -0,0 +1,74 @@
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_optimal_two_pool_output_symmetric() {
+        // Identical pools → d=1 → 50/50 split is always optimal.
+        let (fraction, _) =
+            optimal_two_pool_output(10_000.0, 10_000.0, 10_000.0, 10_000.0, 1_000.0);
+        assert!(
+            (fraction - 0.5).abs() < 1e-9,
```
dianacarvalho1: maybe we can add a little function that does the equality (just for readability). something like:
```
fn f64_equals(x: f64, other: f64) -> bool {
   (x - other).abs() <1e-9
}
```

dianacarvalho1 (approved):
Thank you @louise-poole ! Tiny tiny comment only, otherwise it looks good to me! 🙏🏼

PR #213, opened 2026-05-26
feat(split-routing): add simulation utils for split-route optimisation

tamaralipows:
Implement hop-by-hop path simulation, marginal price computation, gas-deduplicated objective function, and degraded market builder for the upcoming split-routing algorithm (ENG-5845).

### DIFF
--- a/fynd-core/src/algorithm/split_primitives.rs
+++ b/fynd-core/src/algorithm/split_primitives.rs
@@ -12,7 +12,7 @@ use tycho_simulation::tycho_common::{
     Bytes,
 };
 
-use crate::{algorithm::AlgorithmError, feed::market_data::MarketState, types::ComponentId};
+use crate::{algorithm::AlgorithmError, feed::market_data::MarketDataView, types::ComponentId};
 
 pub(crate) struct HopDescriptor {
     pub(crate) component_id: ComponentId,
@@ -43,7 +43,7 @@ pub(crate) struct SimResult {
     pub(crate) marginal_price_product: f64,
 }
 
-/// Pool state overrides for reused pools in subsequent simulation/route searches.
+/// Pool state overrides for passing degraded states to `find_single_route`.
 #[derive(Default)]
 pub(crate) struct MarketOverrides(HashMap<ComponentId, Box<dyn ProtocolSim>>);
 
@@ -71,6 +71,12 @@ impl MarketOverrides {
     pub(crate) fn get(&self, id: &ComponentId) -> Option<&dyn ProtocolSim> {
         self.0.get(id).map(|b| b.as_ref())
     }
+}
+
+/// Wraps a sim so that `get_amount_out().gas` is always zero while amounts
+/// remain correct. Use for pools whose gas is already accounted for elsewhere.
+pub(crate) fn wrap_zero_gas(sim: Box<dyn ProtocolSim>) -> Box<dyn ProtocolSim> {
+    Box::new(ZeroGasSim(sim))
 }
 
 /// Wrapper that delegates all [`ProtocolSim`] calls unchanged except
@@ -265,14 +271,12 @@ pub(crate) fn fractions_to_amounts(
 /// near-zero input.
 pub(crate) fn compute_marginal_price_product(
     hops: &[HopDescriptor],
-    market: &MarketState,
-    overrides: &MarketOverrides,
+    market: &MarketDataView,
 ) -> Result<f64, AlgorithmError> {
     let mut product = 1.0;
     for hop in hops {
-        let sim = overrides
-            .get(&hop.component_id)
-            .or_else(|| market.get_simulation_state(&hop.component_id))
+        let sim = market
+            .get_simulation_state(&hop.component_id)
             .ok_or_else(|| AlgorithmError::DataNotFound {
                 kind: "simulation state",
                 id: Some(hop.component_id.clone()),
@@ -291,22 +295,19 @@ pub(crate) fn compute_marginal_price_product(
 /// Simulates a path hop-by-hop, threading output of each hop as input to the
 /// next.
 ///
-/// Checks `overrides` before falling back to the live market state for each
-/// hop. Returns the final output amount, raw gas sum, and marginal price
-/// product.
+/// Uses the overlay-aware `MarketDataView` for each hop's simulation state.
+/// Returns the final output amount, raw gas sum, and marginal price product.
 pub(crate) fn simulate_path(
     hops: &[HopDescriptor],
     amount_in: &BigUint,
-    market: &MarketState,
-    overrides: &MarketOverrides,
+    market: &MarketDataView,
 ) -> Result<SimResult, AlgorithmError> {
     let mut current_amount = amount_in.clone();
     let mut total_gas: u64 = 0;
 
     for hop in hops {
-        let sim = overrides
-            .get(&hop.component_id)
-            .or_else(|| market.get_simulation_state(&hop.component_id))
+        let sim = market
+            .get_simulation_state(&hop.component_id)
             .ok_or_else(|| AlgorithmError::DataNotFound {
                 kind: "simulation state",
                 id: Some(hop.component_id.clone()),
@@ -324,7 +325,7 @@ pub(crate) fn simulate_path(
         current_amount = result.amount;
     }
 
-    let marginal_price_product = compute_marginal_price_product(hops, market, overrides)?;
+    let marginal_price_product = compute_marginal_price_product(hops, market)?;
 
     Ok(SimResult { amount_out: current_amount, gas: total_gas, marginal_price_product })
 }
@@ -335,8 +336,7 @@ pub(crate) fn evaluate_total_output(
     paths: &[&[HopDescriptor]],
     fractions: &[f64],
     total_amount: &BigUint,
-    market: &MarketState,
-    overrides: &MarketOverrides,
+    market: &MarketDataView,
 ) -> Result<(BigUint, u64), AlgorithmError> {
     let amounts = fractions_to_amounts(total_amount, fractions)
         .map_err(|e| AlgorithmError::Other(e.to_string()))?;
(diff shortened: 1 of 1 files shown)


fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -261,6 +261,167 @@ pub(crate) fn fractions_to_amounts(
     Ok(amounts)
 }
 
+/// Product of spot prices along a path — approximates the exchange rate at
+/// near-zero input.
+pub(crate) fn compute_marginal_price_product(
```
tamaralipows, line 272: The task says

> Note: takes the live market, not overrides. Spot prices represent the zero-flow exchange rate, which doesn't change with pool state — only actual swap simulation needs overrides.

I respected that in this PR - but I don't understand it. Why wouldn't we care about the overrides here - wouldn't that potentially impact the marginal price product?

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -261,6 +261,167 @@ pub(crate) fn fractions_to_amounts(
...
+    Ok(SimResult { amount_out: current_amount, gas: total_gas, marginal_price_product })
+}
+
+/// Objective function for [`golden_section_search`].
+///
+/// Simulates all paths at their current fractions and returns
+/// `(total_amount_out, total_gas)`. `paths[i]` corresponds to `fractions[i]`.
+///
+/// Gas is deduplicated: pools shared across multiple paths are counted once.
+pub(crate) fn evaluate_total_output(
```
tamaralipows, line 340: The task said:

> Gas deduplication: pools shared across multiple paths are counted once. The function collects all unique component_ids across all hops/paths and sums each pool's gas cost exactly once.

Why would we do this? Two hops on the same pool (imagine a three-token pool) should still could as two hops and count gas twice - so I don't think this is proper (maybe for grouped swaps? But we take care of that during encoding). I removed this condition - feel free to explain a reason for keeping it.

dianacarvalho1: I think this could be true if it was two exactly the same hops in different paths? like that we trade USDC -> WETH on the same pool twice in different paths and when executing we could collapse this into one actual trade? This seems very complex though and not sure if included in the rest of the plan. Other than this I don't see a reason to do this either 😕

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -261,6 +261,167 @@ pub(crate) fn fractions_to_amounts(
...
+/// For example, if the current solution splits 1000 USDC→ETH across:
+///   - Path 1: USDC→WETH via Uniswap (600 USDC)
+///   - Path 2: USDC→WBTC→WETH via Curve+Balancer (400 USDC)
+///
+/// this function simulates both swaps and returns overrides where Uniswap,
+/// Curve, and Balancer all reflect their post-swap reserves.
+///
+/// Paths are processed in order so shared pools accumulate the effects of
+/// all prior paths.
+pub(crate) fn build_degraded_market(
```
dianacarvalho1: not a fan of this naming.. maybe `build_post_swap_market` would be more obvious? degraded makes it sound like something bad

dianacarvalho1 (commented):
Thank you @tamaralipows ! 🙏🏼 only have tiny comments. One general though: why is this logicu just solo util functions and not methods of some struct? Afaiu there isn't such a struct that would make sense to have them there right now, but it would be something like a Path? I'm missing some overall context in these whole tasks, so don't take my comments so seriously (I'm just wondering)

fynd-core/src/algorithm/split_primitives.rs
louise-poole: Good question - I am inclined to agree with you. Maybe @kayibal can weigh in here on what the plan intended? If we have the overrides available, what's the harm in using them?

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -261,6 +261,167 @@ pub(crate) fn fractions_to_amounts(
...
+/// next.
+///
+/// Checks `overrides` before falling back to the live market state for each
+/// hop. Returns the final output amount, raw gas sum, and marginal price
+/// product.
+pub(crate) fn simulate_path(
+    hops: &[HopDescriptor],
+    amount_in: &BigUint,
+    market: &MarketState,
+    overrides: &MarketOverrides,
```
louise-poole, line 307: Another thought here - Alan recently merged MarketDataView as an overlay aware market state... Maybe we should update our plan to use that instead of duplicating a lot of this market + overrides logic ourselves?

fynd-core/src/algorithm/split_primitives.rs
louise-poole: I agree with removing it. We don't plan to make the algorithm aware of grouping logic/protocols so should skip this entirely. Having this also assumes people using our router/execution...

fynd-core/src/algorithm/split_primitives.rs
louise-poole: Yes! Agreed!!! Though this is technically not building the full market and is building only the overrides. Maybe the name should reflect that rather? `build_market_overrides`?

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -261,6 +261,167 @@ pub(crate) fn fractions_to_amounts(
...
+            current_amount = result.amount;
+            states.insert(hop.component_id.clone(), result.new_state);
+        }
+    }
+
+    let mut overrides = MarketOverrides::empty();
+    for (id, sim) in states {
+        overrides = overrides.with_override(id, sim);
+    }
+    overrides
```
louise-poole: Probably would be nice to have a batch version of `with_override` on `MarketOverrides`? Or a constructor that takes in a batch of overrides immediately?

fynd-core/src/algorithm/split_primitives.rs
kayibal: Note that in the context of the current swap, we are indeed building a market that has degraded. I find this naming better.

fynd-core/src/algorithm/split_primitives.rs
kayibal: The gas needs to be correctly deduplicated (this is important!).

https://propeller-heads.atlassian.net/wiki/spaces/DEFI/pages/3592749057/Path-Level+Frank-Wolfe+Split+Routing+Algorithm#Handling-Shared-Pools

Please note: that this has nothing to do with grouping protocols but instead is required since we use a simplification of the problem, where basically during search we only define splits at the initial token. I reocmmend reading the whole proposal which gives a high level overview of the algorithm.

@dianacarvalho1 got it right! Note this is medium complexity but avoids complexity of creating multiple simplex vectors in the solution space, instead you just need to create one. Motivation see [here](https://propeller-heads.atlassian.net/wiki/spaces/DEFI/pages/3592749057/Path-Level+Frank-Wolfe+Split+Routing+Algorithm#Flow-Representation)

fynd-core/src/algorithm/split_primitives.rs
kayibal: I am not sure it works 100% for this use case, but if it does, that would be great

fynd-core/src/algorithm/split_primitives.rs
kayibal: I agree, we need them available here.

fynd-core/src/algorithm/split_primitives.rs
louise-poole: Ohhhh - I think I see. To clarify quick, we're talking about pre-split point pools no? P1 we swap tokens A -> B, then P2 takes 30% of B and swaps to C and P3 takes 70% of B and swaps to C. These are shown as 2 split paths in our current design: A -> (P1) -> B -> (P2) -> C and  A -> (P1) -> B -> (P3) -> C, but P1 was before we logically 'split off' to different pools, so was done in one hop and should not be counted twice even though it appears in both paths.

fynd-core/src/algorithm/split_primitives.rs
louise-poole: Either way, it is building only the overrides for the pools touched and not a full market... I find saying 'build market' misleading when we do not return the market state, but only some overrides.

tamaralipows: `build_post_swap_overrides`?

fynd-core/src/algorithm/split_primitives.rs
tamaralipows: In commit [daf5642](https://github.com/propeller-heads/fynd/pull/213/commits/daf56421d1f25f099cafcaa66fb32ca6e632512c) I've changed it to make the methods take `MarketDataView` instead. I actually like this. It makes the tests a bit more complex but I like that it deduplicated this logic in all those functions.

One thing I'm unsure about that would be clearer later, is that I kept returning the `MarketOverride` struct in `build_degraded_market` (instead of updating the `MarketDataView` and returning it) since it didn't seem to make sense to add an overlay there just to read it and pass it to `find_single_route` later. I will find out in that task if that's really the best way to do it, but please feel free to correct me if you see an immediate problem with this.

fynd-core/src/algorithm/split_primitives.rs
tamaralipows: I think this might be irrelevant after [daf5642](https://github.com/propeller-heads/fynd/pull/213/changes/daf56421d1f25f099cafcaa66fb32ca6e632512c) since we are adding to overrides as we are looping? Pls double check

tamaralipows:
@dianacarvalho1 

> Afaiu there isn't such a struct that would make sense to have them there right now, but it would be something like a Path?

Exactly this. I was envisioning perhaps adding this in later tasks but I didn't want to speculate too much especially while still getting comfortable with this code and not having touched any higher level tasks yet.

fynd-core/src/algorithm/split_primitives.rs
kayibal: Yeah it looks actually really nice but I am afraid it may not work quite as we want it to work... So MarketDataView can change anytime. It contains locks internally. It is mean to be shared between worker pools / threads. Algorithms usually work with &MarketState which can be extracted from MarketDataView using extract_subset and should be seen as a snapshot of MarketDataView at a specific point in time. 

After thinking about it a bit more, I think we shouldn't use it here. Alone the fact that it contains locks we don't need disqualifies it imo.

Basically the mental model is:

- SolveRequest comes in
- Use extract_subset to get a snapshot on the market data and hold any locks in MarketDataView only very briefly
- Use &MarketState freely without locks etc. to work on that data
- &MarketState can not be mutated while I hold it

louise-poole (approved):
Thanks! Looks good 👌

PR #219, opened 2026-05-28
test: implement TestScenario harness with named split_scenarios

louise-poole:
Created a TestScenario harness with 5 named test scenarios for split routing.

### DIFF
--- a/fynd-core/src/algorithm/split_test_harness.rs
+++ b/fynd-core/src/algorithm/split_test_harness.rs
@@ -1,7 +1,7 @@
 //! Test helpers for split-routing algorithm split_scenarios.
 
-use tycho_simulation::tycho_core::{models::token::Token, simulation::protocol_sim::ProtocolSim};
 use num_bigint::{BigInt, BigUint};
+use tycho_simulation::tycho_core::{models::token::Token, simulation::protocol_sim::ProtocolSim};
 
 use crate::{
     algorithm::test_utils::setup_market_unweighted, feed::market_data::MarketData,
@@ -45,18 +45,9 @@ pub(crate) struct ScenarioPool {
 
 /// A self-contained algorithm test case with pre-computed bounds.
 ///
-/// ## Bounds
-///
-/// Both bounds are net values (gross - gas cost). Gas cost uses the test market's fixed
-/// assumptions:
-/// - gas price: 100 wei/gas (set in `setup_market_unweighted`)
-/// - token price: 1 output-token = 1 ETH, so gas cost in token units = gas_units × 100
-///
-/// `lower_bound` is the BF single-route net output — the minimum the algorithm must produce.
-/// `analytical_optimum` is the best net output for the simplified model (quality target, not a
-/// hard ceiling). Equals `lower_bound` when splitting is not beneficial.
-///
-/// Both are hardcoded from the scenario's fixed reserves and stay stable as regression targets.
+/// Both bounds are net output amounts (gross minus gas cost), hardcoded from the scenario's fixed
+/// reserves. Gas cost uses the test market's fixed assumptions: 100 wei/gas, 1 output-token = 1
+/// ETH.
 #[allow(dead_code)]
 pub(crate) struct TestScenario {
     pub name: &'static str,
@@ -65,10 +56,10 @@ pub(crate) struct TestScenario {
     pub token_in: Token,
     pub token_out: Token,
     pub trade_amount: BigUint,
-    /// BF single-route net output. Algorithm must produce `net_output >= lower_bound`.
+    /// Floor: the algorithm must produce at least this much net output.
     pub lower_bound: BigInt,
-    /// Best net output for the simplified model. Equals `lower_bound` when splitting is not
-    /// beneficial.
+    /// Target: the best net output achievable under the scenario's simplified pool model. A
+    /// quality ceiling to measure against, not a hard constraint.
     pub analytical_optimum: BigInt,
 }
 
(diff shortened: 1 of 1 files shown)


fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -25,6 +33,326 @@ pub(crate) fn optimal_two_pool_output(
...
+            token_out: token_b,
+            trade_amount: BigUint::from(200_000u64) * &one_eth,
+            lower_bound: BigUint::from(166_666_666_666_666_666_666_666u128),
+            analytical_optimum: BigUint::from(176_470_588_235_294_097_103_232u128),
+        }
+    }
+
+    /// S3: tiny trade with high per-pool gas; gas overhead outweighs the split benefit.
+    ///
+    /// `analytical_optimum`: equals `lower_bound`. Only two pools exist so no diamond is possible,
```
tamaralipows: It's not clear to me how the diamond scenario is relevant to this case? How would it help?

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -25,6 +33,326 @@ pub(crate) fn optimal_two_pool_output(
...
+    /// S5: A→B (one pool) → C (two parallel pools); bottleneck is at the B→C hop.
+    ///
+    /// PathFrankWolfe discovers two complete paths [P_AB, P_BC1] and [P_AB, P_BC2]. Because both
+    /// share P_AB, `build_split_route` emits one combined A→B swap followed by split B→C swaps,
+    /// with P_AB's gas counted once.
+    ///
+    /// `lower_bound`: best single 2-hop route A→B→C through the larger B→C pool.
+    /// `analytical_optimum`: only one A→B pool (P_AB) exists so the B amount is fixed;
+    /// `optimal_two_pool_output` gives the exact optimum for splitting that B across the two B→C
+    /// pools. No diamond is possible.
```
tamaralipows: Also not sure why the diamond is relevant here. Should we add a test case for the diamond since it is mentioned so much? haha

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -25,6 +33,326 @@ pub(crate) fn optimal_two_pool_output(
...
+    pub(crate) fn multi_hop_bottleneck() -> TestScenario {
```
tamaralipows, line 290: What about the following scenario (excuse the formatting):
```
       / ─ USV2 ─ \         / ─ USV2 ─ DAI
WETH ─             ─ USDC─
       \ ─ USV3 ─ /         \ ─ USV3 ─ DAI
```
The way our router treats this is that it pools all the USDC after the first split, then splits from the total USDC into the second split. It would be nice to have this guardrail that Fynd also treats that properly?

tamaralipows (commented):
Thank you! Just a few questions!

fynd-core/src/algorithm/split_test_harness.rs
louise-poole: Urgh - AI slop that fell through. Soz, it was obsessed with the diamond pattern being a possible way of beating the analytical optimum and it wanted to emphasise all cases it didn't apply... I thought I had manually removed it all 👀

fynd-core/src/algorithm/split_test_harness.rs
louise-poole: Urgh - so mad it was added back and I missed it 😭 It's not relevant at all

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -25,6 +33,326 @@ pub(crate) fn optimal_two_pool_output(
...
+    pub token_1: Token,
+    pub token_2: Token,
+    pub sim: Box<dyn ProtocolSim>,
+}
+
+/// A self-contained algorithm test case with pre-computed bounds.
+///
+/// `lower_bound` is the BF single-route output — the algorithm under test must produce ≥ this.
+/// `analytical_optimum` is the best output derivable from the simplified model used in the
+/// scenario (see each constructor for how it is computed). It is a quality target, not a hard
```
dianacarvalho1: it bothers me a bit that these variables just describe something about the attribute not what the attribute actually is supposed to represent 😅 like that these are both amount outs right? one is the lower bound of amount out from BF and the other is the analytical optimum

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -62,4 +390,14 @@ mod tests {
             "marginal prices should equalise at the optimum: {marginal_1} vs {marginal_2}"
         );
     }
+
+    #[test]
+    fn test_scenario_market_builds_without_panic() {
```
dianacarvalho1, line 471: amazing test 😂

fynd-core/src/algorithm/split_test_harness.rs
dianacarvalho1: I agree that this is a relevant scenario @tamaralipows ! But afaiu this isn't actually tested here no? 🤔 like that the same pools are merged nicely, it would only test that the amount out is nice

dianacarvalho1 (approved):
Thank you @louise-poole ! Only have a tiny comment

fynd-core/src/algorithm/split_test_harness.rs
louise-poole: I know 😂 but it was described in the task and helped me avoid marking things as dead code so I kept it.

fynd-core/src/algorithm/split_test_harness.rs
louise-poole: I've added gas considerations to the TestScenarios so correctly merging pools would have a impact on the amount out. 
I'll make a new double-split scenario for the case Tamara gave above.

PR #220, opened 2026-05-28
feat(split routing): Implement build_split_route 

tamaralipows:
[Task](https://propeller-heads.atlassian.net/browse/ENG-5846)

Assembles a Route from split-route path allocations, deduplicating shared pool hops into a single combined swap. Emits swaps in BFS topological order (from the first token) with the tycho-execution remainder convention (last swap per split point gets split=0.0).

### DIFF
--- a/fynd-core/src/algorithm/split_primitives.rs
+++ b/fynd-core/src/algorithm/split_primitives.rs
@@ -22,6 +22,22 @@ pub(crate) struct HopDescriptor {
     pub(crate) component_id: ComponentId,
     pub(crate) token_in: Token,
     pub(crate) token_out: Token,
+    /// Per-hop output amount, populated by the solving algorithm.
+    pub(crate) amount_out: BigUint,
+    /// Per-hop gas estimate, populated by the solving algorithm.
+    pub(crate) gas: BigUint,
+}
+
+impl HopDescriptor {
+    pub(crate) fn new(component_id: ComponentId, token_in: Token, token_out: Token) -> Self {
+        Self { component_id, token_in, token_out, amount_out: BigUint::ZERO, gas: BigUint::ZERO }
+    }
+
+    pub(crate) fn with_amounts(mut self, amount_out: BigUint, gas: BigUint) -> Self {
+        self.amount_out = amount_out;
+        self.gas = gas;
+        self
+    }
 }
 
 /// A fully-simulated path allocation.
@@ -444,6 +460,8 @@ struct SplitSwap {
     hop: HopDescriptor,
     split: f64,
     amount_in: BigUint,
+    amount_out: BigUint,
+    gas: BigUint,
 }
 
 /// Merge shared hops across paths, summing their flow fractions, and return
@@ -461,15 +479,20 @@ fn merge_shared_hops(paths: &[PathAllocation]) -> HashMap<Bytes, Vec<SplitSwap>>
                 hop.token_out.address.clone(),
             );
             hops.entry(key)
-                .and_modify(|h| h.split += path.flow_fraction)
+                .and_modify(|h| {
+                    h.split += path.flow_fraction;
+                    h.amount_out += &hop.amount_out;
+                })
                 .or_insert(SplitSwap {
-                    hop: HopDescriptor {
-                        component_id: hop.component_id.clone(),
-                        token_in: hop.token_in.clone(),
-                        token_out: hop.token_out.clone(),
-                    },
+                    hop: HopDescriptor::new(
+                        hop.component_id.clone(),
+                        hop.token_in.clone(),
+                        hop.token_out.clone(),
+                    ),
                     split: path.flow_fraction,
                     amount_in: BigUint::ZERO,
+                    amount_out: hop.amount_out.clone(),
+                    gas: hop.gas.clone(),
                 });
         }
     }
@@ -531,8 +554,8 @@ pub(crate) fn build_split_route(
 ) -> Result<Route, AlgorithmError> {
     let mut hops_by_token = merge_shared_hops(paths);
 
-    let mut queue = VecDeque::new();
-    queue.push_back(order.token_in().clone());
+    let mut pending_tokens = VecDeque::new();
+    pending_tokens.push_back(order.token_in().clone());
 
     let mut available: HashMap<Bytes, BigUint> = HashMap::new();
     available.insert(order.token_in().clone(), order.amount().clone());
@@ -540,11 +563,10 @@ pub(crate) fn build_split_route(
     let mut swaps = Vec::new();
     let mut route_tokens: HashMap<Bytes, Token> = HashMap::new();
     let mut visited: HashSet<Bytes> = HashSet::new();
-    let mut state_overrides = MarketOverrides::empty();
 
-    // BFS from the source token: at each token, simulate outgoing hops,
-    // build Swap objects, and track post-swap pool states for downstream hops.
-    while let Some(token_addr) = queue.pop_front() {
+    // BFS from the source token: at each token, assemble Swap objects from
+    // pre-computed hop amounts (populated by the solving algorithm).
+    while let Some(token_addr) = pending_tokens.pop_front() {
         // Converging paths (e.g. B→D and C→D) add D to the queue twice — skip duplicates.
         if !visited.insert(token_addr.clone()) {
             continue;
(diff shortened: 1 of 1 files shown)


fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -12,12 +12,32 @@ use tycho_simulation::tycho_common::{
...
+impl HopDescriptor {
+    pub(crate) fn new(component_id: ComponentId, token_in: Token, token_out: Token) -> Self {
+        Self { component_id, token_in, token_out, amount_out: BigUint::ZERO, gas: BigUint::ZERO }
+    }
+
+    pub(crate) fn with_amounts(mut self, amount_out: BigUint, gas: BigUint) -> Self {
+        self.amount_out = amount_out;
+        self.gas = gas;
+        self
+    }
```
tamaralipows, line 40: I've added this here to avoid re-simulating inside the `build_split_route` method, since we need the amount out and gas in there. Simulating twice would be wasteful since we already have the correct values

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -915,4 +1051,369 @@ mod tests {
             .amount;
         assert_eq!(degraded_out, BigUint::from(163u64));
     }
+
+    // ==================== merge / allocate Tests ====================
```
tamaralipows, line 1070: These tests weren't mentioned in the task description, they're supplementary since I decided to use helpers

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -12,12 +12,32 @@ use tycho_simulation::tycho_common::{
...
     pub(crate) token_out: Token,
+    /// Per-hop output amount, populated by the solving algorithm.
+    pub(crate) amount_out: BigUint,
+    /// Per-hop gas estimate, populated by the solving algorithm.
+    pub(crate) gas: BigUint,
+}
+
+impl HopDescriptor {
+    pub(crate) fn new(component_id: ComponentId, token_in: Token, token_out: Token) -> Self {
+        Self { component_id, token_in, token_out, amount_out: BigUint::ZERO, gas: BigUint::ZERO }
```
dianacarvalho1: so ZERO means that it isn't set? can we use Option instead to make it more obvious please? (I'm sensing that @louise-poole will ask for this too 👀)

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -436,12 +456,188 @@ pub(crate) fn build_post_swap_overrides(
...
+                    h.amount_out += &hop.amount_out;
+                })
+                .or_insert(SplitSwap {
+                    hop: HopDescriptor::new(
+                        hop.component_id.clone(),
+                        hop.token_in.clone(),
+                        hop.token_out.clone(),
+                    ),
+                    split: path.flow_fraction,
+                    amount_in: BigUint::ZERO,
```
dianacarvalho1, line 512: why is amount in zero? 🤔

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -436,12 +456,188 @@ pub(crate) fn build_post_swap_overrides(
...
+    let mut visited: HashSet<Bytes> = HashSet::new();
+
+    // BFS from the source token: at each token, assemble Swap objects from
+    // pre-computed hop amounts (populated by the solving algorithm).
+    while let Some(token_addr) = pending_tokens.pop_front() {
+        // Converging paths (e.g. B→D and C→D) add D to the queue twice — skip duplicates.
+        if !visited.insert(token_addr.clone()) {
+            continue;
+        }
+        // Terminal tokens (e.g. the final output) have no outgoing swaps.
```
dianacarvalho1, line 593: Terminal tokens 🪦

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -915,4 +1047,371 @@ mod tests {
             .amount;
         assert_eq!(degraded_out, BigUint::from(163u64));
     }
+
+    // ==================== merge / allocate Tests ====================
+
+    #[test]
+    fn test_merge_shared_hops_combines_fractions() {
+        // Two paths share the first hop A→B via P1; second hops diverge.
```
dianacarvalho1, line 1074: can we have a little diagram? 🥺

fynd-core/src/algorithm/split_primitives.rs
dianacarvalho1: ah because it will be filled only in `apply_remainder_convention` right?

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -436,12 +456,188 @@ pub(crate) fn build_post_swap_overrides(
...
+                .unwrap_or(std::cmp::Ordering::Equal)
+        });
+    }
+    hops_by_token
+}
+
+/// Assign each hop in a group its input amount and final split value using the
+/// tycho-execution remainder convention: last hop (smallest fraction) gets
+/// `split = 0.0`.
+fn apply_remainder_convention(
```
dianacarvalho1: this function actually does more than what the name describes:
1. apply the remainder convention
2. normalises the amounts
3. sets the amount ins

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -915,4 +1047,371 @@ mod tests {
...
+        }
+    }
+
+    #[test]
+    fn test_build_split_route_shared_first_pool() {
+        // Two paths sharing pool P1 at A→B, diverging at B→C (P2 vs P3).
+        //
+        //                  P2 (price=3) --> C
+        //                 /
+        //  A -- P1 (2) --+
```
dianacarvalho1:
```suggestion
        //  A -- P1 (2) --B
```

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -915,4 +1047,371 @@ mod tests {
...
+            *ab_swap.amount_in(),
+            BigUint::from(1000u64),
+            "A→B swap amount_in should equal sum of both paths"
+        );
+        assert_eq!(
+            *ab_swap.amount_out(),
+            BigUint::from(2000u64),
+            "A→B amount_out should be sum of per-path outputs (1400+600)"
+        );
+        assert_eq!(*ab_swap.split(), 0.0, "only swap at A→B should be remainder");
```
dianacarvalho1: > "only swap at A→B should be remainder"

sounds a bit strange, maybe "A→B is the sole swap in its group, so it gets the remainder convention (split = 0.0)"

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -915,4 +1047,371 @@ mod tests {
...
+    #[test]
+    fn test_build_split_route_source_level_split_different_intermediates() {
+        // Paths A→B→Z and A→C→Z: source-level split with different
+        // intermediate tokens.
+        //
+        //       pool_ab --> B -- pool_bz --> Z
+        //      /
+        //  A --
+        //      \
+        //       pool_ac --> C -- pool_cz --> Z
```
dianacarvalho1:
```suggestion
        //       pool_ab --> B -- pool_bz 
        //      /                                         \
        //  A --                                           Z
        //      \                                         /
        //       pool_ac --> C -- pool_cz
```

dianacarvalho1 (commented):
Thank you @tamaralipows 🙏🏼 only have smol comments!

fynd-core/src/algorithm/split_primitives.rs
tamaralipows: Damnit i was literally gonna do that and then forgot 😭 you're right lol thank you

fynd-core/src/algorithm/split_primitives.rs
louise-poole: I don't see where this is used tbh? Other than in tests... but I suspect they're just ignored by the code. `build_split_route` looks like it uses `split_swap.amount_out` etc.

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -436,12 +456,188 @@ pub(crate) fn build_post_swap_overrides(
...
+            let key: HopKey = (
+                hop.component_id.clone(),
+                hop.token_in.address.clone(),
+                hop.token_out.address.clone(),
+            );
+            hops.entry(key)
+                .and_modify(|h| {
+                    h.split += path.flow_fraction;
+                    h.amount_out += &hop.amount_out;
+                })
```
louise-poole, line 503: Tiny suggestion: might be nice to add a comment here that we purposefully do not increase the gas here as it's assumed that swapping more on the same pool does not increase the gas usage compared to swapping less (or something better worded lol)

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -436,12 +456,188 @@ pub(crate) fn build_post_swap_overrides(
...
+) -> Vec<SplitSwap> {
+    let len = hops.len();
+    let fraction_total: f64 = hops.iter().map(|h| h.split).sum();
+
+    let normalized: Vec<f64> = hops
+        .iter()
+        .map(|h| h.split / fraction_total)
+        .collect();
+    let amounts = fractions_to_amounts(total_available, &normalized)
+        .unwrap_or_else(|_| vec![total_available.clone()]);
```
louise-poole, line 551: Would be nice if we had a unit test for this behaviour: no splits (total goes to the single hop path)

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -915,4 +1047,371 @@ mod tests {
...
+        assert_eq!(swaps[2].component_id(), "pool_bz");
+        assert_eq!(*swaps[2].split(), 0.0);
+        assert_eq!(*swaps[2].amount_in(), BigUint::from(1200u64));
+        assert_eq!(*swaps[2].amount_out(), BigUint::from(4800u64));
+
+        assert_eq!(swaps[3].component_id(), "pool_cz");
+        assert_eq!(*swaps[3].split(), 0.0);
+        assert_eq!(*swaps[3].amount_in(), BigUint::from(1200u64));
+        assert_eq!(*swaps[3].amount_out(), BigUint::from(6000u64));
+    }
```
louise-poole, line 1474: Interesting - so the following hops after a split are also ordered according to the first split hops order (largest to smallest split fraction). Might be nice to note this here.

louise-poole: Also - I recall in the `Route` struct definition the swaps vec's doc assumes sequential swaps. This will need to be updated.

EDIT: ignore this - I've left a comment on @Troshchk 's PR where it's more appropriate to update this.

louise-poole (commented):
Nice! The core logic looks sound 👌 Just left a couple suggestions

fynd-core/src/algorithm/split_primitives.rs
Troshchk: Where did you leave the comment @louise-poole ?

Troshchk (approved):
Thank you!

fynd-core/src/algorithm/split_primitives.rs
tamaralipows: @louise-poole It's used in `merge_shared_hops` to create the final `SplitSwap` objects

fynd-core/src/algorithm/split_primitives.rs
tamaralipows: correct! added a comment

dianacarvalho1 (approved):
merci merci! 🙏🏼

PR #222, opened 2026-05-29
feat: add split route validation

Troshchk:
- Route::validate() now detects split routes and dispatches to validate_split_route()
  - Adds BFS connectivity, dead-end detection, and cycle detection

### DIFF
--- a/fynd-core/src/types/quote.rs
+++ b/fynd-core/src/types/quote.rs
@@ -2217,6 +2217,24 @@ mod tests {
     }
 
     #[test]
+    fn test_validate_split_topology_with_all_zero_splits() {
+        // Diamond topology but all split fractions are 0.0.
+        // is_split() returns false, so validate_sequential_route() runs
+        // and rejects the route (connectivity breaks at the second A→ swap).
+        //   ┌──[0.0]── B ──┐
+        // A │               D
+        //   └──[0.0]── C ──┘
+        let swaps = vec![
+            make_split_swap(0x01, 0x02, 0.0), // A→B
+            make_split_swap(0x01, 0x03, 0.0), // A→C
+            make_swap(0x02, 0x04, 490, 480),  // B→D
+            make_swap(0x03, 0x04, 490, 480),  // C→D
+        ];
+        let route = Route::new(swaps, HashMap::new());
+        assert!(route.validate().is_err());
+    }
+
+    #[test]
     fn test_is_split_detection() {
         // A → B                     (no split)
         let route_single = make_route(vec![(0x01, 0x02)]);
--- a/fynd-core/src/types/quote.rs
+++ b/fynd-core/src/types/quote.rs
@@ -1356,6 +1356,13 @@ fn validate_cycles(
     terminal_token: &Address,
 ) -> Result<(), RouteValidationError> {
     let is_round_trip = first_token == terminal_token;
+    if is_round_trip && swaps_by_token_in.len() <= 1 {
+        return Err(RouteValidationError::UnsupportedCycle {
+            token: first_token.clone(),
+            first: first_token.clone(),
+            last: terminal_token.clone(),
+        });
+    }
     let mut seen_back_edge_group = false;
     let mut earlier_inputs: HashSet<&Address> = HashSet::new();
     for (token_in, group) in swaps_by_token_in {
@@ -1978,6 +1985,20 @@ mod tests {
             make_split_swap(0x01, 0x06, 0.0),                // A→F (remainder)
             make_swap(0x05, 0x06, 270, 260),                 // E→F
             make_swap(0x06, 0x01, 810, 800),                 // F→A
+        ];
+        let route = Route::new(swaps, HashMap::new());
+        let err = route.validate().unwrap_err();
+        assert!(matches!(err, RouteValidationError::UnsupportedCycle { .. }));
+    }
+
+    #[test]
+    fn test_validate_split_single_group_round_trip() {
+        //   ┌──[50%]──┐
+        // A │         │ A   ERROR: single group cycling back to start
+        //   └──[rem]──┘
+        let swaps = vec![
+            make_split_swap(0x01, 0x01, 0.5), // A→A
+            make_split_swap(0x01, 0x01, 0.0), // A→A (remainder)
         ];
         let route = Route::new(swaps, HashMap::new());
         let err = route.validate().unwrap_err();
--- a/fynd-core/src/types/quote.rs
+++ b/fynd-core/src/types/quote.rs
@@ -1830,7 +1830,9 @@ mod tests {
 
     #[test]
     fn test_validate_interior_split_valid() {
-        // A→B sequential, then B→C parallel with remainder convention
+        //     ┌──[50%]──┐
+        // A → B         C
+        //     └──[rem]──┘
         let swaps = vec![
             make_split_swap(0x01, 0x02, 0.0),
             make_split_swap(0x02, 0x03, 0.5),
@@ -1842,7 +1844,9 @@ mod tests {
 
     #[test]
     fn test_validate_interior_split_no_remainder() {
-        // A→B, then B→C parallel but last swap has nonzero split (no remainder)
+        //     ┌──[50%]──┐
+        // A → B         C   ERROR: last swap has split=0.3, not 0.0
+        //     └──[30%]──┘
         let swaps = vec![
             make_split_swap(0x01, 0x02, 0.0),
             make_split_swap(0x02, 0x03, 0.5),
@@ -1855,7 +1859,9 @@ mod tests {
 
     #[test]
     fn test_validate_interior_split_sum_exceeds_one() {
-        // A→B, then B→C parallel with explicit splits summing >= 1.0
+        //     ┌──[60%]──┐
+        // A → B──[50%]──C   ERROR: 0.6 + 0.5 ≥ 1.0
+        //     └──[rem]──┘
         let swaps = vec![
             make_split_swap(0x01, 0x02, 0.0),
             make_split_swap(0x02, 0x03, 0.6),
@@ -1869,7 +1875,9 @@ mod tests {
 
     #[test]
     fn test_validate_source_split_valid() {
-        // Two parallel legs from A through different intermediates, converging on D
+        //   ┌──[50%]── B ──┐
+        // A │               D
+        //   └──[rem]── C ──┘
         let swaps = vec![
             make_split_swap(0x01, 0x02, 0.5),
             make_split_swap(0x01, 0x03, 0.0),
@@ -1882,8 +1890,9 @@ mod tests {
 
     #[test]
     fn test_validate_diamond_split_valid() {
-        // Diamond: A splits to B and D via different intermediates, both converge on C.
-        // A→B (60%), B→C, A→D (remainder), D→C
+        //   ┌──[60%]── B ──┐
+        // A │               C
+        //   └──[rem]── D ──┘
         let swaps = vec![
             make_split_swap(0x01, 0x02, 0.6), // A→B, 60%
             make_swap(0x02, 0x03, 590, 580),  // B→C
@@ -1896,7 +1905,7 @@ mod tests {
 
     #[test]
     fn test_validate_split_single_swap_nonzero_split() {
-        // A single swap for a given token_in must have split == 0.0
+        // A ──[50%]── B   ERROR: single swap must have split=0.0
         let swaps = vec![make_split_swap(0x01, 0x02, 0.5)];
         let route = Route::new(swaps, HashMap::new());
         let err = route.validate().unwrap_err();
@@ -1905,7 +1914,9 @@ mod tests {
 
     #[test]
     fn test_validate_split_dead_end() {
-        // A splits to B and C, but the paths don't converge — B→D diverges
+        //   ┌──[50%]── B → D   ERROR: dead end, D not consumed
+        // A │
+        //   └──[rem]── C → E   (terminal)
         let swaps = vec![
             make_split_swap(0x01, 0x02, 0.5), // A→B
             make_split_swap(0x01, 0x03, 0.0), // A→C
@@ -1919,7 +1930,9 @@ mod tests {
 
     #[test]
     fn test_validate_split_cycle() {
-        // Both split paths converge back to B, creating a cycle
+        //     ┌──[50%]── C ──┐
+        // A → B               → B   ERROR: C→B and D→B cycle back
+        //     └──[rem]── D ──┘
         let swaps = vec![
             make_swap(0x01, 0x02, 1000, 990),                // A→B
             make_swap(0x02, 0x03, 990, 980).with_split(0.5), // B→C
@@ -1934,7 +1947,9 @@ mod tests {
 
     #[test]
     fn test_validate_split_round_trip_valid() {
-        // A splits to B and C, both converge on D, then D→A (round-trip).
+        //   ┌──[50%]── B ──┐
+        // A │               D → A   (round-trip back to start)
+        //   └──[rem]── C ──┘
         let swaps = vec![
             make_split_swap(0x01, 0x02, 0.5), // A→B
             make_split_swap(0x01, 0x03, 0.0), // A→C
@@ -1948,9 +1963,12 @@ mod tests {
 
     #[test]
     fn test_validate_split_multi_back_edge_groups() {
-        // A→B→D→A→F→A and A→C→D→E→F→A — two groups (D and F) produce
-        // back-edges to A, which means A is visited more than twice.
-        // A→F merges into group A so splits must account for all three.
+        //   ┌──[30%]── B ──┐       ┌──[rem]──┐
+        // A ┤               D──→A──┤          F──→A
+        //   └──[30%]── C ──┘       └──── E ──┘
+        //                    ↑                  ↑
+        //              back-edge 1        back-edge 2
+        //         ERROR: two groups cycle back to A
         let swaps = vec![
             make_split_swap(0x01, 0x02, 0.3),                // A→B
             make_split_swap(0x01, 0x03, 0.3),                // A→C
@@ -1969,8 +1987,9 @@ mod tests {
 
     #[test]
     fn test_validate_split_terminal_is_group_input() {
-        // A→C→B→C: terminal C is also a group input, creating a cycle.
-        // Not a round-trip (first=A ≠ terminal=C).
+        //     ┌──[50%]── B ──┐
+        // A → C               → C   ERROR: B→C and D→C cycle back
+        //     └──[rem]── D ──┘       (not a round-trip: first=A ≠ terminal=C)
         let swaps = vec![
             make_swap(0x01, 0x03, 1000, 990),                // A→C
             make_swap(0x03, 0x02, 990, 980).with_split(0.5), // C→B
@@ -2200,15 +2219,17 @@ mod tests {
 
     #[test]
     fn test_is_split_detection() {
-        // Single hop, no split
+        // A → B                     (no split)
         let route_single = make_route(vec![(0x01, 0x02)]);
         assert!(!route_single.is_split());
 
-        // Multi-hop, no split
+        // A → B → C                 (no split)
         let route_multi = make_route(vec![(0x01, 0x02), (0x02, 0x03)]);
         assert!(!route_multi.is_split());
 
-        // Interior split: A→B single, B→C through two parallel pools
+        //     ┌──[60%]──┐
+        // A → B         C           (interior split)
+        //     └──[rem]──┘
         let swaps_interior = vec![
             make_swap(0x01, 0x02, 1000, 990),                // A→B, no split
             make_swap(0x02, 0x03, 594, 580).with_split(0.6), // B→C via P2, 60%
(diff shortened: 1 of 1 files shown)


fynd-core/src/types/quote.rs
```diff
@@ -1180,6 +1183,153 @@ impl Route {
 
         Ok(())
```
louise-poole, line 1195: For consistency/neatness can we move this logic to a `validate_sequential_route` helper fn?

fynd-core/src/types/quote.rs
```diff
@@ -1180,6 +1183,172 @@ impl Route {
...
+        self.swaps.iter().any(|s| s.split > 0.0)
+    }
+
+    /// Validates a split route.
+    ///
+    /// Groups swaps by `token_in`, then checks:
+    /// - Single-swap groups: `split` must be `0.0`
+    /// - Multi-swap groups: all but the last must have `0.0 < split < 1.0`, the last must have
+    ///   `split == 0.0` (remainder), and the sum must be strictly less than `1.0`
+    /// - BFS connectivity: outputs of each group connect to inputs of the next
```
louise-poole: I think it'd be useful for this behaviour to be documented on the Route struct itself too (currently the vec<Swaps> doc assumes sequential swaps). It's important for future algorithms to adhere to it.

fynd-core/src/types/quote.rs
```diff
@@ -1180,6 +1183,172 @@ impl Route {
...
+                    }
+                }
+            }
+        }
+        for (token_in, _) in &groups {
+            if !reachable.contains(token_in) {
+                return Err(RouteValidationError::InvalidSplit {
+                    reason: format!(
+                        "group input {token_in} is not reachable from \
+                         the start token {first_token} (reachable: {reachable:?})",
```
louise-poole: I wonder if this error message prints out nicely. In your test for this error can you verify the error string is easy to read? Essentially add the error string itself to the asserts. We don't want addresses to truncate or diplay as raw bytes etc.

fynd-core/src/types/quote.rs
```diff
@@ -1180,6 +1183,172 @@ impl Route {
...
+        let mut group_index: HashMap<Address, usize> = HashMap::new();
+        let mut groups: Vec<(Address, Vec<&Swap>)> = Vec::new();
+        for swap in &self.swaps {
+            if let Some(&idx) = group_index.get(&swap.token_in) {
+                groups[idx].1.push(swap);
+            } else {
+                group_index.insert(swap.token_in.clone(), groups.len());
+                groups.push((swap.token_in.clone(), vec![swap]));
+            }
+        }
```
louise-poole: Can you add a comment explaining how they're grouped? At first read my brain just assumed 'group' was a split path and took a little bit for me to go back and realise it was grouped by input token.

fynd-core/src/types/quote.rs
```diff
@@ -1180,6 +1183,172 @@ impl Route {
...
+                for swap in &groups[idx].1 {
+                    if reachable.insert(&swap.token_out) {
+                        queue.push_back(&swap.token_out);
+                    }
+                }
+            }
+        }
+        for (token_in, _) in &groups {
+            if !reachable.contains(token_in) {
+                return Err(RouteValidationError::InvalidSplit {
```
louise-poole, line 1268: Actually shouldn't this be a `DisconnectedSwaps` error?

fynd-core/src/types/quote.rs
```diff
@@ -1180,6 +1183,172 @@ impl Route {
...
+                        ),
+                    });
+                }
+            }
+        }
+
+        // Cycle detection: no swap output may match an earlier group's input.
+        // Exception: in a round-trip (first == terminal) exactly one group may
+        // produce back-edges to first_token. Multiple groups doing so means token
+        // is visited more than twice (e.g. A→…→A→…→A).
```
louise-poole: Why can only one group route back to the start token? What happens with the split paths: A -> B -> C -> A and A -> B -> D -> A. Groups with input tokens D and C both go back to start token.

louise-poole (commented):
Looking good! Some complex checks there, well done 👌

fynd-core/src/types/quote.rs
tamaralipows: I think this var could be named better (my Claude also took a liking to "group" and "queue" the other day, but I've added to my local config that he should avoid such general var names). In general if we can avoid comments by having better named vars we should go that route

fynd-core/src/types/quote.rs
```diff
@@ -1180,6 +1183,172 @@ impl Route {
 
         Ok(())
     }
+
+    /// Returns `true` if the route contains any swaps with non-zero split
+    /// fractions.
+    pub fn is_split(&self) -> bool {
+        self.swaps.iter().any(|s| s.split > 0.0)
+    }
```
tamaralipows, line 1202: I think this is bad. Like what if we have multiple splits in a branch that are 0.0 - this should probably be caught in the validation and fail, but we would skip it because of this if check (please also add a test for this case)

fynd-core/src/types/quote.rs
Troshchk: Could you clarify this, please?
Do you mean move everything sequential related from `validate()` to a helper function `validate_sequential_route()`?

fynd-core/src/types/quote.rs
tamaralipows: Yeah this should be a valid case

fynd-core/src/types/quote.rs
```diff
@@ -1180,6 +1183,172 @@ impl Route {
...
+                            "swap output {} is a dead end — not consumed by any \
+                             later group and not the terminal token {terminal_token}",
+                            swap.token_out
+                        ),
+                    });
+                }
+            }
+        }
+
+        // Cycle detection: no swap output may match an earlier group's input.
```
tamaralipows: I would also like to add that these should have at least one hop? i.e. if there only one hop and it has a cycle, there's likely an error

fynd-core/src/types/quote.rs
```diff
@@ -1180,6 +1183,172 @@ impl Route {
...
+                    reason: format!(
+                        "split sum for token_in {token_in} must be \
+                         < 1.0, got {sum}"
+                    ),
+                });
+            }
+        }
+
+        // BFS connectivity: starting from the first group's token_in,
+        // expand through swap outputs to verify all group inputs are reachable.
```
tamaralipows: This function is quite massive. It'd perhaps be better to break all these into helper methods and call them here:
- `validate_split_amounts`
- `validate_bfs_connectivity`
- `validate_dead_ends`
- `validate_cycles`

Something like that?

fynd-core/src/types/quote.rs
```diff
@@ -1180,6 +1183,172 @@ impl Route {
...
+        let non_first_input_tokens: HashSet<&Address> = groups
+            .iter()
+            .skip(1)
+            .map(|(token_in, _)| token_in)
+            .collect();
+
+        // Dead-end detection: every swap output must either feed a later group
+        // or be the terminal output token.
+        for (_, group) in &groups {
+            for swap in group {
```
tamaralipows: Couldn't we do something like `for swap in groups.values().flatten()` for better readability

fynd-core/src/types/quote.rs
```diff
@@ -1594,6 +1769,169 @@ mod tests {
...
+            make_split_swap(0x02, 0x03, 0.0),
+        ];
+        let route = Route::new(swaps, HashMap::new());
+        let err = route.validate().unwrap_err();
+        assert!(matches!(err, RouteValidationError::InvalidSplit { .. }));
+    }
+
+    #[test]
+    fn test_validate_source_split_valid() {
+        // Two parallel legs from A through different intermediates, converging on D
```
tamaralipows: Let's add nice diagrams to the tests to see at a glance. Words don't land the same way

fynd-core/src/types/quote.rs
```diff
@@ -1180,6 +1183,172 @@ impl Route {
...
+                });
+            }
+        }
+
+        // BFS connectivity: starting from the first group's token_in,
+        // expand through swap outputs to verify all group inputs are reachable.
+        let first_token = &groups[0].0;
+        let mut reachable: HashSet<&Address> = HashSet::new();
+        reachable.insert(first_token);
+        let mut queue: VecDeque<&Address> = VecDeque::new();
```
tamaralipows: Also not a fan of the general "queue" var name

tamaralipows (commented):
Thanks! Some comments

fynd-core/src/types/quote.rs
Troshchk: the error looks like: 
`InvalidSplit {
        reason: "group input 0x0303030303030303030303030303030303030303 is not reachable from the start token 0x0101010101010101010101010101010101010101 (reachable: [0x0101010101010101010101010101010101010101, 0x0202020202020202020202020202020202020202])",
    }`
    
 How do you think it should be displayed instead?

fynd-core/src/types/quote.rs
Troshchk: I think `DisconnectedSwaps` is designed for the sequential swaps coz it contains first_out and second_in.
I have added this one another error `DisconnectedGroup`.

fynd-core/src/types/quote.rs
Troshchk: This case is valid coz A is the initial token, I was talking about the case when:
A -> B -> D -> A -> F -> A
A -> C -> D -> E -> F -> A

In this case A is visited twice and afaik it should not be allowed. Is this assumption not correct?

fynd-core/src/types/quote.rs
louise-poole: Yes, sorry not sure why it didn't highlight the code I was trying to highlight. 😓 Yes, that's exactly what I meant

fynd-core/src/types/quote.rs
louise-poole: Nice! No that's fine then. Was concerned it would display in raw bytes format, but this looks good!

fynd-core/src/types/quote.rs
louise-poole: Yes, one path having 3 instances of the start token is an invalid case (as you have said). But this is a path-based check and not a group-based check. Multiple groups can produce the start token, but not have internal cycles (like the example I gave).

fynd-core/src/types/quote.rs
Troshchk: This function is basically from the ticket description.
I think, if there are multiple splits, than it would be rejected by `validate_sequential_route` with `DisconnectedSwaps` error. Do you think it is sufficient?
As an alternative I can rewrite `is_split` to detect split swaps differently. Lmk what you think.
I have added a test for this

tamaralipows: Ah I see you are calling `validate_sequential_route` right after no matter what then. All good!
Just make sure there is a test for that case

fynd-core/src/types/quote.rs
```diff
@@ -1594,6 +1828,202 @@ mod tests {
...
+            make_split_swap(0x01, 0x03, 0.0),
+            make_swap(0x02, 0x04, 490, 480),
+            make_swap(0x03, 0x04, 490, 480),
+        ];
+        let route = Route::new(swaps, HashMap::new());
+        assert!(route.validate().is_ok());
+    }
+
+    #[test]
+    fn test_validate_diamond_split_valid() {
```
dianacarvalho1: these two tests are doing the same thing

fynd-core/src/types/quote.rs
```diff
@@ -1594,6 +1828,202 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_validate_split_multi_back_edge_groups() {
+        //   ┌──[30%]── B ──┐       ┌──[rem]──┐
+        // A ┤               D──→A──┤          F──→A
+        //   └──[30%]── C ──┘       └──── E ──┘
+        //                    ↑                  ↑
+        //              back-edge 1        back-edge 2
+        //         ERROR: two groups cycle back to A
```
dianacarvalho1: do we need such a complex example? 🤔 what is being tested here? it should fail already after the first group no?

fynd-core/src/types/quote.rs
```diff
@@ -1809,6 +2239,58 @@ mod tests {
...
+        // A │               D
+        //   └──[0.0]── C ──┘
+        let swaps = vec![
+            make_split_swap(0x01, 0x02, 0.0), // A→B
+            make_split_swap(0x01, 0x03, 0.0), // A→C
+            make_swap(0x02, 0x04, 490, 480),  // B→D
+            make_swap(0x03, 0x04, 490, 480),  // C→D
+        ];
+        let route = Route::new(swaps, HashMap::new());
+        assert!(route.validate().is_err());
```
dianacarvalho1: can we assert with which error it fails please?

dianacarvalho1 (commented):
Thank you @Troshchk ! 🙏🏼 only have small comments concerning the tests!

fynd-core/src/types/quote.rs
Troshchk: I have removed this tests case and created a [cooldown task](https://propeller-heads.atlassian.net/browse/ENG-6015) to investigate routes, where the same token appears multiple times.

fynd-core/src/types/quote.rs
```diff
@@ -1180,6 +1194,198 @@ impl Route {
...
+        let mut token_in_to_index: HashMap<Address, usize> = HashMap::new();
+        let mut swaps_by_token_in: Vec<(Address, Vec<&Swap>)> = Vec::new();
+        for swap in &self.swaps {
+            if let Some(&idx) = token_in_to_index.get(&swap.token_in) {
+                swaps_by_token_in[idx].1.push(swap);
+            } else {
+                token_in_to_index.insert(swap.token_in.clone(), swaps_by_token_in.len());
+                swaps_by_token_in.push((swap.token_in.clone(), vec![swap]));
+            }
+        }
```
tamaralipows, line 1224: Why don't we just make `swaps_by_token_in` a HashMap right away instead of having it be a tuple and then indexing like this?

fynd-core/src/types/quote.rs
```diff
@@ -1180,6 +1194,198 @@ impl Route {
...
+        validate_split_amounts(&swaps_by_token_in)?;
+        validate_bfs_connectivity(&swaps_by_token_in, &token_in_to_index)?;
+
+        let first_token = &swaps_by_token_in[0].0;
+        let terminal_token = &self.swaps[self.swaps.len() - 1].token_out;
+        let non_first_input_tokens: HashSet<&Address> = swaps_by_token_in
+            .iter()
+            .skip(1)
+            .map(|(token_in, _)| token_in)
+            .collect();
```
tamaralipows: Maybe it would be more organized to move this inside of `validate_dead_ends`?

fynd-core/src/types/quote.rs
```diff
@@ -1180,6 +1194,198 @@ impl Route {
...
+                token_in_to_index.insert(swap.token_in.clone(), swaps_by_token_in.len());
+                swaps_by_token_in.push((swap.token_in.clone(), vec![swap]));
+            }
+        }
+
+        validate_split_amounts(&swaps_by_token_in)?;
+        validate_bfs_connectivity(&swaps_by_token_in, &token_in_to_index)?;
+
+        let first_token = &swaps_by_token_in[0].0;
+        let terminal_token = &self.swaps[self.swaps.len() - 1].token_out;
```
tamaralipows: And maybe it would also be more organized to move this inside of `validate_cycles` so this validate method looks really clean, it's just the four validations consecutively with any filtering happening inside the sub-methods

fynd-core/src/types/quote.rs
```diff
@@ -1180,6 +1194,198 @@ impl Route {
...
+            for swap in &swaps_by_token_in[idx].1 {
+                if reachable.insert(&swap.token_out) {
+                    tokens_to_visit.push_back(&swap.token_out);
+                }
+            }
+        }
+    }
+    for (token_in, _) in swaps_by_token_in {
+        if !reachable.contains(token_in) {
+            return Err(RouteValidationError::DisconnectedBranchCollection {
```
tamaralipows: I would maybe call this error something like `UnreachableInputToken` to differentiate more from the dead end scenario

fynd-core/src/types/quote.rs
```diff
@@ -1180,6 +1194,198 @@ impl Route {
...
+    swaps_by_token_in: &[(Address, Vec<&Swap>)],
+    non_first_input_tokens: &HashSet<&Address>,
+    terminal_token: &Address,
+) -> Result<(), RouteValidationError> {
+    for swap in swaps_by_token_in
+        .iter()
+        .flat_map(|(_, swaps)| swaps)
+    {
+        if !non_first_input_tokens.contains(&swap.token_out) && &swap.token_out != terminal_token {
+            return Err(RouteValidationError::InvalidSplit {
```
tamaralipows: It's a bit weird that the `DisconnectedBranchCollection` but this one is lumped into `InvalidSplit` - we should maybe also make a separate error?

tamaralipows (commented):
Yaaay thank you! This looks great! I'm happy about the tests, with the really nice diagrams. Super readable and quick to check that all the necessary cases are there.

My comments are just extremely minor things now. Let me know if you agree!

fynd-core/src/types/quote.rs
Troshchk: Because we want to preserve the order of insertion, which hashmap does not guarantee. 
We rely on this order to obtain the first token and also during cycle validation

fynd-core/src/types/quote.rs
Troshchk: Could you clarify, please? These are already different: the dead ends return `InvalidSplit` as an error and `DisconnectedBranchCollection` is only returned here

fynd-core/src/types/quote.rs
Troshchk: Changed to `DeadEndOutput`

fynd-core/src/types/quote.rs
Troshchk: I think I understood what you meant
 I have changed the error in dead ends to be `DeadEndOutput` and for this error -- since we are checking for connectivity, I would be leaning to keep it as is `DisconnectedBranchCollection` or may be `DisconnectedInputToken`
 Lmk what you think

louise-poole: `DisconnectedBranchCollection` is leaking internal var names and not using our publicly understood verbage. Could be confusing to users.
`DisconnectedInputToken` makes it sound like the _input token_ alone doesn't link to the rest of the solution. Meanwhile we're trying to say that it's a _split group_ that is disconnected from the rest of the path.
May I suggest simply `DisconnectedSplit`?

louise-poole: Or another suggestion to bring some consistency to the error names:
`DisconnectedBranchCollection` -> `DisconnectedSplitStart`
`DeadEndOutput` -> `DisconnectedSplitEnd`

fynd-core/src/types/quote.rs
```diff
@@ -1212,6 +1412,38 @@ pub enum RouteValidationError {
...
+        /// Human-readable description of the violation.
+        reason: String,
+    },
+    /// A swap output token is a dead end — not consumed by any later branch
+    /// collection and not the terminal token.
+    #[non_exhaustive]
+    #[error(
+        "dead-end output: {token_out} is not consumed by any later branch \
+         collection and is not the terminal token {terminal_token}"
+    )]
```
louise-poole, line 1434: Can you update the error message to not refer to branch collection? That's an internally used term and will be hard for users to understand/reason out.

fynd-core/src/types/quote.rs
```diff
@@ -1180,6 +1194,192 @@ impl Route {
...
+    Ok(())
+}
+
+/// Dead-end detection: every swap output must either feed a later branch
+/// collection or be the terminal output token.
+fn validate_dead_ends(
+    swaps_by_token_in: &[(Address, Vec<&Swap>)],
+    swaps: &[Swap],
+) -> Result<(), RouteValidationError> {
+    let terminal_token = &swaps[swaps.len() - 1].token_out;
```
louise-poole: Small note here:
`swaps[swaps.len() - 1]` would error if the swaps slice was ever empty. It's better to use `swaps.last()` for this.

fynd-core/src/types/quote.rs
```diff
@@ -1212,6 +1412,38 @@ pub enum RouteValidationError {
         /// Last token in the route.
         last: Address,
     },
+    /// A split route has invalid split fractions.
+    #[error("invalid split route: {reason}")]
+    InvalidSplit {
```
louise-poole, line 1424: Can you add the `#[non_exhaustive]` flag here please?

louise-poole (commented):
Looking great! Just some final minor comments, then I think we can merge 👍

fynd-core/src/types/quote.rs
Troshchk: I like your last suggestion!
@tamaralipows lmk if also agree

fynd-core/src/types/quote.rs
Troshchk: There is a check above ensuring that the swaps are not empty:
``` 
if self.swaps.is_empty() {
            return Err(RouteValidationError::EmptyRoute);
        }
```

I di dit this way to avoid `expect()`, do you still prefer to use last()?

louise-poole: Yes, but that check isn't in this fn... This fn is trusting that the caller already did the check. Since we know we're the only caller it might be fine to keep this assumption, but please at least document it in a docstring? Ortherwise yes, I'd prefer to use the safer `.last()` and return an error if it errors (not `expect`... you're right to want to avoid that).

louise-poole: Side note: in the current context using `.expect()` or using `swaps[swaps.len() - 1]` would both result in a panic if ever triggered. We want to avoid both and rather return an error.

fynd-core/src/types/quote.rs
tamaralipows: Sounds good!

tamaralipows (approved):
✨ Thank you! ✨

louise-poole (approved):
Niiiice!

PR #223, opened 2026-06-01
test: add ScenarioResult and generic evaluate_scenario harness

louise-poole:
Added:
- ScenarioResult with `assert_meets_lower_bound`, `assert_beats_lower_bound`, and `within_pct_of_optimum`. 
- `evaluate_scenario` generic over any Algorithm — caller supplies market and graph manager so weighted and unweighted algorithms both work.

### DIFF
--- a/fynd-core/src/algorithm/split_test_harness.rs
+++ b/fynd-core/src/algorithm/split_test_harness.rs
@@ -92,8 +92,8 @@ pub(crate) struct ScenarioResult {
     pub net_output: BigInt,
     /// Best single-route net output. `net_output` must be >= this.
     pub lower_bound: BigInt,
-    /// Net analytical optimum for the scenario's simplified pool model. A quality target, not a
-    /// hard ceiling.
+    /// Net analytical optimum for the scenario's simplified pool model. A reference value for
+    /// measuring quality — the algorithm is not required to reach it.
     pub analytical_optimum: BigInt,
     /// Number of swaps consuming `token_in`. 1 for single-route, N for a split.
     pub path_count: usize,
@@ -112,10 +112,11 @@ impl ScenarioResult {
     }
 
     /// Returns true if `net_output >= (100 - threshold_pct)% of analytical_optimum`.
+    /// `threshold_pct` is a whole-number percentage (e.g. `5` means within 5% of optimum).
     #[allow(dead_code)]
     pub fn within_pct_of_optimum(&self, threshold_pct: u32) -> bool {
-        self.net_output.clone() * BigInt::from(100u32) >=
-            self.analytical_optimum.clone() * BigInt::from(100 - threshold_pct)
+        &self.net_output * BigInt::from(100u32) >=
+            &self.analytical_optimum * BigInt::from(100 - threshold_pct)
     }
 
     fn assert_bound(&self, passes: bool, op: &str) {
--- a/fynd-core/src/algorithm/split_test_harness.rs
+++ b/fynd-core/src/algorithm/split_test_harness.rs
@@ -1,11 +1,17 @@
 //! Test helpers for split-routing algorithm split_scenarios.
 
 use num_bigint::{BigInt, BigUint};
-use tycho_simulation::tycho_core::{models::token::Token, simulation::protocol_sim::ProtocolSim};
+use num_traits::Zero;
+use tycho_simulation::tycho_core::{
+    models::{token::Token, Address},
+    simulation::protocol_sim::ProtocolSim,
+};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


Troshchk (approved):
Thank you! Lgtm

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -64,40 +70,135 @@ pub(crate) struct TestScenario {
...
+    }
+
+    #[allow(dead_code)]
+    pub fn assert_beats_lower_bound(&self) {
+        self.assert_bound(self.net_output > self.lower_bound, ">");
+    }
+
+    /// Returns true if `net_output >= (100 - threshold_pct)% of analytical_optimum`.
+    #[allow(dead_code)]
+    pub fn within_pct_of_optimum(&self, threshold_pct: u32) -> bool {
```
dianacarvalho1, line 117: pct is percentage right?

dianacarvalho1 (approved):
Thank you @louise-poole 🙏🏼

fynd-core/src/algorithm/split_test_harness.rs
louise-poole: Yes - I'll update the doc string to make this clearer 👍

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -64,40 +70,135 @@ pub(crate) struct TestScenario {
...
+/// Results from running one algorithm against one scenario.
+#[allow(dead_code)]
+pub(crate) struct ScenarioResult {
+    pub scenario_name: &'static str,
+    pub algorithm_name: String,
+    /// Gross output minus gas costs. Can be negative when gas exceeds proceeds.
+    pub net_output: BigInt,
+    /// Best single-route net output. `net_output` must be >= this.
+    pub lower_bound: BigInt,
+    /// Net analytical optimum for the scenario's simplified pool model. A quality target, not a
```
tamaralipows: What does the hard ceiling mean?

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -64,40 +70,135 @@ pub(crate) struct TestScenario {
...
+
+    #[allow(dead_code)]
+    pub fn assert_beats_lower_bound(&self) {
+        self.assert_bound(self.net_output > self.lower_bound, ">");
+    }
+
+    /// Returns true if `net_output >= (100 - threshold_pct)% of analytical_optimum`.
+    #[allow(dead_code)]
+    pub fn within_pct_of_optimum(&self, threshold_pct: u32) -> bool {
+        self.net_output.clone() * BigInt::from(100u32) >=
```
tamaralipows: Small thing but I don't think clones are necessary here, we could use &

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -432,11 +533,62 @@ pub(crate) mod split_scenarios {
...
     }
 
+    fn bf_default() -> BellmanFordAlgorithm {
+        BellmanFordAlgorithm::with_config(AlgorithmConfig::default()).unwrap()
+    }
+
+    // ==================== evaluate_scenario tests ====================
+
+    #[tokio::test]
+    async fn evaluate_bellman_ford_passes_lower_bound_on_all_scenarios() {
```
tamaralipows, line 550: Just want to vent about my pet peeve, no need to change, but I find it super annoying how claude puts the entire outcome of the test in the test name 😭

tamaralipows (approved):
Thank you!

fynd-core/src/algorithm/split_test_harness.rs
louise-poole: Agreed!!!! I find this crazy annoying. Fought it lots and recently gave up. You finding it annoying too has given me new energy to keep fighting the good fight 😆

tamaralipows: I'm actually realizing now that this is very similar to the test specified in https://propeller-heads.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-5842, so actually the majority of that test task is pretty much done. So I just adapted it a little bit to check for the exact output, consider gas in the algo, increase hops, (and change the name) in my PR. I haven't removed this original test though, we probably can remove it I guess?

fynd-core/src/algorithm/split_test_harness.rs
louise-poole: Hard ceiling means that you can never do more than this value. I was trying to emphasise that it's a target and not a cap.... will reword it better though 👍

PR #224, opened 2026-06-01
feat(BF): Extract find_single_route sync method with MarketOverrides

dianacarvalho1:
Combined two tasks: [ENG-5849](https://propeller-heads.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-5849) and [ENG-5850](https://propeller-heads.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-5850)
- Extract find_single_route sync method with MarketOverrides
- Use it in find_best_route
- Add test_find_single_route_with_state_overrides and test_find_route_options_default_is_no_overrides
- Update helper visibility

[ENG-5849]: https://propeller-heads.atlassian.net/browse/ENG-5849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ENG-5850]: https://propeller-heads.atlassian.net/browse/ENG-5850?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/fynd-core/src/algorithm/bellman_ford.rs
+++ b/fynd-core/src/algorithm/bellman_ford.rs
@@ -84,7 +84,6 @@ pub(crate) enum RouteScoringMode {
 
 /// Per-call overrides for `find_single_route`.
 #[derive(Default)]
-#[allow(dead_code)]
 pub(crate) struct FindRouteOptions {
     /// Pool state overrides: degrade or zero-gas specific pools without modifying market data.
     pub(crate) overrides: MarketOverrides,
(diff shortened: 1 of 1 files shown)


Troshchk (approved):
Thank you @dianacarvalho1

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -370,373 +636,130 @@ impl BellmanFordAlgorithm {
...
-                            );
-                            continue;
-                        }
-                    };
-
-                    let candidate_cumul_gas = &cumul_gas[u_idx] + &result.gas;
+        let mut token_nodes: HashSet<NodeIndex> = HashSet::new();
+        let mut component_ids: HashSet<ComponentId> = HashSet::new();
+        let mut visited = HashSet::new();
+        let mut queue = VecDeque::new();
```
tamaralipows: I don't like this ambiguous queue var name that I see showing up in all our PRs thanks to claudio :( I know it was here before but can we do better, like say what the queue is supposed to hold

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -370,373 +636,130 @@ impl BellmanFordAlgorithm {
...
+        route: &Route,
+        gas_price: &BigUint,
+        token_prices: Option<&TokenGasPrices>,
+        spot_product: &[f64],
+        node_address: &HashMap<NodeIndex, Address>,
+        token_in_node: NodeIndex,
+    ) -> BigInt {
+        let Some(last_swap) = route.swaps().last() else {
+            return BigInt::from(amount_out.clone());
+        };
```
tamaralipows: This means the route has no swaps no? Shouldn't we error?

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -370,373 +636,130 @@ impl BellmanFordAlgorithm {
...
-            tokens
-                .entry(token_in.address.clone())
-                .or_insert_with(|| token_in.clone());
-            tokens
-                .entry(token_out.address.clone())
-                .or_insert_with(|| token_out.clone());
+        let total_gas = route.total_gas();
+
+        if gas_price.is_zero() {
+            warn!("missing gas price, returning gross amount_out");
```
tamaralipows, line 756: Should we be okay with this actually? (I don't know the answer)
I would rather error/not consider that path?

tamaralipows (commented):
Thanks @dianacarvalho1! Just some small comments. Also I'd say this is moreso a refactor than a feat?

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -222,6 +221,273 @@ impl BellmanFordAlgorithm {
         })
     }
 
+    /// Runs the SPFA relaxation loop and reconstructs the best route from a pre-built context.
+    ///
+    /// This is the repeatable, synchronous solve phase. Call it multiple times with different
+    /// `opts.overrides` to evaluate alternative pool states without redoing the setup in `ctx`.
+    /// Overrides shadow the corresponding pool in `ctx.market_data` for both relaxation and route
+    /// construction.
+    pub(crate) fn find_single_route(
```
louise-poole, line 250: Small note for readability - I find this really long... we could split this with some helper fns. Maybe extract a `run_spfa` fn?

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -1552,4 +1575,73 @@ mod tests {
...
+            .unwrap();
+        assert_eq!(overridden.route().swaps()[0].amount_out(), &BigUint::from(1000u64));
+
+        assert!(
+            overridden.route().swaps()[0].amount_out() < normal.route().swaps()[0].amount_out()
+        );
+    }
+
+    #[tokio::test]
+    async fn test_find_route_options_default_is_no_overrides() {
```
louise-poole: I agree with @tamaralipows 's sentiment (in other PRs) that the very verbose Clauded test names suck 😆 I'm going to update our repo claude docs to try help this situation... for now can you shorten this? 
```suggestion
    async fn test_find_route_options_default() {
```

fynd-core/src/algorithm/bellman_ford.rs
louise-poole: Hmmm... I don't think a missing gas price is worth crippling the solver over. At this stage, roughly accounting for gas allows us to make a more informed decision on which route to choose overall. Without gas price we're still able to make a decision... it's not ideal, but not as bad as rendering the solver as unusable.

louise-poole (approved):

fynd-core/src/algorithm/bellman_ford.rs
tamaralipows: That's true if the gas price is equally missing for all the swaps but I thought this meant that it could be missing for just one? 
Like if it's only missing for one or a few, that could force us to choose those swaps over other swaps where the gas is available, which could give bad prices. I guess that's not as bad as potentially crippling the solver if gas is unavailable for a long time though?

fynd-core/src/algorithm/bellman_ford.rs
dianacarvalho1: true.. this should actually never happen.. let me try to make it better

fynd-core/src/algorithm/bellman_ford.rs
dianacarvalho1: I think if the gas price is ever missing, then it will be missing for all so this should be ok!

fynd-core/src/algorithm/bellman_ford.rs
dianacarvalho1: yup! splitting this into smaller methods!

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -89,6 +89,18 @@ pub(crate) struct FindRouteOptions {
     pub(crate) overrides: MarketOverrides,
 }
 
+/// Output of the SPFA relaxation pass: per-node best-path arrays.
+struct SpfaResult {
```
tamaralipows: `SPFAResult`?

tamaralipows (approved):
Thank yoooouuuu!

fynd-core/src/algorithm/bellman_ford.rs
louise-poole: Also note - gas price is only missing before the first fetch succeeds. After that we'll always have a gas price, though it might be outdated.

PR #226, opened 2026-06-02
feat: add flashblock support

louise-poole:
Add partial blocks flag to enable consuming flashblocks.

Other improvements:
- move gas price fetch off the hot path. At the faster block rate of flashblocks this small delay waiting for gas price compounded exponentially and fynd could not keep up with consuming the block messages. Now gas price fetch is async and only fires on a new tycho message if one isn't already in flight.


fynd-core/src/feed/tycho_feed.rs
```diff
@@ -376,29 +376,12 @@ impl TychoFeed {
...
-                })?;
-
-            signal_rx.await.map_err(|e| {
-                DataFeedError::GasPriceFetcherError(format!(
-                    "Failed to receive gas price refresh confirmation: {}",
-                    e
-                ))
-            })?;
+    /// Signals the gas price worker to refresh. Non-blocking: skipped if the worker is
+    /// already processing a previous signal (channel capacity 5).
```
kayibal: Why is this channel capacity set to 5? Shouldn't it be 1?

fynd-core/src/feed/tycho_feed.rs
```diff
@@ -202,9 +202,9 @@ impl TychoFeed {
                         Some(msg) => {
                             trace!("Received message from protocol stream: {:?}", msg);
                             let msg = msg.map_err(|e| DataFeedError::StreamError(e.to_string()))?;
-                            // Refresh gas price before broadcasting the event so that
-                            // ComputationManager has gas price available when it starts computing.
-                            self.refresh_gas_price().await?;
+                            // Signal the gas price worker to refresh concurrently. Best-effort:
+                            // skipped if the worker is already busy with a previous fetch.
+                            self.signal_gas_price_refresh();
```
kayibal: I wonder if we even should bother tying gas price updates to the block, could just run every x seconds instead. I think the assumption new_block -> new_gas_price is wrong, the next blocks estimates priority fee keeps updating all the time continuously

louise-poole: I had a similar thought while doing this, but didn't want to take on bigger design decisions unless it was necessary. However, the fact that you also thought this is good enough for me to implement it - will do so shortly.

kayibal: Since you are already changing this part of the code and it should make things simpler, I agree let's just do it now

kayibal (approved):

PR #229, opened 2026-06-02
feat: PathFrankWolfe structs

tamaralipows:
[Task](https://propeller-heads.atlassian.net/browse/ENG-5853)
- Assumed expects are okay in `computation_requirements` since all algorithms have this and this is only at startup 
- Remove "fraction" from param names `max_probe` and `min_split`. It was confusing that it said fraction but was f64

### DIFF
--- a/fynd-core/src/algorithm/path_frank_wolfe.rs
+++ b/fynd-core/src/algorithm/path_frank_wolfe.rs
@@ -38,35 +38,22 @@ impl Default for PathFrankWolfeConfig {
 /// and optimises the input split across them using a Frank-Wolfe loop.
 pub struct PathFrankWolfeAlgorithm {
     inner: BellmanFordAlgorithm,
+    // Used by the split-routing loop (not yet implemented).
+    #[allow(dead_code)]
     config: PathFrankWolfeConfig,
 }
 
 impl PathFrankWolfeAlgorithm {
     /// Creates a new `PathFrankWolfeAlgorithm`.
-    ///
-    /// # Errors
-    ///
-    /// Returns `AlgorithmError::InvalidConfiguration` if the underlying
-    /// `BellmanFordAlgorithm` rejects the algorithm config.
-    pub(crate) fn new(
-        algorithm_config: AlgorithmConfig,
-        config: PathFrankWolfeConfig,
-    ) -> Result<Self, AlgorithmError> {
-        let inner = BellmanFordAlgorithm::with_config(algorithm_config)?;
-        Ok(Self { inner, config })
+    pub(crate) fn new(algorithm_config: AlgorithmConfig, config: PathFrankWolfeConfig) -> Self {
+        let inner = BellmanFordAlgorithm::with_config(algorithm_config);
+        Self { inner, config }
     }
+}
 
-    /// Creates a new `PathFrankWolfeAlgorithm` with default configs for both
-    /// algorithm and PFW tuning parameters.
-    #[allow(dead_code)]
-    pub(crate) fn new_with_defaults() -> Result<Self, AlgorithmError> {
+impl Default for PathFrankWolfeAlgorithm {
+    fn default() -> Self {
         Self::new(AlgorithmConfig::default(), PathFrankWolfeConfig::default())
-    }
-
-    /// Returns a reference to the PFW-specific tuning config.
-    #[allow(dead_code)]
-    pub(crate) fn pfw_config(&self) -> &PathFrankWolfeConfig {
-        &self.config
     }
 }
 
@@ -80,16 +67,13 @@ impl Algorithm for PathFrankWolfeAlgorithm {
 
     async fn find_best_route(
         &self,
-        graph: &Self::GraphType,
-        market: MarketData,
-        label: Option<StateLabel>,
-        derived: Option<SharedDerivedDataRef>,
-        order: &Order,
+        _graph: &Self::GraphType,
+        _market: MarketData,
+        _label: Option<StateLabel>,
+        _derived: Option<SharedDerivedDataRef>,
+        _order: &Order,
     ) -> Result<RouteResult, AlgorithmError> {
-        // Delegate to inner BF until the split-routing loop is implemented.
-        self.inner
-            .find_best_route(graph, market, label, derived, order)
-            .await
+        unimplemented!("PathFrankWolfe split-routing loop not yet implemented")
     }
 
     fn computation_requirements(&self) -> ComputationRequirements {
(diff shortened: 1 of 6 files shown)


fynd-core/src/worker_pool/registry.rs
```diff
@@ -20,15 +20,19 @@ use tokio::sync::broadcast;
...
+    },
     derived::{events::DerivedDataEvent, SharedDerivedDataRef},
     feed::{events::MarketEvent, market_data::MarketData},
     types::internal::SolveTask,
     worker_pool::worker::SolverWorker,
 };
 
 /// List of available built-in algorithm names (for registry-based dispatch).
-pub(crate) const AVAILABLE_ALGORITHMS: &[&str] = &["most_liquid", "bellman_ford"];
+pub(crate) const AVAILABLE_ALGORITHMS: &[&str] =
```
dianacarvalho1: are we sure we want to make the algorithm already available?? 😕 tbh I would prefer to add it here as a last step after everything is implemented (it might lead to some confusion from users who just get an intermediate version)

dianacarvalho1 (commented):
Thank you @tamaralipows ! 🙏🏼 I have a concern

fynd-core/src/worker_pool/registry.rs
tamaralipows: good catch!

Troshchk: Would we merge all PRs separately or to this branch?

tamaralipows: No @Troshchk, Diana meant that as soon as we add something to `AVAILABLE_ALGORITHMS`, it becomes "visible" for the users, which wouldn't be good, since it's not actually complete yet.

Troshchk: I understand -- my question was related to that. I meant: would it be okay to keep it if this branch acts as an integration branch and only gets merged into main once all the remaining pieces are implemented?

tamaralipows: Yeah in that case it would be okay, but that often causes more effort with constantly updating the branch with main and solving conflicts

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -0,0 +1,128 @@
...
+
+impl Default for PathFrankWolfeConfig {
+    fn default() -> Self {
+        Self { max_paths: 4, max_probe: 0.25, min_split: 0.05, line_search_evals: 12 }
+    }
+}
+
+/// Split-routing algorithm that discovers multiple paths via Bellman-Ford
+/// and optimises the input split across them using a Frank-Wolfe loop.
+pub struct PathFrankWolfeAlgorithm {
```
louise-poole, line 39: I wonder if it's easy enough to make this generic over the algorithm... 👀 I suppose that's a huge refactor because we're using BellmanFord specific fns like solving a single route etc... 
[no action required, just thinking out aloud]

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -0,0 +1,128 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns `AlgorithmError::InvalidConfiguration` if the underlying
+    /// `BellmanFordAlgorithm` rejects the algorithm config.
+    pub(crate) fn new(
+        algorithm_config: AlgorithmConfig,
+        config: PathFrankWolfeConfig,
+    ) -> Result<Self, AlgorithmError> {
+        let inner = BellmanFordAlgorithm::with_config(algorithm_config)?;
```
louise-poole: Lol - I just had a look at the implementation logic for `BellmanFordAlgorithm::with_config` and the Result return in wholly unnecessary - it literally only returns an Ok(), never an error. Can you update that and therefore update this constructor too? It can return Self without Result wrapper.

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -0,0 +1,128 @@
...
+        Ok(Self { inner, config })
+    }
+
+    /// Creates a new `PathFrankWolfeAlgorithm` with default configs for both
+    /// algorithm and PFW tuning parameters.
+    #[allow(dead_code)]
+    pub(crate) fn new_with_defaults() -> Result<Self, AlgorithmError> {
+        Self::new(AlgorithmConfig::default(), PathFrankWolfeConfig::default())
+    }
```
louise-poole: Following the previous comment, we can then have:
```suggestion
    impl Default for PathFrankWolfeAlgorithm {
        fn default() -> Self {
            Self::new(AlgorithmConfig::default(), PathFrankWolfeConfig::default())
        }
    }
```

louise-poole: (feel free to implement `Default` for BellmanFordAlgorithm to make this `Default` even simpler.

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    /// Returns a reference to the PFW-specific tuning config.
+    #[allow(dead_code)]
+    pub(crate) fn pfw_config(&self) -> &PathFrankWolfeConfig {
```
louise-poole: Is this fn necessary? 👀  If so, can it have a clearer name?

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -0,0 +1,128 @@
...
+        graph: &Self::GraphType,
+        market: MarketData,
+        label: Option<StateLabel>,
+        derived: Option<SharedDerivedDataRef>,
+        order: &Order,
+    ) -> Result<RouteResult, AlgorithmError> {
+        // Delegate to inner BF until the split-routing loop is implemented.
+        self.inner
+            .find_best_route(graph, market, label, derived, order)
+            .await
```
louise-poole: Strictly speaking, I'd prefer we just have `unimplemented!()` here. If we have this internal fallback it's easy to miss that we haven't actually implemented it yet. Or in the very least we need a `TODO` comment.

fynd-core/src/algorithm/path_frank_wolfe.rs
louise-poole: If it's just for testing purposes could you put it behind the test flag?

louise-poole (commented):
Thanks! A couple minor suggestions here quick

fynd-core/src/algorithm/path_frank_wolfe.rs
tamaralipows: Oh gosh good catch

fynd-core/src/algorithm/path_frank_wolfe.rs
tamaralipows: Yeah, it would need a trait like `SingleRouteAlgorithm` that exposes `build_context` + `find_single_route` - those are BF-specific today and not part of Algorithm. Doable but not worth it until we have a second algorithm that could also serve as a path finder imo. Want to avoid premature generalization for now.

louise-poole: It is indeed out of scope. I'm just thinking ahead to the Algorithm competition project - this could be a useful tool for those designing a new algorithm.

louise-poole (approved):
Looks good! 👍

PR #230, opened 2026-06-02
feat(PFW): compute_probe_amount and compute_average_price_impact

tamaralipows:
[Task](https://propeller-heads.atlassian.net/browse/ENG-5854)

**Context:** In the PFW algorithm, after we route the full trade through the best single path (via BF), we want to know: is there a second path worth splitting to? We can't just re-run BF at the full trade amount - BF would find the same path again because nothing changed. Instead, we:                                                                                   
                                                                                                                                                                                                               
  1. "Degrade" the pools on the first path (simulate their post-swap state, as if the trade already went through)                                                                                                  
  2. Run BF again with a small probe amount on all pools, including those degraded pools, in oder to discover a candidate second path. 
  
**Note:** This part of the task is not relevant yet, but we should keep in mind that when calling these methods:
  
> gas_cost_output_tokens is obtained from ctx.gas_price_wei and the output token's price from ctx.token_prices. If token price data is unavailable, fall back to total_amount *  0.10 (10% of trade) and always return Some.

### DIFF
--- a/fynd-core/src/algorithm/path_frank_wolfe.rs
+++ b/fynd-core/src/algorithm/path_frank_wolfe.rs
@@ -7,7 +7,13 @@
 
 use std::time::Duration;
 
-use super::{Algorithm, AlgorithmConfig, AlgorithmError, BellmanFordAlgorithm};
+use num_bigint::BigUint;
+use num_traits::ToPrimitive;
+
+use super::{
+    split_primitives::{split_amount, PathAllocation},
+    Algorithm, AlgorithmConfig, AlgorithmError, BellmanFordAlgorithm,
+};
 use crate::{
     derived::{computation::ComputationRequirements, SharedDerivedDataRef},
     feed::market_data::{MarketData, StateLabel},
@@ -89,6 +95,65 @@ impl Algorithm for PathFrankWolfeAlgorithm {
     }
 }
 
+/// Computes the minimum probe amount from the initial route's price impact.
+///
+/// Returns `None` when the probe exceeds `config.max_probe × total_amount`,
+/// signalling that splitting is not worthwhile.
+#[allow(dead_code)]
+fn compute_probe_amount(
+    total_amount: &BigUint,
+    price_impact: f64,
+    gas_cost_output_tokens: f64,
+    config: &PathFrankWolfeConfig,
+) -> Option<BigUint> {
+    if price_impact <= 0.0 {
+        return None;
+    }
+    let gas_floor = gas_cost_output_tokens / price_impact;
+
+    let probe = BigUint::from(gas_floor.ceil() as u128);
+    let (max_probe_amount, _remainder) = split_amount(total_amount, config.max_probe);
+    if probe > max_probe_amount {
+        return None;
+    }
+
+    Some(probe)
+}
+
+/// Flow-fraction-weighted average price impact across all active paths.
+///
+/// Per-path price impact measures how much the realized output falls short
+/// of the ideal (marginal-price) output. Paths are weighted by their share of
+/// total flow, not averaged equally — in a 95/5 split, the result is 95%
+/// determined by the big path's price impact.
+#[allow(dead_code)]
+fn compute_average_price_impact(paths: &[PathAllocation]) -> Result<f64, AlgorithmError> {
+    let mut weighted_price_impact = 0.0;
+    for path in paths {
+        let amount_in = path.amount_in.to_f64().ok_or_else(|| {
+            AlgorithmError::Other(format!("amount_in too large for f64: {}", path.amount_in))
+        })?;
+        let amount_out = path
+            .amount_out
+            .to_f64()
+            .ok_or_else(|| {
+                AlgorithmError::Other(format!("amount_out too large for f64: {}", path.amount_out))
+            })?;
+        let ideal_out = amount_in * path.marginal_price_product;
+        if ideal_out <= 0.0 {
+            return Err(AlgorithmError::Other(format!(
+                "non-positive ideal output ({ideal_out}) from amount_in={amount_in}, \
+                 marginal_price_product={}",
+                path.marginal_price_product
+            )));
+        }
+
+        let price_impact = 1.0 - amount_out / ideal_out;
+        weighted_price_impact += path.flow_fraction * price_impact;
+    }
+    Ok(weighted_price_impact)
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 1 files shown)


fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -105,6 +111,65 @@ impl Algorithm for PathFrankWolfeAlgorithm {
...
+            .ok_or_else(|| {
+                AlgorithmError::Other(format!("amount_out too large for f64: {}", path.amount_out))
+            })?;
+        let ideal_out = amount_in * path.marginal_price_product;
+        if ideal_out <= 0.0 {
+            return Err(AlgorithmError::Other(format!(
+                "non-positive ideal output ({ideal_out}) from amount_in={amount_in}, \
+                 marginal_price_product={}",
+                path.marginal_price_product
+            )));
```
tamaralipows: Please let me know if you think a better error would fit here

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -105,6 +111,65 @@ impl Algorithm for PathFrankWolfeAlgorithm {
...
+#[allow(dead_code)]
+fn compute_probe_amount(
+    total_amount: &BigUint,
+    price_impact: f64,
+    gas_cost_output_tokens: f64,
+    config: &PathFrankWolfeConfig,
+) -> Option<BigUint> {
+    if price_impact <= 0.0 {
+        return None;
+    }
```
dianacarvalho1: in which case is this possible? 🤔

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -125,4 +190,158 @@ mod tests {
...
+            PathAllocation {
+                hops: vec![],
+                flow_fraction: 0.5,
+                amount_in: BigUint::from(50_000u64),
+                amount_out: BigUint::from(95_238u64),
+                marginal_price_product: 2.0,
+            },
+        ];
+
+        let third = 1.0 / 3.0;
```
dianacarvalho1, line 292: nice variable 😂

dianacarvalho1 (approved):
Thank you @tamaralipows ! 🙏🏼 it looks good to me! only have 1 question for my understanding

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -105,6 +111,65 @@ impl Algorithm for PathFrankWolfeAlgorithm {
     }
 }
 
+/// Computes the minimum probe amount from the initial route's price impact.
+///
+/// Returns `None` when the probe exceeds `config.max_probe × total_amount`,
+/// signalling that splitting is not worthwhile.
+#[allow(dead_code)]
+fn compute_probe_amount(
```
dianacarvalho1: why isn't this a method of the Algorithm struct? same for the compute_average_price_impact

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -105,6 +111,65 @@ impl Algorithm for PathFrankWolfeAlgorithm {
...
+    price_impact: f64,
+    gas_cost_output_tokens: f64,
+    config: &PathFrankWolfeConfig,
+) -> Option<BigUint> {
+    if price_impact <= 0.0 {
+        return None;
+    }
+    let gas_floor = gas_cost_output_tokens / price_impact;
+
+    let probe = BigUint::from(gas_floor.ceil() as u128);
```
Troshchk: would it make sense to rename it to `probe_amount` for consistency with `max_probe_amount`?

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -105,6 +111,65 @@ impl Algorithm for PathFrankWolfeAlgorithm {
...
+            AlgorithmError::Other(format!("amount_in too large for f64: {}", path.amount_in))
+        })?;
+        let amount_out = path
+            .amount_out
+            .to_f64()
+            .ok_or_else(|| {
+                AlgorithmError::Other(format!("amount_out too large for f64: {}", path.amount_out))
+            })?;
+        let ideal_out = amount_in * path.marginal_price_product;
+        if ideal_out <= 0.0 {
```
Troshchk: Could you give an example when this can happen, please?

Troshchk (commented):
Thank you!
I have a general question: why do we degrade the pools on the first path, as if the full trade already went through? Why would not we use max_probe, for example?

fynd-core/src/algorithm/path_frank_wolfe.rs
tamaralipows: The pool has a flat pricing curve (e.g. stable-swap pools within their peg range). A verrryyy real example: WETH-ETH unwrapping.

fynd-core/src/algorithm/path_frank_wolfe.rs
tamaralipows: It shouldn't - that's why I return an error. This would indicate a bug in the algorithm, and we would rather error/recover gracefully and have time to investigate, than to have a panic (thanks to the division-by-zero) that crashes the whole system and renders it useless.

fynd-core/src/algorithm/path_frank_wolfe.rs
tamaralipows: I've left them free because neither function needs `&self` - they're pure computations with explicit inputs.  Keeping them as free functions also makes tests simpler: you can call them directly without constructing a full `PathFrankWolfeAlgorithm` (which in turn requires setting up `BellmanFordAlgorithm`). 

I can also see the organization benefit of moving them into the algorithm though - so please see commit [c8308f7](https://github.com/propeller-heads/fynd/pull/230/commits/c8308f702e8f102b91e8b001452a09d9efe98ba7) and let me know if you like that better!

tamaralipows:
> I have a general question: why do we degrade the pools on the first path, as if the full trade already went through? Why would not we use max_probe, for example?

@Troshchk There is more context about this in the story. This is actually a core property of the Frank-Wolfe method. In particular this line:

> Ask: "If I were already routing everything through this path, where would BF send a small probe trade?" BF sees the pool states after the big trade, so it finds a path that possibly avoids the now-depleted pools. This is the candidate second path.

Keep in mind this step doesn't determine the split amount - it only finds candidate paths. We degrade by the full amount because that's what Frank-Wolfe requires: evaluate the descent direction at the current allocation (initially 100% on path 1). If we degraded by less, BF would see a less-stressed pool and might miss better alternatives or pick the wrong one.                                                   
                                                                                                                                                          
The actual split fraction is decided afterwards by golden-section line search (step 2f), which finds the optimal balance between the existing allocation and the new candidate. So this full degradation is just "where should I look next?", not "how much should I send there?"

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -68,6 +74,69 @@ impl PathFrankWolfeAlgorithm {
...
+                    ))
+                })?;
+            let ideal_out = amount_in * path.marginal_price_product;
+            if ideal_out <= 0.0 {
+                return Err(AlgorithmError::Other(format!(
+                    "non-positive ideal output ({ideal_out}) from \
+                     amount_in={amount_in}, \
+                     marginal_price_product={}",
+                    path.marginal_price_product
+                )));
```
louise-poole: I think it's unclear what 'ideal' means to a user that just sees the error and not the code that triggered it. Maybe we can just say something like "got negative estimated amount out"?

Or - let's error on what the actual cause of this is? We'll only get a negative ideal amount if `amount_in` or `marginal_price_product` are negative. I think specifically checking and flagging those would be more useful for us 👀

louise-poole (approved):
Small error message feedback, otherwise approved! Looks good 🚀

louise-poole (approved):
Nice! Happy with this 👍

dianacarvalho1 (approved):
✨ ✨ ✨

PR #231, opened 2026-06-02
feat: correct Polygon gas token and add per-chain default RPC URLs

tvinagre:
## What

Three chain-correctness fixes, surfaced while running Fynd on Polygon with defaults.

- **Polygon gas token**: map `Chain::Polygon` to wrapped WPOL (`0x0d50..1270`) instead of the native POL placeholder (`0x..1010`). The placeholder isn't traded in any AMM pool, so token-gas-price path discovery found nothing, the price table stayed empty, and every quote fell back to gross output (gas never subtracted), logging `no gas price for output token`.
- **Per-chain default RPC**: replace the single Ethereum-only `DEFAULT_RPC_URL` constant with `default_rpc_url(chain)`. Defaults use keyless dRPC endpoints (private by default), with the official data seed for BSC. The prior llamarpc and polygon-rpc.com defaults returned 526/401 on the gas-price call.
- **Chain-aware default TVL**: source the default min-TVL from tycho-common's `Chain::default_tvl_threshold(TvlThresholdTier::Low)` instead of a single chain-agnostic constant. The old `10` native-token default meant ~10 POL (~$1) on Polygon and admitted nearly every pool; the chain-aware value targets a ~$20k floor per chain (Polygon 200k POL, BSC 32 BNB, ETH chains unchanged at 10). `--min-tvl` is now optional; when unset the chain default applies.

## Why

Running Fynd on Polygon with defaults produced empty net-of-gas values, gas-fetch failures, and a near-zero liquidity filter. All defaults now work one-click; users wanting performance can still pass `--rpc-url` / `--min-tvl`. Sourcing the TVL thresholds from tycho-common keeps Fynd's filter in lockstep with the indexer rather than reproducing the values.


zizou0x (approved):

fynd-core/src/types/constants.rs
```diff
@@ -35,10 +35,10 @@ lazy_static! {
             0xBD, 0x0D, 0x56, 0xF3, 0x52, 0x41, 0x52, 0x3F, 0xBA, 0xB1,
         ]));
 
-        // Polygon - POL (0x0000000000000000000000000000000000001010)
+        // Polygon - WPOL/WMATIC (0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270)
         map.insert(Chain::Polygon, Address::from([
-            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
-            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x10, 0x10,
+            0x0D, 0x50, 0x0B, 0x1D, 0x8E, 0x8E, 0xF3, 0x1E, 0x21, 0xC9,
+            0x9D, 0x1D, 0xB9, 0xA6, 0x44, 0x4D, 0x3A, 0xDF, 0x12, 0x70,
```
brunoguerios, line 41: Why are Addresses bytes instead of strings? Is it related to non-evm chains?

tvinagre: A couple of reasons:
  -  Hexstrings are annoying to work with, and they're essentially a human-readable bytes representation . You have to always normalize between lowercase, missing 0x, parse to make sure there are non-hex chars, so working strictly with Bytes makes it easier. Two strings for the same address can compare unequal. A Bytes blob makes no assumption about length or address format, so the same type works for chains with different address widths/encodings without a string-format contract.
  
- Performance. bytes::Bytes is reference-counted and cheap to clone (no buffer copy), and comparison/hashing over 20 bytes beats parsing/validating hex repeatedly. Address values are cloned and hashed all over the code, so the performance impact is expected to be relevant.

 - Storage efficiency. On diesel, Bytes maps to the SQL Binary type. Postgres stores 20 raw bytes instead of a ~42-char hex string — smaller rows, smaller indexes, and faster equality comparison on the columns that get joined/looked up constantly. (This was Claude's info, I didn't check if we're setting the size correctly on Tycho but the information is still relevant).

- Human-readability is preserved at the boundary. The newtype carries hex serde (#[serde(with = "hex_bytes")], hex_bytes.rs:28), with Display/Debug/LowerHex all emitting 0x… (hex_bytes.rs:45-61). So JSON/WebSocket payloads (the dto layer) still show hex strings to consumers, while in-memory and in-DB it stays binary. You get the readable wire format without paying the string cost internally.

PR #232, opened 2026-06-03
fix: apply curve_pool_filter to vm:curve exchange

mev15:
`vm:curve` was registered with no pool filter (`None`), unlike other VM-based exchanges such as `vm:balancer_v2`. This applies `curve_pool_filter`, which excludes Curve pools whose token types the simulation does not support (non-zero `asset_types` / `asset_type`).

### DIFF
--- a/fynd-core/src/feed/protocol_registry.rs
+++ b/fynd-core/src/feed/protocol_registry.rs
@@ -9,7 +9,10 @@ use tycho_simulation::{
             ekubo::state::EkuboState,
             ekubo_v3::{self, state::EkuboV3State},
             erc4626::state::ERC4626State,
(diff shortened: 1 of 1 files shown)


louise-poole:
Thanks for your contribution! However we actually applied some curve-specific improvements to tycho a little while ago and as a result the curve stream no longer needs a dedicated filter for unsupported pools. We will be deprecating and removing that curve filter in the near future. I have closed this PR for now, however if you are having ongoing problems with curve pools please feel free to reach out to us so we can fix it at the source.

louise-poole:
Quick follow up: [here](https://github.com/propeller-heads/tycho/pull/1062) is the PR that we'll merge shortly on the main tycho repo that removes this filter. Appreciate you flagging this with your above contribution 🙏  This will no doubt help avoid future users being misled by the outdated examples we had.

PR #233, opened 2026-06-03
fix: pass TLS setting to protocol stream builder

mev15:
## Problem

The `--disable-tls` flag was forwarded to token loading (`load_all_tokens`) but not to the `ProtocolStreamBuilder`. The WebSocket stream kept using `wss://` against a plaintext Tycho server and failed the TLS handshake ("wrong version number").

## Fix

Add `.no_tls(!use_tls)` to the stream builder, matching the flag already passed to token loading.


louise-poole (approved):
Nice catch, thanks for this!

PR #234, opened 2026-06-03
feat(PFW): implement find_candidate_path and is_duplicate_path 

dianacarvalho1:
Done:
- Adds two Frank-Wolfe path-discovery primitives to PathFrankWolfeAlgorithm: find_candidate_path and is_duplicate_path
- Also replaces MarketOverrides::with_zero_gas(id, sim) with with_zero_gas(id): the old form required cloning the degraded sim out of the map to pass it back in; the new form removes the entry, wraps it, and reinserts — no clone, same intent.

### DIFF
--- a/fynd-core/src/algorithm/path_frank_wolfe.rs
+++ b/fynd-core/src/algorithm/path_frank_wolfe.rs
@@ -12,7 +12,9 @@ use num_traits::ToPrimitive;
 
 use super::{
     bellman_ford::{BellmanFordContext, FindRouteOptions},
-    split_primitives::{build_post_swap_overrides, split_amount, HopDescriptor, PathAllocation},
+    split_primitives::{
+        build_post_swap_overrides, split_amount, HopDescriptor, PathAllocation, SimulatedHop,
+    },
     Algorithm, AlgorithmConfig, AlgorithmError, BellmanFordAlgorithm,
 };
 use crate::{
@@ -138,7 +140,7 @@ impl PathFrankWolfeAlgorithm {
     /// Pools already present in `current_allocations` are promoted to zero-gas so their
     /// committed gas cost is not counted again as marginal cost for the new path.
     ///
-    /// Returns an ordered sequence of [`HopDescriptor`]s representing the discovered path,
+    /// Returns an ordered sequence of [`SimulatedHop`]s representing the discovered path,
     /// or an error if no route exists.
     #[allow(dead_code)]
     pub(crate) fn find_candidate_path(
@@ -146,7 +148,7 @@ impl PathFrankWolfeAlgorithm {
         ctx: &BellmanFordContext,
         current_allocations: &[PathAllocation],
         probe_amount: &BigUint,
-    ) -> Result<Vec<HopDescriptor>, AlgorithmError> {
+    ) -> Result<Vec<SimulatedHop>, AlgorithmError> {
         let mut overrides = build_post_swap_overrides(current_allocations, &ctx.market_data);
 
         // Pools committed in the current solution are executed once on-chain — their gas is
@@ -157,9 +159,9 @@ impl PathFrankWolfeAlgorithm {
         for alloc in current_allocations {
             for hop in &alloc.hops {
                 overrides = overrides.with_zero_gas(
-                    hop.component_id.clone(),
-                    hop.token_in.address.clone(),
-                    hop.token_out.address.clone(),
+                    hop.descriptor.component_id.clone(),
+                    hop.descriptor.token_in.address.clone(),
+                    hop.descriptor.token_out.address.clone(),
                 );
             }
         }
@@ -212,8 +214,15 @@ impl PathFrankWolfeAlgorithm {
                         kind: "token",
                         id: Some(format!("{:?}", swap.token_out())),
                     })?;
-                Ok(HopDescriptor::new(swap.component_id().to_string(), token_in, token_out)
-                    .with_amounts(swap.amount_out().clone(), swap.gas_estimate().clone()))
+                Ok(SimulatedHop {
+                    descriptor: HopDescriptor::new(
+                        swap.component_id().to_string(),
+                        token_in,
+                        token_out,
+                    ),
+                    amount_out: swap.amount_out().clone(),
+                    gas: swap.gas_estimate().clone(),
+                })
             })
             .collect()
     }
@@ -229,7 +238,7 @@ impl PathFrankWolfeAlgorithm {
     /// the common segment.
     #[allow(dead_code)]
     pub(crate) fn is_duplicate_path(
-        candidate: &[HopDescriptor],
+        candidate: &[SimulatedHop],
         existing: &[PathAllocation],
     ) -> bool {
         existing.iter().any(|alloc| {
@@ -239,9 +248,9 @@ impl PathFrankWolfeAlgorithm {
                     .iter()
                     .zip(candidate.iter())
                     .all(|(a, b)| {
-                        a.component_id == b.component_id &&
-                            a.token_in.address == b.token_in.address &&
-                            a.token_out.address == b.token_out.address
+                        a.descriptor.component_id == b.descriptor.component_id &&
+                            a.descriptor.token_in.address == b.descriptor.token_in.address &&
+                            a.descriptor.token_out.address == b.descriptor.token_out.address
                     })
         })
     }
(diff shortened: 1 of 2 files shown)


fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -137,6 +138,121 @@ impl PathFrankWolfeAlgorithm {
...
+        &self,
+        ctx: &BellmanFordContext,
+        current_allocations: &[PathAllocation],
+        probe_amount: &BigUint,
+    ) -> Result<Vec<HopDescriptor>, AlgorithmError> {
+        let paths: Vec<&PathAllocation> = current_allocations.iter().collect();
+        let mut overrides = build_post_swap_overrides(&paths, &ctx.market_data);
+
+        // Pools committed in the current solution are executed once on-chain — their gas is
+        // already priced into the combined transaction. Zero out marginal gas so BF doesn't
```
louise-poole: FMI, what is meant by "marginal gas"? 👀 Can we just call this "protocol gas"?

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -137,6 +138,121 @@ impl PathFrankWolfeAlgorithm {
...
+        }
+
+        let token_in = ctx
+            .node_address
+            .get(&ctx.token_in_node)
+            .cloned()
+            .ok_or_else(|| AlgorithmError::DataNotFound {
+                kind: "token_in address",
+                id: Some(format!("{:?}", ctx.token_in_node)),
+            })?;
```
louise-poole, line 176: Since the id we're returning isn't the address we're looking for, but rather the node idx we're looking at, can we make that clearer in what we put for 'kind'? Maybe we just say 'token_in node index'?

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -137,6 +138,121 @@ impl PathFrankWolfeAlgorithm {
...
+    }
+
+    /// Returns `true` if `candidate` has the same ordered sequence of pool component IDs as any
+    /// existing allocation.
+    ///
+    /// Paths that share only a prefix but diverge at a later hop are **not** duplicates — the
+    /// shared hops are handled by `build_split_route`, which emits a single combined swap for
+    /// the common segment.
+    #[allow(dead_code)]
+    pub(crate) fn is_duplicate_path(
```
Troshchk, line 240: Here, we only compare by the component ids -- is it enough? Should we also compare the token in and token out?

Troshchk (commented):
Only one q from my side

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -137,6 +138,121 @@ impl PathFrankWolfeAlgorithm {
...
+            .ok_or_else(|| AlgorithmError::DataNotFound {
+                kind: "token_in address",
+                id: Some(format!("{:?}", ctx.token_in_node)),
+            })?;
+        let token_out = ctx
+            .node_address
+            .get(&ctx.token_out_node)
+            .cloned()
+            .ok_or_else(|| AlgorithmError::DataNotFound {
+                kind: "token_out address",
```
louise-poole: Same here.

fynd-core/src/algorithm/path_frank_wolfe.rs
louise-poole: Very good point - no it is not enough to do this by id only. We should indeed check the tokens too.

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -137,6 +138,121 @@ impl PathFrankWolfeAlgorithm {
...
+    /// or an error if no route exists.
+    #[allow(dead_code)]
+    pub(crate) fn find_candidate_path(
+        &self,
+        ctx: &BellmanFordContext,
+        current_allocations: &[PathAllocation],
+        probe_amount: &BigUint,
+    ) -> Result<Vec<HopDescriptor>, AlgorithmError> {
+        let paths: Vec<&PathAllocation> = current_allocations.iter().collect();
+        let mut overrides = build_post_swap_overrides(&paths, &ctx.market_data);
```
louise-poole: I think this is the only place we use `build_post_swap_overrides` no? if that is the case, surely we can alter `build_post_swap_overrides`'s signature to fit better the data we have here and avoid intermediate variable conversions? We could change it to accept `&[PathAllocation]` instead of `Vec<&PathAllocation>`.

louise-poole (commented):
Thanks! Looks good - mainly some small comments, plus I agree with what @Troshchk said.

fynd-core/src/algorithm/path_frank_wolfe.rs
Troshchk: "Marginal" here means the additional cost of including a pool in the route -- not just the protocol's base gas. I am not sure about renaming since the distinction matters.                             
                                                                                                                                                                                                   
I'm wondering whether zeroing gas by component_id alone is correct. A few cases to consider:                                                                                                                                                                                                                                                                                                   
1. Same pool, same token pair, across paths: merge_shared_hops merges these into one on-chain swap, so gas is paid once. Zeroing is correct.                                                  
2. Same pool, different token pairs (e.g. P1 as A->B in one path and P1 as A->C in another) -- these are not merged in merge_shared_hops, so gas would be paid twice on-chain. Is that true?
But with_zero_gas zeros by component_id alone, which would incorrectly skip the second swap's gas.                                                                                               
3. Same pool, same tokens, different hops in two paths (e.g. P1 at layer 1 in path A and layer 2 in path B) — merge_shared_hops keys on (component_id, token_in, token_out) regardless of position, so these merge into one swap. Is this correct?

Troshchk: after discussing with @tamaralipows 
Case 2: not correct, should be zeroed based on component id and tokens
Case 3: for now leaving it as todo, @tamaralipows will be looking into this

tamaralipows: Case 3: After my latest PR - this must also be zeroed!

louise-poole: Nicely spotted with those edge cases! @Troshchk 

> Case 2: not correct, should be zeroed based on component id and tokens
Yes - case 2 should not be zeroed

louise-poole: > "Marginal" here means the additional cost of including a pool in the route

At this point in our solver only the protocol specific gas is considered - we do not take token transfers or router overhead etc into account yet. If you have a look at what `with_zero_gas` does it merely zeros out this protocol gas we estimate in `get_amount_out`... this I think I recall we usually refer to as 'protocol gas' in our codebase? I could be wrong though... Whatever term we use to refer to this gas is the term I'd prefer we use here (for consistency and general codebase readability).

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -157,38 +157,37 @@ impl PathFrankWolfeAlgorithm {
...
+            }
         }
 
         let token_in = ctx
             .node_address
             .get(&ctx.token_in_node)
             .cloned()
             .ok_or_else(|| AlgorithmError::DataNotFound {
-                kind: "token_in address",
+                kind: "token_in node address",
```
louise-poole: Strictly speaking it's a node index we have here and failed to find in the map keys (`ctx.token_in_node`), not the node address (what we were trying to look up in the map, but node index is missing).
As I wrote that sentence I could feel how confusing it was 😅 Basically the `id` we're including in the error message is an index, not an address. So I'd prefer if the 'kind' wording reflected that (not even mention address).

louise-poole (approved):
A veeeery tiny comment from me, otherwise approved!

fynd-core/src/algorithm/path_frank_wolfe.rs
Troshchk: @tamaralipows Thank you -- case 3 stays zeroed than, no change needed here anymore
@louise-poole okok, let me check the naming in the codebase

Troshchk: You are right, I have renamed it

fynd-core/src/algorithm/path_frank_wolfe.rs
Troshchk: okok, fixed

tamaralipows (approved):
Thank you! Looks good!

PR #235, opened 2026-06-03
refactor: split HopDescriptor into descriptor and SimulatedHop

tamaralipows:
Separates simulation results (amount_out, gas) from the hop topology (component_id, token_in, token_out). This makes it clear at the type level which functions need simulation data and which only need routing descriptors, and eliminates the Option unwrapping that was needed when both lived on the same struct.

### DIFF
--- a/clients/openapi.json
+++ b/clients/openapi.json
@@ -7,7 +7,7 @@
       "name": "MIT",
       "identifier": "MIT"
     },
-    "version": "0.59.0"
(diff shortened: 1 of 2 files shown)


louise-poole (approved):
YESSS!! A thing of beauty 😍

dianacarvalho1 (approved):
🧹 🧹 💅🏼 💅🏼 thank you @tamaralipows!

PR #236, opened 2026-06-03
refactor(split routing): rename group to branch_collection

tamaralipows:
Avoids confusion with the "grouped swaps" concept in tycho-execution.


Troshchk (approved):
Thank you @tamaralipows!
I will approve now, so you can merge whenever the CIs will be passing

PR #238, opened 2026-06-04
fix: clear stale edge weights when spot price is unavailable

louise-poole:
When spot price computation fails, `update_edge_weights_with_derived` was skipping those edges, leaving stale `DepthAndPrice` data in place. The pool stayed selectable for routing with an outdated price, causing on-chain reverts.

Set `edge_data.data = None` for edges where `from_sim_and_derived` returns `None` instead of skipping them.


Troshchk (approved):
Thank you!

tamaralipows (approved):
Thank youuu

PR #239, opened 2026-06-04
fix(split routing): topological order for cross-depth hops 

tamaralipows:
## Problem, summarized:
The original two paths:
```
WETH ───▶ USDC ─── pool_a ──▶ DAI                    (shorter path)
WETH ───▶ USDT ────────▶ USDC ── pool_a ──▶ DAI      (longer path)
```

The merged paths:
```
WETH ──┬────────────▶ USDC ─── pool_a ──▶ DAI
       │                ▲
       └────▶ USDT ─────┘
                                       
```
Our merging algo was emitting the USDC  ─ pool_a ─▶ DAI before emitting the  USDT ─▶ USDC swap. This would cause unintended consequences on our router side, since only the amount for WETH ─▶ USDC would be inputted into the final USDC ─▶ DAI swap, ignoring the USDT ─▶ USDC amount.

Related addition to the Tycho docs: https://github.com/propeller-heads/tycho/pull/1068

## Solution
Replace naive BFS in build_split_route with [Kahn's topological sort.](https://www.geeksforgeeks.org/dsa/topological-sorting-indegree-based-solution/)

## Summary of Kahn's topological sort:
1. Count incoming edges for every node (the "in-degree")                                                                           
2. Enqueue all nodes with in-degree 0 — they have no dependencies                
3. Loop: pop a node from the queue, "emit" it, and decrement the in-degree of each of its neighbors. Whenever a neighbor's in-degree reaches 0, enqueue it                                                                                                    
4. Done when the queue is empty. If any nodes were never enqueued, the graph has a dependency cycle (A → B → A → C) — distinct from a valid trade cycle (A → B → A) where the trade starts and ends at the same token     

## Consequence
The USDC ─── pool_a ──▶ DAI will now be **one swap** meaning we should **only count gas once**. The swap will be emitted only after the inflow swaps are emitted. Our router is happy.

## Important Note: Unsupported edge case
The following case is still not supported, because Pool A and Pool B cause a deadlock from being swapped in opposite orders in different branches:
```
WETH ─┬─ USDC ─ (Pool A) ─ DAI ─ PEPE ─ (Pool B) ─ UNI ─ WBTC
      └─ PEPE ─ (Pool B) ─ UNI ─ USDC ─ (Pool A) ─ DAI ─ WBTC
```
This is a pretty rare case and any solutions I found resulted in a massive refactor. For now, I've added a test that ensures this results in an error. Let me know if you agree or if you think this is worth spending time to support.

### DIFF
--- a/fynd-core/src/algorithm/split_primitives.rs
+++ b/fynd-core/src/algorithm/split_primitives.rs
@@ -621,8 +621,8 @@ fn assign_splits_and_amounts(
 /// intermediate token:
 ///
 /// ```text
-/// Path 1 (2 hops): WETH -> USDC -> DAI          (via Pool A)
-/// Path 2 (3 hops): WETH -> USDT -> USDC -> DAI  (via Pool A)
+/// Path 1 (2 hops): WETH -> USDC -(pool A)-> DAI
+/// Path 2 (3 hops): WETH -> USDT -> USDC -(pool A)-> DAI
 /// ```
 ///
 /// Pool A (USDC→DAI) is merged into one swap. If USDC were visited before
--- a/fynd-core/src/algorithm/split_primitives.rs
+++ b/fynd-core/src/algorithm/split_primitives.rs
@@ -18,6 +18,7 @@ use crate::{
     types::{ComponentId, Order, Route, Swap},
 };
 
+#[derive(Clone)]
 pub(crate) struct HopDescriptor {
     pub(crate) component_id: ComponentId,
     pub(crate) token_in: Token,
@@ -44,6 +45,7 @@ impl HopDescriptor {
 ///
 /// One path in the current split solution, with the fraction of total `amount_in`
 /// currently allocated to it. All fractions across allocations sum to 1.0.
+#[derive(Clone)]
 pub(crate) struct PathAllocation {
     pub(crate) hops: Vec<HopDescriptor>,
     /// Fraction of total input on this path (0 < f <= 1).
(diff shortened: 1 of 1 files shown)


fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -566,31 +566,82 @@ fn assign_splits_and_amounts(
...
+/// before emitting Pool A's swap. This extends to downstream splits too:
+///
+/// ```text
+/// Path 1: WETH -> USDC -> DAI (Pool A) -> PEPE (Pool B)  (0.5)
+/// Path 2: WETH -> USDC -> DAI (Pool A) -> PEPE (Pool C)  (0.5)
+/// Path 3: WETH -> USDT -> USDC -> DAI (Pool A) -> PEPE (Pool B or C)
+/// ```
+///
+/// The DAI→PEPE split between Pool B and Pool C must wait until all DAI
+/// has been produced (from both paths through the merged Pool A swap).
```
tamaralipows, line 640: Docstring getting too long. Considering removing this example.

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -566,31 +566,82 @@ fn assign_splits_and_amounts(
...
 pub(crate) fn build_split_route(
     paths: &[PathAllocation],
```
tamaralipows, line 642: It's okay that we are already taking the deduplicated (after `merged_shared_hops`) paths here, because in-degree isn't tracking how many times Pool A appears - it's tracking how many distinct swaps produce a given token. We still have this information with the given data.

Troshchk (approved):
Thank you! Looks good to me

fynd-core/src/algorithm/split_primitives.rs
Troshchk: I also think that the example above should be enough for understanding, but leaving it up to you

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -596,6 +598,46 @@ fn assign_splits_and_amounts(
...
+/// Note: the TychoRouter contract *could* support interleaved splits
+/// (partial consume, more inflows, consume rest), but that would require
+/// an extra swap on the same pool, spending more gas.
+///
+/// For example, given paths of different lengths that converge on the same
+/// intermediate token:
+///
+/// ```text
+/// Path 1 (2 hops): WETH -> USDC -> DAI          (via Pool A)
+/// Path 2 (3 hops): WETH -> USDT -> USDC -> DAI  (via Pool A)
```
dianacarvalho1:
```suggestion
/// Path 1 (2 hops): WETH -> USDC -(pool A)-> DAI
/// Path 2 (3 hops): WETH -> USDT -> USDC -(pool A)-> DAI
```

dianacarvalho1 (approved):
Thank you @tamaralipows 🙏🏼 This is extremely clear and simple! Great job, had a good time reviewing this! 🐬

PR #242, opened 2026-06-05
fix(split routing): validate paths before merge_shared_hops

dianacarvalho1:
## Problem:
`merge_shared_hops` keys on `(pool, token_in, token_out)` — if a path revisits a token `(e.g. USDC → WETH → USDT → USDC → WETH)`, it silently merges the two `USDC→WETH` hops into one, corrupting the route

## Fix:
Added `PathAllocation::validate()` with cycle detection, called before `merge_shared_hops` to catch invalid paths before they're corrupted

## Why not move all validation before merge: 
Split fractions, BFS connectivity, and dead-end checks operate on the assembled Swap objects — those fractions are computed during `build_split_route` and don't exist on the raw `PathAllocation` inputs

## Duplicate cycle detection: 
`validate_cycles` still runs on the assembled Route as a safety net — the extra compute is negligible (one pass over 3-6 swaps with hash lookups). We can also choose to remove this - let me know if you'd prefer that.

Original PR from @tamaralipows: https://github.com/propeller-heads/fynd/pull/240

### DIFF
--- a/fynd-core/src/algorithm/split_primitives.rs
+++ b/fynd-core/src/algorithm/split_primitives.rs
@@ -61,9 +61,9 @@ impl PathAllocation {
     /// A token appearing more than once means `merge_shared_hops` would
     /// incorrectly collapse distinct hops into one. The only exception is
     /// a round-trip where the final output equals the first input.
-    pub(crate) fn validate(&self) -> Result<(), AlgorithmError> {
+    pub(crate) fn validate_token_cycles(&self) -> Result<(), AlgorithmError> {
         if self.hops.is_empty() {
-            return Ok(());
+            return Err(AlgorithmError::Other("path has no hops".to_string()));
         }
         let first_token = &self.hops[0].token_in.address;
         let mut seen = HashSet::new();
@@ -602,7 +602,7 @@ pub(crate) fn build_split_route(
     order: &Order,
 ) -> Result<Route, AlgorithmError> {
     for path in paths {
-        path.validate()?;
+        path.validate_token_cycles()?;
     }
     let mut hops_by_token = merge_shared_hops(paths)?;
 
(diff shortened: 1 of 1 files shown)


fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -750,6 +783,104 @@ mod tests {
...
+    #[test]
+    fn test_path_allocation_validate_empty_hops() {
+        let path = PathAllocation {
+            hops: vec![],
+            flow_fraction: 1.0,
+            amount_in: BigUint::from(100u64),
+            amount_out: BigUint::from(100u64),
+            marginal_price_product: 1.0,
+        };
+        assert!(path.validate().is_ok());
```
dianacarvalho1: why is an empty path valid? 🤔

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -55,6 +55,36 @@ pub(crate) struct PathAllocation {
     pub(crate) marginal_price_product: f64,
 }
 
+impl PathAllocation {
+    /// Validates that this path does not revisit any token.
+    ///
+    /// A token appearing more than once means `merge_shared_hops` would
+    /// incorrectly collapse distinct hops into one. The only exception is
+    /// a round-trip where the final output equals the first input.
+    pub(crate) fn validate(&self) -> Result<(), AlgorithmError> {
```
dianacarvalho1: I would suggest renaming this to be more explicit: it does not perform full validation, it only checks for forbidden cycles. Maybe `validate_token_cycles`, `validate_token_revisits`?

dianacarvalho1 (commented):
Thank you @tamaralipows 🙏🏼 I agree with your reasoning! Only have small comments

louise-poole (approved):
Looks good! Agreed with Diana's comments, otherwise I've approved.

Troshchk (approved):
Thank you @tamaralipows! 
Agree with the earlier comments, otherwise looks good to me 💫 
Also thank you for addressing it so quickly

PR #245, opened 2026-06-08
feat: implement optimize_step_size, apply_step, and the find_best_route main loop

Troshchk:
- Implement the Path Frank-Wolfe main loop in find_best_route: iteratively discovers parallel paths via BF probes, optimizes split fractions with golden-section line search, and re-simulates allocations each step
- Early exits on low price impact, duplicate paths, small step size, or timeout
- returns the split route only if it beats the initial single-path result

### DIFF
--- a/fynd-core/src/algorithm/path_frank_wolfe.rs
+++ b/fynd-core/src/algorithm/path_frank_wolfe.rs
@@ -226,26 +226,29 @@ impl PathFrankWolfeAlgorithm {
     }
 
     /// Computes the gas cost of a route in output-token units as `f64`.
-    fn gas_cost_output_tokens(route: &Route, ctx: &BellmanFordContext) -> f64 {
+    fn gas_cost_output_tokens(
+        route: &Route,
+        ctx: &BellmanFordContext,
+    ) -> Result<f64, AlgorithmError> {
         let gas_price = match &ctx.gas_price_wei {
             Some(gp) if !gp.is_zero() => gp,
-            _ => return 0.0,
+            _ => return Ok(0.0),
         };
-        let last_swap = match route.swaps().last() {
-            Some(s) => s,
-            None => return 0.0,
-        };
+        let last_swap = route
+            .swaps()
+            .last()
+            .ok_or_else(|| AlgorithmError::Other("route has no swaps".to_string()))?;
         let price = match ctx
             .token_prices
             .as_ref()
             .and_then(|tp| tp.get(last_swap.token_out()))
         {
             Some(p) if !p.denominator.is_zero() => p,
-            _ => return 0.0,
+            _ => return Ok(0.0),
         };
         let gas_cost_wei = &route.total_gas() * gas_price;
         let gas_cost_tokens = &gas_cost_wei * &price.numerator / &price.denominator;
-        gas_cost_tokens.to_f64().unwrap_or(0.0)
+        Ok(gas_cost_tokens.to_f64().unwrap_or(0.0))
     }
 
     /// Converts a `Route` (from BF's initial solve) into a single `PathAllocation`.
@@ -416,10 +419,14 @@ impl PathFrankWolfeAlgorithm {
 
     /// Computes `net_amount_out` for a split route, mirroring
     /// `BellmanFordAlgorithm::compute_net_amount_out`.
-    fn compute_split_net_amount_out(route: &Route, ctx: &BellmanFordContext) -> BigInt {
-        let Some(last_swap) = route.swaps().last() else {
-            return BigInt::zero();
-        };
+    fn compute_split_net_amount_out(
+        route: &Route,
+        ctx: &BellmanFordContext,
+    ) -> Result<BigInt, AlgorithmError> {
+        let last_swap = route
+            .swaps()
+            .last()
+            .ok_or_else(|| AlgorithmError::Other("route has no swaps".to_string()))?;
         let output_token = last_swap.token_out();
         let total_out: BigUint = route
             .swaps()
@@ -428,9 +435,9 @@ impl PathFrankWolfeAlgorithm {
             .map(|s| s.amount_out().clone())
             .fold(BigUint::zero(), |acc, x| acc + x);
 
-        let gas_cost = Self::gas_cost_output_tokens(route, ctx);
+        let gas_cost = Self::gas_cost_output_tokens(route, ctx)?;
         let gas_cost_tokens = BigUint::from(gas_cost.ceil() as u128);
-        BigInt::from(total_out) - BigInt::from(gas_cost_tokens)
+        Ok(BigInt::from(total_out) - BigInt::from(gas_cost_tokens))
     }
 
     /// Returns `true` if `candidate` has the same ordered sequence of
@@ -492,7 +499,7 @@ impl Algorithm for PathFrankWolfeAlgorithm {
             vec![Self::route_to_allocation(single_path_result.route(), order, &ctx)?];
 
         // Compute gas cost and initial probe.
-        let gas_cost = Self::gas_cost_output_tokens(single_path_result.route(), &ctx);
+        let gas_cost = Self::gas_cost_output_tokens(single_path_result.route(), &ctx)?;
         let total_amount = order.amount();
         let initial_pi = Self::compute_average_price_impact(&allocations)?;
         if self
@@ -560,7 +567,7 @@ impl Algorithm for PathFrankWolfeAlgorithm {
             .gas_price_wei
             .clone()
             .unwrap_or_default();
-        let split_net = Self::compute_split_net_amount_out(&split_route, &ctx);
+        let split_net = Self::compute_split_net_amount_out(&split_route, &ctx)?;
         let split_result = RouteResult::new(split_route, split_net, gas_price);
 
         if split_result.net_amount_out() > single_path_result.net_amount_out() {
--- a/fynd-core/src/algorithm/path_frank_wolfe.rs
+++ b/fynd-core/src/algorithm/path_frank_wolfe.rs
@@ -288,6 +288,13 @@ impl PathFrankWolfeAlgorithm {
             })
             .collect::<Result<_, AlgorithmError>>()?;
 
+        if hops.is_empty() {
+            return Err(AlgorithmError::DataNotFound {
+                kind: "swap",
+                id: Some("route contains no swaps".to_string()),
+            });
+        }
+
         let descriptors: Vec<HopDescriptor> = hops
             .iter()
             .map(|h| h.descriptor.clone())
--- a/fynd-core/src/algorithm/path_frank_wolfe.rs
+++ b/fynd-core/src/algorithm/path_frank_wolfe.rs
@@ -836,7 +836,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_pi_exit_criterion_stops_loop_early() {
+    async fn test_pi_exit_criterion_with_high_gas() {
         // Three distinct pools so BF can always find a new (non-duplicate) path.
         // High gas costs relative to trade size mean that after the first split
         // lowers PI, `compute_probe_amount` returns None before iteration 2 can
@@ -1328,7 +1328,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_split_beats_single_route() {
+    async fn test_split_vs_single_route() {
         // Large trade through two parallel pools should produce more output than
         // routing everything through just one.
         let token_a = token(0x01, "A");
@@ -1529,7 +1529,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_timeout_returns_partial_result() {
+    async fn test_timeout_mid_iteration() {
         // With a generous timeout the algo splits across all pools.
         // With a near-zero timeout it returns fewer paths, proving the FW loop
         // was cut short while still producing a valid result.
--- a/fynd-core/src/algorithm/path_frank_wolfe.rs
+++ b/fynd-core/src/algorithm/path_frank_wolfe.rs
@@ -503,7 +503,7 @@ impl Algorithm for PathFrankWolfeAlgorithm {
             return Ok(single_path_result);
         }
 
-        // Frank-Wolfe loop — discover up to max_paths - 1 additional paths.
+        // Step 2: Frank-Wolfe loop — discover up to max_paths - 1 additional paths.
         for iteration in 1..self.config.max_paths {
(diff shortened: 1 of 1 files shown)


fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -227,6 +225,231 @@ impl PathFrankWolfeAlgorithm {
...
 
+    /// Computes the gas cost of a route in output-token units as `f64`.
+    ///
+    /// Returns `0.0` when gas price or token prices are unavailable
+    fn gas_cost_output_tokens(route: &crate::types::Route, ctx: &BellmanFordContext) -> f64 {
+        let gas_price = match &ctx.gas_price_wei {
+            Some(gp) if !gp.is_zero() => gp,
+            _ => return 0.0,
+        };
+        let last_swap = match route.swaps().last() {
```
tamaralipows: We should probably error since this is different than just a missing gas price?

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -227,6 +225,231 @@ impl PathFrankWolfeAlgorithm {
...
+            _ => return 0.0,
+        };
+        let gas_cost_wei = &route.total_gas() * gas_price;
+        let gas_cost_tokens = &gas_cost_wei * &price.numerator / &price.denominator;
+        gas_cost_tokens.to_f64().unwrap_or(0.0)
+    }
+
+    /// Converts a `Route` (from BF's initial solve) into a single `PathAllocation`.
+    fn route_to_allocation(
+        route: &crate::types::Route,
```
tamaralipows: Can we import Route at the top instead of accessing like this

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -227,6 +225,231 @@ impl PathFrankWolfeAlgorithm {
...
+                    .collect()
+            })
+            .collect();
+        let candidate_descriptors: Vec<HopDescriptor> = candidate
+            .iter()
+            .map(|h| h.descriptor.clone())
+            .collect();
+        let overrides = MarketOverrides::empty();
+
+        let evaluate_split = |gamma: f64| -> f64 {
```
tamaralipows: Maybe let's make this a separate fn instead of defining it inline?

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -227,6 +225,231 @@ impl PathFrankWolfeAlgorithm {
...
+            amount_in: order.amount().clone(),
+            amount_out,
+            marginal_price_product,
+        })
+    }
+
+    /// Golden-section search over the split fraction `gamma ∈ [0, 1]`.
+    ///
+    /// At each probe point, builds trial fractions (existing paths scaled by
+    /// `1 − gamma`, candidate at `gamma`) and evaluates the combined output via
```
tamaralipows: Not a big fan of the "gamma" param name - is there a better way to describe what's being done here? it's the split amount right?

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -227,6 +225,231 @@ impl PathFrankWolfeAlgorithm {
...
+                Ok((total_output, _gas)) => total_output.to_f64().unwrap_or(0.0),
+                Err(_) => 0.0,
+            }
+        };
+
+        golden_section_search(evaluate_split, 0.0, 1.0, self.config.line_search_evals)
+    }
+
+    /// Applies a Frank-Wolfe step: shifts `gamma` fraction of flow to the
+    /// candidate path, re-simulates all paths, and prunes negligible allocations.
```
tamaralipows: What exactly does negligible allocations mean - can we clarify the docstring a bit

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -495,10 +835,80 @@ mod tests {
         assert!((pi - 0.14).abs() < 1e-10, "expected 0.14, got {pi}");
     }
 
-    // TODO(ENG-5856): requires the main loop to verify early exit behavior.
-    #[test]
-    #[ignore]
-    fn test_pi_exit_criterion_stops_loop_early() {}
+    #[tokio::test]
+    async fn test_pi_exit_criterion_stops_loop_early() {
```
tamaralipows: I know this test is here from before but let's rename it to not include the outcome (same with the other added tests)

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -758,4 +1168,430 @@ mod tests {
...
+        assert_eq!(ids.len(), 3, "expected 3 paths, got {ids:?}");
+        assert!(ids.contains(&"P1"), "missing P1");
+        assert!(ids.contains(&"P2"), "missing P2");
+        assert!(ids.contains(&"P3"), "missing P3");
+    }
+
+    #[tokio::test]
+    async fn test_shared_pool_degradation() {
+        // Two routes share an interior pool: A→B via P1, then B→C via P_shared.
+        // A second route: A→B via P2, then B→C via P_shared.
```
tamaralipows: Diagrams for the tests? :)

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -272,10 +494,104 @@ impl Algorithm for PathFrankWolfeAlgorithm {
...
+                None => {
+                    debug!(iteration, pi, "probe exceeds cap, stopping");
+                    break;
+                }
+            };
+
+            let candidate = match self.find_candidate_path(&ctx, &allocations, &probe_amount) {
+                Ok(c) => c,
+                Err(e) => {
+                    debug!(iteration, ?e, "no candidate path found, stopping");
```
tamaralipows: Hmmm are we sure we don't want to return the error in this case? Like it seems the `AlgorithmError::DataNotFound` would just be ignored - is that okay? Maybe its fine as long as we find another candidate? 

@louise-poole do you know?

tamaralipows (commented):
Thank you! I think the logic looks good - only comments about readability :)

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -227,6 +225,231 @@ impl PathFrankWolfeAlgorithm {
...
+            .token_prices
+            .as_ref()
+            .and_then(|tp| tp.get(output_token))
+        {
+            Some(p) if !p.denominator.is_zero() => p,
+            _ => return BigInt::from(total_out),
+        };
+
+        let gas_cost_wei = &route.total_gas() * gas_price;
+        let gas_cost_tokens = &gas_cost_wei * &price.numerator / &price.denominator;
```
louise-poole: This looks like duplicate logic from `gas_cost_output_tokens` no? Can we not use that fn here instead of reimplementing this?

fynd-core/src/algorithm/path_frank_wolfe.rs
louise-poole: Agreed - a route with no swaps should error (even if we think the callsite will never pass such a route)

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -227,6 +225,231 @@ impl PathFrankWolfeAlgorithm {
...
+    fn route_to_allocation(
+        route: &crate::types::Route,
+        order: &Order,
+        ctx: &BellmanFordContext,
+    ) -> Result<PathAllocation, AlgorithmError> {
+        let tokens = route.tokens();
+        let hops: Vec<SimulatedHop> = route
+            .swaps()
+            .iter()
+            .map(|swap| {
```
louise-poole, line 264: This would create an empty vec if the route has no swaps... we should error in that case please.

fynd-core/src/algorithm/path_frank_wolfe.rs
louise-poole: Personally I don't mind this in-line fn here. Makes me think back to functional programming days 😂  Play around a bit and see what you think looks neatest (readability is important and unfortunately this style is not natural to object-oriented programming).

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -227,6 +225,231 @@ impl PathFrankWolfeAlgorithm {
...
+        let candidate_descriptors: Vec<HopDescriptor> = candidate
+            .iter()
+            .map(|h| h.descriptor.clone())
+            .collect();
+        let overrides = MarketOverrides::empty();
+
+        let evaluate_split = |gamma: f64| -> f64 {
+            let mut trial_fractions: Vec<f64> = current_allocations
+                .iter()
+                .map(|a| a.flow_fraction * (1.0 - gamma))
```
louise-poole: Also 'gamma' is very generic - can we call it `split` or `split_fraction` or something to add more context?

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -227,6 +225,231 @@ impl PathFrankWolfeAlgorithm {
...
+        golden_section_search(evaluate_split, 0.0, 1.0, self.config.line_search_evals)
+    }
+
+    /// Applies a Frank-Wolfe step: shifts `gamma` fraction of flow to the
+    /// candidate path, re-simulates all paths, and prunes negligible allocations.
+    fn apply_step(
+        &self,
+        allocations: &mut Vec<PathAllocation>,
+        candidate: &[SimulatedHop],
+        gamma: f64,
```
louise-poole: I think in the plan this was called `split`... is there a reason for the rename?

fynd-core/src/algorithm/path_frank_wolfe.rs
louise-poole: No need for an error here. If we reach here it means an initial path was already found and no new/competing route can be found in addition to it. This is a valid scenario and means we should simply stop looking and work with what was already found.

Maybe we can make it clearer by making the debug message "no additional candidate path found, stopping further searches"? Or "all candidate paths already found, stopping".

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -272,10 +494,104 @@ impl Algorithm for PathFrankWolfeAlgorithm {
...
+        if self
+            .compute_probe_amount(total_amount, initial_pi, gas_cost)
+            .is_none()
+        {
+            debug!(pi = initial_pi, gas_cost, "price impact too low to justify splitting");
+            return Ok(single_path_result);
+        }
+
+        // Frank-Wolfe loop — discover up to max_paths - 1 additional paths.
+        for iteration in 1..self.config.max_paths {
```
louise-poole: This is "Step 2" no? Searching for additional candidate paths? In your fn comments you have a "Step 1" and "Step 3" but no "Step 2" comment....

louise-poole (commented):
The logic looks great! Just some minor comments, otherwise I'm very happy with this 👌

fynd-core/src/algorithm/path_frank_wolfe.rs
Troshchk: I was wondering -- is Route with empty swaps valid in any case? Should be may be ensure, that Route contains swap on the level of the struct instead of explicitly checking it?

fynd-core/src/algorithm/path_frank_wolfe.rs
Troshchk: It's not the amount, but a fraction of the total amount to move to the candidate path.
I have renamed it to `step_size`, lmk what you think

fynd-core/src/algorithm/path_frank_wolfe.rs
louise-poole: I fully agree with that, good thinking 👍

fynd-core/src/algorithm/path_frank_wolfe.rs
Troshchk: I have experimented with it a bit, but this implementation seems better to me
I kept it inline because extracting it would mean passing 5 captured variables as explicit parameters to a function. 
Also, I think, the evaluate logic and the golden_section_search call belong together -- splitting them forces you to scroll between two places to understand a single operation. 
On top of that, we'd still need a closure(or another thin wrapper) since `golden_section_search` takes i`mpl Fn(f64) -> f64` as an input.
Lmk what you think

fynd-core/src/algorithm/path_frank_wolfe.rs
Troshchk: true, removed

fynd-core/src/algorithm/path_frank_wolfe.rs
Troshchk: I have renamed to the `step_size` to be inline with the fn name, wdyt?

fynd-core/src/algorithm/path_frank_wolfe.rs
Troshchk: I think split sounded to generic for me, but I agree gamma is no better.
I have renamed to `step_size`

fynd-core/src/algorithm/path_frank_wolfe.rs
tamaralipows: ok ok I accept! Thanks for checking!

fynd-core/src/algorithm/path_frank_wolfe.rs
Troshchk: Fixed, we will return errors now

fynd-core/src/algorithm/path_frank_wolfe.rs
tamaralipows: I'm a bit confused because the function is called `evaluate_split` but takes `step_size` as input. So are we actually evaluating the `split_step_size` and not the split itself?

louise-poole: I take it we evaluate the full split after applying the incremental change (step). So yeah, that would be `split_step_size`. Could either add a comment to clarify this or update the var name.

louise-poole (approved):
Thanks! Looks good! Excited to see this finally all come together and tested 🚀

fynd-core/src/algorithm/path_frank_wolfe.rs
tamaralipows: I think a comment would be good enough here!

tamaralipows (approved):
Thank you so much! Just the little comment and all is good :))

PR #246, opened 2026-06-09
feat: wire module declarations and registry entry

Troshchk:
- Added "path_frank_wolfe" to AVAILABLE_ALGORITHMS                                                                                   
- Added test_registry_has_path_frank_wolfe to verify the name is in AVAILABLE_ALGORITHMS
- Added test_registry_spawns_path_frank_wolfe to verify the registry spawner creates workers without panicking

### DIFF
--- a/fynd-core/src/worker_pool/registry.rs
+++ b/fynd-core/src/worker_pool/registry.rs
@@ -354,11 +354,6 @@ mod tests {
     }
 
     #[test]
-    fn test_registry_has_path_frank_wolfe() {
-        assert!(AVAILABLE_ALGORITHMS.contains(&"path_frank_wolfe"));
-    }
-
-    #[test]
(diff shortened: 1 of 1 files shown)


fynd-core/src/worker_pool/registry.rs
```diff
@@ -351,4 +352,39 @@ mod tests {
 
         let _ = shutdown_tx.send(());
     }
+
+    #[test]
+    fn test_registry_has_path_frank_wolfe() {
+        assert!(AVAILABLE_ALGORITHMS.contains(&"path_frank_wolfe"));
+    }
```
tamaralipows: I know it was in the plan but I don't think this is necessary

tamaralipows (approved):
Thank you!

louise-poole (approved):
Thanks!

PR #247, opened 2026-06-10
feat: add e2e tests

Troshchk:
- Add end-to-end tests for PathFrankWolfeAlgorithm using the existing harness 
- Add e2e helper functions to validate route and calldata
- Remove `#[allow(dead_code)]` from items which are now used
- Fix identified bug in apply_step():
Previously, only the total path amount_out was updated after finding the optimal split -- per hop amount_out values remained stale from initial discovery. Now each hop is updated to match the final allocation.


fynd-core/src/algorithm/pfw_e2e_tests.rs
louise-poole, line 1: Eish - I see a huge oversight in the plan for this task: it should have pointed you to use the test harness set up earlier in the project (see [here](https://github.com/propeller-heads/fynd/blob/main/fynd-core/src/algorithm/split_test_harness.rs#L170)). There are already a bunch of scenarios ([here](https://github.com/propeller-heads/fynd/blob/main/fynd-core/src/algorithm/split_test_harness.rs#L236)) that have been created so you shouldn't need to design/create your own.

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -683,4 +699,67 @@ mod tests {
...
     }
+
+    // ==================== PFW scenario tests ====================
+
+    #[tokio::test]
+    async fn pfw_symmetric_split() {
+        let scenario = split_scenarios::symmetric_split();
+        let (market, gm) = scenario.build_market();
+        let result = evaluate_scenario(&pfw_default(), &scenario, market, gm).await;
+        result.assert_beats_lower_bound();
```
tamaralipows: I notice in the task that it says to assert that `route.validate() passes`
Is there a reason it's not done here - or is it included in one of these other assertions?

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -683,4 +699,67 @@ mod tests {
...
+        assert_eq!(result.path_count, 2, "expected 2 paths");
+        assert!(result.within_pct_of_optimum(5), "symmetric split should be within 5% of optimum");
```
tamaralipows: It also says to assert input and output tokens, I guess that is also important?

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -683,4 +699,67 @@ mod tests {
...
+    #[tokio::test]
+    async fn pfw_multi_hop_bottleneck() {
+        let scenario = split_scenarios::multi_hop_bottleneck();
+        let (market, gm) = scenario.build_market();
+        let result = evaluate_scenario(&pfw_default(), &scenario, market, gm).await;
+        result.assert_beats_lower_bound();
+    }
+
+    #[tokio::test]
+    async fn pfw_double_split() {
```
tamaralipows: So if I understand correctly, this corresponds to `test_split_vs_single_route_comparison` in the task? Because the task says

> results appear in test output for manual review.

Should we add a print here so we can easily see it in CI? Not sure how important it is.

tamaralipows (commented):
Thanks @Troshchk, I think this looks correct. I'm only a bit confused about which parts of the task are staying the same and which parts of the task we are consciously deciding are not important. Maybe I misunderstood something

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -683,4 +699,67 @@ mod tests {
...
+    }
+
+    #[tokio::test]
+    async fn pfw_no_alternative_path() {
+        let scenario = split_scenarios::no_alternative_path();
+        let (market, gm) = scenario.build_market();
+        let result = evaluate_scenario(&pfw_default(), &scenario, market, gm).await;
+        result.assert_meets_lower_bound();
+        assert_eq!(result.path_count, 1, "single-pool scenario should produce exactly 1 path",);
+    }
```
louise-poole: There are a lot of similarities between these tests... I think we could design this better so that we can just have 1 test that loops through `split_scenarios::all()`. We can probably adjust the ScenarioResult's assert helper to do a smart check here... the assert helper can check if `scenario.analytical_optimum == scenario.lower_bound` and if so, the result must meet lower bound, otherwise must beat lower bound.

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -683,4 +699,67 @@ mod tests {
...
+        let result = evaluate_scenario(&pfw_default(), &scenario, market, gm).await;
+        result.assert_beats_lower_bound();
+        assert_eq!(result.path_count, 2, "expected 2 paths");
+        let primary = result.split_fractions[0];
+        assert!(
+            primary > 0.55,
+            "asymmetric pools: larger pool should get >55%, got {primary} \
+             (fractions {:?})",
+            result.split_fractions,
+        );
```
louise-poole: Then these special asserts can be done like you have here in a stand-alone test. The main test loop should do a generic lower bound and optimum check.

fynd-core/src/algorithm/split_test_harness.rs
louise-poole: Also please assert `result.within_pct_of_optimum` too? For 5% off of optimum.

louise-poole: Lastly - there are `#[allow(dead_code)]` flags on those assert helpers... can you remove them now that they're in use?

fynd-core/src/algorithm/split_test_harness.rs
louise-poole: I think the task was badly designed and not paid much attention to. All test scenarios are doing a version of a split vs single route comparison. We have a `within_pct_of_optimum` check on the scenario result struct that we can call to ensure the result is close to optimum without need for manual review.

fynd-core/src/algorithm/split_test_harness.rs
Troshchk: The `evaluate_scenario` returns `ScenarioResult`, not route, so I cannot validate it here.
As an alternative, I can do validation inside the `evaluate_scenario`, that might be good to validate the bf routes too, what do you think?

Troshchk: I have added validation by expanding ScenarioResult with route field

fynd-core/src/algorithm/split_test_harness.rs
Troshchk: Added assert for it too

fynd-core/src/algorithm/split_test_harness.rs
Troshchk: I have updated the ticket description to reflect what was actually required in this task.
The benchmarking part is not in the scope of this task. The other relevant checks are added

fynd-core/src/algorithm/split_test_harness.rs
Troshchk: Fixed all

fynd-core/src/algorithm/split_test_harness.rs
Troshchk: I have only kept stand-alone test for the case not covered in the path_frank_wolfe.rs tests

fynd-core/src/algorithm/split_test_harness.rs
tamaralipows: > All test scenarios are doing a version of a split vs single route comparison.

Affff derpppp. That was the whole point of using BF as baseline. Of course.
Late night brainfart. Thank you!

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -125,23 +125,30 @@ pub(crate) struct ScenarioResult {
     /// Number of swaps consuming `token_in`. 1 for single-route, N for a split.
     pub path_count: usize,
     /// Split fraction per `token_in` swap. All-zero if the algorithm does not set `with_split`.
+    #[allow(dead_code)]
```
Troshchk: Are we planning to use split_fractions later? If not, I would remove it.
split_fractions are currently calculated only for token_in. Is that intended?

tamaralipows: I think we can remove it if we aren't using it for assertions

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -125,23 +125,30 @@ pub(crate) struct ScenarioResult {
...
     }
 
+    /// Asserts `>= lower_bound` when optimum equals it, `> lower_bound` otherwise.
+    pub fn assert_passes_lower_bound(&self) {
+        if self.analytical_optimum == self.lower_bound {
+            self.assert_meets_lower_bound();
+        } else {
+            self.assert_beats_lower_bound();
+        }
+    }
```
tamaralipows, line 145: Why is this necessary? `meets_lower_bound` is already `>=` (could maybe be named better)

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -683,4 +725,150 @@ mod tests {
...
+
+    fn assert_route_is_valid(route: &types::quote::Route, scenario_name: &str) {
+        route
+            .validate()
+            .unwrap_or_else(|e| panic!("'{scenario_name}': validate failed: {e}"));
+    }
+
+    /// Converts a route to a Solution, encodes it, and verifies the calldata
+    /// contains the correct token_in, token_out, and amount_in.
```
tamaralipows: Oh you really checked the encoded calldata, I (and I think the original task) was just referring to whether the route has the expected tokens. I don't think it's necessary to re-test our encoding (it's already thoroughly done in tycho-execution)

fynd-core/src/algorithm/split_test_harness.rs
```diff
@@ -683,4 +725,150 @@ mod tests {
...
+    }
+
+    fn real_encoder() -> Encoder {
+        let registry = SwapEncoderRegistry::new(Chain::Ethereum)
+            .add_default_encoders(None)
+            .expect("registry should build");
+        Encoder::new(Chain::Ethereum, registry).expect("encoder should build")
+    }
+
+    fn assert_route_is_valid(route: &types::quote::Route, scenario_name: &str) {
```
tamaralipows: This is just a passthrough fn, I don't think it's necessary, we can just call validate wherever this is called instead

fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -421,6 +421,18 @@ impl PathFrankWolfeAlgorithm {
             alloc.amount_in = alloc_amount_in;
             alloc.amount_out = sim.amount_out;
             alloc.marginal_price_product = sim.marginal_price_product;
+
+            // Sync per-hop amount_out/gas with the freshly simulated values.
+            // Without this, hops retain stale values from initial discovery,
+            // causing build_split_route to emit wrong swap amounts
```
tamaralipows: For my info, what exactly changes in the fresh simulation that makes this stale? Maybe we can clarify in this doc? Is it the gas estimation?

tamaralipows (commented):
Thanks - it's looking good! Just a few questions/suggestions 🙏

fynd-core/src/algorithm/split_test_harness.rs
Troshchk: There are scenarios, where we want the result be strictly better, that lower bound(when we expect a split). For other scenarios we are fine with at least meeting the lower bound, when split will not be happening(like when gas is too high, for example)

fynd-core/src/algorithm/split_test_harness.rs
louise-poole: Agreed - let's remove it

fynd-core/src/algorithm/path_frank_wolfe.rs
louise-poole: It's the split optimisation. `amount_out` and `gas` are populated with the initial pre-optimised split results and need to be updated to reflect the final split chosen.
@Troshchk if what I've said here is correct, can you update the comment to explain it too?

Troshchk: Hop amount_out was never updated before this change, only the route amount_out.
Because of that, the `merge_shared_hops` output was incorrect, since it would use old values for the hop amount_out and sum them.

fynd-core/src/algorithm/split_test_harness.rs
Troshchk: I am not sure I understand what you mean by `the route has the expected tokens`.
There are asserts for solution fields, is this what you mean?

louise-poole: I think it was only through this encoder test that you saw the stale `amount_out` and `gas` bug right? I agree with Tamara, we should avoid re-testing the encoder like this. If we change something on the encoder we'll forget to update it here (this is a non-obvious dependency/hidden maintenance). It was good you did this as a sanity check, but as a permanent unit test let's test the bug scenario directly and avoid the encoder?

Troshchk: I have removed the part that tested encoding.
The initial issue with the stale hop `amount_out` was caught using `within_pct_of_optimum`. It is an indirect way to detect this bug, but since all downstream data is derived from the stale values, I think this is the best place to catch this kind of issue. For that reason, I will not be adding an additional test here.

tamaralipows (approved):
Thank you!

louise-poole (approved):
Nice! Thanks!!!

PR #248, opened 2026-06-10
feat: scaffold permissioned surplus pools

kayibal:
## What this is

A **scaffold** (structure only, `todo!()` bodies) for routing through permissioned "fair-flow hook" pools that only Fynd can access. The trade executes through the permissioned pool, the user is quoted the best public-market rate, and the protocol captures the surplus (`egAmount`) above that rate. On-chain hook contracts and permissioned-pool ingestion are out of scope; this is the off-chain solving side.

Open it to review the structure and the design encoded in the doc comments — no logic is implemented yet.

## What's in the diff

- **`fynd-core/src/feed/permission.rs`** (new) — `PermissionPolicy` (an `is_permissioned(&ProtocolComponent) -> bool` predicate), `ComponentScope { ExcludePermissioned, IncludeAll }`, and `filter_topology` / `filter_component_ids` helpers (`todo!()`).
- **`worker.rs` / `pool.rs` / `registry.rs`** — thread a per-worker permission context so public pools filter permissioned components out of their local graph in `initialize_graph` and `process_event`. The shared `MarketState` is not duplicated.
- **`worker_pool_router/mod.rs`** — `PoolRole { Public, Surplus }` on `SolverPoolHandle`; a role-aware `combine_with_surplus` wired into `quote()`. Its doc comment carries the full per-leg attribution math (pro-rata haircut, the concavity proof that the user is never quoted worse than the public route, rounding/clamp rules).
- **`types/quote.rs`** — `SurplusInfo` (order-level reporting aggregate) on `OrderQuote`, and a per-leg `committed_amount_out` on `Swap` that the encoder reads to derive the hook's `maxExchangeRate`. Both are `#[serde(skip)]` and never reach the public `/v1/quote` DTO.
- **`solver.rs`** — `PoolConfig.role` (serde) + `FyndBuilder::permission_policy(...)`; `assemble_components` tags each pool with its role and scope.
- **`encoder.rs`**, **`fynd-rpc-types`**, **`worker_pools.toml`** — a TODO to read the committed amount into the encode path, a NOTE that surplus fields stay out of the DTO, and a commented `[pools.surplus]` example.

## Not implemented (next, on top of this scaffold)

- `combine_with_surplus` body (surplus-vs-public selection + per-leg stamping).
- Role-aware response gating in `solve_order` (wait for a public **and** the surplus candidate, not just the first response).
- `filter_topology` / `filter_component_ids` bodies.
- Encoder hook calldata / signature.

### DIFF
--- a/fynd-core/src/encoding/encoder.rs
+++ b/fynd-core/src/encoding/encoder.rs
@@ -58,11 +58,11 @@ impl TryFrom<&OrderQuote> for Solution {
             SolveError::FailedEncoding("successful quote must have a route".to_string())
         })?;
 
-        // TODO: when this quote routes through a permissioned pool, read
-        // `quote.committed_amount_out()` and `quote.eg_amount()` here and carry them into the
-        // encoded `Solution` so the on-chain fair-flow hook can be parameterised. The user is
-        // committed to `committed_amount_out` while the executed route yields the surplus. Hook
-        // calldata/signature encoding is a separate later extension and is out of scope here.
+        // TODO: when a swap in this route is permissioned, read its `Swap::committed_amount_out`
+        // and carry it into the encoded `Solution` so the on-chain permissioned hook can be
+        // parameterised — the user is committed to that amount while the executed route yields the
+        // surplus. Hook calldata/signature encoding is a separate later extension, out of scope
+        // here.
 
         let token_in = route
             .input_token()
--- a/fynd-core/src/encoding/encoder.rs
+++ b/fynd-core/src/encoding/encoder.rs
@@ -58,6 +58,12 @@ impl TryFrom<&OrderQuote> for Solution {
             SolveError::FailedEncoding("successful quote must have a route".to_string())
         })?;
 
+        // TODO: when this quote routes through a permissioned pool, read
+        // `quote.committed_amount_out()` and `quote.eg_amount()` here and carry them into the
+        // encoded `Solution` so the on-chain fair-flow hook can be parameterised. The user is
+        // committed to `committed_amount_out` while the executed route yields the surplus. Hook
+        // calldata/signature encoding is a separate later extension and is out of scope here.
+
         let token_in = route
             .input_token()
             .ok_or_else(|| SolveError::FailedEncoding("route has no input token".to_string()))?;
--- a/fynd-core/src/feed/permission.rs
+++ b/fynd-core/src/feed/permission.rs
@@ -6,11 +6,17 @@
 //! each worker's local graph topology/events through a `PermissionPolicy` according to its
 //! `ComponentScope` — the shared `MarketState` is never duplicated.
 
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::{HashMap, HashSet},
+    sync::Arc,
+};
 
 use tycho_simulation::tycho_common::models::{protocol::ProtocolComponent, Address};
 
-use crate::{feed::market_data::MarketState, types::ComponentId};
+use crate::{
+    feed::{events::MarketEvent, market_data::MarketState},
+    types::ComponentId,
+};
 
 /// Classifies a [`ProtocolComponent`] as permissioned (Fynd-exclusive) or public.
 ///
@@ -58,38 +64,98 @@ pub enum ComponentScope {
     IncludeAll,
 }
 
-/// Filters a full topology map to the components visible under `scope`.
+/// Per-worker permission scoping: which components a worker may ingest into its local graph.
 ///
-/// Returned by public workers as a strict subset of the input; surplus workers receive the input
-/// unchanged.
-pub fn filter_topology(
+/// Bundles a [`ComponentScope`] with an optional [`PermissionPolicy`]. All graph filtering for a
+/// worker flows through this type, so the worker never reasons about permissioned-ness itself — it
+/// just hands its topology and incoming events here.
+#[derive(Clone, Debug)]
+pub(crate) struct PermissionContext {
+    /// Scope governing whether permissioned components are filtered out.
     scope: ComponentScope,
-    policy: &PermissionPolicy,
-    market: &MarketState,
-    topology: HashMap<ComponentId, Vec<Address>>,
-) -> HashMap<ComponentId, Vec<Address>> {
-    // TODO: when scope == ExcludePermissioned, drop entries whose ComponentId resolves (via
-    // market.get_component(id)) to a ProtocolComponent for which policy.is_permissioned is true.
-    // When scope == IncludeAll, return `topology` unchanged.
-    let _ = (scope, policy, market, topology);
-    todo!("filter permissioned components from the worker's initial topology")
+    /// Predicate classifying components as permissioned. `None` ⇒ no filtering is ever applied.
+    policy: Option<PermissionPolicy>,
 }
 
-/// Filters a list of changed component ids to those visible under `scope`.
-///
-/// Used on the incremental-event path (added/updated/removed component ids) so that a public
-/// worker never ingests a permissioned component mid-stream.
-pub fn filter_component_ids(
-    scope: ComponentScope,
-    policy: &PermissionPolicy,
-    market: &MarketState,
-    ids: &[ComponentId],
-) -> Vec<ComponentId> {
-    // TODO: when scope == ExcludePermissioned, keep only ids whose ProtocolComponent (via
-    // market.get_component(id)) is NOT permissioned per policy.is_permissioned. When scope ==
-    // IncludeAll, return all ids unchanged.
-    let _ = (scope, policy, market, ids);
-    todo!("filter permissioned component ids from an incremental market event")
+impl PermissionContext {
+    /// Default context: see every component, apply no permission filtering.
+    pub(crate) fn include_all() -> Self {
+        Self { scope: ComponentScope::IncludeAll, policy: None }
+    }
+
+    /// Context derived from a pool's role: a scope plus the (optional) classifying policy.
+    pub(crate) fn new(scope: ComponentScope, policy: Option<PermissionPolicy>) -> Self {
+        Self { scope, policy }
+    }
+
+    /// Returns the policy to enforce, or `None` when this worker filters nothing (scope is
+    /// `IncludeAll`, or no policy was configured).
+    fn active_policy(&self) -> Option<&PermissionPolicy> {
+        match (self.scope, &self.policy) {
+            (ComponentScope::ExcludePermissioned, Some(policy)) => Some(policy),
+            _ => None,
+        }
+    }
+
+    /// Filters a full topology map to the components this worker may see.
+    ///
+    /// Public workers drop permissioned components; surplus workers (and workers with no policy)
+    /// receive the map unchanged.
+    pub(crate) fn filter_topology(
+        &self,
+        market: &MarketState,
+        topology: HashMap<ComponentId, Vec<Address>>,
+    ) -> HashMap<ComponentId, Vec<Address>> {
+        let Some(policy) = self.active_policy() else {
+            return topology;
+        };
+        // TODO: drop entries whose ComponentId resolves (via market.get_component(id)) to a
+        // ProtocolComponent for which policy.is_permissioned is true.
+        let _ = (policy, market);
+        todo!("filter permissioned components from the worker's initial topology")
+    }
+
+    /// Restricts a market event to the components this worker may see.
+    ///
+    /// Public workers drop permissioned component ids from the added/updated/removed lists so a
+    /// permissioned component is never ingested mid-stream; other workers see the event unchanged.
+    pub(crate) fn scope_event(&self, market: &MarketState, event: MarketEvent) -> MarketEvent {
+        let Some(policy) = self.active_policy() else {
+            return event;
+        };
+        let MarketEvent::MarketUpdated { added_components, removed_components, updated_components } =
+            event;
+
+        let added_ids: Vec<ComponentId> = added_components
+            .keys()
+            .cloned()
+            .collect();
+        let permitted_added: HashSet<ComponentId> = self
+            .filter_component_ids(policy, market, &added_ids)
+            .into_iter()
+            .collect();
+        let added_components = added_components
+            .into_iter()
+            .filter(|(id, _)| permitted_added.contains(id))
+            .collect();
+        let removed_components = self.filter_component_ids(policy, market, &removed_components);
+        let updated_components = self.filter_component_ids(policy, market, &updated_components);
+
+        MarketEvent::MarketUpdated { added_components, removed_components, updated_components }
+    }
+
+    /// Keeps only the component ids that are NOT permissioned under `policy`.
+    fn filter_component_ids(
+        &self,
+        policy: &PermissionPolicy,
+        market: &MarketState,
+        ids: &[ComponentId],
+    ) -> Vec<ComponentId> {
+        // TODO: keep only ids whose ProtocolComponent (via market.get_component(id)) is NOT
+        // permissioned per policy.is_permissioned.
+        let _ = (policy, market, ids);
+        todo!("filter permissioned component ids from an incremental market event")
+    }
 }
 
 #[cfg(test)]
@@ -97,7 +163,7 @@ mod tests {
     use super::*;
     use crate::{
         algorithm::test_utils::{component, token},
-        feed::market_data::MarketState,
+        feed::{events::MarketEvent, market_data::MarketState},
     };
 
     /// A predicate that treats the `vm:permissioned` protocol system as Fynd-exclusive.
@@ -136,32 +202,36 @@ mod tests {
         let market = market_with(vec![public_component("pub-1"), permissioned_component("perm-1")]);
         let topology = market.component_topology();
 
-        let public_view = filter_topology(
-            ComponentScope::ExcludePermissioned,
-            &policy,
-            &market,
-            topology.clone(),
-        );
+        let public = PermissionContext::new(ComponentScope::ExcludePermissioned, Some(policy));
+        let public_view = public.filter_topology(&market, topology.clone());
         assert!(public_view.contains_key("pub-1"));
         assert!(!public_view.contains_key("perm-1"));
 
         let surplus_view =
-            filter_topology(ComponentScope::IncludeAll, &policy, &market, topology.clone());
+            PermissionContext::include_all().filter_topology(&market, topology.clone());
         assert_eq!(surplus_view.len(), topology.len());
     }
 
     #[test]
     #[ignore = "scaffold: filter helpers are todo!()"]
-    fn filter_component_ids_excludes_permissioned() {
+    fn scope_event_excludes_permissioned() {
         let policy = permissioned_protocol_policy();
         let market = market_with(vec![public_component("pub-1"), permissioned_component("perm-1")]);
-        let ids = vec!["pub-1".to_string(), "perm-1".to_string()];
+        let event = MarketEvent::MarketUpdated {
+            added_components: HashMap::from([
+                ("pub-1".to_string(), vec![]),
+                ("perm-1".to_string(), vec![]),
+            ]),
+            removed_components: vec!["pub-1".to_string(), "perm-1".to_string()],
+            updated_components: vec!["pub-1".to_string(), "perm-1".to_string()],
+        };
 
-        let public_ids =
-            filter_component_ids(ComponentScope::ExcludePermissioned, &policy, &market, &ids);
-        assert_eq!(public_ids, vec!["pub-1".to_string()]);
-
-        let surplus_ids = filter_component_ids(ComponentScope::IncludeAll, &policy, &market, &ids);
-        assert_eq!(surplus_ids, ids);
+        let public = PermissionContext::new(ComponentScope::ExcludePermissioned, Some(policy));
+        let MarketEvent::MarketUpdated { added_components, removed_components, updated_components } =
+            public.scope_event(&market, event);
+        assert!(added_components.contains_key("pub-1"));
+        assert!(!added_components.contains_key("perm-1"));
+        assert_eq!(removed_components, vec!["pub-1".to_string()]);
+        assert_eq!(updated_components, vec!["pub-1".to_string()]);
     }
 }
--- a/fynd-core/src/feed/permission.rs
+++ b/fynd-core/src/feed/permission.rs
@@ -1,10 +1,11 @@
 //! Per-worker permission scoping for permissioned components.
 //!
-//! "Permissioned" means accessible only to Fynd: such components must never appear in a normal
-//! public quote, yet a dedicated worker is allowed to route through them to capture the surplus
-//! they offer above the best public-market rate. Isolation is achieved by filtering each worker's
-//! local graph topology/events through its `PermissionContext` — the shared `MarketState` is
-//! never duplicated.
+//! "Permissioned" means accessible only to Fynd: such components must never enter the route a
+//! public pool returns, yet the surplus pool is allowed to route through them to capture the
+//! surplus they offer above the best public-market rate. Both pool kinds serve the same request;
+//! they differ only in which liquidity their workers may route through. Isolation is achieved by
+//! filtering each worker's local graph topology/events through its `PermissionContext` — the
+//! shared `MarketState` is never duplicated.
 
 use std::{
     collections::{HashMap, HashSet},
--- a/fynd-core/src/feed/permission.rs
+++ b/fynd-core/src/feed/permission.rs
@@ -1,10 +1,10 @@
-//! Per-worker permission scoping for permissioned ("fair-flow hook") pools.
+//! Per-worker permission scoping for permissioned components.
 //!
-//! Some pools are exclusive to Fynd: they must never appear in a normal public quote, yet a
-//! dedicated "surplus" worker is allowed to route through them to capture the surplus
-//! (`egAmount`) they offer above the best public-market rate. Isolation is achieved by filtering
-//! each worker's local graph topology/events through a `PermissionPolicy` according to its
-//! `ComponentScope` — the shared `MarketState` is never duplicated.
+//! "Permissioned" means accessible only to Fynd: such components must never appear in a normal
+//! public quote, yet a dedicated worker is allowed to route through them to capture the surplus
+//! they offer above the best public-market rate. Isolation is achieved by filtering each worker's
+//! local graph topology/events through a `PermissionPolicy` according to its `ComponentScope` — the
+//! shared `MarketState` is never duplicated.
 
 use std::{
     collections::{HashMap, HashSet},
@@ -18,7 +18,7 @@ use crate::{
     types::ComponentId,
 };
 
-/// Classifies a [`ProtocolComponent`] as permissioned (Fynd-exclusive) or public.
+/// Classifies a [`ProtocolComponent`] as permissioned or public.
 ///
 /// The predicate is supplied by the caller rather than hard-coded against an id or protocol name,
 /// so the notion of "permissioned" can evolve (e.g. a hook address allowlist) without touching the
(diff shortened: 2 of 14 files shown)


fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -160,10 +186,28 @@ impl WorkerPoolRouter {
...
+
+        // Rank quotes for each order (sorted by refined amount_out_net_gas descending).
+        // When a surplus pool is configured, fold the surplus candidate into the public reference
+        // via `combine_with_surplus`; otherwise keep the plain public-only ranking.
         let ranked_quotes: Vec<Vec<OrderQuote>> = order_responses
             .into_iter()
-            .map(|responses| self.rank_quotes(&responses, request.options()))
+            .map(|responses| {
+                if has_surplus_pool {
+                    vec![combine_with_surplus(&responses, &pool_roles, request.options())]
```
kayibal: **Surplus path drops the price-guard fallback list**

`combine_with_surplus` returns a single-element `vec![quote]`, but the price-guard path (`PriceGuard::validate` -> `select_first_valid`) walks a *ranked list* per order to find the first candidate that passes external validation. With a surplus pool active, a failed validation has nothing to fall back to (stamped `PriceCheckFailed`, no recovery), unlike the public-only path which keeps the full list.

The types still line up (`Vec<OrderQuote>`), so this is easy to miss once the body is implemented. Worth deciding at scaffold time: return the combined winner *prepended to* the remaining ranked public candidates, or revisit the routing/price-guard contract.

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -160,10 +186,28 @@ impl WorkerPoolRouter {
...
+            .any(|r| *r == PoolRole::Surplus);
+
+        // Rank quotes for each order (sorted by refined amount_out_net_gas descending).
+        // When a surplus pool is configured, fold the surplus candidate into the public reference
+        // via `combine_with_surplus`; otherwise keep the plain public-only ranking.
         let ranked_quotes: Vec<Vec<OrderQuote>> = order_responses
             .into_iter()
-            .map(|responses| self.rank_quotes(&responses, request.options()))
+            .map(|responses| {
+                if has_surplus_pool {
```
kayibal: **Combiner forks `rank_quotes` fallback logic**

This branch routes *every* order through `combine_with_surplus`, even when no surplus candidate responded, bypassing `rank_quotes` and its `max_gas` filtering, multi-candidate ranking, and `NoRouteFound`/`Timeout`/`NotReady` placeholder + metrics. The TODO in `combine_with_surplus` says it must replicate that placeholder behavior — duplicating non-trivial logic in two places.

Consider having `combine_with_surplus` delegate to `rank_quotes` for the fallback and only override the winner, so the seam doesn't fork that logic.

kayibal (commented):
🤖  Scaffold review — interface & plan focus. Two seams worth deciding before `combine_with_surplus` is implemented; both stem from the surplus path replacing rather than augmenting the existing ranked-candidate flow. Not blockers for the scaffold.

kayibal (commented):
some smaller issues

fynd-core/src/types/quote.rs
```diff
@@ -701,6 +701,50 @@ impl SingleOrderQuote {
...
+/// committed_amount_out`, in the order's `token_out`). Present only on quotes selected from the
+/// surplus pool; absent for pure public quotes.
+///
+/// This struct is the *reporting aggregate*. The *actionable* per-pool commitment the encoder
+/// signs lives on the permissioned [`Swap::committed_amount_out`] leg, because the on-chain hook
+/// captures surplus per pool (in that pool's `token_out`), which only equals the order-level
+/// `eg_amount` when the permissioned pool is the route's terminal leg.
+///
+/// These values must survive into the encoding path so the on-chain hook can be parameterised, but
+/// are intentionally never exposed on the public `/v1/quote` HTTP response.
```
kayibal: HTTP api mentioned in core - this paragraph is out of place/unnecessary.

fynd-core/src/types/quote.rs
```diff
@@ -752,6 +796,12 @@ pub struct OrderQuote {
     /// The state overlay this quote was computed against.
     /// When no overlay was requested this is the block number of the base state at solve time.
     solved_against: StateLabel,
+    /// Surplus captured when this quote executes through a permissioned pool.
+    ///
+    /// `None` for pure public quotes. Skipped during (de)serialization: it must reach the
+    /// in-process encoder but is deliberately not part of the public HTTP response DTO.
```
kayibal: If this is informational only, (on quote level) then it should need to reach encoding. 

Http api mention in core out of place.

fynd-core/src/worker_pool/worker.rs
```diff
@@ -57,6 +59,34 @@ where
...
+
+/// Per-worker permission scoping: which components the worker may ingest into its local graph.
+#[derive(Clone, Debug)]
+pub(crate) struct PermissionContext {
+    /// Scope governing whether permissioned components are filtered out.
+    scope: ComponentScope,
+    /// Predicate classifying components as permissioned. `None` ⇒ no filtering is ever applied
+    /// (equivalent to `IncludeAll`).
+    policy: Option<PermissionPolicy>,
+}
```
kayibal: Why is this not part of feed/permission.rs?

fynd-core/src/worker_pool/worker.rs
```diff
@@ -102,7 +141,21 @@ where
...
+                    // topology so they can never surface in a public quote.
+                    crate::feed::permission::filter_topology(
+                        self.permission.scope,
+                        policy,
+                        market.base_market_state(),
+                        topology,
+                    )
+                }
+                // IncludeAll (surplus worker) or no policy ⇒ original unfiltered behaviour.
+                _ => topology,
```
kayibal: This logic should be encapsulated under the permission module. filter topology could become a method on PermissionContext

fynd-core/src/worker_pool_router/mod.rs
kayibal: Fixed in cf3a161. combine_with_surplus now returns Vec<OrderQuote>: the public ranking (from rank_quotes) is the fallback chain, and the surplus winner is prepended to it, so the price guard keeps its full fallback list.

fynd-core/src/worker_pool_router/mod.rs
kayibal: Fixed in cf3a161. The public ranking is delegated to rank_quotes at the call site and passed into combine_with_surplus, which now only overlays the surplus winner -- the max_gas filtering, placeholder construction, and metrics are no longer duplicated.

fynd-core/src/types/quote.rs
kayibal: Fixed in cf3a161. Dropped the HTTP / v1/quote paragraph from the SurplusInfo doc.

fynd-core/src/types/quote.rs
kayibal: Fixed in cf3a161. The order-level surplus is now framed as informational only (observability); it no longer claims to reach the encoder -- the per-leg Swap::committed_amount_out is what the encoder reads. The HTTP DTO mention is removed.

fynd-core/src/worker_pool/worker.rs
kayibal: Fixed in cf3a161. PermissionContext moved into feed/permission.rs.

fynd-core/src/worker_pool/worker.rs
kayibal: Fixed in cf3a161. Filtering is now encapsulated as methods on PermissionContext (filter_topology and scope_event); the worker just calls self.permission.filter_topology(...) / scope_event(...) and no longer reasons about scope/policy.

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -469,6 +549,80 @@ impl WorkerPoolRouter {
     }
 }
 
+/// Overlays the surplus winner onto the ranked public fallback list for one order.
```
tamaralipows, line 560: Question on the surplus capture model: I'm understanding that the user gets pinned to the public-market rate and the protocol captures `egAmount = realized - committed`. I understand this is a different way of doing the same thing that the OG proposal suggested: the pool reduces its fee to beat the next best alternative, and the protocol charges a portion of the delta. I am mostly interested in ensuring that **the user gets a better price**, not the same public price (so this way we would also win more trades in quoting competitions). 

Are you planning on capping this surplus by some delta so that the price is still better? I don't see that in this PR

fynd-core/src/feed/permission.rs
```diff
@@ -0,0 +1,237 @@
...
+/// so the notion of "permissioned" can evolve (e.g. a hook address allowlist) without touching the
+/// routing core.
+#[derive(Clone)]
+pub struct PermissionPolicy {
+    /// Returns `true` when the component is permissioned and therefore excluded from public
+    /// quotes.
+    is_permissioned: Arc<dyn Fn(&ProtocolComponent) -> bool + Send + Sync>,
+}
+
+impl PermissionPolicy {
```
tamaralipows, line 34: I like that this allows flexibility, but it honestly took a while to understand with this abstraction. Do we actually have plans to support more PropAMM types than Ekubo, or is this possibly a premature generalization? Would this be bad to add in the future if necessary? I guess that would make backwards compatibility harder...

kayibal: Really? What exactly did you find hard to understand here? Basically it's just a function that given a component decides weather it's permissioned or public. 

Happy for a suggestion to make this easier.

Note there a couple of types around this since this needs to be exposed all the way to the toml and for consistency we have different types per layer.

fynd-core/src/worker_pool_router/mod.rs
kayibal: Exactly atm, here the user get's in theory the same price (in practice probably minimally better due to the rounding here being in favour of the user). We can adjust this very easily to also give him a better price in this function.

I don't think though that this would significantly change the plan, we just need to decide the max bps improvement we want to forward to the user and add it in here.

fynd-core/src/feed/permission.rs
tamaralipows: Alright I'm gonna do another review pass and focus on readability

fynd-core/src/encoding/encoder.rs
```diff
@@ -77,6 +77,12 @@ impl TryFrom<&OrderQuote> for Solution {
             SolveError::FailedEncoding("successful quote must have a route".to_string())
         })?;
 
+        // TODO: when this quote routes through a permissioned pool, read
+        // `quote.committed_amount_out()` and `quote.eg_amount()` here and carry them into the
+        // encoded `Solution` so the on-chain fair-flow hook can be parameterised. The user is
```
tamaralipows: Not a fan of using so much different terminology to describe basically the same thing - could we eliminate the "fair-flow" vocab and call it the permissioned hook?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -469,6 +549,80 @@ impl WorkerPoolRouter {
...
+/// permissioned leg. We use a linear (no-inverse-simulation) approximation that is provably
+/// conservative for concave AMMs:
+///
+/// ```text
+/// committed_amount_out_leg = leg.amount_out * committed_route_out / realized_route_out   (ceil)
+/// eg_amount_leg            = leg.amount_out − committed_amount_out_leg                    (floor)
+/// ```
+///
+/// Equivalently, the leg's realized output is scaled down by the same ratio as the whole-route
+/// haircut — i.e. `eg_route` is converted from the order's `token_out` into the leg's `token_out`
```
tamaralipows: Can we use more concrete language that "haircut"

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -469,6 +549,80 @@ impl WorkerPoolRouter {
...
+/// by dividing through the realized downstream price `realized_route_out / leg.amount_out`.
+///
+/// ## Why this never shorts the user
+///
+/// Let `f` map the downstream segment's input (the leg's `token_out`) to the order's final
+/// `token_out`. For standard AMMs `f` is concave and increasing, and a composition of such legs
+/// stays concave. Capturing `Δ = eg_route / p_down` (with `p_down = f(x₀)/x₀`) drops the user's
+/// final output by `f(x₀) − f(x₀−Δ)`. Define `h(eg) = eg − [f(x₀) − f(x₀ − eg/p_down)]`. Then
+/// `h(0) = 0`, `h(f(x₀)) = f(0) = 0`, and `h'' = f''/p_down² < 0`, so `h` is concave with both
+/// endpoints zero ⟹ `h ≥ 0` throughout. Hence the true drop never exceeds `eg_route`, so the user
```
tamaralipows: Would appreciate removing the proof from here and focusing on what the function does

fynd-rpc-types/src/lib.rs
```diff
@@ -1700,6 +1700,9 @@ mod conversions {
     }
 
     impl From<fynd_core::OrderQuote> for OrderQuote {
+        // NOTE: `eg_amount` and `committed_amount_out` (permissioned-pool surplus) are
+        // intentionally NOT mapped onto this public response DTO. They are internal values that
+        // reach only the encoder; exposing them would leak Fynd-exclusive surplus to clients.
```
tamaralipows: We are also using permissioned and Fynd-exclusive interchangeably, I think sticking to permissioned would be better (maybe just one comment somewhere explaining it means Fynd-exclusive)

fynd-core/src/feed/permission.rs
```diff
@@ -0,0 +1,237 @@
+//! Per-worker permission scoping for permissioned ("fair-flow hook") pools.
+//!
+//! Some pools are exclusive to Fynd: they must never appear in a normal public quote, yet a
+//! dedicated "surplus" worker is allowed to route through them to capture the surplus
+//! (`egAmount`) they offer above the best public-market rate. Isolation is achieved by filtering
```
tamaralipows: I see we also use surplus and egAmount interchangeably - let's stick to surplus

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -45,19 +45,40 @@ use crate::{
...
+///
+/// Public pools route only through public liquidity and provide the committed (quoted) reference
+/// output. The single surplus pool additionally routes through permissioned pools and may beat the
+/// public reference, in which case the protocol captures the surplus (`egAmount`).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PoolRole {
+    /// Routes through public liquidity only. Establishes the committed reference output.
+    Public,
+    /// Routes through all liquidity including permissioned pools; source of surplus quotes.
+    Surplus,
```
tamaralipows: I'd call this All - Surplus is rather referring to the outcome of the routing

fynd-core/src/types/quote.rs
```diff
@@ -701,6 +701,47 @@ impl SingleOrderQuote {
...
+/// The user is quoted and committed to the best *public*-market output
+/// (`committed_amount_out`), while the trade actually executes through a permissioned pool that
+/// yields more. The protocol captures the difference (`eg_amount = realized_surplus_output −
+/// committed_amount_out`, in the order's `token_out`). Present only on quotes selected from the
+/// surplus pool; absent for pure public quotes.
+///
+/// This is the *reporting aggregate* (observability only). The *actionable* per-pool commitment
+/// the encoder signs lives on the permissioned [`Swap::committed_amount_out`] leg, because the
+/// on-chain hook captures surplus per pool (in that pool's `token_out`), which only equals the
+/// order-level `eg_amount` when the permissioned pool is the route's terminal leg.
```
tamaralipows: A few docs like this are a bit text-heavy, I think the point can be reached in two sentences

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -45,19 +45,40 @@ use crate::{
     QuoteOptions, QuoteRequest, QuoteStatus, SolveError, SolveParams,
 };
 
+/// The role a solver pool plays in a quote.
+///
+/// Public pools route only through public liquidity and provide the committed (quoted) reference
+/// output. The single surplus pool additionally routes through permissioned pools and may beat the
+/// public reference, in which case the protocol captures the surplus (`egAmount`).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PoolRole {
```
tamaralipows, line 60: I think a lot of confusion comes from the word "pool" meaning two different things though the docs in this PR:                                                                                                   
                                                                                                                                                                     
  1. AMM pool 
  2. Worker pool - a group of solver workers that route across many AMM pools

I guess this is a Fynd-wide problem. 

Even a simple naming convention like "component" for AMM pools and "pool" for worker pools (which the codebase already partially does with ProtocolComponent) would reduce confusion

tamaralipows (commented):
Thanks @kayibal - the structure itself looks sound and pretty close to what I envisioned. Just a couple comments here that could improve understandability. If we take care of those I think the generalization is okay 👍

fynd-core/src/encoding/encoder.rs
kayibal: Done in 5de1b18. Dropped the 'fair-flow' vocab throughout; the on-chain mechanism is now just the permissioned hook.

fynd-core/src/worker_pool_router/mod.rs
kayibal: Done in 5de1b18. Replaced 'haircut' with concrete wording: the leg's realized output is reduced by the same proportion as the order-level reduction.

fynd-core/src/worker_pool_router/mod.rs
kayibal: Done in 5de1b18. Removed the formula block and the conservativeness proof from the doc -- it now describes what the function does and points to the design plan, where the derivation and the user-never-short-changed bound live.

fynd-rpc-types/src/lib.rs
kayibal: Done in 5de1b18. Standardized on 'permissioned'; a single note in feed/permission.rs explains it means accessible only to Fynd.

fynd-core/src/feed/permission.rs
kayibal: Done in 5de1b18. Prose now uses 'surplus' throughout; egAmount only survives as the eg_amount field/getter name.

fynd-core/src/worker_pool_router/mod.rs
kayibal: Done in 5de1b18. Renamed PoolRole::Surplus to PoolRole::All (the role names the liquidity scope; surplus is the outcome). The user-facing 'surplus' config value still maps to it.

fynd-core/src/types/quote.rs
kayibal: Done in 5de1b18. Trimmed SurplusInfo (and the other surplus docs) down to two sentences.

fynd-core/src/worker_pool_router/mod.rs
kayibal: Agreed -- this is a Fynd-wide naming issue beyond this PR. In the docs touched here I now call AMM pools 'components' and reserve 'pool' for worker pools (the PoolRole doc spells out 'a group of workers'). Happy to file a follow-up for a repo-wide convention.

fynd-core/src/feed/permission.rs
```diff
@@ -0,0 +1,237 @@
+//! Per-worker permission scoping for permissioned components.
+//!
+//! "Permissioned" means accessible only to Fynd: such components must never appear in a normal
+//! public quote, yet a dedicated worker is allowed to route through them to capture the surplus
+//! they offer above the best public-market rate. Isolation is achieved by filtering each worker's
```
louise-poole: To cross check my understanding here: 'normal public quote' = quote request and 'dedicated worker to route' = solve request?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -45,19 +45,40 @@ use crate::{
     QuoteOptions, QuoteRequest, QuoteStatus, SolveError, SolveParams,
 };
 
+/// The role a solver pool (a group of workers) plays in a quote.
+///
+/// A `Public` pool routes only through public liquidity and provides the committed (quoted)
```
louise-poole: I think this makes it more obvious:
```suggestion
/// A `Public` worker routes only through public liquidity and provides the committed (quoted)
```
The worker itself does the routing so makes sense to refer to it on that level here (to avoid the pool vs pool confusion)

fynd-core/src/solver.rs
```diff
@@ -119,6 +122,40 @@ fn parse_connector_tokens(
...
 }
 
+/// Role a configured pool plays in routing: public-only or permissioned-inclusive.
+///
+/// Serialized in lowercase (`"public"` / `"surplus"`) in `worker_pools.toml`. Maps to the router's
+/// `PoolRole` and the worker's `ComponentScope`: a public pool excludes permissioned components; a
+/// surplus pool routes through all liquidity to capture surplus above the public rate.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum PoolRoleConfig {
```
louise-poole: `PoolRole` and `PoolRoleConfig` look veeeery similar. Could they maybe just be merged into the same enum? `All` was more clear to me than `Surplus` (which is all liquidity?).

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -283,6 +359,11 @@ impl WorkerPoolRouter {
                             // Extract the OrderQuote from SingleOrderQuote
                             quotes.push((pool_name.clone(), single_quote.order().clone()));
 
+                            // TODO: make this gating role-aware for surplus quotes. The surplus
+                            // route can only be priced once BOTH at least one public candidate
+                            // (the committed reference) AND the surplus pool have responded (or the
+                            // deadline elapses). A plain count-based early return may fire before
+                            // the surplus pool reports, silently dropping the surplus opportunity.
```
louise-poole, line 373: This seems important. If someone uses a 'fastest' config that always returns the first response we'll rarely (if ever) be able to showcase the surplus pool option. Maybe for the count-based early return we should have some special case handling (surplus pool + public pool both count towards just 1 response).

louise-poole (commented):
Looks very solid to me! I have no design feedback, only smaller comments 👍

fynd-core/src/types/quote.rs
```diff
@@ -752,6 +787,14 @@ pub struct OrderQuote {
...
     /// When no overlay was requested this is the block number of the base state at solve time.
     solved_against: StateLabel,
+    /// Order-level surplus summary, populated when this quote executes through a permissioned
+    /// pool.
+    ///
+    /// Informational only (observability); the value the encoder acts on is the per-leg
+    /// [`Swap::committed_amount_out`]. `None` for pure public quotes. `#[serde(skip)]` — internal
+    /// reporting data, not part of the wire format.
+    #[serde(skip)]
+    surplus: Option<SurplusInfo>,
```
louise-poole, line 797: Observability where? I think I missed where the surplus part is actually consumed/used? Is this just for logging purposes? Does Fynd care about the surplus? I assumed Fynd only cared that we're be able to work it to get a better rate on that pool and doesn't care so much about how much the pool earns from it no?

fynd-core/src/feed/permission.rs
```diff
@@ -0,0 +1,237 @@
...
+/// Bundles a [`ComponentScope`] with an optional [`PermissionPolicy`]. All graph filtering for a
+/// worker flows through this type, so the worker never reasons about permissioned-ness itself — it
+/// just hands its topology and incoming events here.
+#[derive(Clone, Debug)]
+pub(crate) struct PermissionContext {
+    /// Scope governing whether permissioned components are filtered out.
+    scope: ComponentScope,
+    /// Predicate classifying components as permissioned. `None` ⇒ no filtering is ever applied.
+    policy: Option<PermissionPolicy>,
+}
```
louise-poole: Quick thought here: `PermissionPolicy` defines *how* we filter, `ComponentScope` defines *what* we filter and `PermissionContext` combines the two. I think this could be made simpler by merging the how and what together a bit? So something like:
```rust
  pub(crate) enum PermissionContext {
      /// See everything — default, no filtering.
      IncludeAll,
      /// Drop components matching this policy.
      ExcludePermissioned(PermissionPolicy),
  }
```

fynd-core/src/worker_pool/pool.rs
```diff
@@ -45,6 +46,10 @@ pub struct WorkerPoolConfig {
     algorithm_config: AlgorithmConfig,
     /// Task queue capacity (maximum number of pending tasks).
     task_queue_capacity: usize,
+    /// Permission scope for this pool's workers (default: include all, no filtering).
+    component_scope: ComponentScope,
+    /// Permission predicate. `None` ⇒ no filtering regardless of scope.
+    permission_policy: Option<PermissionPolicy>,
```
louise-poole: Looks like they're highly coupled so my comment above feels even more valid to me now.

fynd-core/src/feed/permission.rs
tamaralipows: See latest commit - just a note that I had to make this pub enum though because of the pub fns where it's called

tamaralipows: Before, the public builder had two methods taking the two pub pieces (ComponentScope, PermissionPolicy), and built PermissionContext internally - so the context never showed up in a public signature. Merging those two methods into one pub fn permission(PermissionContext) put the context itself into the public signature.

fynd-core/src/feed/permission.rs
tamaralipows: So neither phrase maps to a request type (unless I misunderstood what you meant?). The sentence is about routing scope per pool, applied to the same request:

- "normal public quote" - the output a regular user gets, which must reflect only public liquidity (the public-role pools, scoped ExcludePermissioned).
- "a dedicated worker routes through them" - the surplus pool (PoolRole::All, scoped IncludeAll), which can route through permissioned components to beat that reference.

The wording might be a bit ambiguous - I'll reword it to talk about public pools vs. the surplus pool directly.

fynd-core/src/types/quote.rs
tamaralipows: You've got it - Fynd acts on the committed output, not the surplus amount. I believe this is indeed just meant for logging/metrics. But we do care about the surplus amount itself: we want to show LPs how much they can earn from our pools, and SurplusInfo could be what backs that in the future.

fynd-core/src/worker_pool_router/mod.rs
tamaralipows: Agreed - the role-aware gating is listed under "Not implemented" in the PR description, to land with the `combine_with_surplus` body rather than in this scaffold (it's included in the implementation plan though: see [here](https://propeller-heads.atlassian.net/browse/ENG-6133)).

fynd-core/src/feed/permission.rs
tamaralipows: Okay I rebased to update with main (sorry) but just know the latest three commits are mine

fynd-core/src/feed/permission.rs
louise-poole: Yeah, after going through the rest of the scaffold the wording here made sense. Buuuut not ideal for someone to need to read a lot of code to understand a docstring correctly 👀  Thanks for the reword!

fynd-core/src/feed/permission.rs
louise-poole: Thanks! Looks much cleaner to me now! 🙏 

> put the context itself into the public signature

Yeah, but I think this is okay. It felt very flimsy to have 2 highly coupled optional params that needed to be passed and if only 1 was passed it was not usable. Makes it so much cleaner for it to be passed in 1 param.

fynd-core/src/worker_pool/pool.rs
```diff
@@ -45,6 +46,8 @@ pub struct WorkerPoolConfig {
     algorithm_config: AlgorithmConfig,
     /// Task queue capacity (maximum number of pending tasks).
     task_queue_capacity: usize,
+    /// Permission scoping for this pool's workers (default: include all, no filtering).
+    permission: PermissionContext,
```
Troshchk, line 50: Would it be safer to only include public by default?

PR #250, opened 2026-06-16
feat: Read router fees from on-chain FeeCalculator

dianacarvalho1:
Fynd now reads router fees from the FeeCalculator contract at startup and refreshes them every 5 minutes via a background fetcher, replacing the previously hardcoded fee values.

Done:
- Resolve the FeeCalculator address from the Tycho Router, then read the default router fees, per-client fee overrides, and the fee-unit precision scale (MAX_FEE_BPS).
- Apply fees per order by client address, falling back to the default fee when the client has no override. Uses the client fee receiver to identify the client if set, if not it uses the order sender.
- Encode using only fetched on-chain values; return an error instead of encoding when fees are not yet loaded.
- Add RouterFees/FeeRates types and a SharedRouterFees handle shared between the encoder and the fetcher
  - Track the last successful fetch time on RouterFees
- Gate Solver::wait_until_ready and GET /v1/health on router fees having been loaded at least once.
- Return 503 from /v1/health when the last successful fetch is over an hour old, so a liveness probe restarts the service.

### DIFF
--- a/fynd-core/src/encoding/encoder.rs
+++ b/fynd-core/src/encoding/encoder.rs
@@ -1,4 +1,4 @@
-use std::sync::{Arc, OnceLock};
+use std::sync::Arc;
 
 use alloy::{
     primitives::{aliases::U48, keccak256, Address, Keccak256, U160, U256},
@@ -20,34 +20,13 @@ use tycho_execution::encoding::{
 };
 use tycho_simulation::tycho_common::{models::Chain, Bytes};
 
-use crate::{EncodingOptions, FeeBreakdown, OrderQuote, QuoteStatus, SolveError, Transaction};
+use crate::{
+    encoding::router_fees::{FeeRates, RouterFees, SharedRouterFees},
+    EncodingOptions, FeeBreakdown, OrderQuote, QuoteStatus, SolveError, Transaction,
+};
 
 /// Canonical Permit2 contract address — identical on all EVM chains.
 pub const PERMIT2_ADDRESS: &str = "0x000000000022D473030F116dDEE9F6B43aC78BA3";
-
-static ROUTER_FEE_ON_OUTPUT_BPS: OnceLock<u64> = OnceLock::new();
-
-fn router_fee_on_output_bps() -> u64 {
-    *ROUTER_FEE_ON_OUTPUT_BPS.get_or_init(|| {
-        std::env::var("ROUTER_FEE_BPS")
-            .ok()
-            .and_then(|v| {
-                v.parse()
-                    .map_err(|e| {
-                        tracing::warn!(
-                            value = %v,
-                            error = %e,
-                            "ROUTER_FEE_BPS is not a valid u64; using default (10)"
-                        );
-                    })
-                    .ok()
-            })
-            .unwrap_or(10)
-    })
-}
-
-/// Router's share of the client fee: 2000 basis points (20%).
-const ROUTER_FEE_ON_CLIENT_FEE_BPS: u64 = 2000;
 
 /// Encodes solution into tycho compatible transactions.
 ///
@@ -56,10 +35,12 @@ const ROUTER_FEE_ON_CLIENT_FEE_BPS: u64 = 2000;
 ///   compatible transactions
 /// * `chain` - Chain to be used.
 /// * `router_address` - Address of the Tycho Router contract on this chain.
+/// * `router_fees` - Router fee configuration, refreshed from chain by a background fetcher.
 pub struct Encoder {
     tycho_encoder: Box<dyn TychoEncoder>,
     chain: Chain,
     router_address: Bytes,
+    router_fees: SharedRouterFees,
 }
 
 impl TryFrom<&OrderQuote> for Solution {
@@ -149,12 +130,21 @@ impl Encoder {
                 .build()?,
             chain,
             router_address,
+            router_fees: SharedRouterFees::default(),
         })
     }
 
     /// Returns the Tycho Router contract address for this chain.
     pub fn router_address(&self) -> &Bytes {
         &self.router_address
+    }
+
+    /// Returns the shared router fee handle this encoder reads on every encode.
+    ///
+    /// Pass it to a [`RouterFeeFetcher`](crate::encoding::fee_fetcher::RouterFeeFetcher)
+    /// to keep the fees in sync with the on-chain FeeCalculator.
+    pub fn router_fees(&self) -> SharedRouterFees {
+        self.router_fees.clone()
     }
 
     /// Encodes order solutions for execution.
@@ -199,13 +189,27 @@ impl Encoder {
             .tycho_encoder
             .encode_solutions(solutions)?;
 
+        // Require real, on-chain fee values
+        let router_fees = self
+            .router_fees
+            .snapshot()
+            .ok_or_else(|| {
+                SolveError::FailedEncoding(
+                    "router fees not yet loaded from the on-chain FeeCalculator; cannot encode"
+                        .to_string(),
+                )
+            })?;
         for (encoded_solution, (idx, solution)) in encoded_solutions
             .into_iter()
             .zip(to_encode)
         {
             quotes[idx].set_gas_estimate(encoded_solution.estimated_gas().clone());
-            let (transaction, fee_breakdown) =
-                self.encode_tycho_router_call(encoded_solution, &solution, &encoding_options)?;
+            let (transaction, fee_breakdown) = self.encode_tycho_router_call(
+                encoded_solution,
+                &solution,
+                &encoding_options,
+                &router_fees,
+            )?;
             quotes[idx].set_transaction(transaction);
             quotes[idx].set_fee_breakdown(fee_breakdown);
         }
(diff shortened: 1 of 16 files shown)


fynd-core/src/encoding/router_fees.rs
```diff
@@ -0,0 +1,170 @@
...
+};
+
+use tycho_simulation::tycho_common::Bytes;
+
+/// Legacy basis-points denominator: client fees on the wire use 10,000 = 100%.
+///
+/// This is the calldata convention between Fynd and the router (`clientFeeBps`), independent
+/// of the FeeCalculator's internal precision. The contract scales `clientFeeBps` into its own
+/// fee units by `max_fee_units / LEGACY_BPS_DENOMINATOR`.
+pub const LEGACY_BPS_DENOMINATOR: u64 = 10_000;
```
dianacarvalho1, line 21: note that when we resolve this mismatch between the router and the fee calculator (and do a new router deployment) Fynd won't be able to automatically update, the clients will need to update their Fynd version

fynd-core/src/encoding/fee_fetcher.rs
```diff
@@ -0,0 +1,420 @@
...
+        assert_eq!(default_rates.max_fee_units(), fees.max_fee_units());
+
+        println!(
+            "mainnet router fees: max_fee_units={}, default_on_output={}, \
+             default_on_client_fee={}, custom_clients={}",
+            fees.max_fee_units(),
+            default_rates.on_output(),
+            default_rates.on_client_fee(),
+            fees.custom_client_count(),
+        );
```
louise-poole, line 419: We don't usually have a println in our tests... couldt hese not just be asserts?

fynd-core/src/encoding/router_fees.rs
```diff
@@ -0,0 +1,170 @@
...
+    use super::*;
+
+    const SCALE: u64 = 100_000_000;
+
+    fn client(byte: u8) -> Bytes {
+        Bytes::from(vec![byte; 20])
+    }
+
+    #[test]
+    fn test_fess_for_unknown_client() {
```
louise-poole:
```suggestion
    fn test_fees_for_unknown_client() {
```

fynd-core/src/encoding/router_fees.rs
```diff
@@ -0,0 +1,170 @@
...
+        let fees = RouterFees::new(SCALE, 100_000, 20_000_000, HashMap::new());
+
+        assert_eq!(fees.fees_for(&client(0xAA)), FeeRates::new(100_000, 20_000_000, SCALE));
+    }
+
+    #[test]
+    fn test_fess_for_known_client() {
```
louise-poole:
```suggestion
    fn test_fees_for_known_client() {
```

fynd-core/src/encoding/encoder.rs
```diff
@@ -199,13 +189,27 @@ impl Encoder {
...
 
+        // Require real, on-chain fee values
+        let router_fees = self
+            .router_fees
+            .snapshot()
+            .ok_or_else(|| {
+                SolveError::FailedEncoding(
+                    "router fees not yet loaded from the on-chain FeeCalculator; cannot encode"
+                        .to_string(),
+                )
```
louise-poole: If we will fail to solve unless router fees have been fetched, maybe this should be checked in the readiness checker/health endpoint?
Or let's default to a hardcoded fallback value?

fynd-core/src/encoding/encoder.rs
```diff
@@ -402,28 +414,48 @@ impl Encoder {
...
+        let scaled_client_fee = client_fee_bps as u64 * fee_rates.fee_units_per_bps();
+        let fee_on_output = fee_rates.on_output() as u64;
+        let fee_on_client_fee = fee_rates.on_client_fee() as u64;
+
+        if scaled_client_fee + fee_on_output > max_fee_units || fee_on_client_fee > max_fee_units {
+            return Err(EncodingError::FatalError(format!(
+                "combined client fee ({client_fee_bps} bps) and router fee on output \
+                 ({fee_on_output} fee units) exceed 100%; the router would revert"
+            )));
+        }
```
Troshchk, line 441: Will this erro be accurate also for `fee_on_client_fee > max_fee_units` ?

fynd-core/src/solver.rs
```diff
@@ -65,6 +65,8 @@ pub mod defaults {
     pub const TVL_BUFFER_RATIO: f64 = 1.1;
     /// How often the gas price is refreshed from the RPC node.
     pub const GAS_REFRESH_INTERVAL: Duration = Duration::from_secs(30);
+    /// How often router fees are refreshed from the on-chain FeeCalculator contract.
+    pub const ROUTER_FEE_REFRESH_INTERVAL: Duration = Duration::from_secs(300);
```
Troshchk, line 69: I was wondering, would there be any other way to determine, when we should fees instead of relying on the time interval?
As an example, can the `FeeCalculator` emit some event, that will trigger refreshing fees?

fynd-core/src/encoding/fee_fetcher.rs
```diff
@@ -0,0 +1,420 @@
...
+                        "router fees refreshed from on-chain FeeCalculator"
+                    );
+                    self.shared_fees.set(fees);
+                }
+                Err(e) => {
+                    warn!(
+                        error = %e,
+                        "failed to refresh router fees from chain; keeping previous values"
+                    );
+                }
```
Troshchk, line 132: Should there be some field indicating, that fees are stale?
Also, just for my understanding, encoding may be relying on the stale fees, are we ok with that? Why wouldn't we return error in case of the stale fees?

Troshchk (commented):
Thank you @dianacarvalho1! 💫 
I left a couple of questions for my understanding

fynd-core/src/encoding/fee_fetcher.rs
```diff
@@ -0,0 +1,420 @@
...
+    interface IFeeCalculator {
+        function MAX_FEE_BPS() external view returns (uint32);
+        function getRouterFeeOnOutput() external view returns (uint32);
+        function getRouterFeeOnClientFee() external view returns (uint32);
+        function getAllClientFees(uint256 start, uint256 count)
+            external view returns (address[] memory clients, CustomFees[] memory fees);
+    }
+}
+
+/// Custom-fee entries requested per `getAllClientFees` call. Each entry is three ABI words,
```
louise-poole: Is it really just 3 words per entry? CustomFees has 4 fields no? Plus the address is 1 more. Either way it's well within the node response limits, I just suspect the details of this comment are off.

louise-poole (commented):
Just some small comments - otherwise looks good! 👍

fynd-core/src/solver.rs
louise-poole: That would be cleanest, but then we'd need to set up a websocket connection to consume the events. Not sure the overhead is worth it for something as small as this. This call once every 5 mins doesn't add much load overall and is regular enough to be sufficient imo.

fynd-core/src/encoding/fee_fetcher.rs
louise-poole: Fees will very rarely change onchain. It likely is sufficient enough to just fetch the fees at start-up and never update them again to be honest. BUUUUUT we want to play with the fees a little for now and to be sure they're having an affect that we can observe onchain, we need all users to be using the latest fee within at least an hour of us changing them.

Troshchk: Hmm, I think currently we do not have a way to enforce the latest fee is used(within an hour).
So may be we should add some timestamp indicating when the fees were updated and error on encoding if the fees we updated more than an hour ago?

fynd-core/src/solver.rs
dianacarvalho1: @Troshchk this exact point was discussed [here](https://propellerswap.slack.com/archives/C05LQALE9L1/p1781273719096199)

fynd-core/src/encoding/fee_fetcher.rs
dianacarvalho1: I think it's a bit of an overkill, but I can add a "staleness" check for the fees

fynd-core/src/encoding/fee_fetcher.rs
dianacarvalho1: I didn't want to have asserts because the contract fees might actually change for real and then we have a broken test 😕

fynd-core/src/encoding/router_fees.rs
dianacarvalho1: proof of humanity right?

fynd-core/src/encoding/encoder.rs
dianacarvalho1: initially I had the hardcoded fallback value but I didn't like that because it could just be relying on the hardcoded fallback forever and that's it. I will add this to the health endpoint!

fynd-core/src/solver.rs
Troshchk: thanks!

Troshchk (approved):
Thank you!

fynd-core/src/encoding/fee_fetcher.rs
```diff
@@ -0,0 +1,421 @@
...
+                    default_fee_on_output
+                };
+                let on_client_fee = if fees.hasCustomFeeOnClientFee {
+                    fees.feeBpsOnClientFee
+                } else {
+                    default_fee_on_client_fee
+                };
+                custom_fees
+                    .insert(Bytes::from(client.as_slice().to_vec()), (on_output, on_client_fee));
+            }
```
tamaralipows, line 213: ✅

tamaralipows (approved):
Thank you! I've double checked the logic and it all looks perfect. Glad this has also been added to the health check

fynd-rpc/src/api/handlers.rs
```diff
@@ -114,7 +114,9 @@ pub(crate) async fn health(state: web::Data<AppState>) -> HttpResponse {
         .health_tracker()
         .gas_price_stale()
         .await;
-    let is_healthy = data_fresh && derived_data_ready && !gas_stale;
+    let fees_ready = state.health_tracker().fees_ready();
+    let fees_stale = state.health_tracker().fees_stale();
```
louise-poole: Honestly the staleness check probably shouldn't be here... 
stale prices = non-ideal, but can be tolerated
Blocking fynd from from quoting because of this is a drastic overreaction. At most we should log a warning.

tamaralipows (approved):
I re-approve!

PR #252, opened 2026-06-17
fix: include native/wrapped tokens in derive-connector-tokens

tamaralipows:
The derive-connector-tokens command now always includes the chain's native and wrapped native tokens in the output, even if they fall outside the top-N or min-pool-count filters. This prevents users from accidentally blocking zero-cost wrap/unwrap hops when using the generated connector_tokens list.


dianacarvalho1 (approved):
Thank you @tamaralipows ! 🙏🏼 it would be good to add a test (won't insist though)

louise-poole (approved):

tamaralipows:
Tried quickly adding a test but doing this easily involved a little refactor that I don't have time to double check right now (since there is no existing test). Merging since it's quite a straightforward change

PR #256, opened 2026-06-22
docs: add PathFrankWolfe algorithm documentation

louise-poole:


docs/algorithms/path-frank-wolfe.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Stage 2: Frank-Wolfe loop
+
+### Probe amount
+
+Before each iteration, the algorithm computes a **probe amount**: the minimum trade size where an additional path would pay for its gas cost. This is `gas_cost / price_impact`. If price impact has fallen enough (because prior splits already reduced it), the probe exceeds the configured `max_probe` cap and the loop stops.
+
+### Finding a candidate path
+
+To find the next candidate path, the algorithm constructs a **degraded market state** that reflects the current allocation:
```
dianacarvalho1:
```suggestion
To find the next candidate path, the algorithm constructs an **updated market state** that reflects the current allocation:
```
😅

Troshchk (approved):
Thank you!

docs/algorithms/path-frank-wolfe.md
```diff
@@ -0,0 +1,137 @@
...
+
+Both paths use pool P1. On-chain, P1 is called once with the combined WETH input; P2 and P3 each receive a fraction of the resulting USDC. The route builder:
+
+1. **Merges shared hops**: identifies hops with the same `(pool, token_in, token_out)` across all paths and combines them into a single swap
+2. **Assigns split fractions**: sorts within each branch by flow fraction (largest first); the last swap in each branch receives `split = 0.0` (the TychoRouter "use remaining balance" convention)
+3. **Topological ordering** (Kahn's algorithm): swaps are emitted only after all upstream swaps producing their input token are done. This is necessary when paths of different lengths converge on the same intermediate token — the shared downstream pool must wait for all inflows to complete before its swap is emitted
+
+Gas for shared pools is counted once, not once per path.
+
+## Configuration
```
dianacarvalho1, line 93: should we add this to the default worker pools config as well?

dianacarvalho1 (approved):
Thank you @louise-poole ! It looks good!

docs/algorithms/path-frank-wolfe.md
louise-poole: Technically it is degraded in terms of the current swap. Like attempting to solve the same swap on the new market state will give a worst route now. I think 'updated' doesn't relay that well and 'degraded' does (not a hill I would die on though if someone feels strongly otherwise).

docs/algorithms/path-frank-wolfe.md
louise-poole: Sure! I think we only want 1 algo in the default so we can replace BF there I guess? But then we're making the assumption people by default want the slower but deeper algo though...
I was tempted to leave this and Cairoline handle the fast/best/deep defaults in their Fynd Auto project.

docs/algorithms/path-frank-wolfe.md
dianacarvalho1: aaff for me "degraded" has a bad connotation, like something is wrong with it now. That's the only reason why I don't like the use of this word here. But I won't push

louise-poole: I'll change it to 'post-swap' -> makes it clear what is different about the new market state but without bad connotations 👀

PR #258, opened 2026-06-26
fix(Route): Enforce no empty swaps on Router::new()

dianacarvalho1:


fynd-core/src/types/quote.rs
```diff
@@ -1726,12 +1736,14 @@ mod tests {
...
 
+    // Constructs a `Route` directly, bypassing `Route::new`'s non-empty check, so
+    // accessor behaviour on empty routes (reachable via deserialization) stays testable.
     fn make_route(swaps: Vec<(u8, u8)>) -> Route {
         let swaps: Vec<Swap> = swaps
             .into_iter()
             .map(|(a, b)| make_swap(a, b, 1000, 990))
             .collect();
-        Route::new(swaps, HashMap::new())
+        Route { swaps, tokens: HashMap::new() }
```
Troshchk: If we enforce to not have empty swaps in the Route, why do we need to keep it testable?

Troshchk (commented):
Thank you @dianacarvalho1! 🤩 
I like the change and it looks good to me, only left one question!

fynd-core/src/types/quote.rs
dianacarvalho1: when we create a Route through deserialization it is still possible to have empty swaps

Troshchk: but shouldn't we also have a check the result after the deserialization and make sure it is not empty?

dianacarvalho1: actually.. I think deserialisation never truly happens inside of Fynd. Route is only a response type. Enforcing at new closes the only realistic source (algorithm code) without touching the public API

Troshchk: Should be keep Route::new() than in this case?

dianacarvalho1: I don't understand the question.. Should we keep Route::new() in general? yes, it is still used everywhere in the algorithms. Or do you mean in the case of this test?  edit: I changed the test back to using `Route::new()`

Troshchk (approved):
Thank you!

louise-poole (approved):
Looks good! Thanks!

PR #260, opened 2026-06-30
fix(worker): Call Route.validate() in the SolverWorker

dianacarvalho1:
This way we make sure that the Route returned by whatever algorithm is compatible with what tycho-execution expects

### DIFF
--- a/fynd-core/src/worker_pool/worker.rs
+++ b/fynd-core/src/worker_pool/worker.rs
@@ -217,6 +217,19 @@ where
                 let gas_price = result.gas_price().clone();
                 let route = result.into_route();
 
+                if let Err(err) = route.validate() {
+                    error!(
+                        order_id = %order.id(),
+                        algorithm = self.algorithm.name(),
+                        error = %err,
+                        "algorithm produced an invalid route"
+                    );
+                    return Err(SolveError::AlgorithmError(format!(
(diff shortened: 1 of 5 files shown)


Troshchk (approved):
Thank you @dianacarvalho1!

fynd-core/src/worker_pool/worker.rs
```diff
@@ -217,6 +217,19 @@ where
...
                 let route = result.into_route();
 
+                if let Err(err) = route.validate() {
+                    error!(
+                        order_id = %order.id(),
+                        algorithm = self.algorithm.name(),
+                        error = %err,
+                        "algorithm produced an invalid route"
+                    );
+                    return Err(SolveError::AlgorithmError(format!(
```
louise-poole, line 227: Quick thought: this will just drop that worker pool's solution... so if someone is using just 1 worker pool then they'd get no solution for this quote. Could you add a validate step in the algorithm's route selection process? So if a route fails validation, then the next best route is chosen instead.

dianacarvalho1: you are right! adding it for ML and PFW. BF can't actually produce invalid routes so I will skip it there

dianacarvalho1: addressed it!

PR #262, opened 2026-06-30
feat(tools): fynd-tools-common shared crate + audit tidy (ENG-6121)

kayibal:
Foundational shared crate `fynd-tools-common` (audit + Hindsight helpers), erc20-overrides tidy, and docker dependency-cache fix. Base of the Hindsight stack (ENG-6120). Stacked PR #1/6; base `main`.


tools/erc20-overrides/src/lib.rs
```diff
@@ -34,13 +39,37 @@ pub const MAX_PROBE_SLOT: u64 = 20;
...
+///
+/// Computed as `keccak256(abi.encode(uint256(keccak256("openzeppelin.storage.ERC20")) - 1)) &
+/// ~bytes32(uint256(0xff))`. Tokens using OpenZeppelin Upgradeable 5.x store their `_balances`
+/// mapping here instead of slot 0.
+const OZ_V5_BALANCES_NS: B256 =
+    B256::new(alloy::hex!("52c63247e1f47db19d5ce0460030c497f067ca4cebf71ba98eeadabe20bace00"));
+
+/// OZ v5 `_allowances` is field 1 in `ERC20Storage`, so namespace + 1.
+const OZ_V5_ALLOWANCES_NS: B256 =
+    B256::new(alloy::hex!("52c63247e1f47db19d5ce0460030c497f067ca4cebf71ba98eeadabe20bace01"));
```
kayibal, line 52: 🤖 **OZ v5 namespace constants are unverified by their derivation.** `OZ_V5_BALANCES_NS`/`OZ_V5_ALLOWANCES_NS` are hardcoded precomputed keccak values. The two tests (`oz_v5_*_differs_from_standard`) only assert these differ from the standard mapping slots — they pass even if a constant is transcribed wrong. A typo would silently make OZ v5 detection miss and fall through to `bail!`, with no test catching it. Add a test that recomputes `keccak256(abi.encode(uint256(keccak256("openzeppelin.storage.ERC20")) - 1)) & ~bytes32(uint256(0xff))` and asserts equality, locking the constant to its documented formula.

tools/common/src/fynd.rs
```diff
@@ -0,0 +1,237 @@
...
+use fynd_client::{
+    EncodingOptions, FyndClient, Order, OrderSide, QuoteOptions, QuoteParams, QuoteStatus,
+};
+use num_bigint::BigUint;
+
+use crate::aggregator::{AggregatorCalldata, AggregatorClient, AggregatorQuote, AggregatorStatus};
+
+/// Some trade datasets use 0xeeee…eeee as a sentinel for native ETH; Fynd expects ZERO_ADDRESS.
+const ETH_SENTINEL: &str = "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee";
+const ZERO_ADDRESS: &str = "0x0000000000000000000000000000000000000000";
```
tamaralipows: Could be in constants.rs

tools/benchmark/src/audit/execute.rs
```diff
@@ -0,0 +1,382 @@
...
+        return (None, None, None);
+    };
+    let Some(token_out) = parse_address_hex(token_out_hex) else {
+        warn!("eth_call: invalid token_out address {token_out_hex}");
+        return (None, None, None);
+    };
+
+    // Aggregators use 0xeeee…eeee as a sentinel for native ETH; normalise to zero address
+    // so the ERC-20 slot probe is skipped and the ETH balance path is used.
+    let eth_sentinel = Address::from([0xeeu8; 20]);
```
tamaralipows: We already have `ETH_SENTINEL` that we can use

tools/common/src/fynd.rs
```diff
@@ -0,0 +1,237 @@
...
+    amount: &str,
+    timeout_ms: u64,
+    sender_hex: Option<&str>,
+    encoding: Option<EncodingOptions>,
+) -> anyhow::Result<QuoteParams> {
+    let sender = parse_addr(sender_hex.unwrap_or("0x0000000000000000000000000000000000000001"))?;
+    let order = Order::new(
+        parse_addr(fynd_addr(token_in))?,
+        parse_addr(fynd_addr(token_out))?,
+        BigUint::from_str(amount).unwrap_or_default(),
```
tamaralipows: This should likely error if we can't parse the string, else it will mess up the resolving to give a 0 input

tools/common/src/fynd.rs
```diff
@@ -0,0 +1,237 @@
...
+
+/// Map a Fynd [`QuoteStatus`] onto the shared [`AggregatorStatus`].
+pub fn fynd_status_to_agg(status: QuoteStatus) -> AggregatorStatus {
+    match status {
+        QuoteStatus::Success => AggregatorStatus::Success,
+        QuoteStatus::NoRouteFound | QuoteStatus::InsufficientLiquidity => AggregatorStatus::NoRoute,
+        QuoteStatus::Timeout | QuoteStatus::NotReady | QuoteStatus::PriceCheckFailed => {
+            AggregatorStatus::NoAmount
+        }
+    }
```
tamaralipows, line 135: I'm worried about the QuoteStatus::Timeout | QuoteStatus::NotReady mapping to AggregatorStatus::NoAmount since it might bias Fynd's true coverage metric.

tools/common/tests/eth_call_integration.rs
```diff
@@ -0,0 +1,53 @@
...
+
+/// Canonical WETH9 contract on Ethereum mainnet.
+const WETH: &str = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2";
+/// `deposit()` selector.
+const DEPOSIT_CALLDATA: [u8; 4] = [0xd0, 0xe3, 0x0d, 0xb0];
+/// 1 ETH in wei.
+const ONE_ETH_WEI: &str = "1000000000000000000";
+
+#[tokio::test]
+#[ignore = "requires a live Ethereum RPC; run with --ignored and RPC_URL set"]
```
tamaralipows, line 23: We should probably consider wiring the RPC urls into the CI eventually (no action needed just thinking out loud)

tools/common/tests/eth_call_integration.rs
```diff
@@ -0,0 +1,53 @@
...
+async fn eth_simulate_weth_deposit_returns_exact_amount() {
+    let Ok(rpc_url) = std::env::var("RPC_URL") else {
+        eprintln!("skipping eth_call fork test: RPC_URL not set");
+        return;
+    };
```
tamaralipows: Afraid of silent failing here when developing locally (minor thing)

tamaralipows (commented):
Thank you - only some minor thoughts, likely more things will come up when reviewing the next stories' PRs but we can just solve those there

tools/benchmark/src/audit/execute.rs
```diff
@@ -0,0 +1,382 @@
...
+
+/// One participant's raw quote outcome, keyed by participant name.
+type RawResult = (String, anyhow::Result<AggregatorQuote>);
+/// `(eth_call_amount_out, eth_call_diff_bps, eth_call_gas_used)`.
+type EthCallResult = (Option<String>, Option<f64>, Option<u64>);
+
+/// Baseline (Fynd) figures used to compute every other participant's diffs.
+struct Baseline {
+    amount: String,
+    net_gas: Option<String>,
```
kayibal: Numerical values are represented as string.

tools/benchmark/src/audit/execute.rs
```diff
@@ -0,0 +1,382 @@
...
+/// when `fee_bps == 0` or the amount cannot be parsed.
+fn apply_fee_bps(amount: &str, fee_bps: u32) -> String {
+    if fee_bps == 0 {
+        return amount.to_string();
+    }
+    amount
+        .parse::<BigUint>()
+        .ok()
+        .map(|v| (v * (10_000u32 + fee_bps) / 10_000u32).to_string())
+        .unwrap_or_else(|| amount.to_string())
```
kayibal: Unnecessary failure point here because `Baseline::amount` is of String type. You should make this BigInt right away instead and only fail if we can't decode it.

tools/benchmark/src/audit/execute.rs
```diff
@@ -0,0 +1,382 @@
...
+) -> (Option<f64>, Option<f64>, Option<f64>) {
+    if idx == 0 || !baseline.ok {
+        return (None, None, None);
+    }
+    let Ok(q) = result else {
+        return (None, None, None);
+    };
+    if !q.is_success() {
+        return (None, None, None);
+    }
```
kayibal, line 313: Why do we have this path if we already have a `AggregatorQuote` wrapped in Result?

tools/benchmark/src/audit/execute.rs
kayibal: Agreed, find the correct constant and use it instead of redeclaring it here.

tools/benchmark/src/audit/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+        let sem = Arc::new(Semaphore::new(self.args.concurrency));
+        let mut jset: JoinSet<Result<TradeResult>> = JoinSet::new();
+
+        for &(pair_idx, amount_idx) in chunk.iter() {
+            let pair = self.pairs[pair_idx].clone();
+            let amount = pair.amounts[amount_idx].clone();
+            let participants = self.participants.to_vec();
+            let sem = sem.clone();
+            let block_sample = block_idx + 1;
+            let aggregator_delay_ms = self.args.aggregator_delay_ms;
```
kayibal, line 278: What if the amount of quotes we query times the aggregator delay results in a number > block time. Then our quotes will not be really comparable since the aggregators indexing infra might execute those trades on the next block. I think we need to put at least warn the user about it or ideally structure our chunks so that this can not happen.

tools/common/src/bps.rs
```diff
@@ -0,0 +1,176 @@
...
+    Some((b - o) / o * 10_000.0)
+}
+
+/// Compare baseline vs another participant net-of-gas, in basis points.
+///
+/// Derives the token-per-gas-unit price from the baseline's `(raw - net_gas) / gas_units`,
+/// then applies it to the other participant's gas units. Returns `None` when either gas
+/// figure is zero, when `baseline_net_gas` is absent, or when gas exceeds 5% of output.
+/// Call this twice with reported and on-chain gas values to keep the two measurements
+/// independent.
```
kayibal, line 25: This is a hack for lacking gas price in out_token denomination and it should be mentioned here in the docs.

tools/common/src/swap_simulation.rs
kayibal, line 1: This is mainly for simulating swaps onchain (with eth_call), I think the module name can be improved upon here.

tools/common/src/eth_call.rs
```diff
@@ -0,0 +1,577 @@
...
+
+/// Sentinel address for the inline ETH-balance reader injected into `eth_simulateV1` state.
+/// Chosen to be obviously synthetic and not a real mainnet contract.
+const ETH_BALANCE_READER_ADDR: Address = Address(alloy::primitives::FixedBytes([
+    0xf0, 0xf0, 0xf0, 0xf0, 0xf0, 0xf0, 0xf0, 0xf0, 0xf0, 0xf0, 0xf0, 0xf0, 0xf0, 0xf0, 0xf0, 0xf0,
+    0xf0, 0xf0, 0xf0, 0xf0,
+]));
+
+/// EVM bytecode: CALLER BALANCE PUSH1 0 MSTORE PUSH1 32 PUSH1 0 RETURN.
+/// Returns the ETH balance of the calling address as a uint256.
```
kayibal:
```suggestion
/// Returns the native ETH balance of the calling address as a uint256.
```

tools/common/src/fynd.rs
kayibal: amount should never be parsed as string in the first place. We should convert to BigInt as soon as possible.

tamaralipows (approved):
Thanks!

PR #263, opened 2026-06-30
feat(hindsight): decode on-chain aggregator swaps (ENG-6122)

kayibal:
Decode settled aggregator swaps from chain data, with Allium verification. Includes decoder reliability: decline ambiguous multi-token nets, and decode CoW/batch settlements via the order maker rather than the solver sender. Stacked PR #2/6; base `eng-6121`.


tools/hindsight/src/decoder/match_receipts.rs
```diff
@@ -0,0 +1,177 @@
...
+    provider: &P,
+    logs: &[Log],
+    native: &[(Address, Address, U256)],
+    exclude: &[Address],
+    names: &HashMap<Address, &'static str>,
+    code_cache: &mut HashMap<Address, bool>,
+) -> Option<(Address, (Address, U256, Address, U256))> {
+    // Prefer externally-owned accounts; pools and routers carry code.
+    let mut maker = None;
+    for (candidate, trade) in maker_candidates(logs, native, exclude, names) {
```
kayibal, line 72: 🤖 **Batch settlements decode only one maker.** This loop returns the first candidate with a clean two-token net, so a CoW batch that settles multiple retail orders in one tx contributes a single decoded trade — the rest surface as 'Allium only' gaps and the decoder systematically under-counts batch trades. If decoding one maker per batch is intentional for v0, make it explicit (doc or `warn!`); the batch/CoW path is the focus of this PR.

tools/hindsight/src/decoder/match_receipts.rs
```diff
@@ -0,0 +1,177 @@
...
+        }
+        maker.get_or_insert((candidate, trade));
+    }
+    maker
+}
+
+/// Addresses with a clean two-token net swap, excluding the zero address, the
+/// excluded addresses, and known registry contracts. Ordered by address for
+/// deterministic selection.
+fn maker_candidates(
```
kayibal, line 84: 🤖 **Maker selection has no settlement-tied tiebreak.** When several non-excluded EOAs each net to a clean two-token swap (two independent makers, or a maker plus an EOA counterparty), the winner is just the first in `maker_candidates`' address-ordered iteration. The EOA-vs-contract filter disambiguates maker-vs-pool but not maker-vs-maker, so a correct-looking decode can attribute the wrong account's flow. Consider tying the choice to the actual settlement rather than address order.

tools/hindsight/src/decoder/trace.rs
```diff
@@ -0,0 +1,247 @@
...
+        if child.error.is_some() {
+            continue;
+        }
+        let Some(to) = child.to else { continue };
+        if to == entry_point || to == sender {
+            continue;
+        }
+        let value = child.value.unwrap_or_default();
+        if best.is_none_or(|(_, best_value)| value > best_value) {
+            best = Some((to, value));
```
kayibal, line 121: 🤖 **Aggregator fallback degenerates to 'first child' for ERC-20 routes.** `largest_external_call` ranks the entry point's child calls by native ETH value, but a token→token route through an unknown router carries no native value, so every candidate ties at zero and the first child wins. That attributes the swap to an essentially arbitrary contract; the label is then diffed against Allium, so the mis-attribution reads as a spurious aggregator mismatch rather than a clean 'unknown'. Prefer returning unknown when all candidates tie at zero value.

tools/hindsight/src/decoder/test_utils.rs
tamaralipows, line 1: I'd call this test_utils to match the rest of the repo

tools/hindsight/src/decoder/match_receipts.rs
```diff
@@ -0,0 +1,177 @@
...
+/// address is treated as a contract so it is not mistaken for an EOA maker.
+async fn is_contract<P: Provider>(
+    provider: &P,
+    address: Address,
+    cache: &mut HashMap<Address, bool>,
+) -> bool {
+    if let Some(is_contract) = cache.get(&address) {
+        return *is_contract;
+    }
+    let is_contract = match provider.get_code_at(address).await {
```
tamaralipows, line 126: Do you think it's a safe assumption that no providers are making use of EIP-7702?

tools/hindsight/src/decoder/net.rs
```diff
@@ -0,0 +1,217 @@
...
+        if to == tracked {
+            *received
+                .entry(Address::ZERO)
+                .or_default() += value;
+        }
+    }
+
+    for log in logs {
+        let primitive = to_primitive_log(log);
+        let Ok(transfer) = Transfer::decode_log(&primitive) else {
```
tamaralipows, line 42: We could just move the contents of `to_primitive_log` into `decode_log` to avoid having to call it every time before `decode_log`

tools/hindsight/src/decoder/mod.rs
```diff
@@ -0,0 +1,147 @@
...
+        let Some((tracked, (token_in, amount_in, token_out, amount_out))) = decoded else {
+            warn!(
+                tx = %receipt.transaction_hash,
+                client = %label(entry_point, &names),
+                "no token or native ETH flow found"
+            );
+            continue;
+        };
+
+        let aggregator =
```
tamaralipows, line 130: I think the case where the aggregator is unknown is probably a different case as when the solver is also the aggregator. We should differentiate here and just say unknown venue probably?

tools/hindsight/src/decoder/match_receipts.rs
```diff
@@ -0,0 +1,177 @@
...
+/// externally-owned maker. Known registry contracts and the excluded
+/// addresses (filler, entry point) never qualify.
+pub(crate) async fn find_maker_trade<P: Provider>(
+    provider: &P,
+    logs: &[Log],
+    native: &[(Address, Address, U256)],
+    exclude: &[Address],
+    names: &HashMap<Address, &'static str>,
+    code_cache: &mut HashMap<Address, bool>,
+) -> Option<(Address, (Address, U256, Address, U256))> {
```
tamaralipows, line 69: Complicated tuple - replace with named struct?

tools/hindsight/src/decoder/registry.rs
```diff
@@ -0,0 +1,105 @@
+use std::collections::HashMap;
+
+use alloy::primitives::{address, Address};
+
+/// Known aggregator routers — the venue that actually settles a swap.
+/// Start with top aggregators — expand later.
+pub(crate) fn known_aggregators() -> HashMap<Address, &'static str> {
```
tamaralipows: Minor thing but these are rebuilt on every call... could replace with a lazylock

tools/hindsight/src/decoder/mod.rs
```diff
@@ -0,0 +1,147 @@
...
+    registry::{known_aggregators, known_names, label},
+    trace::{attribute_aggregator, collect_native_transfers, fetch_trace},
+};
+
+/// A decoded aggregator trade: what token went in, what came out.
+///
+/// Native ETH is represented as [`Address::ZERO`].
+#[derive(Debug, Clone, serde::Serialize)]
+pub(crate) struct DecodedTrade {
+    pub tx_hash: String,
```
tamaralipows: Should prob use `TxHash` here, then we can avoid some `to_lowercase` later. We are already using `Address` below

tools/hindsight/src/decoder/net.rs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+/// Net the sent and received balances into a single swap.
+///
+/// Returns `None` unless exactly one token nets out and exactly one token nets in. A net with more
+/// than one token on a side is a batch settlement (e.g. a CoW solver settling many orders, where
+/// the tracked sender is the solver, not a trader), not a single comparable swap. Amounts across
+/// tokens with different decimals are not comparable, so guessing a "dominant" leg would pair
+/// unrelated tokens — declining keeps the re-solve comparison honest.
+fn net_trade(
```
tamaralipows, line 64: I'm a bit sus of this fn. In the TychoRouter for example if we are doing USDC -> WETH -> DAI but the second hop is an RFQ, there might be tokens left after the first hop since RFQs will not consume positive slippage (the quote is for an exact amount). 

Other examples this could potentially fail: rebasing tokens that cause rounding errors, protocols taking fees in fixed tokens

I guess this is at least worth noting as a limitation in the docstring or adding a warn log because these will be silently rejected right now

tools/hindsight/src/main.rs
```diff
@@ -0,0 +1,234 @@
...
+    Decode(DecodeArgs),
+    /// Decode and compare against Allium's aggregator_trades ground truth.
+    Verify(VerifyArgs),
+}
+
+#[derive(Args)]
+struct DecodeArgs {
+    /// Ethereum RPC URL
+    #[arg(long, env = "ETH_RPC_URL")]
+    rpc_url: String,
```
tamaralipows: Let's just call it `RPC_URL`

tools/hindsight/src/main.rs
```diff
@@ -0,0 +1,234 @@
...
+
+async fn run_decode(args: DecodeArgs) -> anyhow::Result<()> {
+    let provider = provider_from(&args.rpc_url)?;
+    let blocks = resolve_blocks(&provider, args.block, args.range.as_deref()).await?;
+
+    let mut all_trades = Vec::new();
+    for block_number in &blocks {
+        info!(block = block_number, "decoding aggregator trades");
+        let start = Instant::now();
+        let trades = decoder::decode_block(&provider, *block_number).await?;
```
tamaralipows: This will abort the whole run if the block fails to decode iiuc, not ideal

tools/hindsight/src/main.rs
```diff
@@ -0,0 +1,234 @@
...
+    /// Range of blocks to verify (e.g. 21000000-21000010)
+    #[arg(long, conflicts_with = "block")]
+    range: Option<String>,
+
+    /// Allium API key
+    #[arg(long, env = "ALLIUM_API_KEY")]
+    allium_key: String,
+
+    /// Allium saved query ID (parameterized by block_number)
+    #[arg(long, env = "ALLIUM_QUERY_ID", default_value = "vGDdbPGNAxCmcCJaYQPH")]
```
tamaralipows: Not super important since it's just for development but this query id won't work on anyone else's workspace - we would have to manually add and register the sql query here if we want to enable this on any workspace. 

 ```
SELECT project, protocol, token_sold_address, token_sold_symbol, token_sold_amount, usd_sold_amount, token_bought_address, token_bought_symbol, token_bought_amount, usd_bought_amount, sender_address, to_address, transaction_hash, transaction_fees_usd, block_number, block_timestamp, log_index FROM ethereum.dex.aggregator_trades WHERE block_number = {{block_number}} ORDER BY log_index
```

If easy for Claude to add then cool, but if the decoder is already working somewhat fine and doesn't need this Allium verify, i'd just remove the default value

File: tools/hindsight/src/decoder/verify_decoder.rs [hunk 14]
```diff
@@ -0,0 +1,452 @@
...
+    ours: &DecodedTrade,
+    rows: &[&AlliumRow],
+    tolerance_bps: f64,
+    decimals: &mut HashMap<Address, u8>,
+    detail: &mut Vec<String>,
+) -> bool {
+    let [row] = rows else {
+        detail.push(format!("amount not compared ({} Allium legs for this tx)", rows.len()));
+        return true;
+    };
```
tamaralipows, line 251: So Allium shows each leg of a split swap as a separate swap - we should add a comment there to explain this else this is confusing

tamaralipows (commented):
Few comments. Will maybe fine more after playing with it myself - thanks!

tamaralipows:
Another thing: we will still need to ensure this runs on all chains but I guess we can do it as a final step

tools/hindsight/src/decoder/match_receipts.rs
kayibal: It's not very widely used so I think yes this assumptions is relatively safe.

tools/hindsight/src/decoder/mod.rs
kayibal: Not sure I guess having some address is probably better than having nothing? But in general I think this code needs to be reworked quite extensively.

tools/hindsight/src/decoder/net.rs
```diff
@@ -0,0 +1,217 @@
...
+
+/// Determine what the tracked address sent and received, netting
+/// intermediate hops, from ERC-20 Transfer logs and native ETH transfers.
+pub(crate) fn decode_trade(
+    logs: &[Log],
+    native_transfers: &[(Address, Address, U256)],
+    tracked: Address,
+) -> Option<(Address, U256, Address, U256)> {
+    let mut sent: HashMap<Address, U256> = HashMap::new();
+    let mut received: HashMap<Address, U256> = HashMap::new();
```
kayibal, line 27: Wouldn't it make sense to work with I256 here instead and just have a single map? Basically we would be aggregating net flows. I wonder if that would potentially simplify things?

tools/hindsight/src/decoder/net.rs
kayibal: I don't think so because Transfer struct is generated by a macro

tools/hindsight/src/decoder/net.rs
```diff
@@ -0,0 +1,217 @@
...
+/// Convert an RPC log to a primitive log for event decoding.
+pub(crate) fn to_primitive_log(log: &Log) -> PrimitiveLog {
+    PrimitiveLog::new_unchecked(log.address(), log.topics().to_vec(), log.data().data.clone())
+}
+
+/// Determine what the tracked address sent and received, netting
+/// intermediate hops, from ERC-20 Transfer logs and native ETH transfers.
+pub(crate) fn decode_trade(
+    logs: &[Log],
+    native_transfers: &[(Address, Address, U256)],
```
kayibal, line 23: Is there any undocumented assumption here on the receiver/address of the logs here?

tools/hindsight/src/decoder/net.rs
kayibal: Yeah I am pretty sure that in practice this is generating a lot of artifacts and noise since the decoded trade is just garbage.

tools/hindsight/src/decoder/net.rs
```diff
@@ -0,0 +1,217 @@
...
+            net_sent.insert(token, s - r);
+        } else if r > s {
+            net_received.insert(token, r - s);
+        }
+    }
+
+    let (&token_in, &amount_in) = net_sent
+        .iter()
+        .next()
+        .filter(|_| net_sent.len() == 1)?;
```
kayibal, line 95: This is an extremely weird way to express that net_sent needs to be exactly of size 1...

tools/hindsight/src/decoder/match_receipts.rs
kayibal: Documented as an explicit v0 limitation on `find_maker_trade` (one maker per batch; the rest surface as "Allium only" gaps) in bbf808b. A settlement-tied fix is part of a deferred decode/attribution rework.

tools/hindsight/src/decoder/match_receipts.rs
kayibal: Documented as a v0 limitation (maker-vs-maker resolved by address order) alongside the batch note in bbf808b. A settlement-tied tiebreak is deferred to the decode/attribution rework.

tools/hindsight/src/decoder/test_utils.rs
kayibal: Renamed `test_support` → `test_utils` in ea66f91.

tools/hindsight/src/decoder/match_receipts.rs
kayibal: Agreed. Documented on `is_contract` as a v0 limitation — a 7702-delegated maker EOA carries code and is dropped — in bbf808b, since 7702 isn't widely used yet.

tools/hindsight/src/decoder/registry.rs
kayibal: Moved the aggregator/client/name registries to `LazyLock` statics returning `&'static` maps, built once per process — 6f00726.

tools/hindsight/src/decoder/mod.rs
kayibal: `DecodedTrade.tx_hash` (and `TxComparison.tx_hash`) are now `TxHash`; verify keys its trade maps by `TxHash` directly (parsing Allium's hash string once), dropping the `to_lowercase` normalisation — 74938ca.

tools/hindsight/src/main.rs
kayibal: Renamed to `RPC_URL` in 1eae9bb, and swept the same rename through the Resolve/Monitor args up the stack so the whole tool uses one env var.

tools/hindsight/src/main.rs
kayibal: Decode and verify now log a warning for the failing block and continue the range instead of aborting — 36e4642.

tools/hindsight/src/main.rs
kayibal: Dropped the hardcoded default: the query id is now required, and the underlying `SELECT` is in the arg's `--help` so anyone can register it in their own Allium workspace — 1eae9bb.

tools/hindsight/src/decoder/verify_decoder.rs
kayibal: Added a comment explaining Allium records each leg of a split swap as its own row, so amounts are only compared when Allium also reported a single leg — bbf808b.

tools/hindsight/src/decoder/trace.rs
kayibal: Deferring to a dedicated decode/attribution rework rather than a partial fix — the venue attribution needs reworking more broadly, so keeping the current best-effort address for now.

tools/hindsight/src/decoder/net.rs
kayibal: 🤖 `decode_log` is generated by the `sol!` macro, so we can't fold `to_primitive_log` into it. A thin `decode_transfer` wrapper is possible, but I'm rolling this into the decode rework rather than adding it piecemeal now.

tools/hindsight/src/decoder/mod.rs
kayibal: 🤖  Deferring — keeping an address (better than nothing) for now. The unknown-venue vs solver-is-aggregator distinction belongs to the decode/attribution rework.

tools/hindsight/src/decoder/match_receipts.rs
kayibal: 🤖  Deferring the named-struct refactor: it changes `decode_trade`'s return, which the decode rework restructures anyway (a signed net-flow map), so folding it in there to avoid churn.

tools/hindsight/src/decoder/net.rs
kayibal: 🤖  Agreed this generates noise today. Deferring to the decode rework, which will handle RFQ leftovers / rebasing / fixed-fee-token nets rather than silently declining.

tools/hindsight/src/decoder/net.rs
kayibal: 🤖  Good idea — a single signed `HashMap<Address, I256>` is the plan for the decode rework; deferring so it lands with the rest of that work.

tools/hindsight/src/decoder/net.rs
kayibal: Deferring; I'll document the log/receiver assumptions as part of the decode rework.

tools/hindsight/src/decoder/net.rs
kayibal: 🤖  Agreed it's convoluted; the size-1 extraction gets rewritten in the decode rework (signed net-flow map).

tamaralipows (approved):

PR #264, opened 2026-06-30
feat(hindsight): re-solve engine + resolve subcommand (ENG-6123)

kayibal:
Re-solve decoded trades through a running Fynd over HTTP and compare against the settled amount (bps verdict). Includes the partial-fill→coverage-miss reclassification (`served`). Stacked PR #3/6; base `eng-6122`.

### DIFF
--- a/tools/hindsight/src/main.rs
+++ b/tools/hindsight/src/main.rs
@@ -1,4 +1,5 @@
 mod decoder;
+mod resolve;
 
 use std::time::Instant;
 
@@ -21,6 +22,8 @@ enum Command {
     Decode(DecodeArgs),
     /// Decode and compare against Allium's aggregator_trades ground truth.
     Verify(VerifyArgs),
+    /// Decode a block's trades and re-solve each through a running Fynd instance.
+    Resolve(ResolveArgs),
 }
 
 #[derive(Args)]
@@ -83,6 +86,33 @@ struct VerifyArgs {
     json: bool,
 }
 
+#[derive(Args)]
+struct ResolveArgs {
+    /// Ethereum RPC URL (used to decode the block's settled trades)
+    #[arg(long, env = "RPC_URL")]
+    rpc_url: String,
+
+    /// Base URL of a running Fynd solver to re-solve trades through
+    #[arg(long, env = "FYND_URL", default_value = "http://localhost:3000")]
+    fynd_url: String,
+
+    /// Block number to re-solve (latest if omitted)
+    #[arg(long)]
+    block: Option<u64>,
+
+    /// Range of blocks to re-solve (e.g. 21000000-21000010)
+    #[arg(long, conflicts_with = "block")]
+    range: Option<String>,
+
+    /// Per-quote timeout in milliseconds for Fynd
+    #[arg(long, default_value_t = 10_000)]
+    timeout_ms: u64,
+
+    /// Output as JSON instead of human-readable
+    #[arg(long)]
+    json: bool,
+}
+
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
     tracing_subscriber::fmt()
@@ -95,6 +125,17 @@ async fn main() -> anyhow::Result<()> {
     match Cli::parse().command {
         Command::Decode(args) => run_decode(args).await,
         Command::Verify(args) => run_verify(args).await,
+        Command::Resolve(args) => {
+            resolve::run::run(
+                &args.rpc_url,
+                &args.fynd_url,
+                args.block,
+                args.range.as_deref(),
+                args.timeout_ms,
+                args.json,
+            )
+            .await
+        }
     }
 }
 
@@ -150,14 +191,14 @@ async fn run_verify(args: VerifyArgs) -> anyhow::Result<()> {
     Ok(())
 }
 
-fn provider_from(rpc_url: &str) -> anyhow::Result<impl Provider> {
+pub(crate) fn provider_from(rpc_url: &str) -> anyhow::Result<impl Provider> {
     let url: reqwest::Url = rpc_url
         .parse()
         .with_context(|| format!("invalid RPC URL: {rpc_url}"))?;
     Ok(ProviderBuilder::new().connect_http(url))
 }
 
-async fn resolve_blocks<P: Provider>(
+pub(crate) async fn resolve_blocks<P: Provider>(
     provider: &P,
     block: Option<u64>,
     range: Option<&str>,
--- a/tools/hindsight/src/resolve/compare.rs
+++ b/tools/hindsight/src/resolve/compare.rs
@@ -1,0 +1,113 @@
+//! Pure comparison math between a Fynd re-solve and the on-chain settled amount.
+
+use alloy::primitives::U256;
+use fynd_tools_common::bps::raw_bps_diff;
+use serde::Serialize;
+
+use crate::resolve::Outcome;
+
+/// Basis-point deltas of a Fynd quote against the settled amount (positive = Fynd better).
+#[derive(Debug, Clone, Copy, PartialEq, Serialize)]
+pub(crate) struct Deltas {
+    /// `fynd amount_out` vs settled, ignoring gas.
+    pub raw_bps: Option<f64>,
+    /// `fynd amount_out_net_gas` vs settled — Fynd's output after its own gas cost.
+    pub net_bps: Option<f64>,
+}
+
+impl Deltas {
+    const NONE: Self = Self { raw_bps: None, net_bps: None };
+}
+
+/// Win/loss/unsolvable classification for a single trade, judged at one block state.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize)]
+#[serde(rename_all = "lowercase")]
+pub(crate) enum Verdict {
+    /// Fynd would have produced strictly more output than settled (net of gas).
+    Win,
+    /// Fynd's output was equal to or worse than settled, or it could not be compared.
+    Loss,
+    /// Fynd could not solve the trade at all.
+    Unsolvable,
+}
+
+/// Compute raw and net-of-gas bps deltas of `outcome` against `settled_amount_out`.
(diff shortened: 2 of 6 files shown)


tools/hindsight/src/resolve/compare.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+/// Classify a trade by its net-of-gas delta at the given outcome (top-of-block is the default
+/// reference). Fynd wins only when it delivers strictly more output after gas.
+pub(crate) fn verdict(outcome: &Outcome, settled_amount_out: U256) -> Verdict {
+    match compare(outcome, settled_amount_out).net_bps {
+        Some(d) if d > 0.0 => Verdict::Win,
+        Some(_) => Verdict::Loss,
+        None => Verdict::Unsolvable,
+    }
```
kayibal, line 106: 🤖 **Verdict compares net-of-gas Fynd output against gross settled output.** `amount_out_net_gas` (Fynd output minus Fynd's own gas) is compared to the on-chain `settled_amount_out`, which is gross — the settled swap's gas was paid separately in ETH and never subtracted. The baseline is gross but Fynd's side is net: an asymmetric comparison that systematically penalizes Fynd in the headline verdict. The `Verdict::Win` doc ('strictly more output than settled (net of gas)') implies a symmetric net comparison the code doesn't perform. If this is an accepted approximation pending N-1/N gas accounting, state it at the verdict boundary so a 'net' win isn't read as symmetric.

tools/hindsight/src/resolve/run.rs
```diff
@@ -0,0 +1,276 @@
...
+pub(crate) fn summarize(comparisons: &[Comparison]) -> Summary {
+    let mut raw: Vec<f64> = Vec::new();
+    let mut net: Vec<f64> = Vec::new();
+    let mut summary = Summary { total: comparisons.len(), ..Default::default() };
+
+    for cmp in comparisons {
+        match cmp.verdict {
+            Verdict::Win => summary.wins += 1,
+            Verdict::Loss => summary.losses += 1,
+            Verdict::Unsolvable => summary.unsolvable += 1,
```
kayibal, line 152: 🤖 **Coverage-miss signal is collapsed into the unsolvable bucket.** A partial fill reclassified by `served()` as `Outcome::Unsolvable("partial fill: …")` is counted here in the same `summary.unsolvable` total as genuine unsolvables (missing token, timeout, no route). The served/coverage-miss distinction this PR introduces is then unrecoverable in the summary — a reader can't tell a coverage miss from an un-quotable pair. A distinct counter, or a documented decision to merge them, keeps the signal.

tools/hindsight/src/resolve/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+#[derive(Debug, Clone, Serialize)]
+pub(crate) struct Comparison {
+    pub tx_hash: String,
+    pub block_number: u64,
+    pub client: String,
+    pub aggregator: String,
+    pub token_in: Address,
+    pub token_out: Address,
+    pub amount_in: U256,
+    pub settled_amount_out: U256,
```
tamaralipows, line 59: Would it perhaps be better just to have a ref of `DecodedTrade` here instead of restating these fields

tools/hindsight/src/resolve/run.rs
```diff
@@ -0,0 +1,276 @@
...
+        .with_timeout(Duration::from_millis(timeout_ms))
+        .with_retry(RetryConfig::new(1, Duration::from_millis(0), Duration::from_millis(0)))
+        .build_quote_only()
+        .map_err(|e| anyhow::anyhow!("failed to build Fynd client: {e}"))?;
+    let resolver =
+        FyndReSolver { aggregator: FyndAggregator::new(Arc::new(client), timeout_ms, 0.0) };
+
+    let mut comparisons = Vec::new();
+    for block_number in &blocks {
+        let trades = decode_block(&provider, *block_number).await?;
```
tamaralipows: Whole run will die here on a block decode failure

tools/hindsight/src/resolve/run.rs
```diff
@@ -0,0 +1,276 @@
...
+    }
+    match summary.median_net_bps {
+        Some(d) => println!("  median net-gas: {d:+.2} bps"),
+        None => println!("  median net-gas: n/a"),
+    }
+    println!("{}", "=".repeat(60));
+}
+
+/// Decode `blocks`, re-solve every trade through the Fynd instance at `fynd_url`, and report.
+pub(crate) async fn run(
```
tamaralipows, line 201: I worry that when this is run locally there is no guardrail against accidentally resolving a past block on the present fynd state (which would produce nonsensical values). Afaiu this isn't ensured in the step-controller PR either - it adds monitor, which avoids the problem by stepping the solver to the actual block, but resolve itself still has no guard

tools/hindsight/src/resolve/run.rs
```diff
@@ -0,0 +1,276 @@
...
+        .as_deref()
+        .and_then(parse_u256)
+    else {
+        return Outcome::Unsolvable("missing amount_out".to_string());
+    };
+    let amount_out_net_gas = quote
+        .amount_out_net_gas
+        .as_deref()
+        .and_then(parse_u256)
+        .unwrap_or(amount_out);
```
tamaralipows: This might bias the net gas results in Fynd's favour if no net gas result is found (I guess if gas is unavailable for that block for some reason?) 
Don't expect it to happen often but might be worth noting...

tools/hindsight/src/main.rs
```diff
@@ -73,6 +76,33 @@ struct VerifyArgs {
     json: bool,
 }
 
+#[derive(Args)]
+struct ResolveArgs {
+    /// Ethereum RPC URL (used to decode the block's settled trades)
+    #[arg(long, env = "ETH_RPC_URL")]
```
tamaralipows: `RPC_URL` in preparation for supporting other chains

tools/hindsight/src/resolve/run.rs
tamaralipows: Fynd actually doesn't support partial fills... If this is a future consideration we should state it clearly

tamaralipows (commented):
Thank you, mostly looks good - mainly confused by the mentions of nonexistent partial fills 😬

tools/hindsight/src/resolve/compare.rs
kayibal: I think this is ok. It's on the conservative side of things. Should be clearly documented though that this is on purpose. I think trying to isolate how much the trade costed on-chain is hard, especially if the trade is somewhere deeper in the call stack and there are other things happening in the transaction.

tools/hindsight/src/resolve/run.rs
kayibal: It doesn't support price constrained partial fills, but if liquidity is limited it may return a partial trade, which is what this is referring to.

tools/hindsight/src/resolve/run.rs
kayibal: I don't like that this is just silently ignored. Would at least expect a warning here.

tools/hindsight/src/resolve/mod.rs
kayibal: Hm no I have a slight preference to restate them, keeps things focused to only the values we need here, keep things separate from the decoding layer. What would be the advantage of reusing it?

tools/hindsight/src/resolve/compare.rs
kayibal: Documented at the verdict boundary (`verdict()` doc + `Verdict::Win`) that comparing net-of-gas Fynd output against the gross settled amount is intentional and conservative: the settled swap's gas was paid separately in ETH and isn't subtracted, isolating it from a deeper call stack is hard, so a `Win` under-counts rather than over-counts Fynd's edge. Symmetric N-1/N gas accounting is noted as a follow-up. — 09a2551

tools/hindsight/src/resolve/run.rs
kayibal: Split the signal out: added an `Outcome::Partial` state and a `Verdict::CoverageMiss` so a liquidity-limited partial route is counted on its own `coverage miss` line in the summary, separate from genuine unsolvables. Also reworded "partial fill" → "partial route" to match @tamaralipows's point — Fynd doesn't do price-constrained partial fills; this is the thin-liquidity case where it covers only part of the requested size, which is now stated in the `served`/`Outcome::Partial` docs. — e4e040c

tools/hindsight/src/resolve/run.rs
kayibal: Fixed: `run()` now warns and continues on a block decode failure instead of aborting the whole range, matching the `decode`/`verify` drivers. — 9ea4a8c

tools/hindsight/src/resolve/run.rs
kayibal: Added the guardrail: `resolve` fetches chain head and warns when the newest requested block trails head by more than 100 blocks (~20 min on Ethereum), since Fynd solves at current state and comparing that to a much older settled amount is meaningless. It's a warning rather than a hard block, so the near-head default path is unaffected and an intentional recent range still runs. — 7bd4869

tools/hindsight/src/resolve/run.rs
kayibal: No longer silent: `quote_to_outcome` now warns when a quote is missing `amount_out_net_gas` before falling back to the raw `amount_out`, noting the net-of-gas delta is biased in Fynd's favour for that trade. — 1a6bb16

tools/hindsight/src/main.rs
kayibal: Already `RPC_URL` on this branch — the env-var rename was swept through `Resolve`/`Verify`/`Decode` in the 6122 change this PR is based on, so `ResolveArgs` uses `#[arg(long, env = "RPC_URL")]`. Marking this outdated.

tamaralipows (approved):

PR #265, opened 2026-06-30
feat(hindsight): metrics, USD valuation & /metrics exporter (ENG-6124)

kayibal:
Prometheus metrics (trades, net-of-gas bps) with a dedicated /metrics exporter and Grafana dashboard. USD valuation via Fynd token prices anchored to USD through stablecoins, plus per-client observed volume and client-uplift (losses floored at zero). Supersedes the earlier stablecoin-leg/±band USD approach. Stacked PR #4/6; base `eng-6123`.

### DIFF
--- a/tools/hindsight/src/telemetry.rs
+++ b/tools/hindsight/src/telemetry.rs
@@ -4,7 +4,6 @@
 //! and serve its rendered output on a dedicated port via Actix.
 
 use actix_web::{web, App, HttpResponse, HttpServer, Responder};
-use alloy::primitives::Address;
 use metrics::{
     counter, describe_counter, describe_gauge, describe_histogram, gauge, histogram, Unit,
 };
@@ -34,12 +33,6 @@ pub(crate) fn outcome_label(verdict: Verdict) -> &'static str {
     }
 }
 
-/// Pair label as lowercase `token_in->token_out` addresses. Symbol resolution is a future
-/// enhancement; raw addresses keep label cardinality deterministic.
-pub(crate) fn pair_label(token_in: Address, token_out: Address) -> String {
-    format!("{token_in:#x}->{token_out:#x}")
-}
-
 /// Fraction of trades that were comparable (solvable) out of the total seen.
 pub(crate) fn coverage_ratio(total: usize, comparable: usize) -> f64 {
     if total == 0 {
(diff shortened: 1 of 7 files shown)


tools/hindsight/src/telemetry.rs
```diff
@@ -0,0 +1,318 @@
...
+}
+
+/// Record a single re-solve comparison. `prices` is the solver's token-price snapshot used to value
+/// savings in USD; an empty map disables USD recording.
+pub(crate) fn record(cmp: &Comparison, chain: &str, prices: &usd::PriceMap) {
+    counter!(
+        TRADES_TOTAL,
+        "client" => cmp.client.clone(),
+        "aggregator" => cmp.aggregator.clone(),
+        "pair" => pair_label(cmp.token_in, cmp.token_out),
```
tamaralipows: I vaguely remember having had issues before with memory since prometheus creates a separate timeseries for each combination of the label values here afaik. This might get crazy for a lot of token pairs over a long time... perhaps we should keep this to a fixed set, drop this column entirely, or find a different way to record this? Otherwise we can keep our eyes open for issues in the future (not entirely sure it's the same issue).

With our integration test I think this happened when we were recording every single revert reason - we eventually just had to group them into the top X revert reasons, and then "other", if I remember correctly.

tamaralipows (commented):
Thank you - only one thing stuck out to me as a red flag here, but not sure if it's a real issue

tamaralipows (approved):

tools/hindsight/src/telemetry.rs
kayibal: Good catch — you're right, this is unbounded: token pairs are effectively limitless on mainnet, and multiplied by client/aggregator/chain/outcome a long `monitor` run would blow up Prometheus series cardinality. (The old `pair_label` doc claiming addresses keep cardinality "deterministic" conflated deterministic with bounded.)

Dropped the `pair` label from `hindsight_trades_total` entirely (and removed the `pair_label` helper) in fba2823. The counter now carries only client/aggregator/chain/outcome — all bounded, small sets. The per-pair breakdown already lives in the JSONL comparison output (each row has token_in/token_out), which is the right place for that high-cardinality slice; Prometheus keeps the aggregate view. Went with a clean drop over a top-X + "other" allowlist to avoid maintaining a pair set whose long tail would collapse into "other" anyway.

PR #266, opened 2026-06-30
docs(hindsight): tool README (ENG-6125)

kayibal:
README: setup, subcommands, configuration, metrics + Grafana, USD methodology, limitations. Stacked PR #5/6; base `eng-6124`.

### DIFF
--- a/tools/hindsight/README.md
+++ b/tools/hindsight/README.md
@@ -1,0 +1,128 @@
+# Hindsight
+
+Hindsight measures how Fynd would have performed on **real, settled aggregator swaps**. For each
+block it decodes the aggregator trades that actually executed on-chain, re-solves the same
+`(token_in, token_out, amount_in)` through Fynd, and compares Fynd's output against what settled.
+
+It is the live counterpart to the `fynd-benchmark audit` tool: where `audit` compares Fynd against
+other aggregators' *quotes*, Hindsight compares Fynd against what those aggregators *actually
+delivered* on-chain.
+
+## Build
+
+```bash
+cargo build -p hindsight --release
+```
+
+The binary is `hindsight`. Run any subcommand with `--help` for the full option list.
+
+## Subcommands
+
+### `decode` — decode settled aggregator swaps
+
+Decodes every aggregator trade that settled in a block (or range) straight from on-chain data over
+a standard RPC. No external data provider is required.
+
+```bash
+hindsight decode --rpc-url "$ETH_RPC_URL" --block 21000000
+hindsight decode --rpc-url "$ETH_RPC_URL" --range 21000000-21000010 --json
+```
+
+Each decoded trade carries two levels of attribution (see [Methodology](#methodology)): the
+**client** that initiated the trade and the **aggregator** that settled it.
+
+### `verify` — check the decoder against Allium
+
+Decodes a block locally and diffs every transaction against [Allium](https://docs.allium.so/)'s
+`aggregator_trades` ground-truth dataset (tokens, amounts in bps, aggregator attribution),
+producing a worklist of gaps and mismatches.
+
+```bash
+hindsight verify --rpc-url "$ETH_RPC_URL" \
+  --allium-key "$ALLIUM_API_KEY" \
+  --range 21000000-21000010 \
+  --tolerance-bps 50
+```
+
+### `resolve` — re-solve through Fynd and compare
+
+Decodes a block's trades and re-solves each through a **running Fynd instance**, reporting how
+Fynd compares to what settled.
+
+```bash
+# Requires a healthy Fynd solver reachable at --fynd-url.
+hindsight resolve \
+  --rpc-url "$ETH_RPC_URL" \
+  --fynd-url http://localhost:3000 \
+  --range 21000000-21000005
+
+# Serve Prometheus metrics while running (keeps serving until ctrl-c):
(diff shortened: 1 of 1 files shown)


tools/hindsight/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Configuration
+
+| Variable / flag | Used by | Purpose |
+|---|---|---|
+| `ETH_RPC_URL` / `--rpc-url` | all | Ethereum RPC endpoint for decoding |
+| `FYND_URL` / `--fynd-url` | `resolve` | Base URL of a running Fynd solver (default `http://localhost:3000`) |
+| `ALLIUM_API_KEY` / `--allium-key` | `verify` | Allium API key |
+| `--chain` | `resolve` | Chain label applied to metrics (default `ethereum`) |
+| `--metrics-port` | `resolve` | Serve Prometheus `/metrics` on this port |
```
kayibal, line 74: 🤖 **Config table marks `--chain`/`--metrics-port` as `resolve`-only.** Both flags exist on `MonitorArgs` too, and `monitor` is the only command that actually emits USD metrics — so scoping `--metrics-port` to `resolve` is misleading. Update the 'Used by' column to include `monitor`.

tools/hindsight/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Metrics & dashboard
+
+With `--metrics-port`, `resolve` exposes Prometheus metrics at `/metrics`:
+
+| Metric | Type | Labels |
+|---|---|---|
+| `hindsight_trades_total` | counter | `client`, `aggregator`, `pair`, `chain`, `outcome` |
+| `hindsight_savings_bps` | histogram | `client`, `aggregator`, `chain` |
+| `hindsight_savings_usd` | histogram | `client`, `aggregator`, `chain` |
```
kayibal, line 85: 🤖 **Metrics section attributes USD metrics to `resolve`, but only `monitor` emits them.** `resolve` passes an empty `PriceMap`, so `hindsight_savings_usd` is never emitted there — only `monitor` produces USD metrics. The table also omits two emitted metrics, `hindsight_improvement_usd` and `hindsight_volume_usd`. Scope this section to `monitor` for the USD metrics and add the two missing rows.

tools/hindsight/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+- **Two execution modes.** `resolve` compares at the chain's *current* state (re-solving through a
+  separate running Fynd over HTTP). `monitor` provides the full top-of-block / back-of-block range
+  via an in-process stepped solver. `monitor`'s end-to-end run is exercised by a gated integration
+  test that needs a reachable Tycho endpoint (`TYCHO_URL` + `TYCHO_API_KEY`); the comparison/range
+  logic is unit-tested independently.
+- **USD savings need a stablecoin leg.** Savings are valued in USD (`hindsight_savings_usd`) when
+  either leg is a known stablecoin: the output leg is taken at peg directly, or the input leg
+  provides the trade's USD notional and the output delta is valued at the settled trade's implied
+  price. Fynd exposes no public token→USD conversion, so trades with no stablecoin leg are reported
```
kayibal, line 121: 🤖 **Stale limitation — USD valuation no longer needs a stablecoin leg.** `usd.rs` values any output token Fynd has priced (anchoring ETH→USD via stablecoins), and the test `output_leg_non_stable_token_priced_via_eth` proves a non-stablecoin pair (USDC→WETH) gets a USD value. This paragraph describes the superseded stablecoin-leg-only approach and understates USD coverage. Update it to the ETH-anchored model.

tools/hindsight/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+**Client vs aggregator.** A trade has two attribution levels: the **client** (the platform that
+initiated it, e.g. Relay or Matcha — who Fynd is compared against) and the **aggregator** (the solver
+the client routed through, e.g. 1inch or 0x — who to blame when a trade settled worse than Fynd).
+When a user swaps directly through an aggregator router, client and aggregator are the same.
+
+**Comparison.** Fynd's output is compared to the settled amount as a basis-point delta, both raw and
+net of Fynd's estimated gas. A trade is a **win** when Fynd's net-of-gas output strictly exceeds the
+settled amount, a **loss** otherwise, and **unsolvable** when Fynd cannot produce a quote (missing
+token in Tycho, insufficient liquidity, timeout).
```
tamaralipows, line 109: This README doesn't express the two ways to read the savings, which is the key thing a client needs. A short note here distinguishing `hindsight_savings_usd` (signed, losses included - "if you routed everything through Fynd") from `hindsight_improvement_usd` (losses floored at zero - "incremental value of adding Fynd, since a client keeps the incumbent when Fynd loses") could be useful

tools/hindsight/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+```bash
+hindsight decode --rpc-url "$ETH_RPC_URL" --block 21000000
+hindsight decode --rpc-url "$ETH_RPC_URL" --range 21000000-21000010 --json
+```
+
+Each decoded trade carries two levels of attribution (see [Methodology](#methodology)): the
+**client** that initiated the trade and the **aggregator** that settled it.
+
+### `verify` — check the decoder against Allium
```
tamaralipows, line 34: These don't mention that we will have to have saved the query to Allium

tamaralipows (commented):
Thank you - I might have more comments here depending on your answer to my partial fill question on the other PR

PR #267, opened 2026-06-30
feat(hindsight): live two-state monitor + fynd-core feed/step-controller

kayibal:
fynd-core: feed survives a transient absence of MarketEvent subscribers (was killing the whole solver), plus BlockStepController solver/feed wiring. hindsight: live in-process two-state monitor (top-of-block N-1 / back-of-block N) via the step controller, requesting encoding so each quote carries calldata, appending every improvement (complete quote + calldata + block + settled tx) to --improvements-jsonl, with a 30-minute block-settle timeout to ride out Tycho stream resync pauses. Stacked PR #6/6; base `eng-6125`. Note: BlockStepController wiring overlaps #243; rebase if that lands first.


fynd-core/src/feed/tycho_feed.rs
```diff
@@ -439,6 +442,397 @@ impl TychoFeed {
         Ok(())
     }
 
+    /// Like [`run_with_pending`](Self::run_with_pending) but also gates each block behind a
+    /// [`BlockStepController`].
+    ///
+    /// Both the [`PendingBlockProcessor`] and the [`BlockStepController`] are delivered before
+    /// the first block is processed. Only valid when at least one non-RFQ protocol is configured.
+    pub(crate) async fn run_with_pending_and_step_controller(
```
kayibal: 🤖 **Four near-identical feed driver methods.** `run_with_pending_and_step_controller` (here) and `run_with_step_controller` (line 657) each duplicate ~150 lines of token-loading, stream-building, RFQ spawn, and a byte-for-byte identical `tokio::select!` driver loop already in `run`/`run_with_pending`. The shared loop and the 'all-RFQ → error' + 'load tokens → send Err on channel(s)' preamble are extractable into one helper taking the built stream plus optional `controller_tx`/`pending_tx`. As-is, a fix to the select! loop (like the broadcast-send change in this PR) must be hand-applied in four places.

tools/hindsight/src/resolve/monitor.rs
```diff
@@ -0,0 +1,427 @@
...
+    async fn advance(&self) -> anyhow::Result<()> {
+        let before = self.current_block().await;
+        self.controller
+            .trigger_next_block()
+            .map_err(|_| anyhow::anyhow!("block stream ended"))?;
+
+        // Deterministic barrier: wait until the solver applies a block strictly newer than
+        // `before`.
+        let deadline = Instant::now() + BLOCK_SETTLE_TIMEOUT;
+        loop {
```
kayibal, line 134: 🤖 **Back-of-block solve isn't pinned to the target block.** `advance()` releases a block and waits only until the applied block satisfies `now > before`, but the decoded trades are for `target = top_block + 1` regardless of which block the Tycho stream actually applies next. On a reorg/gap/resync the back-of-block solve runs against a state that isn't `target`, silently comparing `back` against a different block's trades. `peek_next_block()` exposes the block number — gating on `applied == target` (or logging/skipping when `now != target`) makes the pairing trustworthy. The top-of-block N-1 headline is unaffected.

tools/hindsight/src/resolve/monitor.rs
```diff
@@ -0,0 +1,427 @@
...
+    }
+}
+
+fn order_quote_to_outcome(quote: &OrderQuote) -> Outcome {
+    if quote.status() != QuoteStatus::Success {
+        return Outcome::Unsolvable(format!("{:?}", quote.status()));
+    }
+    // Capture the complete quote (route, per-hop pools/amounts, and the encoded transaction) so
+    // improvements can be dumped for later investigation.
+    let quote_json = serde_json::to_string(quote).ok();
```
tamaralipows: Conversion to string only to convert later back to a json? Is that just to keep the datatypes of `SolvedAmount` simple?

tools/hindsight/src/telemetry.rs
```diff
@@ -143,6 +153,90 @@ pub(crate) fn record(cmp: &Comparison, chain: &str, prices: &usd::PriceMap) {
     }
 }
 
+/// Record a two-state range, anchored on the headline (top-of-block) state. `prices` is the
+/// solver's token-price snapshot used to value savings in USD; an empty map disables USD recording.
+pub(crate) fn record_range(range: &RangeComparison, chain: &str, prices: &usd::PriceMap) {
```
tamaralipows: Near identical to the `record` method

tools/hindsight/src/telemetry.rs
```diff
@@ -143,6 +153,90 @@ pub(crate) fn record(cmp: &Comparison, chain: &str, prices: &usd::PriceMap) {
...
+        histogram!(
+            SAVINGS_BPS,
+            "client" => range.client.clone(),
+            "aggregator" => range.aggregator.clone(),
+            "chain" => chain.to_string(),
+        )
+        .record(bps);
+    }
+
+    if let Outcome::Solved(solved) = &range.top.outcome {
```
tamaralipows: What about `range.back`?

tools/hindsight/src/main.rs
```diff
@@ -113,6 +116,54 @@ struct ResolveArgs {
     json: bool,
 }
 
+#[derive(Args)]
+struct MonitorArgs {
+    /// Ethereum RPC URL (used to decode each block's settled trades and feed the solver)
+    #[arg(long, env = "ETH_RPC_URL")]
```
tamaralipows: `RPC_URL`

tools/hindsight/src/resolve/monitor.rs
```diff
@@ -0,0 +1,427 @@
...
+        }
+        let elapsed_s = start.elapsed().as_secs_f64();
+        crate::telemetry::record_block_seconds(elapsed_s);
+
+        total_trades += ranges.len();
+        comparable_trades += ranges
+            .iter()
+            .filter(|r| matches!(r.verdict, Verdict::Win | Verdict::Loss))
+            .count();
+        crate::telemetry::record_coverage(total_trades, comparable_trades);
```
tamaralipows: Can we move these imports to the top of the file

tamaralipows (commented):
Few comments for now

tools/hindsight/src/resolve/monitor.rs
```diff
@@ -0,0 +1,427 @@
...
+        };
+        let target = top_block + 1;
+
+        let trades = match decode_block(&provider, target).await {
+            Ok(trades) => trades,
+            Err(e) => {
+                warn!(block = target, "decode failed, skipping block: {e}");
+                adapter.advance().await?;
+                continue;
+            }
```
tamaralipows, line 298: `target` is `current_block + 1`, which tracks the Tycho stream's tip, but we decode it over a separate HTTP RPC. When that RPC lags the Tycho stream, `decode_block` returns "block N not found" and we skip. I hit this when running locally: the monitor logged "not found" ~2s after each block's timestamp and skipped every block in a row, so the run produced no comparisons at all, just warnings.

1. We should likely distinguish this race condition from a genuine decode failure, and retry with a backoff
2. When we do skip, we should maybe count the skipped blocks.

Logs for reference:
```
2026-07-02T16:48:24.968303Z  WARN decode failed, skipping block: block 25446121 not found block=25446121
2026-07-02T16:48:37.016314Z  WARN decode failed, skipping block: block 25446122 not found block=25446122
2026-07-02T16:48:50.024229Z  WARN decode failed, skipping block: block 25446123 not found block=25446123
2026-07-02T16:49:02.217211Z  WARN decode failed, skipping block: block 25446124 not found block=25446124
2026-07-02T16:49:13.088667Z  WARN decode failed, skipping block: block 25446125 not found block=25446125
2026-07-02T16:49:24.917008Z  WARN decode failed, skipping block: block 25446126 not found block=25446126
```

tools/hindsight/src/resolve/monitor.rs
tamaralipows, line 1: I've been looking deeper into this. Right now Hindsight has two commands that overlap but give different data, and I'm not sure that split should exist. We want to make this as easy as possible for users to run and receive somewhat similar results to our public grafana dashboard. 

`resolve` only compares at the chain tip afaiu (limitation of using FyndClient)... but we want top-of-block and back-of-block in both local and cluster runs. Wouldn't it be better to have one engine run two ways: locally it writes a full JSON, in the cluster it feeds the Grafana dashboard, same info in both, and not having two commands to choose from? What do you think? Or am I missing something

kayibal: I agree resolve should not exist. What we actually want is monitor.

fynd-core/src/feed/tycho_feed.rs
```diff
@@ -549,9 +943,12 @@ impl TychoFeed {
...
 
-            self.event_tx
-                .send(market_update_event)
-                .map_err(|e| DataFeedError::EventChannelError(e.to_string()))?;
+            // A broadcast send fails only when no receivers are currently subscribed. The market
+            // state was already updated above; this event is just a notification, so a transient
+            // absence of subscribers must not kill the feed — that would stop the whole solver.
+            if let Err(e) = self.event_tx.send(market_update_event) {
+                warn!(error = %e, "no market-event subscribers; skipping notification");
+            }
```
kayibal: I think this can lead to our graph structs on the worker to become out of date. I'd prefer to revert this change for now and instead debug in the future what causes all subscribers to exit.

fynd-core/src/solver.rs
```diff
@@ -913,7 +917,148 @@ impl FyndBuilder {
...
+    /// a block before it is decoded. Dropping the controller ungates the stream so it runs to its
+    /// natural end.
+    ///
+    /// Only valid when at least one non-RFQ protocol is configured.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`SolverBuildError`] if any component fails to initialize, all protocols are RFQ,
+    /// or the step-controller channel closes before the controller is delivered.
+    pub async fn build_with_step_controller(
```
kayibal, line 939: feature gate under experimental. I'm unsure about the api atm,

tools/hindsight/src/decoder/mod.rs
```diff
@@ -21,11 +21,21 @@ use tracing::warn;
...
 
+/// Sub-type of a Relay solver-initiated rebalancing fill (see [`net::decode_relay_rebalance`]).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Serialize)]
+#[serde(rename_all = "lowercase")]
+pub(crate) enum RelayFill {
+    /// Output delivered to an external recipient — a cross-chain order fill.
+    External,
+    /// Output returns to Relay's fee collector — an internal inventory rebalance.
+    Internal,
+}
```
kayibal: This seems purely informational and it is not really relevant for us so it should be removed.

tools/hindsight/src/decoder/mod.rs
```diff
@@ -127,8 +176,13 @@ pub(crate) async fn decode_block<P: Provider>(
...
 
-        let aggregator =
-            attribute_aggregator(&root, entry_point, sender, aggregators).unwrap_or(entry_point);
+        // Back out any input-side fee a known client collector skimmed before the swap, so the
+        // re-solve compares Fynd against the amount actually routed (not the user's gross spend).
+        let client_fee = fee_to_collectors(logs, &native, &fee_collectors)
+            .get(&token_in)
+            .copied()
+            .filter(|fee| !fee.is_zero());
+        let amount_in = client_fee.map_or(amount_in, |fee| amount_in.saturating_sub(fee));
```
kayibal: We also need to consider amount_out fee taking here.

tools/hindsight/src/resolve/monitor.rs
kayibal: Unlikely to happen but probably worth a warning line

tools/hindsight/src/resolve/monitor.rs
```diff
@@ -0,0 +1,664 @@
...
+    let bytes: &[u8] = address.as_ref();
+    (bytes.len() == 20).then(|| Address::from_slice(bytes))
+}
+
+/// Append one JSON line per re-solved trade to `writer` — every comparison, not just wins. Each
+/// record carries both block states with their verdict (win/loss/unsolvable), so downstream can
+/// filter to wins for the improvement view or to unsolvables for the coverage worklist (where Fynd
+/// needs to improve). Losses keep their route (what path Fynd took and lost on); unsolvables keep
+/// the reason.
+fn write_comparisons<W: std::io::Write>(
```
kayibal: I think the jsonl writing can be put in a separate module

tools/hindsight/src/resolve/monitor.rs
kayibal: Nice catch and absolutely agree!

fynd-core/src/feed/tycho_feed.rs
kayibal: I would prefer if we removed the run_with_pending methods added in this PR since they are not used atm.

kayibal: Removed the unused `run_with_pending_and_step_controller` and its `build_with_pending_and_step_controller` — no callers — which drops ~270 lines of duplicated stream-driver code and the four-way method fan-out. `run` / `run_with_pending` / `run_with_step_controller` remain. — c63bacf

tools/hindsight/src/resolve/monitor.rs
kayibal: Added a warning line: after `resolve_block_range` the monitor checks the applied block against `target = top+1` and warns when they differ (reorg/gap/resync), so a mispaired back-of-block state is visible. The top-of-block headline is unaffected. — e967b7c

tools/hindsight/src/resolve/monitor.rs
kayibal: Yes — `quote_json: Option<String>` keeps `SolvedAmount` a plain `Serialize` type and, crucially, lets us store a pre-projected *slim* quote. We deliberately never serialize the whole `OrderQuote`: it embeds each hop's `protocol_state`, which dominates size and fails to serialize for vm pools (Curve etc.) — that would drop the route for exactly the deep-liquidity stable trades we care about. So the string is the slim projection, parsed back only when writing the JSONL. (Outdated after the rebase; resolving.)

tools/hindsight/src/telemetry.rs
kayibal: Resolved by removing `resolve`: `record` was the resolve-only path and is deleted, so only `record_range` remains — no more duplication. — 7c00f15

tools/hindsight/src/telemetry.rs
kayibal: Good catch — `range.back` now reaches Prometheus. `record_range` records both states, each tagged with a `state="top"|"back"` label on the trade / savings-bps / savings-usd / improvement-usd metrics (volume stays per-trade, recorded once), with back-of-block USD valued at the back-of-block price snapshot. — 2f6f39c

tools/hindsight/src/main.rs
kayibal: Already `RPC_URL` on this branch (the env-var rename was swept through the whole tool earlier). Marking outdated.

tools/hindsight/src/resolve/monitor.rs
kayibal: Pulled the inline `crate::telemetry::*` / `crate::provider_from` fully-qualified calls up into top-of-file `use` imports. — 6d6eadf

tools/hindsight/src/resolve/monitor.rs
kayibal: Nice catch, and agreed. The monitor now distinguishes the RPC-lag race from a genuine failure: before decoding it waits for the RPC head to reach `target` (bounded retries with backoff), so a first "not found" no longer skips the block. Genuine skips are counted and exposed as `hindsight_skipped_blocks_total`. — 2f63a24

tools/hindsight/src/resolve/monitor.rs
kayibal: Agreed — removed the `resolve` subcommand entirely (run.rs, the CLI command, the HTTP re-solve path, and `telemetry::record`). `monitor` is the one engine now: in-process, two-state (top/back), writing the full JSONL locally and feeding the Grafana metrics in the cluster — same info both ways. — 7c00f15

fynd-core/src/feed/tycho_feed.rs
kayibal: Reverted — the market-event send propagates its error again (restoring `DataFeedError::EventChannelError`) rather than tolerating an absent subscriber, so a dropped event surfaces instead of risking stale worker graphs. The underlying all-subscribers-exit will be debugged separately. — 1d6a760

fynd-core/src/solver.rs
kayibal: Gated the step-controller build path (`build_with_step_controller`, `run_with_step_controller`, the `BlockStepController` re-export, and the `StepControllerChannelClosed` error) behind a new `experimental` fynd-core feature; the hindsight monitor enables it. It compiles out cleanly when the feature is off (verified). — fba9de8

tools/hindsight/src/decoder/mod.rs
kayibal: Removed the `RelayFill` enum and the `relay_fill` field/tag (informational). `decode_relay_rebalance` still decodes Relay solver rebalancing fills as ordinary trades — just untagged. — 461637a

tools/hindsight/src/decoder/mod.rs
kayibal: Handled: a client fee can be skimmed from either side, so alongside the input-side back-out we now add an output-side skim back into `amount_out` (keyed by `token_out` from `fee_to_collectors`), making the settled amount the gross swap output — the like-for-like basis against Fynd's gross output. Recorded as `client_fee_out`. — 1d61a6e

tools/hindsight/src/resolve/monitor.rs
kayibal: Extracted the comparison-record building and the slim quote/route/transaction projection into a new `resolve::jsonl` module; monitor now just drives the solver and delegates output. — 8bb6d82

tamaralipows (approved):
Let's merge this? 👀

PR #268, opened 2026-06-30
feat(derived): make ComputationManager a registry of computations

sunce86:
Replaces `ComputationManager`'s three hardcoded computation fields with a small registry of typed computations run in dependency-stage order. Resolves the registry TODO in `derived/manager.rs`: adding a derived computation no longer means editing the manager struct and hand-writing a `compute_all` phase.

Behavior is preserved for the default set - `spot_prices` runs first, `token_prices` and `pool_depths` run after fresh `spot_prices`, partial successes still satisfy dependents, and failures cascade as before; the existing `derived` tests pass unchanged.

Done:
- `ComputationManager` holds `Vec<Box<dyn ErasedComputation>>`; computations run in stages derived from `DerivedComputation::requirements()`.
- `DerivedComputation` gains defaulted `requirements()` (fresh/stale upstreams) and `persist()` (store the output; defaults to a generic slot). `ErasedComputation` is the object-safe bridge; the typed trait keeps its `Output`/`const ID`.
- `DerivedData` keeps outputs in a `ComputationId`-keyed, type-erased slot map behind the existing typed getters/setters; per-computation failure maps stay typed (their merge logic is output-specific).
- `register` is fallible, rejecting duplicate ids via `ComputationError::DuplicateComputationId`; requirement cycles are reported as `ComputationFailed` rather than run out of order.
- Add scheduler unit tests (chains, diamonds, cycles, stale/unregistered deps) and characterization tests pinning the event sequence (stage order, cascades, partial-success propagation, default-set flow).
- Clarify: the old `compute_all` comment called `PoolDepthComputation` dependency-free, but it hard-requires spot prices; `requirements()` now matches the actual graph (and the `derived/mod.rs` diagram).

`register` stays `pub(crate)`, since it's internal refactor, no public API change. A follow-up could expose computations as a public extension point (like `with_algorithm`).


kayibal (commented):
Thank you @sunce86 ! Looks good to me. It seems we have some issues with CI from external contributors, could you please update your branch? We might just bypass those checks.

sunce86:
Updated. Seems like the CI requires manual approval to run: https://docs.github.com/en/actions/how-tos/manage-workflow-runs/approve-runs-from-forks

kayibal:
@sunce86 Yeah there is an issue with available secrets when ran from forks. I will run those locally. Could you update the `crossbeam-epoch` crate it seems like a new CVE appeared for the version that was locked here previously.

Troshchk (commented):
Thank you, looks good to me too @sunce86!

PR #269, opened 2026-07-02
feat: gas-aware sequential shared-pool evaluation in path_frank_wolfe

Haikane:
## What this does

`path_frank_wolfe` now evaluates and reports split routes with one executable semantics.

- **Sequential shared-pool simulation.** `execute_split_plan` merges shared hops, orders swaps topologically (Kahn), and simulates them once with a threaded post-swap state map. The line search objective (`evaluate_total_output`), candidate discovery state (`build_post_swap_overrides`) and the emitted route (`build_split_route`) all run this same plan. Previously each path was simulated from fresh pool state, so splits that reuse a pool double-counted liquidity and quoted more output than the route can deliver.
- **Net-of-gas objective with an activation rule.** The golden-section line search scores gross output minus gas in output-token terms, and a candidate path is only activated when the split at the chosen step nets more than the current allocation at step 0. A path must pay for its own gas before it gets flow.
- **Single-path fallback on error.** The split search runs in `optimize_split`; any failure inside it returns the valid single-path route instead of failing the solve.
- Merged swaps carry simulated amounts and gas (not per-path sums), equal split fractions get a deterministic tie-break, and `docs/algorithms/path-frank-wolfe.md` describes the new behaviour.

## Evidence

Offline quality benchmark (frozen snapshot, 9,896 aggregator trades, independent re-execution of every returned route):

- Before: 46 routes over-quoted vs. what they deliver on-chain, worst by 1,256 bps (12.6%). After: 0.
- On delivered output, the new version returns a better route on 29 trades (up to +11 bps), an identical route on 5,843, a worse one on 4 (all ≤1.15 bps, rounding-scale), and solves 1 trade the old version failed.
- Solve time unchanged (~1% on the full run).

The over-quoting cases are splits sharing a pool: the old objective preferred routes whose reported output assumed each leg gets fresh reserves. A production price guard would reject those quotes as slippage misses.

## Tests

- New regression tests: shared-pool depletion in `evaluate_total_output`, rejection of a gas-losing candidate in the line search, route-order vs path-order consistency for same-component branches.
- `cargo nextest` (497 fynd-core tests), clippy `-D warnings`, nightly fmt, rustdoc: clean.

### DIFF
--- a/fynd-core/src/algorithm/path_frank_wolfe.rs
+++ b/fynd-core/src/algorithm/path_frank_wolfe.rs
@@ -169,7 +169,7 @@ impl PathFrankWolfeAlgorithm {
         current_allocations: &[PathAllocation],
         probe_amount: &BigUint,
     ) -> Result<Vec<SimulatedHop>, AlgorithmError> {
-        let mut overrides = build_post_swap_overrides(current_allocations, &ctx.market_data);
+        let mut overrides = build_post_swap_overrides(current_allocations, &ctx.market_data)?;
 
         // Pools committed in the current solution are executed once on-chain — their gas is
         // already priced into the combined transaction. Zero out protocol gas so BF doesn't
--- a/fynd-core/src/algorithm/path_frank_wolfe.rs
+++ b/fynd-core/src/algorithm/path_frank_wolfe.rs
@@ -4,12 +4,19 @@
 //! optimally split the input amount across them. The inner BF instance handles
 //! single-path discovery; this module layers on the Frank-Wolfe optimisation
 //! loop to determine the best split fractions.
+//!
+//! The optimisation objective is output net of gas, evaluated by simulating
+//! paths sequentially against a shared post-swap state map — paths that share
+//! a pool see its depleted reserves instead of double-counting liquidity. The
+//! single-path route is a floor: any failure while optimising the split falls
+//! back to it rather than failing the solve.
 
 use std::time::{Duration, Instant};
 
 use num_bigint::{BigInt, BigUint};
 use num_traits::{ToPrimitive, Zero};
 use tracing::debug;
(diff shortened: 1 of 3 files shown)


fynd-core/src/algorithm/path_frank_wolfe.rs
```diff
@@ -447,11 +489,111 @@ impl PathFrankWolfeAlgorithm {
...
 
         Ok(())
     }
 
+    /// Runs the Frank-Wolfe split search seeded from the single-path route.
+    ///
+    /// Returns `Ok(None)` when splitting is not worthwhile: price impact too
+    /// low to cover another path's gas, no second path found, or the split
+    /// route failed validation.
+    fn optimize_split(
```
tamaralipows, line 507: Note to whoever looks at this in future:

> Yes - `optimize_split` is almost entirely relocated code, not new logic. If you diff its body against the old inline version, it's the same Frank-Wolfe search line-for-line. The bulk you're seeing is just that the split search is a lot of code, and it now lives in one function instead of being spilled inline into find_best_route.
> 
>   1. Return type is now: Result<Option<RouteResult>> instead of Result<RouteResult>.
>   2. It exists to create one error boundary.

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -629,6 +611,201 @@ fn assign_splits_and_amounts(
     hops
 }
 
+fn build_in_degree(hops_by_token: &HashMap<Bytes, Vec<SplitSwap>>) -> HashMap<Bytes, usize> {
```
tamaralipows, line 618: Note to whoever reviews in the future - most of the changes in this file are just moving existing things around. For example this was already needed for Kahn's topological sort.

tamaralipows (approved):
Thank you!

PR #270, opened 2026-07-02
feat: add water_fill split-routing algorithm

Haikane:
## Summary

- Add the built-in `split` routing algorithm: a portfolio split router that splits one order across
  multiple parallel paths to reduce price impact on large trades.
- The portfolio returns the best net of four candidates per order: the best single path, an
  incumbent-cost coarse disjoint split (a floor that cannot be starved under a tight timeout), a
  refined 256-chunk disjoint split (two-phase: pick the active path set at coarse granularity where
  the gas-activation gate is correct, then refine over that fixed set, then an exchange-refinement
  pass that shifts sub-chunk amounts between paths, accepting only strictly-improving moves), and a
  shared-pool
  fill-and-spill split with marginal-probe candidate selection (top-8 full-amount paths plus the
  best first-chunk probes of the top-32) that can split at intermediate tokens (tree routes).
- Allocation uses an incremental water-fill: AMMs are path-independent in cumulative input, so the
  next chunk's marginal is probed against a committed pool overlay — O(chunks) instead of
  O(chunks²), which is what funds the 256-chunk grid.
- Candidate discovery is exhaustive BFS unioned with the bounded, amount-aware frontier search
  (direct edges, connector tokens, anchor tokens, native-ETH sentinel), so no candidate class is
  lost to either heuristic.
- Routes are assembled through the shared `build_split_route` primitives (topological swap order,
  tycho-execution remainder-split convention, shared-hop merging, route token map), so every
  returned route encodes and executes.
- Net ranking is gas-aware when derived token gas prices are available; without them it falls back
  to gross output and never waits on derived-data warm-up.
- Unlike `split_bounded` (this branch's previous candidate, now replaced and removed), the
  portfolio always returns a route: its never-lose floor means it can serve as the only pool in a
  deployment instead of requiring a companion single-path pool.

## Solve replay

One solve, animated: discovery, candidate ranking, and the allocation strategies filling the order
chunk by chunk (1M USDC → WBTC; the fill-and-spill tree route forms in the Sankey and wins by
+84.8 bps over the best single path):

animated solve replay, 1M USDC to WBTC

## Split vs Path Frank-Wolfe

Head-to-head against `path_frank_wolfe`, the pre-existing split algorithm. Offline, frozen
snapshot, 10k real aggregator trades, `--max-hops 3 --timeout-ms 5000`, exact integer nets
(common set = 5,878 trades both solved):

| | coverage | wins | losses | ties | aggregate net delta | wins: median / mean bps | p50 / p95 solve |
|---|---:|---:|---:|---:|---:|---:|---:|
| `split` vs `path_frank_wolfe` | 5,891 vs 5,878 | **1,025** | 238 | 4,615 | +3.03e27 | +2.1 / +64.0 | 12.9 / 43.7 ms vs 6.1 / 11.5 ms |

The 238 losses have median 1.0 bps; the tail is concentrated on thin-liquidity pairs where both
algorithms return outputs far off par and PFW loses less (the three losses above 1,000 bps are all
one PYUSD pair). The never-lose floor guarantees `split` beats the best single path, not PFW, so
losses to PFW's continuous line-search allocation are possible on pairs where discovery is thin.

Live confirmation (same setup as the audit below), `split` vs `path_frank_wolfe` on the identical
808 solved trades — win rate is the robust signal since the runs are ~25 min apart:

split vs PFW

## Large-trade routing (offline, frozen snapshot)

Three-column route panels: Path Frank-Wolfe | Split Bounded | Portfolio Split. Output amount and
lane count per algorithm; badge = portfolio's gain over `split_bounded`.

**2,000 WBTC → USDC** — portfolio 26,379,472 USDC (+4,360 bps vs PFW, +4.6 bps vs bounded)

2,000 WBTC to USDC

**5,000 WETH → USDC** — portfolio 7,274,393 USDC (+14.3 bps vs bounded)

5,000 WETH to USDC

**17,000 CVX → USDC** — `split_bounded`'s bounded discovery misranks thin-liquidity paths and
returns a split (8,442 USDC) that loses to the best single path (16,193); the portfolio's
never-lose floor plus wider discovery returns 16,453 USDC (+9,488 bps vs bounded).

17,000 CVX to USDC

**64,993 sUSDe → USDC** — portfolio 54,332 USDC (+79 bps vs bounded, +4,724 bps vs PFW)

64,993 sUSDe to USDC

## Live audit vs external aggregators (2026-07-13)

1,000 trades sampled from the 10k dataset (top 100 pairs × 10 amounts), Ethereum mainnet,
`all_onchain` min_tvl=10, no RFQ liquidity — Fynd routes over ~8 on-chain systems while
KyberSwap/Nordstern use their full universe (Curve, Balancer, RFQ, private MMs). Every successful
quote re-executed via `eth_call` at the quote block.

bps distributions vs aggregators

Gas-adjusted on-chain (right column): median gap −0.4 bps vs Nordstern, −0.2 bps vs KyberSwap
despite the liquidity handicap. The raw columns overstate the aggregators: their reported quotes
exceed what `eth_call` actually returns, while Fynd's gas estimates track on-chain usage.

win/loss analysis
gas economics

## Validation

- `cargo +nightly fmt --check`, `clippy --locked --all --all-features --all-targets -D warnings`,
  rustdoc `-D warnings` all clean; `cargo nextest run -p fynd-core`: 502 passed, 8 skipped.
- Quality guarantees enforced by tests: two-equal-pool split output within 0.1% of the analytical
  two-pool optimum, small-order no-loss vs the single path, tight-timeout (1/5/50 ms) no-loss vs
  the single path.
- The live audit exercised the full production quote + encode path: 0 panics, 0 encode failures
  over 808 solved orders.

## Notes

- Trades whose optimal disjoint legs form token cycles (e.g. one leg USDC→USDT, another USDT→USDC)
  cannot be expressed by the token-merged route assembly; both algorithms fall back to the best
  assemblable candidate there. This is an encoding-model limit, not a regression.
- How this branch evolved: four split variants were developed and benchmarked head-to-head on this
  branch's lineage — `split` (exhaustive enumeration), `split_probe` (live-probed exit ranking),
  `split_bounded` (Penumbra-inspired bounded discovery), and the portfolio router. `split_bounded`
  won the first round and replaced the exhaustive variants; the portfolio was then developed
  against it on the offline routing-quality harness (`exp/split-hardening`) and replaced it after
  winning the head-to-head (727W / 83L / 0T on the offline 10k set, aggregate +5.93e25; coverage
  808 vs 205 live, plus 4 encode failures on `split_bounded`'s side). The branch ships exactly one
  new algorithm in one module: `split_bounded` has been removed, its bounded amount-aware discovery
  survives as the portfolio's candidate-union stage, and all earlier variants remain recoverable
  from branch history (`cfb097f`, `ba4448d`/`f493aad`, `b962469`, `835fc80`, `d7170ff`). The
  pre-existing algorithms (`most_liquid`, `bellman_ford`, `path_frank_wolfe`) are untouched.


Haikane:
What changed since the last PR on this branch (#269):

- #269 made `path_frank_wolfe` executable-correct for shared-pool splits: one sequential simulation model, gas-aware activation, and single-path fallback.
- This PR builds on that by adding a separate production candidate, `split_bounded`, instead of continuing to tune the exhaustive split variants.
- During the branch, three variants were tried head-to-head: `split`, `split_probe`, and `split_bounded`. The first two were intentionally removed after benchmarking; only `split_bounded` remains registered and documented.
- `split_bounded` keeps the useful pieces from the experiments: bounded amount-aware candidate discovery, pool-disjoint plus shared-pool split assembly through the existing split primitives, gas-aware net ranking, and a strict floor where it declines if it cannot beat the best single path net-of-gas.
- The pre-existing algorithms are unchanged: `most_liquid`, `bellman_ford`, and `path_frank_wolfe` stay available. This branch only adds the new built-in algorithm and docs/config references.

Why I think it is ready now:

- The implementation is no longer an experimental fork pile. The losing variants are gone, the registry surface is small, and the PR has one clear production behavior: run `split_bounded` alongside a single-path pool.
- Same-block live benchmarking showed `split_bounded` materially beating PFW on the large/XXL trades it is meant for, while solving near Bellman-Ford latency rather than exhaustive split latency.
- The corrected full 10k offline stack check confirms the production stack effect: best of `{bellman_ford, split_bounded}` beat best of `{bellman_ford, path_frank_wolfe}` on 1,220 trades, lost on 95, tied on 4,944, with equal stack coverage of 6,259 trades.
- Successful-solve timings in that 10k run were also in the right shape: `split_bounded` p50 2.0 ms / p95 4.4 ms, vs PFW p50 6.2 ms / p95 12.2 ms and BF p50 4.9 ms / p95 9.9 ms.
- The asymmetric coverage is intentional: `split_bounded` returns `InsufficientLiquidity` when a split does not beat the best single path net-of-gas, so BF/PFW can cover those orders. That makes the algorithm conservative rather than a replacement for the single-path pool.
- Validation is green in the PR description (`./check.sh`: 780 passed, 9 skipped).

Only caveat: the latest 10k stack check used a fresh 2026-07-09 `uniswap_v2,uniswap_v3` snapshot because the older frozen snapshot artifact was not available locally, so those numbers are internally apples-to-apples but not directly comparable to earlier frozen-snapshot standings.

docs/algorithms/README.md
```diff
@@ -28,10 +32,10 @@ See [Architecture](../ARCHITECTURE.md) for the full system design and [Custom Al
...
+| **Approach**           | Enumerate paths, score by heuristic, simulate top-N                 | Simulate every reachable edge, keep best amounts   | Bellman-Ford path discovery plus Frank-Wolfe split optimization             | Portfolio: best net of best single path, coarse disjoint floor, refined 256-chunk disjoint split, and shared-pool fill-and-spill; exhaustive plus bounded candidate discovery |
+| **Strengths**          | Fast; good at common, high-liquidity pairs                          | Finds non-obvious routes; no heuristic blind spots | Reduces price impact by splitting flow across parallel paths                | Never loses to the single path; covers every order (can run as the only pool); captures the large-trade gains |
+| **Weaknesses**         | Path count explodes at high hop counts; heuristic can misjudge      | Single path only; suboptimal for large trades      | More simulation work per request; overkill for small trades                 | Pays the full candidate-simulation cost on every order (offline 10k p50 ~15 ms vs bounded's ~2 ms) |
```
louise-poole: This weakness worries me for VM protocols... I assume you tested this with native protocols only?

fynd-core/src/algorithm/split.rs
```diff
@@ -0,0 +1,1888 @@
...
+    source_token: &'a Address,
+    start: &'a Instant,
+    timeout_ms: u64,
+}
+
+fn timed_out(start: &Instant, timeout_ms: u64) -> bool {
+    start.elapsed().as_millis() as u64 > timeout_ms
+}
+
+fn default_anchor_tokens() -> &'static HashSet<Address> {
```
louise-poole: What's the difference between anchor and connector tokens? Are anchor tokens just default connector tokens? We should make this configurable then as Fynd should be chain agnostic and chain-specifics should be in config defaults.

fynd-core/src/algorithm/water_fill.rs
```diff
@@ -0,0 +1,1888 @@
...
+            from: order.token_in().clone(),
+            to: order.token_out().clone(),
+            reason: NoPathReason::NoGraphPath,
+        });
+    }
+    Ok((paths, scores))
+}
+
+#[cfg(test)]
+mod tests {
```
louise-poole, line 1481: These tests actually have very little coverage - they test a single topology (two equal WETH/USDC pools) with different orders. Why did you decide to create your own tests instead of use the split_test_harness scenarios already created? You could extend those scenarios if you feel something isn't covered there already. This would give you good test coverage already and clean up this test mod.
Note: to use the test harness you'll need to add a `build_market_weighted` builder to the harness before you can use it for this algo.

fynd-core/src/algorithm/water_fill.rs
```diff
@@ -0,0 +1,1888 @@
...
+        }
+
+        let mut sims = 0usize;
+        while delta >= min_delta && !delta.is_zero() {
+            if start.elapsed().as_millis() as u64 > timeout_ms || sims >= EXCHANGE_MAX_SIMS {
+                break;
+            }
+
+            let mut best: Option<ExchangeMove> = None;
+            for donor in 0..k {
```
louise-poole: I would actually add an early break for exceeding simulations to avoid the `path_net` call if simulations are already exceeded.
```suggestion
            for donor in 0..k {
                if sims >= EXCHANGE_MAX_SIMS { break; }
```

docs/algorithms/split.md
```diff
@@ -0,0 +1,123 @@
...
+pool-disjoint, so a trial re-simulates only the two paths it touches, and the pass is bounded by the
+solve timeout and a hard cap on trial simulations. Because only strictly-improving moves are
+accepted, the result never scores below the water-fill allocation.
+
+## Candidate discovery
+
+Discovery unions two sources so no useful route is dropped:
+
+1. **Exhaustive BFS** enumeration of paths between the input and output tokens (the Most Liquid path
+   finder), ranked by a spot-price × depth heuristic.
```
louise-poole: The dependency on Most Liquid algo feels like a side comment here and is not obvious anywhere else. Someone could alter Most Liquid and not realise how it affects this new algo. Can you add it to the Source Reference section of this doc at least? And maybe make the dependency more clear here (or higher in this doc?) with a link to the Most Liquid docs?

fynd-core/src/algorithm/water_fill.rs
louise-poole, line 1: Please add tracing to this algo? (`#[instrument]`) This is highly beneficial for debugging purposes - especially since we'll now have a hosted Fynd.

fynd-core/src/algorithm/split.rs
```diff
@@ -0,0 +1,1888 @@
+//! Split-routing algorithm (`split`).
```
louise-poole: I'm not a fan of the over generic name 'split'. It's not the only splitting algorithm and does not describe what makes this algo unique/different. path_frank_wolfe also uses splitting, but uses the path_frank_wolfe logic to allocate amounts to the split. With that logic maybe we call this water_fill?

louise-poole (commented):
Looks promising! Just left some initial comments... will do another deeper pass later with fresh eyes.

docs/algorithms/README.md
Haikane: Always run in worker pools with BF. This we should always do. PFW included.

fynd-core/src/algorithm/split.rs
Haikane: this is 4 algorithms in one. the most sophisticated one is fill-and-spill. so you can name it fill-and-spill

louise-poole: Yes, but the logic used to allocate amounts to the routes found by those 4 algos is water-fill. I think it's the more appropriate name (in accordance to how we named path_frank_wolfe) as it's the bridging logic between the portfolio of algos in this implementation.

docs/algorithms/README.md
```diff
@@ -4,7 +4,9 @@ icon: compass
 
 # Overview
 
-Fynd ships two built-in routing algorithms. This section explains the problem they solve and how each one works.
+Fynd ships four built-in routing algorithms: three single-route finders (Most Liquid, Bellman-Ford,
+Path Frank-Wolfe) and the [Water-fill](water_fill.md) portfolio router. This section explains the problem they
```
tamaralipows: The words "portfolio router" don't mean much to me without reading the rest, is there a better way to describe this

docs/algorithms/water_fill.md
```diff
@@ -0,0 +1,140 @@
+---
+icon: route
+---
+
+# Water-fill
+
+`water_fill` is Fynd's production split-routing algorithm: it splits one order across multiple parallel
```
tamaralipows: It won't necessarily always be the production split-routing algo right?

docs/algorithms/water_fill.md
```diff
@@ -0,0 +1,140 @@
...
+
+1. **Best single path** — the highest-net full-order route from the candidate set. This is the
+   floor: the portfolio never returns less than this.
+2. **Coarse disjoint floor** — a single gated coarse water-fill over pool-disjoint paths on the
+   same 20-chunk grid, so it does exactly the incumbent split's allocation work at the same cost.
+   Because it shares the solve clock with that work, a tight timeout cannot starve it into a
+   single-path fallback while a split would still win.
+3. **Refined disjoint split** — a two-phase pool-disjoint allocation: pick the active path set at
+   coarse (20-chunk) granularity where the gas-activation gate is correct, then refine the
+   allocation over that fixed set on a fine 256-chunk grid with no gate (the gas is already
```
tamaralipows: This section of the readme is super context-specific vocab heavy and hard for me to understand right away what's happening

docs/algorithms/water_fill.md
```diff
@@ -0,0 +1,140 @@
...
+shared hops (paths that share a prefix pool produce one combined swap with split downstream legs),
+and attach the route's token map. If a candidate's legs cannot be topologically ordered — for
+example two disjoint legs that form a token cycle — assembly returns nothing and that candidate is
+dropped, so the portfolio falls back to the floor or the single path. Assembly is the same code the
+router encodes on-chain, so every route the portfolio returns encodes.
+
+## Never-lose floor
+
+The portfolio's guarantee is that it never returns less net than the best single path, and it holds
+under time pressure because the coarse disjoint floor does exactly the incumbent's work on the
```
tamaralipows: One thing I always like to do in my PRs is to ask Claude to do a pass over docs using simple, clear, concrete language instead of decorative language, that sometimes reduces a lot of confusion for me. Maybe it would help here?

fynd-core/src/algorithm/water_fill.rs
```diff
@@ -0,0 +1,1885 @@
+//! Water-fill split-routing algorithm (`water_fill`).
+//!
+//! For large orders, price impact makes it better to split the order across several parallel routes
+//! so the marginal price stays low. `WaterFillAlgorithm` is a portfolio router: it builds up to
+//! four split candidates and returns the one with the best net-of-gas output, never worse than the
+//! best single path.
+//!
+//! The candidates are the best single path, a coarse pool-disjoint floor, a refined pool-disjoint
```
tamaralipows: Coarse pool-disjoint floor? Refined pool-disjoint split, fill-and-spill 🥴

tamaralipows (commented):
A few comments on docs right away - will review the actual logic after lunch

fynd-core/src/algorithm/water_fill.rs
```diff
@@ -0,0 +1,1885 @@
...
+    fn simulate_step(
+        path: &Path<DepthAndPrice>,
+        market: &MarketState,
+        overlay: &HashMap<ComponentId, Box<dyn ProtocolSim>>,
+        amount: BigUint,
+    ) -> Option<StepResult> {
+        let mut current = amount;
+        let mut total_gas = BigUint::zero();
+        // Intra-path pool reuse: a pool touched twice in one path must see its own first swap.
+        let mut local: HashMap<ComponentId, Box<dyn ProtocolSim>> = HashMap::new();
```
tamaralipows: var should be named more descriptively

fynd-core/src/algorithm/water_fill.rs
```diff
@@ -0,0 +1,1885 @@
...
+        let price = token_prices?.get(token_out)?;
+        if price.denominator.is_zero() {
+            return None;
+        }
+        Some(total_gas * gas_price_wei * &price.numerator / &price.denominator)
+    }
+
+    /// Greedily selects pool-disjoint path indices from `ranked` (best first), up to `max_paths`.
+    fn select_disjoint(ranked: &[Path<DepthAndPrice>], max_paths: usize) -> Vec<usize> {
+        let mut used: HashSet<&ComponentId> = HashSet::new();
```
tamaralipows: Same here (more descriptive naming) - it's the visited pools or?

fynd-core/src/algorithm/water_fill.rs
```diff
@@ -0,0 +1,1885 @@
...
+                    overlay
+                        .get(component_id)
+                        .map(Box::as_ref)
+                })
+                .or_else(|| market.get_simulation_state(component_id))?;
+            let result = state
+                .get_amount_out(current.clone(), token_in, token_out)
+                .ok()?;
+            total_gas += &result.gas;
+            local.insert(component_id.clone(), result.new_state.clone_box());
```
tamaralipows: I'm concerned about this clone. local exists only for the rare case of a pool touched twice within one path, but we pay a full protocol-state clone_box on every hop regardless. This isn't cheap for univ3 hops with a lot of ticks for example afaik 

Do you think it would be worth/possible to perhaps detect whether the path actually repeats a `component_id` once up front, and only maintain `local` when it does (otherwise skip the `clone_box` entirely)?

fynd-core/src/algorithm/water_fill.rs
```diff
@@ -0,0 +1,1885 @@
...
+    /// Builds one independent leg per path at its allocated amount. `subset` and `alloc` are
+    /// aligned by index; pool-disjoint paths never interfere, so each leg is a real independent
+    /// simulation.
+    fn build_disjoint_legs(
+        &self,
+        ordered: &[Path<DepthAndPrice>],
+        subset: &[usize],
+        alloc: &[BigUint],
+        market: &MarketState,
+        _token_prices: Option<&TokenGasPrices>,
```
tamaralipows: should we even pass this at all?

fynd-core/src/algorithm/water_fill.rs
```diff
@@ -0,0 +1,1885 @@
...
+                    max_hops: self.max_hops,
+                    max_candidates: BOUNDED_DISCOVERY_CANDIDATES,
+                    connector_tokens: self.connector_tokens.as_ref(),
+                    anchor_tokens: &anchor_tokens,
+                    source_token: order.token_in(),
+                    start: &start,
+                    timeout_ms,
+                },
+            );
+            if let Ok((bounded_paths, _)) = bounded {
```
tamaralipows: We are totally swallowing the error here, shouldn't we at least debug log? For example `AlgorithmError::InvalidConfiguration` would be missed

fynd-core/src/algorithm/water_fill.rs
```diff
@@ -0,0 +1,1885 @@
...
+                break;
+            }
+            let Ok(result) = MostLiquidAlgorithm::simulate_path(
+                path,
+                &market,
+                token_prices.as_ref(),
+                amount_in.clone(),
+            ) else {
+                continue;
+            };
```
tamaralipows: Paths that error at the full amount are dropped from the entire candidate set. A path that returns `Err` at the full amount is continued and never enters `full_outputs`. Such a path can thus never receive even a small chunk. Is this intended for this algo?

fynd-core/src/algorithm/water_fill.rs
```diff
@@ -0,0 +1,1885 @@
...
+            ordered,
+            &disjoint,
+            market,
+            gas_price,
+            token_prices,
+            order,
+            start,
+            COARSE_CHUNKS,
+            true,
+        )?;
```
tamaralipows: We are already doing this once before - is this not duplicate work? Just a minor thing though if this causes too much of a refactor

tamaralipows (commented):
Thank you for cleaning up @louise-poole! Just a few comments relating to the code itself :)

louise-poole (approved):

PR #271, opened 2026-07-03
feat(hindsight): rework the decoder, fee-aware comparisons, resilient monitor

tamaralipows:
Reworks the hindsight trade decoder, makes the comparison fee-aware, and lets the monitor survive
tycho feed failures. Together these remove every false-win class found in the monitor datasets:
against the original dataset, impossible ">100% wins" on trades above $25 go from 29 to zero, and
unlabeled clients drop from a third of records to under 5%.

## Structure
- The decoder is a stateful struct: it owns the RPC provider and keeps its contract-code cache
  across blocks instead of rebuilding it every block.
- Venue addresses live in a per-chain registry. Only Ethereum is populated; other chains fail
  fast. Adding a chain is a data change, not a logic change.
- Decode logic is isolated per venue: matching picks a strategy per transaction (sender netting,
  maker finding, Relay, or MetaMask), each in its own module. Shared fee handling lives in one
  helper; a strategy that knows its venue asserts it on the decoded flow.
- The monitor defaults to the same protocols and worker pools as `fynd serve`, so local runs are
  comparable to production.

## Decode fixes

Each was traced to a live mainnet transaction before fixing, and each has a regression test:

- Single swaps carrying residue legs (e.g. RFQ surplus returned mid-route) now decode instead of
  being declined; batch settlements still decline.
- NFT purchases through known routers (e.g. Seaport sweeps via Relay) no longer decode as swaps
  of payment vs change.
- Intent fills with no clean EOA maker are declined instead of guessing a routing contract whose
  leftover token amounts netted as an absurd trade.
- Cross-chain bridge orders (LiFi) are skipped: the real output settles on another chain, and
  pairing the deposit with the leftover refund invented trades at two-hundred-thousand-percent
  improvements.
- Wrap-pair trades (native vs wrapped native) far off 1:1 are declined — a wrap is 1:1 by
  construction, so these are mis-paired cross-chain deposits.
- Unconverted Relay payouts (cross-chain orders settled from same-token inventory) are declined
  instead of pairing a gas top-up as "the output".
- Relay's fee collector trading its own funds no longer doubles the output (its receipts were
  counted as a fee).
- Venue attribution no longer falls back to infrastructure (Permit2) on token-to-token swaps.

## Client fees

Relay and MetaMask charge their fee whichever router they plug in, so comparing Fynd against the
trader's after-fee receipt counted the client's own fee as a Fynd improvement. The decoder now
recognizes the clients' fee wallets and corrects the swap amounts: a fee taken from the input is
subtracted from `amount_in` (it never entered the swap), a fee taken from the output is added to
`amount_out` (the swap produced it). Both sides of the comparison are then the amounts that
actually entered and left routing. The fee itself is kept on each record. MetaMask's fee wallets
and behavior were verified across an on-chain sample; without this, its fee dominated small
trades and produced most of the extreme false wins.

## Attribution and coverage

- MetaMask names its venue in the router calldata; the decoder reads it from there. Trace-based
  attribution could not resolve 46% of MetaMask trades (token-to-token routes moving no native
  value) and remains the fallback for other entry points.
- New registry entries: KyberSwap MetaAggregationRouter v2, OKX DEX Router 6, the LiFi Diamond,
  and the MetaMask Swap Router.
- Solver, filler, and MEV entry points are labeled from public sources (Kipseli, Rizzolver /
  Wintermute, Average Solutions, Barter, Quasi, and others) in a display-only tier that does not
  affect how transactions are matched.

## Monitor resilience

Long runs died when a tycho server-side subscription went silent: the block synchronizer stops
emitting, the remaining subscriptions overflow their buffers ~35 minutes later and are dropped,
and nothing resubscribes. The monitor now runs in supervised sessions — if no block arrives
within 15 minutes or the stream ends, it shuts the solver down and builds a new one with fresh
subscriptions, retrying until one succeeds. Counters, the decoder cache, and the output file
carry across rebuilds; rebuilds are counted in a metric. A 16-hour run survived four such
failures unattended.

## Validation

- 501 decoded Relay trades cross-checked against Relay's own FundsMovement events: every input
  matches Relay's recorded deposit exactly or differs by exactly the fee we remove; 499/501
  outputs match Relay's outbound record within 1%.
- 100 blocks verified against Allium's aggregator_trades: 82% agreement on trades both decoded,
  with our records more precise on native ETH; the gaps drove the registry additions above.
- In the traced MetaMask trade, the decoded gross output matches the venue's actual output to
  the wei, and Fynd's quote for the same trade was within 2 bps — correctly reported as a tie
  instead of a 1,175 bps win.


tools/hindsight/src/decoder/mod.rs
```diff
@@ -61,144 +61,174 @@ pub(crate) struct DecodedTrade {
...
+                    )
+                    .await
+                }
+                Strategy::Relay => {
+                    venues::relay::decode(logs, &native, sender, entry_point, registry)
+                }
+                Strategy::Metamask => {
+                    venues::metamask::decode(logs, &native, sender, &root.input, registry)
                 }
+            };
```
tamaralipows: Note to reviewer - this is the main block where the venue strategy is chosen

tools/hindsight/src/decoder/clients/relay.rs
```diff
@@ -0,0 +1,364 @@
...
+};
+
+/// Decode a Relay-entered transaction.
+///
+/// The common case is a user swap: net the sender's flow, then back the
+/// client fee out of it. When the sender has no net flow the transaction is a
+/// solver-initiated rebalancing fill, decoded by anchoring on the fee
+/// collector instead (Relay funds the swap from it); the collector is the
+/// funding source there, not a skim, so no fee is backed out.
+pub(crate) fn decode(
```
tamaralipows, line 25: Didn't think it was worth to put these in structs since the sigs differ a lot and they're stateless. Open to change it

tools/hindsight/src/decoder/mod.rs
```diff
@@ -61,144 +61,174 @@ pub(crate) struct DecodedTrade {
...
+                // A cross-chain bridge order's real output lands on the destination chain; the
+                // same-chain flow (deposit in, leftover refund out) is not a swap. Filtered
+                // before tracing, so bridge orders cost no trace call.
+                if started_bridge_order(matched.receipt.logs()) {
+                    debug!(
+                        tx = %matched.receipt.transaction_hash,
+                        client = %registry.label(matched.entry_point),
+                        "cross-chain bridge order; skipping"
+                    );
+                    return false;
```
kayibal: I feel this should be handled further down the line. The function here uses a LIFI specific logic to detect bridge orders.

tools/hindsight/src/decoder/net.rs
```diff
@@ -1,104 +1,165 @@
...
+/// 1% of the token's gross flow in the transaction. Both conditions are same-token comparisons, so
+/// no prices are needed, and a real batch leg (all of its token's flow) is never dropped. Residue
+/// on a token that never routed third-party (e.g. a rebasing input token) and fixed-token protocol
+/// fees still decline, logged at debug level.
 fn net_trade(
     sent: &HashMap<Address, U256>,
     received: &HashMap<Address, U256>,
-) -> Option<(Address, U256, Address, U256)> {
+    flows: &HashMap<Address, TokenFlow>,
+) -> Option<NetSwap> {
```
kayibal: Idk about you but I don't like this module nor function. I find it hard to reason through how it works exactly and why the flows are required. What is your opinion on it? I think at least it needs a better module level docstring that explains the strategy and assumptions we have here.

tools/hindsight/src/decoder/registry.rs
```diff
@@ -1,157 +1,273 @@
...
-        // on-chain sample; fee ranges ~1–41 bps depending on Relay's fee tier.
-        address!("0xf70da97812cb96acdf810712aa562db8dfa3dbef"),
-    ])
+/// Per-chain address book for trade decoding: which contracts are aggregators,
+/// clients, batch settlers, and venue infrastructure.
+///
+/// Built once per run via [`Registry::for_chain`]. Only Ethereum is populated
+/// today; supporting another chain means adding a constructor with that
+/// chain's addresses, not touching decode logic.
+pub(crate) struct Registry {
```
kayibal, line 47: When I read about Registry, I was expecting this to be a struct that holds the logic to decode a transaction but I feel this feels more like a config and we should probably parse this from a toml or so...

tools/hindsight/src/decoder/net.rs
```diff
@@ -1,104 +1,165 @@
...
-) -> HashMap<Address, U256> {
-    let mut fees: HashMap<Address, U256> = HashMap::new();
-    if fee_collectors.is_empty() {
-        return fees;
-    }
-    for &(_, to, value) in native_transfers {
-        if fee_collectors.contains(&to) {
-            *fees.entry(Address::ZERO).or_default() += value;
-        }
+    tracked: Address,
```
kayibal: What about swaps, that have a different receiver? Afaik we can't track those with this strategy.

tools/hindsight/src/decoder/fixtures/trace_metamask_1inch_0xe815e2b5.json
tamaralipows, line 1: Rename dir to test fixtures

tools/hindsight/src/decoder/mod.rs
tamaralipows: Good point - moved this to lifi specific module

tools/hindsight/src/decoder/registry.rs
tamaralipows: Moved this into a toml

tools/hindsight/src/decoder/net.rs
tamaralipows: Yeah it's a current limitation - documented this for now. Can quantify just how many swaps this affects tomorrow.

tools/hindsight/src/decoder/net.rs
tamaralipows: Agreed. I think one reason it was hard to follow is that the same work - adding up who sent and received how much of each token - was implemented three times in different ways: here for netting the trader's flow, in the Relay module for its rebalancing fills, and again for fee accounting. This is restructured now: net.rs is gone, and a TransferLedger is built once per transaction (one flat list of every transfer)

Tried to also make this clearer in docstrings too but will keep improving this whole PR tomorrow

tvinagre (approved):
this is such a beautiful code
made my eyes shine
great job @tamaralipows

PR #279, opened 2026-07-08
feat: Bump tycho-simulation and tycho-execution to 0.328.1

tvinagre:
Bumps both tycho crates from 0.313.0 to the latest release 0.328.1.

Both must move together — they share `tycho-common` types and a version mismatch causes compilation errors (`Chain` enum version conflict).

`cargo check --package fynd` passes clean.


tamaralipows (approved):
Thank yooouu

PR #286, opened 2026-07-09
feat: use native CurveState decoder and add vm:bopamm/vm:fermiswap

tvinagre:
### Native Curve decoder
Switch `vm:curve` from `EVMPoolState<PreCachedDB>` (deprecated VM adapter) to the native `CurveState` introduced in tycho-simulation 0.324.0.

The hybrid implementation uses pure-Rust quote math (vendored from MIT-licensed `curve-math`/`curve-adapter`) over VM-indexed storage, which is more performant than running full EVM execution for every swap simulation. The release notes for 0.324.0 explicitly warn when `vm:curve` uses the deprecated VM adapter.

### New protocol decoders
Add `vm:bopamm` and `vm:fermiswap` to the protocol registry — both were introduced in tycho-simulation 0.328.1 but were missing from the match block, causing fynd to log `Skipping unknown protocol` and ignore them.

Both follow the same pattern as `vm:maverick_v2`: `EVMPoolState<PreCachedDB>` with no custom pool filter.

`cargo check --package fynd-core` passes clean.

### DIFF
--- a/fynd-core/src/feed/protocol_registry.rs
+++ b/fynd-core/src/feed/protocol_registry.rs
@@ -6,6 +6,7 @@ use tycho_simulation::{
         engine_db::tycho_db::PreCachedDB,
         protocol::{
             aerodrome_slipstreams::state::AerodromeSlipstreamsState,
+            curve::CurveState,
             ekubo::state::EkuboState,
             ekubo_v3::{self, state::EkuboV3State},
             erc4626::state::ERC4626State,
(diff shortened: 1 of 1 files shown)


zizou0x (approved):

fynd-core/src/feed/protocol_registry.rs
```diff
@@ -112,6 +113,20 @@ pub(crate) fn register_exchanges(
...
+                    None,
+                );
+            }
+            "vm:fermiswap" => {
+                builder = builder.exchange::<EVMPoolState<PreCachedDB>>(
+                    "vm:fermiswap",
+                    tvl_filter.clone(),
+                    None,
+                );
+            }
```
Troshchk, line 129: In the description you mention `EVMPoolState<PreCachedDB>` is deprecated, should something else be used here?

PR #287, opened 2026-07-09
feat(quote): populate price_impact_bps (algorithm value or spot-price fallback)

tvinagre:
## Summary

Populates `OrderQuote.price_impact_bps`, which was declared but never set (always `None`). The value is now computed and returned on every `/v1/quote`:

- **Algorithm-reported when available:** `path_frank_wolfe` already computes a price impact internally (`1 - amount_out/ideal_out`) for its probe heuristic but discarded it. It now attaches that value to the `RouteResult` it returns (both single-path and split results). Split routes only come from Frank-Wolfe, so it self-reports for the case the fallback can't handle.
- **Worker fallback otherwise:** for algorithms that don't report one (`bellman_ford` — the default — and `most_liquid`), the worker computes the impact at quote construction from each swap's live `protocol_state().spot_price()` product and endpoint token decimals from `market_data`. This runs at the single `OrderQuote` construction point in `worker_pool/worker.rs`, so it is algorithm-agnostic.

`price_impact_bps` is **signed** — a favorable execution (better than the spot/mid reference) reports a negative value.

### Fail-safe

Price-impact computation never fails or blocks a quote. Any missing token, non-linear/split route reaching the fallback, spot-price error, non-finite/overflow value, or write-lock contention yields `None`, and the field is simply omitted. The Frank-Wolfe split path uses `.ok()` (not `?`) so an impact error can't fail an otherwise-valid solve.

### Why this replaces the frontend estimate

The swap frontend previously estimated price impact with a reverse round-trip quote, which overstated it badly (e.g. 12.37% on a ~1:1 stablecoin swap, because the reverse leg took a different, worse-liquidity route). Sourcing the value from the router — execution vs. the same spot price the solver uses — is the correct, consistent methodology. (Frontend PR consumes this and drops the estimate.)

## Changes

- `fynd-core/src/types/quote.rs` — `RouteResult.price_impact: Option<f64>` + builder/accessor; `with_price_impact_bps` now used.
- `fynd-core/src/worker_pool/price_impact.rs` (new) — pure math (`price_impact_from_spot_product`, 6 unit tests) + route-level fallback (`spot_price_impact`).
- `fynd-core/src/worker_pool/{mod.rs,worker.rs}` — module registration; wire algorithm-value-else-fallback into the quote.
- `fynd-core/src/algorithm/path_frank_wolfe.rs` — report the computed impact (+ regression test).
- `fynd-core/src/feed/market_data.rs` — doc clarification (`try_read_blocking` now used by the fallback).

## Test Plan

- [x] `cargo test -p fynd-core` — 494 passed, 0 failed
- [x] `cargo clippy -p fynd-core --all-targets -- -D warnings` — clean
- [x] Live: `1 WETH → USDC` → 3 bps; `1000 DAI → USDC` (direct pool) → 2 bps; `1,000,000 DAI → USDC` (routes DAI→WETH→USDC) → ~1971 bps, arithmetically correct for that route
- [ ] Reviewer: confirm fee-inclusive spot-price semantics match Frank-Wolfe (intentional, keeps both consistent)

### DIFF
--- a/fynd-core/src/algorithm/path_frank_wolfe.rs
+++ b/fynd-core/src/algorithm/path_frank_wolfe.rs
@@ -15,7 +15,7 @@ use std::time::{Duration, Instant};
 
 use num_bigint::{BigInt, BigUint};
 use num_traits::{ToPrimitive, Zero};
-use tracing::debug;
+use tracing::{debug, warn};
 use tycho_simulation::tycho_core::models::Address;
 
 use super::{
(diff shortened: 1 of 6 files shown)


tamaralipows (approved):
sparkles in my eyes

PR #288, opened 2026-07-09
feat(hindsight): volume by outcome, per-trade log, solver label fix

tamaralipows:
Telemetry changes backing the dashboard rework (dashboard JSON updated here and provisioned via terraform-infrastructure).

## Changes

- `hindsight_volume_usd` gains an `outcome` label (headline verdict, 4 bounded values) so the dashboard can split volume into solvable/unsolvable. Valuation falls back to the input leg when the output token is unpriced — the output leg is unpriced exactly when the trade is unsolvable (long-tail token), so without the fallback ~43% of unsolvable volume was silently dropped.
- One structured `info!` line per priced comparison (tx, client, solver, tokens, verdict, volume, savings). Loki already ingests pod stdout; the dashboard's new top-trades table reads these fields by regexp.
- Fallback-tier solver attributions are labeled `unknown` in metrics: the fallback labels a record with its entry point — a client like relay, not a solver — which polluted the solver dropdown. The entry-point detail stays in the JSONL.
- The coverage-ratio gauge is removed. It was cumulative since process start, so it ignored the Grafana time picker and reset on every redeploy; the dashboard now derives coverage from `hindsight_trades_total` over the selected range. Nothing else read the gauge.
- Dashboard JSON rework: client + state template variables, back-of-block stat row (conservative bound), stacked trades-by-client and trades-by-solver panels, savings distribution from histogram buckets (replacing the median-USD panel), `state="top"` pinned on queries that previously mixed both block states and double-counted trades.

### DIFF
--- a/tools/hindsight/grafana/hindsight.json
+++ b/tools/hindsight/grafana/hindsight.json
@@ -1,5 +1,5 @@
 {
-  "title": "Hindsight \u2014 Fynd vs settled solver swaps",
+  "title": "Hindsight — Fynd vs settled solver swaps",
   "uid": "hindsight",
   "schemaVersion": 39,
   "timezone": "browser",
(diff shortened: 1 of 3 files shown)


tamaralipows:
Added in bc1b9d32f: the monitor now declares its session unhealthy and rebuilds the solver when it falls more than 100 blocks (~20 min) behind chain head. Covers the wedge seen live today (crippled worker pool, feed technically alive, monitor hours behind) that the dead-feed watchdog can't catch.

PierreMkt (approved):

PR #292, opened 2026-07-09
feat(feed): report fynd_version to Tycho via client metadata

brunoguerios:
Bumps `tycho-simulation` and `tycho-execution` to 0.331.0, which exposes `ProtocolStreamBuilder::add_client_metadata`, and attaches a `fynd_version` entry (`env!("CARGO_PKG_VERSION")`) to every outbound Tycho stream.

The Fynd release version is now sent in the `X-Tycho-Client-Metadata` header on all Tycho connections. `User-Agent` is unchanged. This lets the Tycho server attribute traffic and version usage to Fynd.

The entry is applied at all three stream-build paths — `run`, `run_with_pending`, and `run_with_step_controller` — through a shared `fynd_client_metadata()` helper, so telemetry is consistent regardless of feed mode. `fynd_preset` is deferred until preset resolution exists in Fynd.

Wire-level header emission (WS + RPC, ordering, validation) is covered by tycho-client's own tests; the Fynd-side test pins the reported metadata key and a non-empty version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7756,9 +7756,9 @@ dependencies = [
 
 [[package]]
 name = "tycho-client"
-version = "0.328.1"
+version = "0.331.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "260bfa6e48e757b65a97ad61c188db43d70a6902fb1be86881b611b64613e23e"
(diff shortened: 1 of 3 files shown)


zizou0x (approved):
LGTM

PR #294, opened 2026-07-09
fix(core): stop worker busy-spin when derived-data channel closes

tamaralipows:
## Problem

`SolverWorker::run` handles a closed derived-data channel by warning and continuing:

```rust
Err(broadcast::error::RecvError::Closed) => {
    warn!(self.worker_id, "derived event receiver closed");
    // Continue running - derived data won't update but we can still solve
}
```

`Closed` is permanent — once the sender drops, `recv()` returns it **instantly on every call**. The select loop degenerates into a hot spin: recv → warn → recv, millions of iterations per minute.

## Consequences (observed live in hindsight, twice on 2026-07-09)

- One core pegged inside the tokio runtime; the global logging writer hammered so hard that every log-emitting task in the process contends on it
- All solves starve into timeouts (hindsight went 100% `Timeout` verdicts, fell hours behind chain head)
- `kubectl logs` history destroyed: 1.3M identical warn lines in 2 minutes rotated everything else away
- New WebSocket connections fail on the starved runtime — after tearing down the solver, rebuild attempts failed for 70 minutes while a fresh process connected on its first try

Any fynd-core consumer is exposed, including the API, whenever the derived-data publisher dies.

## Fix

Disable the derived-data select arm after the first `Closed` (`, if !derived_closed` guard). The worker keeps solving with the last derived data it saw — the original intent — without polling a permanently-ready channel.

Regression test captures the worker's log output and asserts the closed channel is handled exactly once; it fails on the old code (thousands of warns in the same window).

Note: this fixes the spin, not whatever kills the derived-data publisher in the first place — that root cause is still unknown and worth its own investigation.

### DIFF
--- a/fynd-core/src/worker_pool/worker.rs
+++ b/fynd-core/src/worker_pool/worker.rs
@@ -426,6 +426,11 @@ where
     {
         info!(self.worker_id, "worker started");
 
+        // Once the derived-data channel closes, its recv() returns Closed instantly on every
+        // call; keeping the arm in the select would turn this loop into a busy spin. The guard
+        // disables the arm so the worker keeps solving with the last derived data it saw.
+        let mut derived_closed = false;
+
         loop {
             tokio::select! {
                 biased; // prioritize events in this order: shutdown, market update, derived data, solve task
@@ -460,7 +465,7 @@ where
                 }
 
                 // Process derived data events (pool depths, token prices)
-                derived_result = derived_event_rx.recv() => {
+                derived_result = derived_event_rx.recv(), if !derived_closed => {
                     match derived_result {
                         Ok(event) => {
                             // Always update tracker with every event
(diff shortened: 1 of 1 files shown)


tamaralipows:
Traced why `Closed` occurs at all: the `DerivedDataEvent` sender is owned by `ComputationManager`, whose `run` task has two designed exits (shutdown signal, market-event channel closed) — neither happened in the observed incidents (workers kept running, so the market channel was alive). That leaves an undesigned exit: a panic inside the manager task. Tokio catches it at the task boundary, and `computation_handle` (solver.rs:835) is only awaited at teardown, so the death is silent until workers see `Closed`.

Which panic, unknown — the warn spam rotated the panic message out of log retention. This fix makes the next occurrence diagnosable (the panic will remain visible in the logs).

Open design question for a follow-up: should a solver whose ComputationManager has died keep serving quotes on frozen derived data indefinitely, or should the handle be monitored and its death treated as fatal like a dead feed?

tamaralipows:
Root cause found and fixed in #295: a NaN spot-price spread hit `partial_cmp().unwrap()` in `token_gas_price.rs:408` (panic messages recovered from Loki, identical in both incidents). That PR is the headline fix; this one keeps any future publisher death from cascading into a spin.

kayibal (approved):
Keeps happening, I understand this is the minimal change necessary to avoid the log. We sould potentially consider completely exiting if derived data is too stale. Also agree that this is out of scope here. So I think we should merge this asap.

PR #295, opened 2026-07-10
fix(core): drop NaN-scored gas-price paths instead of panicking

tamaralipows:
## Problem

`TokenGasPriceComputation` sorts candidate paths by float score with:

```rust
paths.sort_by(|a, b| b.score.partial_cmp(&a.score).unwrap());
```

The score is a spot-price spread (`|forward − 1/reverse|`); degenerate pool math can make it NaN, `partial_cmp` then returns `None`, and the `unwrap` panics.

## Consequences

This is the root cause of both 2026-07-09 hindsight outages (prod ~14:00Z, dev ~22:01Z) — confirmed via the panic messages in Loki, identical in both:

```
thread 'tokio-rt-worker' panicked at fynd-core/src/derived/computations/token_gas_price.rs:408:64
```

The panic kills the ComputationManager task **silently** — its JoinHandle is only awaited at teardown (solver.rs) — which drops the derived-data sender, closes the channel for every worker, and triggered the worker busy-spin fixed in #294. One NaN from one pool took down the deployment. The fynd API shares this code path.

## Fix

Drop NaN-scored candidates before sorting (a NaN score is meaningless as a routing candidate; the affected token simply gets no gas price that block) and sort with `total_cmp`, which is defined for all floats.

Regression test injects NaN spot prices and asserts the computation completes with the path dropped; it panics on the old code.

## Related

- #294 — the worker busy-spin this panic triggered (independent defect, keeps the next publisher death from cascading)
- Follow-up worth discussing: monitor the ComputationManager's JoinHandle so a task death is fatal/visible instead of silent

### DIFF
--- a/fynd-core/src/derived/computations/token_gas_price.rs
+++ b/fynd-core/src/derived/computations/token_gas_price.rs
@@ -404,9 +404,8 @@ impl TokenGasPriceComputation {
             .collect();
 
         // Sort each token's paths: lowest spread last (for popping). A NaN score (degenerate
-        // pool math in the spread computation) is meaningless as a candidate and, worse, makes
-        // partial_cmp-based sorting panic — which killed the ComputationManager task in
-        // production. Drop those candidates and sort with the float total order.
+        // pool math in the spread computation) cannot rank a path and would panic a
+        // partial_cmp-based sort, so drop those candidates and sort with the float total order.
         for paths in paths_by_token.values_mut() {
             paths.retain(|path| !path.score.is_nan());
             paths.sort_by(|a, b| b.score.total_cmp(&a.score));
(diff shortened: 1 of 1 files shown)


fynd-core/src/derived/computations/token_gas_price.rs
```diff
@@ -403,9 +403,12 @@ impl TokenGasPriceComputation {
...
             .collect();
 
-        // Sort each token's paths: lowest spread last (for popping)
+        // Sort each token's paths: lowest spread last (for popping). A NaN score (degenerate
+        // pool math in the spread computation) cannot rank a path and would panic a
+        // partial_cmp-based sort, so drop those candidates and sort with the float total order.
         for paths in paths_by_token.values_mut() {
-            paths.sort_by(|a, b| b.score.partial_cmp(&a.score).unwrap());
+            paths.retain(|path| !path.score.is_nan());
+            paths.sort_by(|a, b| b.score.total_cmp(&a.score));
```
tamaralipows, line 411: partial_cmp can panic and total_cmp can't — and this line already proved a panic here kills the whole deployment. The retain implements the decision "NaN isn't a valid candidate"; total_cmp guarantees the sort can never panic even if a NaN slips through some future path. 

partial_cmp = IEEE float comparison, a partial order (NaN incomparable to everything, even itself). total_cmp = IEEE totalOrder, a total order: −NaN < −∞ < reals (−0 < +0) < +∞ < +NaN. On ordinary numbers they agree exactly

kayibal (approved):
thank you!

PR #300, opened 2026-07-12
feat: Record per-pool queue wait and depth metrics

tvinagre:
## Summary

Quote requests are dispatched to worker pools via an async_channel task queue; each `SolverWorker` runs on a dedicated OS thread. This adds two Prometheus series, recorded at the point a worker picks a task off the queue:

- `worker_pool_queue_wait_seconds{pool=...}` (histogram) — how long the task sat in the queue before a worker picked it up.
- `worker_pool_queue_depth{pool=...}` (gauge) — the number of tasks still queued behind it at pickup time.

Both are labeled by `pool`, so per-pool saturation is visible when multiple pools (different algorithms) are running.

Queue wait growing while solve time stays flat is the leading indicator of worker saturation: it means work is arriving faster than workers can pick it up, before solve latency itself degrades. It is a better early signal than solve-time percentiles, which only move once the algorithm itself is under pressure.

`SolveTask::wait_time()` was already being computed on task pickup but discarded (`let _wait_time = task.wait_time();`); this now records it instead, alongside `async_channel::Receiver::len()` for the depth gauge.

## Changes

- `SolverWorker::new` gains a 5th parameter, `pool_name: String`, threaded through from `WorkerPoolConfig` (`pool.rs`) → `SpawnWorkersParams` (`registry.rs`) → each spawned worker thread, so the metric label reflects the pool's configured name rather than just the algorithm name.
- New `record_task_pickup_metrics` helper in `worker.rs` records both series.
- Added `metrics-util` as a `fynd-core` dev-dependency to assert on recorded metric values via `DebuggingRecorder`.

## Test plan

- [x] `cargo test -p fynd-core task_pickup_metrics_recorded` — new unit test asserting both series are recorded with the correct pool label and values
- [x] `cargo check --all-features`
- [x] `cargo test -p fynd-core worker` — all worker/registry/pool tests pass
- [x] `./check.sh` (fmt, clippy, doc, nextest) — clean, aside from pre-existing integration tests that fail locally because git-lfs isn't installed in this environment (fixture files are LFS pointers, unrelated to this change; confirmed failing on `main` too)
- [ ] Manual verification against a running local solver (not run in this session)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2976,6 +2976,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "endian-type"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c34f04666d835ff5d62e058c3995147c06f42fe86ff053337632bca83e42702d"
+
+[[package]]
 name = "enum-ordinalize"
 version = "4.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3429,6 +3435,7 @@ dependencies = [
  "itertools 0.14.0",
  "lazy_static",
  "metrics",
+ "metrics-util 0.20.4",
  "num-bigint",
  "num-rational",
  "num-traits",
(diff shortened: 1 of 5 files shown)


tamaralipows (approved):
Thank you! LGTM

PR #301, opened 2026-07-13
feat: Add capacity benchmark subcommand

tvinagre:
## Summary

- Adds a `capacity` subcommand to `fynd-benchmark` that measures the highest sustainable request rate within a latency SLO. It walks an RPS ladder (`--ladder start:step:max`, default `5:5:200`), measuring p95 latency and error rate at each step after a warm-up period. A step passes when p95 latency stays at or below `--slo-multiplier` (default 1.2) times an unloaded baseline p95, and the error rate stays under the configured cap. The command reports the highest RPS step that passed, and writes a report with a marker string so it can be located in cluster logs.
- Adds an `--encoding` flag to the benchmark request path. When set, every request attaches the standard encoding options; the flag is per-request and does not require a separate RPC call.
- Ships the `fynd-benchmark` binary in the Docker image. Both `cargo build` invocations in the builder stage now build `fynd-benchmark` alongside `fynd`, `fynd-swap-cli`, and `hindsight`. The second (real-source) build stage copies `tools/benchmark/src/` instead of stubbing it, since `pairs.json` and `trades_sample.json` are embedded into the binary via `include_str!`. The runtime stage copies the compiled binary to `/usr/local/bin/fynd-benchmark`.

## Test plan

- [x] `docker build -t fynd-bench-test .` completes successfully, compiling `fynd-benchmark` in both build stages
- [x] `docker run --rm --entrypoint fynd-benchmark fynd-bench-test capacity --help` prints all capacity flags, including `--ladder`, `--slo-multiplier`, `--encoding`, `--target-label`, and `--output-file`
- [x] `docker run --rm --entrypoint fynd-benchmark fynd-bench-test --help` lists `capacity` alongside the existing subcommands
- [x] existing `fynd` and `fynd-swap-cli` entrypoints still resolve in the built image


tamaralipows (approved):
Thenk yoou

PR #303, opened 2026-07-13
feat(hindsight): detect sandwich attacks around decoded trades

kayibal:
## What

hindsight now detects likely sandwich attacks around decoded solver trades and classifies them apart from real wins and losses.

- New `decoder/sandwich.rs`: for each decoded trade, scans a window of 2 transactions on each side — using the block receipts `decode_block` already fetched, so no extra RPC calls — for a front/back pair that (a) links to one attacker (same sender, or same non-registry-known target contract), (b) both re-emit a log from a pool contract the victim's swap touched, and (c) moves the victim's output token in sandwich direction: a linked entity receives it in the front leg and sends it in the back leg, read from each leg's Transfer logs. A native-ETH output is checked as its wrapped form. Pool candidates exclude token contracts (Transfer/Approval emitters), the wrapped-native token, and Permit2, so transactions sharing only a token or its plumbing cannot produce a fake overlap.
- `DecodedTrade` and `RangeComparison` gain `tx_index` and `sandwich` evidence (`front_tx`, `back_tx`, `attacker`, `pools`); both appear in the JSONL comparison records and the `decode` subcommand output. All fields are additive.
- New `Verdict::Sandwiched`: a flagged trade's solved states are reclassified (the headline follows top-of-block) while bps/USD deltas are still computed and written, so the size of MEV-inflated deltas stays analyzable offline. Unsolved states keep `Unsolvable`/`CoverageMiss` — a sandwich explains the settled price, not a missing route — so coverage shares are unaffected.
- Telemetry: trade/volume metrics gain the `outcome="sandwiched"` label; sandwiched states skip the savings/improvement histograms so the "value of adding Fynd" aggregates exclude MEV-degraded settlements. The per-trade Loki line keeps logging with `verdict=sandwiched`.

Known limitations (false negatives keep polluting savings as all sandwiches did before this feature; false positives exclude real comparisons) are listed in the design spec: `docs/superpowers/specs/2026-07-13-hindsight-sandwich-detection-design.md`.

## Why

Many of the largest "improvement" opportunities in hindsight output were sandwich victims: a frontrun degraded the settled output, so the top-of-block re-solve looked like a large win Fynd could not actually have captured.

### DIFF
--- a/docs/superpowers/specs/2026-07-13-hindsight-sandwich-detection-design.md
+++ b/docs/superpowers/specs/2026-07-13-hindsight-sandwich-detection-design.md
@@ -13,14 +13,15 @@ sandwiches, record the evidence, and classify them apart from real wins and loss
 
 ## Detection methodology
 
-Bracket-pair with pool overlap — the zeromev/EigenPhi heuristic without direction decoding.
+Bracket-pair with pool overlap — the zeromev/EigenPhi heuristic — plus a log-level direction
+check on the attacker's token flow.
 
 Detection runs inside `Decoder::decode_block`, which already fetches every receipt of the block in
 one `eth_getBlockReceipts` call. All signals below come from those receipts; detection makes no
 additional RPC calls.
 
 For a victim trade at transaction index `i`, scan a window of `W = 2` transactions on each side for
-a pair `(front, back)` with `front.index < i < back.index` satisfying both conditions:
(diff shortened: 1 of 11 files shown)


tamaralipows (approved):
Did some of the main fixes I could find. Will continue to iterate on the organization in future PRs - working on some refactors atm and don't want to create conflicts by keeping this open.

PR #304, opened 2026-07-13
feat: taker propAMM surplus routing

Troshchk:
The changes add a second worker pool that routes through "exclusive" liquidity (pools only Fynd can access). If the exclusive route beats the best public route, the protocol captures the difference as surplus while the user gets the same price they would have gotten from public pools.
[Task](https://propeller-heads.atlassian.net/browse/ENG-6133)

- Added permission filtering so public workers never see exclusive components
- Added surplus route comparison and selection logic
- Added role-aware early-return so the router waits for both pools
 - Wired the permission policy from config through to workers and router
 - Added pool role configuration (public / all) to worker pool config


fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -469,6 +601,161 @@ impl WorkerPoolRouter {
...
+    result
+}
+
+/// Returns `true` if the quote has a valid exclusive layout for v1: at least one exclusive
+/// leg, all positioned as terminal legs of their respective paths (no mid-route exclusive legs).
+///
+/// Path boundaries are detected by checking whether the next swap's `token_in` differs from the
+/// current swap's `token_out`. This heuristic is correct for Fynd's sequential route
+/// representation but would need revisiting if routes gain explicit path-boundary markers.
+fn has_valid_exclusive_layout(quote: &OrderQuote, policy: &PermissionPolicy) -> bool {
```
Troshchk: This checks, that the exclusive hop is terminal.
It is a v1 limitation

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -469,6 +601,161 @@ impl WorkerPoolRouter {
...
+    let committed = match public_ranked.first() {
+        Some(q) if q.status() == QuoteStatus::Success => q,
+        _ => return public_ranked,
+    };
+
+    // Find best surplus candidate from All-role pools, respecting max_gas and layout constraints.
+    let best_surplus = responses
+        .quotes
+        .iter()
+        .filter(|(pool, _)| {
```
Troshchk: Pool here is worker_pool

fynd-core/src/feed/mod.rs
```diff
@@ -7,7 +7,7 @@ pub mod events;
 pub(crate) mod gas;
 /// Shared market data store (`MarketState`, `MarketData`).
 pub mod market_data;
-/// Per-worker permission scoping for permissioned components.
+/// Per-worker permission scoping for exclusive components.
```
tamaralipows: You've changed this terminology but there is still a file called `permission.rs` and `PermissionPolicy` so I'm not entirely sure this helped tbh, I feel it was fine before as permissioned because at least within the Fynd code it was consistent?

fynd-core/src/feed/permission.rs
```diff
@@ -0,0 +1,205 @@
+//! Per-worker permission scoping for exclusive components.
+//!
+//! "Exclusive" means accessible only to Fynd: such components must never enter the route a
+//! public pool returns, yet the surplus pool is allowed to route through them to capture the
```
tamaralipows: And here we are calling them surplus pool instead of permissioned or exclusive, can get confusing

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -155,6 +150,9 @@ pub struct WorkerPoolRouter {
     /// Validates solution outputs against external price sources.
     /// Present when the server has price guard enabled; `None` when disabled.
     price_guard: Option<PriceGuard>,
+    /// Predicate identifying exclusive components, used by `combine_with_surplus` to locate the
```
tamaralipows: I don't think we need to put "used by ____" in the docs, if we ever change that fn name or remove it it might get confusing

fynd-core/src/worker_pool_router/mod.rs
tamaralipows: Can we call this route instead of layout

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -576,23 +615,135 @@ fn combine_with_surplus(
...
+    } else {
+        BigUint::ZERO
+    };
+    executed.set_amount_out_net_gas(committed_net_gas);
+
+    let surplus_info = SurplusInfo::new(surplus_amount, committed_route_out.clone());
+    executed = executed.with_surplus(surplus_info);
+
+    debug_assert!(
+        executed.amount_out() >= committed.amount_out(),
```
tamaralipows: Isn't this assert trivially true now?

fynd-core/src/worker_pool_router/mod.rs
tamaralipows: I know this was here initially but this docstring is not very clear to me what this method does. "Overlay" is pretty vague

fynd-core/src/worker_pool/worker.rs
```diff
@@ -61,7 +61,7 @@ where
     /// Permission scoping for this worker's local graph.
     ///
     /// `IncludeAll` (the default) preserves the original non-filtered behaviour. A public worker
-    /// is configured with `ExcludePermissioned(policy)` so permissioned components never enter
+    /// A public worker is configured with `PublicOnly(policy)` so exclusive components never enter
```
tamaralipows: Duplicated "a public worker"

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -576,23 +615,135 @@ fn combine_with_surplus(
...
+    }
+
+    let realized_route_out = surplus_quote.amount_out();
+    let committed_route_out = committed.amount_out();
+
+    if realized_route_out <= committed_route_out {
+        return public_ranked;
+    }
+
+    let surplus_amount = realized_route_out - committed_route_out;
```
tamaralipows: Shouldn't we be calculating this net gas here actually? I'm worried the user might get a worse price after gas is considered

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -576,23 +615,135 @@ fn combine_with_surplus(
...
+
+    // Build the executed quote: clone the surplus candidate, stamp per-leg committed amounts,
+    // pin amount_out to committed, attach SurplusInfo.
+    let mut executed = surplus_quote.clone();
```
tamaralipows: Maybe we should call it something other than "executed", since we technically didn't execute it yet?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -576,23 +615,135 @@ fn combine_with_surplus(
...
+            if policy.is_exclusive(swap.protocol_component()) {
+                // Proportional reduction: committed_leg = leg.amount_out * committed / realized
+                // (rounds down — safe direction: under-capture)
+                let committed_leg = swap.amount_out() * committed_route_out / realized_route_out;
+                swap.set_committed_amount_out(committed_leg);
+            }
+        }
+    }
+
+    executed.set_amount_out(committed_route_out.clone());
```
tamaralipows: Also another comment about this: Should we maybe add a constant BPS buffer to ensure that the user gets strictly a better price? We can start by setting it to 0 so the user gets exactly the regular best output amount, but I think eventually we would like to ensure the user gets the `committed_route_out` and a bit more

fynd-core/src/worker_pool_router/mod.rs
tamaralipows: This also needs a test I think

tamaralipows (commented):
Thank you! Looks promising. A few bigger points to start with, let me know what you think

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -469,6 +596,165 @@ impl WorkerPoolRouter {
...
+        return public_ranked;
+    };
+
+    // The surplus route must beat the committed reference net-of-gas.
+    if surplus_quote.amount_out_net_gas() <= committed.amount_out_net_gas() {
+        return public_ranked;
+    }
+
+    let realized_route_out = surplus_quote.amount_out();
+    let committed_route_out = committed.amount_out();
```
louise-poole: Design thought here: this logic is specific for exact-in orders... Which is fine as we do not yet support exact-out. But I jsuspect that if/when we add that support this area of code will get overlooked. Can you add comment noting the limitation of the current logic?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -469,6 +596,165 @@ impl WorkerPoolRouter {
...
+    public_ranked: Vec<OrderQuote>,
+    permission_policy: Option<&PermissionPolicy>,
+) -> Vec<OrderQuote> {
+    let Some(policy) = permission_policy else {
+        return public_ranked;
+    };
+
+    let committed = match public_ranked.first() {
+        Some(q) if q.status() == QuoteStatus::Success => q,
+        _ => return public_ranked,
```
louise-poole, line 709: Hmmmm - if there was no successful public route, but we are able to find a route using the exclusive pools, then we should return that exclusive pool route no? We could set it to take a default fee (choose what amount_out to pin it to) and act as if it's a normal pool instead of dynamically matching another route. Is this possible?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -40,24 +41,52 @@ use tycho_execution::encoding::{
...
+/// [`PoolConfig`]: crate::PoolConfig
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum PoolRole {
+    /// Routes through public liquidity only. Establishes the committed reference output. Default.
+    #[default]
+    Public,
+    /// Routes through all liquidity, including exclusive components; source of surplus quotes.
+    All,
+}
```
louise-poole: Something about this naming doesn't fully sit right with me. It leaks into the public TOML and seeing 'poolrole = all' isn't so clear to me on what it does unless I know the code itself. Maybe we rename them like this? What do think?
```suggestion
pub enum LiquidityScope {
    /// Routes through public liquidity only. Establishes the committed reference output. Default.
    #[default]
    Public,
    /// Routes through all liquidity, including exclusive components; source of surplus quotes.
    All,
}
```

louise-poole: Or maybe even 
```suggestion
pub enum LiquidityScope {
    /// Routes through public liquidity only. Establishes the committed reference output. Default.
    #[default]
    PublicOnly,
    /// Routes through all liquidity, including exclusive components; source of surplus quotes.
    IncludeExclusive,
}
```

fynd-core/src/feed/permission.rs
```diff
@@ -0,0 +1,205 @@
...
+///
+/// Each worker gets exactly one context for its lifetime, derived from its pool's role. All graph
+/// filtering for a worker flows through this type, so the worker never reasons about
+/// exclusivity itself — it just hands its topology and incoming events here.
+#[derive(Clone, Debug)]
+pub enum PermissionContext {
+    /// See every component — the default. No filtering is applied.
+    IncludeAll,
+    /// See only public components — exclusive ones are filtered out by the attached policy.
+    PublicOnly(PermissionPolicy),
```
louise-poole: I kind of like the scaffold's naming here of `ExcludePermissioned` instead of `PublicOnly`. Open to discuss this though.

fynd-core/src/solver.rs
```diff
@@ -688,10 +741,23 @@ impl FyndBuilder {
...
             let pool_event_rx = tycho_feed.subscribe();
             let derived_rx = derived_event_tx.subscribe();
 
+            // The router-facing role doubles as the per-worker scope: only a public pool with a
+            // policy filters; the surplus pool (and any pool without a policy) includes everything.
+            let pool_role = pool_entry.role();
+            let permission = match (pool_role, permission_policy.clone()) {
+                (PoolRole::Public, Some(policy)) => PermissionContext::PublicOnly(policy),
+                _ => PermissionContext::IncludeAll,
+            };
```
louise-poole: It's probably cleaner/easier to maintain this mapping if we implemented it on one of the enums? Something like:
```rust
  impl LiquidityScope {
      /// The per-worker permission context for a pool of this scope.
      /// Only a `Public` pool with a policy filters; `All` (and any pool without a policy) sees everything.
      pub(crate) fn into_permission(self, policy: Option<PermissionPolicy>) -> PermissionContext {
          match (self, policy) {
              (LiquidityScope::Public, Some(p)) => PermissionContext::PublicOnly(p),
              _ => PermissionContext::IncludeAll,
          }
      }
  }
```

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -469,6 +596,165 @@ impl WorkerPoolRouter {
...
+    result
+}
+
+/// v1 constraint: rejects surplus routes unless every exclusive leg is the terminal leg of its
+/// path. This keeps per-leg surplus attribution exact (no inverse simulation needed). Mid-route
+/// exclusive legs and multiple exclusive legs per path are deferred to a future version.
+///
+/// Path boundaries are detected by checking whether the next swap's `token_in` differs from the
+/// current swap's `token_out`. This heuristic is correct for Fynd's sequential route
+/// representation but would need revisiting if routes gain explicit path-boundary markers.
```
louise-poole: This doc string is clear, but breaks from our usual docstring format. Can you adjust this to have something like:
```rust
  /// Returns `true` only for routes....
  ///
  /// Returns `false` for....
```

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -469,6 +596,165 @@ impl WorkerPoolRouter {
...
+/// - `Swap::committed_amount_out` set on each exclusive leg (consumed by the encoder), and
+/// - an order-level [`SurplusInfo`] attached (observability).
+///
+/// Otherwise `public_ranked` is returned unchanged, so the user is never quoted worse than the
+/// public market.
+///
+/// Each exclusive leg's `committed_amount_out` is its realized output reduced by the same
+/// proportion as the order-level reduction; the protocol captures the difference. The per-leg
+/// attribution formula and the bound that keeps the user at or above the committed reference are
+/// derived in the design plan.
```
louise-poole: I don't think referring to external docs like the design plan is a good idea here. Can you rather extrapolate the info that is relevant and include it directly in this docstring? Or if it's very low level details, then put it as inline comments in the fn body.

fynd-core/src/encoding/encoder.rs
```diff
@@ -58,6 +58,12 @@ impl TryFrom<&OrderQuote> for Solution {
             SolveError::FailedEncoding("successful quote must have a route".to_string())
         })?;
 
+        // TODO: when a swap in this route is exclusive, read its `Swap::committed_amount_out`
```
louise-poole: For future reference, with these sorts of TODOs make sure to reference the jira ticket number the TODO will be addressed in so that this is easily traceable.

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -469,6 +596,165 @@ impl WorkerPoolRouter {
...
+
+    // Build the executed quote: clone the surplus candidate, stamp per-leg committed amounts,
+    // pin amount_out to committed, attach SurplusInfo.
+    let mut executed = surplus_quote.clone();
+
+    if let Some(route) = executed.route_mut() {
+        for swap in route.swaps_mut() {
+            if policy.is_exclusive(swap.protocol_component()) {
+                // Proportional reduction: committed_leg = leg.amount_out * committed / realized
+                // (rounds down — safe direction: under-capture)
```
louise-poole: Rounding down produces over-capture no? The pool will capture more as 'surplus' and the user will get less...

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -952,4 +1238,355 @@ mod tests {
...
+
+    fn surplus_pool_roles() -> HashMap<String, PoolRole> {
+        HashMap::from([
+            ("public_pool".to_string(), PoolRole::Public),
+            ("surplus_pool".to_string(), PoolRole::All),
+        ])
+    }
+
+    #[test]
+    fn combine_prefers_surplus_when_it_beats_public() {
```
louise-poole: Can you add a `combine_with_surplus` test for a split route?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -952,4 +1238,355 @@ mod tests {
...
+        // The surplus winner is at the head: user is quoted the committed public output, protocol
+        // captures the surplus. The public candidate remains as a fallback.
+        assert_eq!(combined.len(), 2);
+        assert_eq!(*combined[0].amount_out(), BigUint::from(900u64));
+        assert_eq!(combined[0].committed_amount_out(), Some(&BigUint::from(900u64)));
+        assert_eq!(combined[0].surplus_amount(), Some(&BigUint::from(50u64)));
+    }
+
+    #[test]
+    fn combine_falls_back_to_public_when_surplus_does_not_beat_it() {
```
louise-poole: I'm not such a fan of test names talking about the result cases instead of the focusing on what is tested. When we feel the need to add the outcomes to test names to add distinction between tests it usually is a good indicator that we could use 1 test with multiple test cases (`rstest`).

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -243,13 +336,24 @@ impl WorkerPoolRouter {
...
+
         let mut quotes = Vec::new();
         let mut failed_solvers: Vec<(String, SolveError)> = Vec::new();
         let mut remaining_pools: HashSet<String> = self
             .solver_pools
             .iter()
             .map(|p| p.name().to_string())
             .collect();
+        let mut has_public_response = false;
+        let mut has_surplus_response = false;
```
louise-poole: Can you add tests that test these early return cases?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -469,6 +596,165 @@ impl WorkerPoolRouter {
...
+        .iter()
+        .filter(|(pool, _)| pool_roles.get(pool) == Some(&PoolRole::All))
+        .filter(|(_, q)| q.status() == QuoteStatus::Success)
+        .filter(|(_, q)| {
+            options
+                .max_gas()
+                .map(|max| q.gas_estimate() <= max)
+                .unwrap_or(true)
+        })
+        .filter(|(_, q)| has_valid_exclusive_route(q, policy))
```
louise-poole, line 724: Can you add a test that these max_gas and has_valid_exclusive_route checks fire correctly and we fall back to the public quote in this case?

louise-poole (commented):
Thanks! This is a huge one so took some time for me to go through properly. Thanks for your patience! Was a good start 👌

fynd-core/src/feed/mod.rs
Troshchk: I have changed the naming as following:
  PermissionPolicy -> ExclusivityPolicy
  PoolRole::All -> PoolRole::ExclusiveAccess
  PermissionContext::IncludeAll -> LiquidityScope::All

  Let me know what you think about it

fynd-core/src/feed/permission.rs
Troshchk: responded [here](https://github.com/propeller-heads/fynd/pull/304#discussion_r3602588814)

fynd-core/src/worker_pool_router/mod.rs
Troshchk: Renamed to winning candidate

fynd-core/src/worker_pool_router/mod.rs
Troshchk: I made a rename and mentioned it [here](https://github.com/propeller-heads/fynd/pull/304#discussion_r3602588814)
Let me know if you agree with it

fynd-core/src/feed/permission.rs
Troshchk: I have renamed it and put the info [here](https://github.com/propeller-heads/fynd/pull/304#discussion_r3602588814), lmk what you think

fynd-core/src/worker_pool_router/mod.rs
Troshchk: Yes, true, we should calculate the surplus_amount net gas, but this is for reporting accuracy. The surplus quote amount_out is still set to the mount_out of the public quote

fynd-core/src/worker_pool_router/mod.rs
Troshchk: moved it in a per leg check

fynd-core/src/worker_pool_router/mod.rs
Troshchk: True, fixed

fynd-core/src/worker_pool_router/mod.rs
Troshchk: I am not sure -- I think the main idea is to not quote worse than the public and we won't have a reference to pin the exclusive route
I would say we shouldn't be doing this in v1

fynd-core/src/worker_pool_router/mod.rs
Troshchk: Should we add it when we know, how we want it set up may be? 
Also, I think the fixed bps might not guarantee it at the moment before we decide how to handle the gas/net committed_amount_out

fynd-core/src/worker_pool_router/mod.rs
Troshchk: Yes, added

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -694,15 +699,27 @@ fn combine_with_surplus(
...
-    let surplus_amount = exclusive_candidate.amount_out_net_gas() - committed.amount_out_net_gas();
+    // Gas the user pays to execute the exclusive route, in output-token terms.
+    let exclusive_gas_cost = exclusive_route_amount_out - exclusive_candidate.amount_out_net_gas();
+
+    // The commitment honors two floors: the quoted amount_out never drops below the public
+    // market's, and the user — who pays the exclusive route's gas — never nets less than the
+    // public route would leave them (public_net + gas). Together with the strict net check
+    // above, these gates guarantee the route covers the committed amount.
+    let committed_amount_out =
+        (committed.amount_out_net_gas() + &exclusive_gas_cost).max(public_amount_out.clone());
```
tamaralipows, line 760: Should we do a max here? What if the public amount out is also extremely gas expensive?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -1748,6 +1768,112 @@ mod tests {
         assert_eq!(combined[0].surplus_amount(), None);
     }
 
+    #[test]
+    fn combine_gas_heavier_exclusive_compensates_user() {
```
tamaralipows: No test outcome in the test name pls

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -1748,6 +1768,112 @@ mod tests {
...
+
+        assert_eq!(combined.len(), 2);
+        assert_eq!(*combined[0].amount_out(), BigUint::from(1090u64));
+        assert_eq!(*combined[0].amount_out_net_gas(), BigUint::from(950u64));
+        assert_eq!(combined[0].surplus_amount(), Some(&BigUint::from(10u64)));
+        assert_eq!(combined[0].committed_amount_out(), Some(&BigUint::from(1090u64)));
+    }
+
+    #[test]
+    fn combine_gas_cheaper_exclusive_keeps_public_amount() {
```
tamaralipows: These could perhaps be parametrized no? Not sure if it would help organization though

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -1748,6 +1768,112 @@ mod tests {
...
+            .route()
+            .expect("surplus quote should have a route");
+        let public_leg = route
+            .swaps()
+            .iter()
+            .find(|s| !policy.is_exclusive(s.protocol_component()))
+            .expect("should have a public swap");
+        assert_eq!(public_leg.committed_amount_out(), None);
+
+        // The public branch (600) can't be skimmed, so the full excess (1100 − 1000 = 100) is
```
tamaralipows: I don't like using words like "skimmed" in docstrings though I know Claude loves it. If you ask him to use more concrete language he will do better though

fynd-core/src/worker_pool_router/mod.rs
tamaralipows: I didn't understand this comment - why wouldn't we be able to add this now? I guess this is done in Louise's PR though, not here, right?

fynd-core/src/worker_pool_router/mod.rs
tamaralipows: I think this is a good idea actually. We can't control the pool price, just the fee, so in the worst case the user gets a price instead of no price at all. We could just specify some sort of `MIN_FEE_ONLY_EXCLUSIVE_ROUTE_FOUND` or something to protect the LPs? And set it really high initially? (That would be in Louise's PR)

fynd-core/src/feed/scope.rs
```diff
@@ -0,0 +1,208 @@
...
+    }
+}
+
+/// Which components a worker may ingest into its local graph.
+///
+/// Each worker gets exactly one scope for its lifetime, derived from its pool's role. All graph
+/// filtering for a worker flows through this type, so the worker never reasons about
+/// exclusivity itself — it just hands its topology and incoming events here.
+#[derive(Clone, Debug)]
+pub enum LiquidityScope {
```
tamaralipows: I'm having trouble understanding just from the docstring the difference between this and `PoolRole`. This is the docstring for `PoolRole`
```
/// The role a solver pool (a group of workers) plays in a quote.
```
Why do we have two separate enums for this? In any case perhaps we should document the first line of the docstring better for first-time readers?

tamaralipows (commented):
Thank you! Just a few comments/questions for now for my understanding

fynd-core/src/worker_pool_router/mod.rs
Troshchk: If we drop max here, then whenever the public route is the gas expensive one we'd end up quoting less than the public market shows
Using max in case of extremely expensive public route should actually result in higher value for the user and under capture for protocol

tamaralipows: Ok I understand - I think we can make this a bit clearer in the docstring (maybe the docstring tries to say this but I didn't understand this?)

fynd-core/src/worker_pool_router/mod.rs
louise-poole: Yes exactly. If we miss a committed amount out to pin it to because there is no public route, then we apply our own default fee to the amount out of the exclusive route and pin it to that. I think this is important to have in v1 already as it is the option that earns our LPs the most per swap (max default fee) and also improves our coverage.

fynd-core/src/feed/scope.rs
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    fn market_with(components: Vec<ProtocolComponent>) -> MarketState {
+        let mut market = MarketState::new();
+        market.upsert_components(components);
+        market
+    }
+
+    #[test]
+    fn is_exclusive_reflects_predicate() {
```
louise-poole: Please make sure all test names start with 'test_'.

fynd-core/src/solver.rs
```diff
@@ -688,10 +743,18 @@ impl FyndBuilder {
...
+        // pools include them. `None` ⇒ no pool filters anything (original behaviour).
+        let exclusivity_policy = self.exclusivity_policy.take();
+        let pools = std::mem::take(&mut self.pools);
+
+        for pool_entry in pools {
             let pool_event_rx = tycho_feed.subscribe();
             let derived_rx = derived_event_tx.subscribe();
 
+            let pool_role = pool_entry.role();
+            let scope = pool_role.liquidity_scope(exclusivity_policy.clone());
```
louise-poole: I think we should have a check here to ensure that there is an exclusivity policy set if a pool role is meant to use it (currently I think we can set the policy to All but have exclusivity default to None?)

fynd-core/src/solver.rs
```diff
@@ -604,6 +643,21 @@ impl FyndBuilder {
         self
     }
 
+    /// Sets the predicate that classifies components as exclusive.
+    ///
+    /// Public pools exclude matching components from their graphs; exclusive-access pools
+    /// include them. Without a policy, no pool filters anything.
+    pub fn exclusivity_policy<F>(mut self, predicate: F) -> Self
```
louise-poole, line 662: Hmmm.... I've just realised we set the exclusivity policy globally and not per-workerpool. This is not ideal for if we have more than one exclusive access liquidities that we want running on separate workers. Though that's a big change. Let's not do it in this PR (maybe not even in v1?).

louise-poole, line 662: Another thing here: this is only exposed programatically on FyndBuilder. We need users to be able to set it when they run using cli (so probably best to add it as a config somewhere?).

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -280,11 +405,29 @@ impl WorkerPoolRouter {
...
+                            // Role-aware early return: when an exclusive-access pool is
+                            // configured, only fire once we have ≥1 public AND the
+                            // exclusive-access pool (so the surplus overlay has both inputs).
+                            // Without an exclusive-access pool, use pure
+                            // count-based gating (original behaviour).
+                            let role_ready = if has_exclusive_access_pool {
+                                has_public_response && has_exclusive_access_response
+                            } else {
+                                true
+                            };
```
louise-poole: Do we want to do an early return here though? I assume we'll often only have 1 exclusive route returned then instead of all that we could find. So `combine_with_surplus` would sort and choose the 'best' exclusive route out of a list of only 1. If we have more than one exclusive liquidity pool this means we'd only take the first found and not the best one.

fynd-core/src/feed/scope.rs
louise-poole: It's because `PoolRole` is the struct derived from the config (has copy and deserialise etc derived) and `LiquidityScope` is the worker-side mechanism that combines the poolRole and the exclusivity policy in one so that it can be used to filter the market graph etc. LiquidityScope cannot be directly derived from the config (does not have Copy or Deserialise as it as a Arc field that prevents this).

Having said all that - this can probably be cleaned up. I don't think the worker pool itself needs to know what it's role is, just what exclusivity policy it was assigned. What I'd do:
- Delete this LiquidityScope that is used within the worker pool only. We'd replace it on the worker pool config with a direct `Option<ExclusivityPolicy>`.
- Rename `PoolRole` to `LiquidityScope` for clarity on the user facing interface (the configs)
- Add the filtering logic that was on PoolRole to `ExclusivityPolicy` (`filter_topology` and `scope_event`).

This cleans upo the duplicated nature of the previous design and only passes to the worker pool that it actually uses to act upon.

fynd-core/src/worker_pool_router/mod.rs
louise-poole: >(That would be in Louise's PR)

I don't think this change is for my PR (encoding). Encoding only encodes the info it's given. Setting the committed amount out is done here solving side so this change should be done here.

fynd-core/src/worker_pool_router/mod.rs
Troshchk: I think the same logic actually applies to the public pools. If we have `min_responses` set, than we will always wait only for that amount of responses(so we keep `min_responses` responses from the fastest pools).
The gate added here just ensures we have at least one response from each public and exclusive if exclusive is enabled.
Actually, if we have `min_responses` set to 3 and 1 public quote and 2 exclusive returned, we will be exiting at that point having 2 exclusive returned.
Also, if `min_responses` we will wait for all pools.
Or are you saying we want to await `min_responses` of public and exclusive each?

fynd-core/src/solver.rs
Troshchk: Based on plan, I assumed it is out of scope:
**Out of scope:** on-chain hook contracts, **permissioned-component ingestion config**, hook calldata/
signature encoding (only guarantee the values *reach* the encoder).
Shall I still add it in this PR?

fynd-core/src/worker_pool_router/mod.rs
Troshchk: Do you think this is addressed by:
```
let committed_amount_out = (committed.amount_out_net_gas() + &exclusive_gas_cost).max(public_amount_out.clone());
```

fynd-core/src/feed/scope.rs
Troshchk: I agree, changed it, looks much cleaner now

fynd-core/src/solver.rs
louise-poole: Interesting - I personally think this is a plan oversight. But we can do it in a separate PR.

fynd-core/src/worker_pool_router/mod.rs
louise-poole: Oh I see! I miss understood this - didn't see the min_responses check after these lines. All good then!

fynd-core/src/feed/exclusivity.rs
```diff
@@ -1,14 +1,14 @@
...
-//! through. Isolation is achieved by filtering each worker's local graph topology/events through
-//! its [`LiquidityScope`] — the shared `MarketState` is never duplicated.
-//!
-//! [`PoolRole::ExclusiveAccess`]: crate::PoolRole::ExclusiveAccess
-//! [`LiquidityScope`]: crate::feed::scope::LiquidityScope
+//! public pool returns, while an exclusive-access pool routes through them to capture the
+//! surplus they offer above the best public-market rate. Both pool kinds serve the same
+//! request; they differ only in which liquidity their workers may route through. Isolation is
+//! achieved by filtering each worker's local graph topology/events through the pool's
+//! [`ExclusivityPolicy`](crate::feed::exclusivity::ExclusivityPolicy) — the shared
```
louise-poole: I'm not a fan of claude's tendency to link structs like this. Can you change this from
`[`ExclusivityPolicy`](crate::feed::exclusivity::ExclusivityPolicy)` to just ``ExclusivityPolicy``

fynd-core/src/feed/exclusivity.rs
```diff
@@ -19,7 +19,8 @@ use crate::{
     types::ComponentId,
 };
 
-/// Classifies a [`ProtocolComponent`] as exclusive or public.
+/// Classifies a [`ProtocolComponent`] as exclusive or public, and filters worker inputs
```
louise-poole: Same here. The square brackets bug me.

fynd-core/src/solver.rs
```diff
@@ -156,10 +156,11 @@ pub struct PoolConfig {
...
     connector_tokens: Option<Vec<String>>,
-    /// Pool liquidity scope: `public_only` (default) or `all` (routes through exclusive
-    /// liquidity as well).
+    /// Pool liquidity scope: `public_only` or `all` (routes through exclusive liquidity as
+    /// well). Setting this key (either value) declares exclusive routing intent and requires an
+    /// exclusivity policy; absent = plain deployment, no filtering.
     #[serde(default)]
-    liquidity_scope: LiquidityScope,
+    liquidity_scope: Option<LiquidityScope>,
 }
```
louise-poole: Why change the default? I think we shouldn't allow `None` here (doesn't make sense for there to be no scope) and should keep the behaviour of defaulting to public only.

Troshchk: Here None is only to identify that there is no scope set in the toml, meaning the exclusivity policy is not expected.
So basically, first we check if any of the pools has liquidity_scope. If yes, than we check the exclusivity_policy and if it does not exist --> error
Otherwise we expect to have the exclusivity_policy.
Here:
```        
if self.exclusivity_policy.is_none() &&
            self.pools
                .iter()
                .any(|p| p.liquidity_scope().is_some())
        {
            return Err(SolverBuildError::LiquidityScopeWithoutPolicy);
        }
```

And later, the liquidity scope default is PublicOnly, but if no liquidity policy -- we don't filter anything:
```
// Explicit `public_only` pools (and unscoped pools, harmlessly) get the policy so
            // their workers filter exclusive components out; `all` pools get `None` and ingest
            // everything.
            let pool_scope = pool_entry
                .liquidity_scope()
                .unwrap_or_default();
            let pool_policy = match pool_scope {
                LiquidityScope::PublicOnly => exclusivity_policy.clone(),
                LiquidityScope::All => None,
            };
 ```
 I think this can actually be misleading, coz we enable case with scope PublicOnly and policy being None.
 
 Do we always expect to have the scope set per pool? If yes, than I can change and add the exclusivity policy config in this PR still

louise-poole: > Here None is only to identify that there is no scope set in the toml

I don't think being able to differentiate between set/not set in the toml is important here. You later use `unwrap_or_default` to map the None to the default LiquidityScope - might as well default it from the start?

> Do we always expect to have the scope set per pool?

No - I expect that if one isn't set then it gets inferred as PublicOnly. `#[serde(default)]` will derive that if the flag isn't explicitly used on the cli.

louise-poole: > I think this can actually be misleading, coz we enable case with scope PublicOnly and policy being None.

I'm not sure I follow here? That's a valid case. We should only check and error on a missing policy if a liquidity scope that needs a policy is set. How is this misleading?

Troshchk: I explained why there is a difference -- coz if no exclusivity_policy is set, that means we cannot filter anything and we than use all pools(including exclusive), so no way to enforce public only. And fallback to PublicOnly when there is no exclusivity_policy is not a valid case.
To use publicOnly, we need to filter the pool's graph and this is why we need to have the exclusivity_policy

louise-poole (approved):
Approved so we can unblock testing. Let's address these remaining open issue in new PRs?

PR #305, opened 2026-07-13
ci: make pre-releases safe to cut from any branch

tvinagre:
GitHub pre-releases currently trigger the same side effects as full releases. This guards three of them on `github.event.release.prerelease`:

- `docker.yaml`: the multi-arch manifest no longer retags `latest` for pre-releases (version and `sha-` tags still push)
- `docker.yaml`: `promote-staging` is skipped for pre-releases — the tag is deployed deliberately instead of auto-promoted
- `release.yaml`: crates.io and npm publish jobs are skipped for pre-releases (manual `workflow_dispatch` still works)

With these guards a pre-release cut from any branch builds and pushes only its own version-tagged image, e.g. `ghcr.io/propeller-heads/fynd:0.90.0-rc.1`. Motivation: pre-release staging deploys of unmerged work (capacity harness, PRs #300/#301) without touching `latest` or the registries.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -108,7 +108,8 @@ jobs:
           tags: |
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
-            type=raw,value=latest
+            # Pre-releases (e.g. 0.90.0-rc.1) must not move the floating tag.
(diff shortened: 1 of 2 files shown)


tamaralipows (approved):
Thanks!

PR #306, opened 2026-07-13
feat: support self-hosted custom chains (quote-only)

brunoguerios:
Adds support for running Fynd against a custom EVM chain served by a self-hosted Tycho, plus
docs for pointing Fynd at a self-hosted Tycho on built-in chains.

## What
- **Deps:** bump Tycho (simulation/execution/client/common) 0.331.0 → 0.334.0 for the
  custom-chain registry APIs.
- **Chain resolution:** `parse_chain` now resolves custom chains via the `TYCHO_CHAINS_CONFIG`
  registry, installed explicitly with a new `--chains-config` flag (also on
  `derive-connector-tokens`). `native_token` resolves the wrapped-native token from the registry
  instead of a hardcoded map (fail-fast preserved for chains with no real native token).
- **Quote-only:** where no Tycho router is deployed, Fynd runs quote-only — `GET /v1/info`
  reports `router_address: null` and an encoding-enabled quote returns `501 Not Implemented`
  (`ENCODING_UNAVAILABLE`). Genuine encoding failures still return `422`. Built-in chains are
  unchanged (encoding fully supported).
- **Clients:** OpenAPI spec + TypeScript and Rust clients regenerated for the now-nullable
  `router_address`.
- **Docs:** how to point Fynd at a self-hosted Tycho, and how to run a custom chain.

Router/executor deployment for a custom chain remains an ops flow, synced with clients on demand.

## Testing
`cargo nextest run --workspace --all-targets --all-features` → 948 passed / 10 skipped.
TS client typecheck + lint + vitest → 152 passed. `cargo fmt --check` clean. No OpenAPI drift.

## Follow-ups (non-blocking)
- TS client `use_vaults_funds` vs Rust `Ok(None)` parity on null router.
- `@fynd/autogen` reference in repo docs is stale (schema generates into `client/src/`).

Related Tycho docs: https://docs.propellerheads.xyz/tycho/for-solvers/self-hosted-evm-chain

### DIFF
--- a/clients/rust/src/client.rs
+++ b/clients/rust/src/client.rs
@@ -988,7 +988,7 @@ where
         let spender_addr = match params.transfer_type {
             UserTransferType::TransferFrom => {
                 let router_address = info.router_address().ok_or_else(|| {
-                    FyndError::Protocol(
+                    FyndError::Config(
                         "server has no router_address; encoding is unavailable on this chain"
                             .into(),
                     )
@@ -2002,7 +2002,13 @@ mod tests {
 
         assert_eq!(info1.chain_id(), 1);
         assert_eq!(info2.chain_id(), 1);
-        assert_eq!(info1.router_address().unwrap().as_ref(), &[0x01u8; 20]);
+        assert_eq!(
+            info1
+                .router_address()
+                .expect("mock info response includes a router address")
+                .as_ref(),
+            &[0x01u8; 20],
+        );
         assert_eq!(info1.permit2_address().as_ref(), &[0x02u8; 20]);
         // MockServer verifies expect(1) on drop.
     }
--- a/clients/rust/src/client.rs
+++ b/clients/rust/src/client.rs
@@ -987,7 +987,13 @@ where
         let info = self.info().await?;
         let spender_addr = match params.transfer_type {
             UserTransferType::TransferFrom => {
-                mapping::bytes_to_alloy_address(info.router_address())?
+                let router_address = info.router_address().ok_or_else(|| {
+                    FyndError::Protocol(
+                        "server has no router_address; encoding is unavailable on this chain"
+                            .into(),
+                    )
+                })?;
+                mapping::bytes_to_alloy_address(router_address)?
             }
             UserTransferType::TransferFromPermit2 => {
                 mapping::bytes_to_alloy_address(info.permit2_address())?
@@ -1996,7 +2002,7 @@ mod tests {
 
         assert_eq!(info1.chain_id(), 1);
         assert_eq!(info2.chain_id(), 1);
-        assert_eq!(info1.router_address().as_ref(), &[0x01u8; 20]);
+        assert_eq!(info1.router_address().unwrap().as_ref(), &[0x01u8; 20]);
         assert_eq!(info1.permit2_address().as_ref(), &[0x02u8; 20]);
         // MockServer verifies expect(1) on drop.
     }
--- a/fynd-core/src/solver.rs
+++ b/fynd-core/src/solver.rs
@@ -344,7 +344,7 @@ struct CustomPoolEntry {
 struct BuiltComponents {
     tycho_feed: TychoFeed,
     gas_price_fetcher: GasPriceFetcher<EthereumRpcClient>,
-    router_fee_fetcher: RouterFeeFetcher,
+    router_fee_fetcher: Option<RouterFeeFetcher>,
     computation_manager: ComputationManager,
     computation_event_rx: broadcast::Receiver<MarketEvent>,
     computation_shutdown_tx: broadcast::Sender<()>,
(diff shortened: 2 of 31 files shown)


clients/rust/src/client.rs
```diff
@@ -1996,7 +2002,7 @@ mod tests {
 
         assert_eq!(info1.chain_id(), 1);
         assert_eq!(info2.chain_id(), 1);
-        assert_eq!(info1.router_address().as_ref(), &[0x01u8; 20]);
+        assert_eq!(info1.router_address().unwrap().as_ref(), &[0x01u8; 20]);
```
tvinagre: even though it's a test, I'd rather have an expect with a reason

clients/rust/src/client.rs
```diff
@@ -987,7 +987,13 @@ where
         let info = self.info().await?;
         let spender_addr = match params.transfer_type {
             UserTransferType::TransferFrom => {
-                mapping::bytes_to_alloy_address(info.router_address())?
+                let router_address = info.router_address().ok_or_else(|| {
+                    FyndError::Protocol(
```
tvinagre: Not sure if this should be `Protocol` or `Config` error. I think the latter but pls check

fynd-core/src/types/constants.rs
```diff
@@ -65,28 +16,34 @@ impl UnsupportedChainError {
...
-///
-/// Returns `UnsupportedChainError` if the chain is not in the registry.
+/// Returns `UnsupportedChainError` if the chain (custom) is not registered.
 pub fn native_token(chain: &Chain) -> Result<Address, UnsupportedChainError> {
-    NATIVE_TOKEN
-        .get(chain)
-        .cloned()
-        .ok_or(UnsupportedChainError { chain: *chain })
+    let address = chain
+        .try_wrapped_native_token()
```
tvinagre, line 26: nice - this could be later made static on tycho-common!

fynd-core/src/types/constants.rs
```diff
@@ -65,28 +16,34 @@ impl UnsupportedChainError {
...
+        .map(|token| token.address)
+        .map_err(|_| UnsupportedChainError { chain: *chain })?;
+
+    // A zero address is a placeholder (e.g. Starknet) meaning the chain has no real
+    // wrapped-native token; treat it as unsupported so callers still fail fast instead of
+    // deriving a gas token from a placeholder address.
+    if address.is_zero() {
+        return Err(UnsupportedChainError { chain: *chain });
```
tvinagre, line 38: I don't understand why this would be unsupported?

tvinagre (commented):
Code looks good, but it's an opinionated interface change, so I'd like to also get @tamaralipows or @louise-poole opinion here

clients/rust/src/client.rs
brunoguerios: Yeah, I think Config is a better fit indeed. Changed ✅

clients/rust/src/client.rs
brunoguerios: ok - fixed it here ✅

fynd-core/src/types/constants.rs
brunoguerios: Starknet doesn't have a wrapped native token, so we set it as the zero address in tycho common. With this refactor of relying in Tycho to provide the wrapped native tokens, Fynd would no longer fail on Starknet as not supported and would forward the zero address as a valid wrapped native token. This guard protects that and keeps current behavior of saying zero address is not a valid wrapped native token.

fynd-core/src/types/constants.rs
brunoguerios: Yeah - we'd have to properly evaluate pros and cons of making this static, but it might indeed make sense. Since it's not in the hot path, I'd say we should be fine as is.

fynd-core/src/types/constants.rs
brunoguerios: I updated the comment to better explain all this - let me know what you think.

fynd-core/src/solver.rs
```diff
@@ -1316,8 +1335,8 @@ pub struct SolverParts {
     computation_shutdown_tx: broadcast::Sender<()>,
     /// Chain this solver is configured for.
     chain: Chain,
-    /// Address of the Tycho Router contract on this chain.
-    router_address: Bytes,
+    /// Address of the Tycho Router contract on this chain, or `None` on a quote-only chain.
```
tamaralipows, line 1338: Is this necessary? Are we planning to have quote-only chains? Curious about the motivation of this - I might have missed this discussion

tamaralipows (approved):
Thanks! All looks good, just a question for my understanding.

Also - today I've merged Plasma as an variant of the Chain enum - could you update this PR with the latest tycho so that this is reflected? Otherwise from what I understand, it will still register Plasma as a custom chain, right?

fynd-core/src/solver.rs
brunoguerios: This enables users to play around with quoting a custom chain even before we're able to deploy a router and executors in it. Otherwise they get completely blocked until contracts are deployed.

brunoguerios:
> Also - today I've merged Plasma as an variant of the Chain enum - could you update this PR with the latest tycho so that this is reflected? Otherwise from what I understand, it will still register Plasma as a custom chain, right?

I updated over here ✅ - but answering your question, if users don't update to the latest Fynd version, they'll have to register Plasma as a custom chain to support it - otherwise it won't be available. If they do update to latest Fynd/Tycho version, they get it as a builtin chain.

PR #308, opened 2026-07-14
refactor: Hindsight clean up

louise-poole:
Refactor and clean up of tools/hindsight:

  - Renamed client → venue throughout — the old name was ambiguous; "venue" matches what the field actually represents
  (tx.to, the order-flow owner where a user initiated their swap).
  - Moved Allium + verify logic to `src/verify/` — was scattered inline; now has a proper module boundary.
  - Added clippy lint config (`[lints.clippy]`) with pedantic + anti-panic lints.
  - Replaced `u256_to_f64` string round-trip with a limb-based conversion — clearly documented assumptions made to avoid throwing errors.
  - Added outcome label to hindsight_savings_bps histogram, enabling histogram_quantile median queries split by win/loss on the dashboard.


tamaralipows (approved):
Thanks! I'll fix clippy and merge this

PR #309, opened 2026-07-14
ci: migrate release to release-v3 (release-plz)

kayibal:
Migrates fynd's release from `release-v2` (semantic-release, pushes to main) to the PR-based `release-v3`, using the **release-plz** engine (auto-selected for Rust).

## Changes
- **`release-plz.toml`** — unified-version workspace config using the single-tag pattern (tags/releases disabled workspace-wide, enabled on the `fynd` crate) so one bare tag (e.g. `0.90.0`) is created per release. `publish = false`, `semver_check = false`.
- **`v2-main-workflow.yaml`** — calls `release-v3`. fynd is a library, so there's no build/deploy here.
- **`release.yaml`** — unchanged publishing (crates + npm on the release event), plus one new step: set the npm client version from the release tag (release-plz bumps `Cargo.toml`/`Cargo.lock` but not the npm `package.json`). Also drops the `workflow_dispatch` trigger: a failed release run is recovered by re-running it (re-runs keep the release event payload), while a fresh dispatch would have no tag and attempt a stale npm version.
- Remove `release.config.js`.

## Flow
- Merge to `main` → release-plz opens/updates a release PR (bumps the workspace version + `Cargo.lock` + `CHANGELOG`).
- Merge that PR → single bare tag + GitHub Release → `release.yaml` publishes crates + npm. No commit to main.

## Depends on (all merged)
- propeller-heads/ci-cd-templates#90 — language-dispatching `release-v3` (release-plz for rust).
- propeller-heads/ci-cd-templates#91 — rust toolchain channel fix for the SHA-pinned dtolnay action.
- The App granted `issues: write`.

## Verify on the first release PR
- release-plz bumps the **shared workspace version** once (not per-crate) and updates `Cargo.lock` for all members — not exercised by the single-crate token-quoter validation.
- The `release` event fires `release.yaml` (crates + npm publish) — fynd's separate-workflow trigger path, also not covered by token-quoter.
- `clients/typescript/client/package.json` is bumped only at publish time, not in-tree (intentional; avoids commits to main).


zizou0x:
**The `workflow_dispatch` trigger path of `release.yaml` no longer works after this migration.**

Since release-plz never bumps `clients/typescript/client/package.json` in-tree (semantic-release used to commit the bump to main), the in-tree version stays permanently stale. On a manual dispatch:

1. `github.event.release` doesn't exist, so `tag_name` is `''`
2. Checkout falls back to `github.sha` (tip of main)
3. The new "Set client version from release tag" step is skipped (`if: ${{ github.event.release.tag_name != '' }}`)
4. `npm publish` attempts the stale `package.json` version, which already exists on npm → 409 Conflict

It fails loudly rather than publishing a wrong version, but it kills the manual re-publish escape hatch (e.g. re-running after an npm registry flake when crates already published) — and it fails at the last step, after install + build.

Suggested fix — give dispatch a tag input and use it wherever `tag_name` is used:

```yaml
on:
  workflow_dispatch:
    inputs:
      tag:
        description: "Release tag to publish (e.g. 0.91.0)"
        required: true

# in publish-npm:
      - name: Checkout
        with:
          ref: ${{ github.event.release.tag_name || inputs.tag || github.sha }}

      - name: Set client version from release tag
        env:
          TAG: ${{ github.event.release.tag_name || inputs.tag }}
        if: ${{ env.TAG != '' }}
        run: npm version "$TAG" --no-git-tag-version --allow-same-version
```

(Routing the tag through `env:` also avoids interpolating event data directly into the `run:` script.)

zizou0x (approved):
LGTM, just one Claude comment, but happy to look into it in the next iteration

louise-poole (approved):
Thank you for this! 🙏

kayibal:
Resolved differently in 94ad867: removed `workflow_dispatch` instead of adding a tag input. A failed release run is recovered by re-running it — GitHub re-runs keep the original `release` event payload, so `tag_name` is still set and the publish behaves exactly as the original run (and "Re-run failed jobs" won't repeat an already-succeeded crates publish). That covers the recovery case the dispatch path was for, so rather than wiring a tag input through checkout + versioning to prop up a broken escape hatch, the trigger is gone. If we ever need to republish an arbitrary old tag, we can add a dispatch-with-tag-input then.

PR #311, opened 2026-07-14
feat: expose Fynd version via /v1/info and build_info metric

brunoguerios:
Implements ENG-6178 (Fynd Auto tech doc §3, steps F1 + F2): makes the Fynd binary version observable. Both changes are additive and backward-compatible.

## Changes

**`/v1/info` version (F1)**
- `InstanceInfo` gains a `version` field (populated from `CARGO_PKG_VERSION` in the `/v1/info` handler). It is `#[serde(default)]`, so a newer client tolerates an older server that predates the field.
- `InstanceInfo` moves from a positional `new(...)` to a builder and is marked `#[non_exhaustive]`, so future instance-metadata fields don't churn call sites.
- The Rust client's `InstanceInfo` and its mapping carry `version` through, so `FyndClient::info()` surfaces it.
- Regenerated `clients/openapi.json` and the TypeScript schema.

**`fynd_build_info` metric (F2)**
- Emits `fynd_build_info{version="x.y.z"} 1` on `/metrics` at startup, behind the `metrics` feature.

## Verification
- Full CI (format, clippy, doc, nextest, OpenAPI/TS drift) passes locally.
- `/v1/info` returns `version`; `/metrics` exposes `fynd_build_info` with the version label.

### DIFF
--- a/fynd-rpc-types/src/lib.rs
+++ b/fynd-rpc-types/src/lib.rs
@@ -1197,6 +1197,7 @@ impl HealthStatus {
 /// Static metadata about this Fynd instance, returned by `GET /v1/info`.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
+#[non_exhaustive]
 pub struct InstanceInfo {
     /// EIP-155 chain ID (e.g. 1 for Ethereum mainnet).
     #[cfg_attr(feature = "openapi", schema(example = 1))]
@@ -1213,12 +1214,22 @@ pub struct InstanceInfo {
         schema(value_type = String, example = "0x000000000022D473030F116dDEE9F6B43aC78BA3")
     )]
     permit2_address: Bytes,
+    /// Fynd binary version (Cargo package version, e.g. "0.89.1").
+    ///
+    /// Defaults to empty when absent so newer clients tolerate older servers that predate it.
+    #[serde(default)]
+    #[cfg_attr(feature = "openapi", schema(example = "0.89.1"))]
+    version: String,
 }
 
 impl InstanceInfo {
-    /// Creates a new instance info.
-    pub fn new(chain_id: u64, router_address: Bytes, permit2_address: Bytes) -> Self {
-        Self { chain_id, router_address, permit2_address }
+    /// Starts building an instance info from the required immutable fields.
+    pub fn builder(
+        chain_id: u64,
+        router_address: Bytes,
+        permit2_address: Bytes,
+    ) -> InstanceInfoBuilder {
+        InstanceInfoBuilder { chain_id, router_address, permit2_address, version: String::new() }
     }
 
     /// EIP-155 chain ID.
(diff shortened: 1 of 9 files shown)


fynd-rpc-types/src/lib.rs
```diff
@@ -1235,6 +1246,38 @@ impl InstanceInfo {
...
+
+    /// Fynd binary version.
+    pub fn version(&self) -> &str {
+        &self.version
+    }
+}
+
+/// Builder for [`InstanceInfo`]. Keeps future `/v1/info` fields cheap to add.
+#[derive(Debug, Clone)]
+pub struct InstanceInfoBuilder {
```
zizou0x, line 1266: I wonder if it makes sense to prepare the struct for future addition? 

`/v1/info` is considered public interface so we should really avoid breaking it. I have a feeling that making it cheap to add fields could encourage devs to do it and break the interface without noticing.

fynd-rpc-types/src/lib.rs
```diff
@@ -1213,12 +1214,22 @@ pub struct InstanceInfo {
         schema(value_type = String, example = "0x000000000022D473030F116dDEE9F6B43aC78BA3")
     )]
     permit2_address: Bytes,
+    /// Fynd binary version (Cargo package version, e.g. "0.89.1").
+    ///
+    /// Defaults to empty when absent so newer clients tolerate older servers that predate it.
```
zizou0x, line 1227: Do old client tolerate new structs too? (ignore unknown fields)

zizou0x (commented):
Just two small questions but looks good!

fynd-rpc-types/src/lib.rs
brunoguerios: If we we expect to expose more instance metadata over time (as the story states) I think making it cheap to add fields is the correct approach.

The builder doesn't make breaking the interface cheaper — only facilitate additive changes, which are the safe direction. Two separate contracts:

- **Wire** (`/v1/info` JSON): the builder is irrelevant here. Adding a response field is backward-compatible either way — old clients ignore it, and `#[serde(default)]` keeps new-client/old-server working. Rename/remove/retype are the breaking edits, and the builder doesn't ease any of those.
- **Rust crate API** (`fynd-rpc-types`/`fynd-client`, semver-gated): this is what the builder buys. With positional `new(chain_id, router, permit2)`, each new field changes the signature → a breaking API change (cargo-semver-checks flags it) plus call-site churn. With `builder()` + `#[non_exhaustive]`, a new field is just an optional setter + getter → additive, non-breaking.

And a shape change can't slip through unnoticed: any change to `InstanceInfo` forces regenerating `openapi.json` + `schema.d.ts` (visible in the PR diff, CI drift-fails otherwise), the semver check gates the Rust side, and a regression test rejects any new required field that's missing a serde default.

I've added a note on the struct spelling these rules out for the next person.

fynd-rpc-types/src/lib.rs
brunoguerios: Yes. `InstanceInfo` has no `#[serde(deny_unknown_fields)]`, so serde ignores unknown fields by default — an old client deserializing a response from a newer server just drops the fields it doesn't recognize. That's the opposite direction from this doc line (which covers new-client/old-server via `#[serde(default)]`), so both directions are safe:

- **old client ← new server** (extra fields) → ignored — covered by `instance_info_deserializes_and_ignores_unknown_fields` (feeds a `future_field`, asserts it's dropped).
- **new client ← old server** (missing field) → defaulted — covered by `instance_info_deserializes_without_version`.

fynd-rpc-types/src/lib.rs
zizou0x: > Adding a response field is backward-compatible either way — old clients ignore it, and #[serde(default)] keeps new-client/old-server working. Rename/remove/retype are the breaking edits, and the builder doesn't ease any of those.

This is assuming that people are using Rust and our DTO structs. Which is not always the case. Adding a response field here isn't backward compatible.

This is why I'm wondering if making the rust crate API flexible with adding new field is a good move given that it would be a breaking change if we do it

brunoguerios: I'm not sure the premise is right though.
Adding fields to the response shouldn't be a breaking change for any other client as well, right?
If I'm not mistaken, they'd have to make it explicitly strict in order to not ignore unknown fields by default.
Do you know if that's the case?

zizou0x: > If I'm not mistaken, they'd have to make it explicitly strict in order to not ignore unknown fields by default.

Hm, this depends on the language/lib AFAIK. But I guess here we're assuming that in most libs the strictness is opt-in so this wouldn't be breaking, which is probably fair for most modern languages. So I guess we can say that we're good on that side.

To be honest I am still not 100% convinced by the value that this builder pattern gives us (vs the code/interface addition) but it's minor, I won't block the PR for it 👍

zizou0x (approved):
LGTM! Thanks!

PR #312, opened 2026-07-14
refactor(hindsight): move chain and protocol knowledge into the address book

tamaralipows:
Preparation for integrating new venues, solvers, and chains: everything the decoder knows about a specific chain or protocol now lives either in the per-chain address book (TOML) or in that protocol's own module. No behavior changes.

Review by commit.

Custom `--registry` files need the new sections and fail loudly at load if they are missing.

### DIFF
--- a/tools/hindsight/src/decoder/registry/ethereum.toml
+++ b/tools/hindsight/src/decoder/registry/ethereum.toml
@@ -1,4 +1,4 @@
-# Ethereum address book for trade decoding: which contracts are solvers, clients, batch
+# Ethereum address book for trade decoding: which contracts are solvers, venues, batch
 # settlers, and venue infrastructure. Loaded by `Registry` (see registry.rs for what each section
 # drives); pass a modified copy via `--registry` to extend it without recompiling.
 
@@ -110,7 +110,7 @@ fee_collectors = [
 
 # MetaMask's router names the solver API behind each swap in a calldata aggregatorId like
 # "oneInchV6FeeDynamic" or "okx6". Each entry maps a lowercase substring of that id to a solver
-# name used in this book; ids that match no entry are recorded as-is.
+# name used in this address book; ids that match no entry are recorded as-is.
 [venues.metamask.solver_aliases]
 oneinch = "1inch"
 zeroex = "0x"
--- a/tools/hindsight/src/decoder/registry/ethereum.toml
+++ b/tools/hindsight/src/decoder/registry/ethereum.toml
@@ -10,6 +10,15 @@ wrapped_native = "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"
 # a solver, not the trader (CoW Protocol Settlement).
 batch_settlers = ["0x9008d19f58aabd9ed0d60971565aa8510560ab41"]
 
+# Stablecoins anchoring the native token to USD for reporting (address = decimals): a priced
+# stablecoin is worth ~$1 per 10^decimals native units, which pins the native token's USD value.
+[usd_stablecoins]
+"0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48" = 6  # USDC
+"0xdac17f958d2ee523a2206206994597c13d831ec7" = 6  # USDT
+"0x6b175474e89094c44da98b954eedeac495271d0f" = 18 # DAI
+"0x4c9edd5852cd905f086c759e8383e09bff1e68b3" = 18 # USDe
+"0xdc035d45d973e3ec169d2276ddab16f1e407384f" = 18 # USDS
+
 # Solver routers — the venue that actually settles a swap.
 [solvers]
 # 1inch v6 and v5 Aggregation Routers
--- a/tools/hindsight/src/decoder/registry/ethereum.toml
+++ b/tools/hindsight/src/decoder/registry/ethereum.toml
@@ -105,3 +105,19 @@ fee_collectors = [
     "0xe3478b0bb1a5084567c319096437924948be1964",
     "0xf326e4de8f66a0bdc0970b79e0924e33c79f1915",
 ]
+
+# MetaMask's calldata solver vocabulary: its router names the solver API behind each swap in an
+# aggregatorId like "oneInchV6FeeDynamic" or "okx6". The first substring below found in the
+# lowercased id maps it to the solver label used in this book — trimming MetaMask's version/fee
+# decoration, not a 1:1 rename. Ids matching nothing pass through as-is: still more informative
+# than a raw executor address, and a signal to extend this table.
+[venues.metamask.solver_aliases]
+oneinch = "1inch"
+zeroex = "0x"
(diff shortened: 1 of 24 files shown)


tools/hindsight/src/decoder/solvers/mod.rs
```diff
@@ -30,6 +30,14 @@ pub(crate) struct SolverQuote {
     pub timestamp: Option<u64>,
 }
 
+/// Match-time veto: order shapes a solver's own logs mark as not same-chain swaps, checked
+/// before the transaction costs a trace. Netting such a transaction would pair its input with a
+/// leftover refund as a phantom trade. Adding a solver's veto is one check here — the matching
+/// in `strategy` stays solver-agnostic.
+pub(crate) fn match_veto(logs: &[Log]) -> Option<&'static str> {
```
Troshchk, line 39: Will there be more logic here? 
If it is only for the cross chain transfers -- may be the name should reflect it?

tools/hindsight/src/decoder/solvers/mod.rs
```diff
@@ -30,6 +30,14 @@ pub(crate) struct SolverQuote {
     pub timestamp: Option<u64>,
 }
 
+/// Match-time veto: order shapes a solver's own logs mark as not same-chain swaps, checked
+/// before the transaction costs a trace. Netting such a transaction would pair its input with a
+/// leftover refund as a phantom trade. Adding a solver's veto is one check here — the matching
+/// in `strategy` stays solver-agnostic.
```
louise-poole: I don't find this very human-readable. Actually I think the first sentence is a bit broken no? Also this doc string is specific to lifi's veto reason, I think it should rather be made generic?

tools/hindsight/src/decoder/registry/ethereum.toml
```diff
@@ -6,6 +6,10 @@
 # intermediary.
 wrapped_native = "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"
 
+# Token-movement infrastructure that traces touch on most swaps without ever being the settling
+# venue: attribution guesses skip these (the wrapped-native token above is always skipped too).
```
louise-poole: I also found this hard to understand at first pass. Something like this feels easier to understand immediately:
```suggestion
# Contracts nearly every swap routes through but that never settle the trade. Hindsight guesses an 
# unknown swap's venue from on-chain activity; without this list those guesses would land on the 
# plumbing instead of the real venue. The wrapped-native token above is skipped the same way.
```

tools/hindsight/src/decoder/registry/ethereum.toml
```diff
@@ -6,6 +6,10 @@
...
+infrastructure = ["0x000000000022d473030f116ddee9f6b43ac78ba3"]
```
louise-poole: Can you use comments to label these contracts?
```suggestion
infrastructure = ["0x000000000022d473030f116ddee9f6b43ac78ba3"] # Permit2
```

tools/hindsight/src/resolve/monitor.rs
```diff
@@ -69,10 +69,6 @@ pub(crate) struct MonitorArgs {
     #[arg(long, env = "TYCHO_URL")]
     pub tycho_url: String,
 
-    /// Chain to monitor (the decoder is Ethereum-only for now)
-    #[arg(long, default_value = "ethereum")]
-    pub chain: String,
-
```
louise-poole, line 75: It makes more sense here in the higher level args than in the RPCArgs no? In case we have any non-rpc chain specific configs we find we have to deal with (maybe making tycho_url defaults etc?).

tools/hindsight/src/decoder/registry/ethereum.toml
```diff
@@ -105,3 +105,19 @@ fee_collectors = [
     "0xe3478b0bb1a5084567c319096437924948be1964",
     "0xf326e4de8f66a0bdc0970b79e0924e33c79f1915",
 ]
+
+# MetaMask's calldata solver vocabulary: its router names the solver API behind each swap in an
+# aggregatorId like "oneInchV6FeeDynamic" or "okx6". The first substring below found in the
+# lowercased id maps it to the solver label used in this book — trimming MetaMask's version/fee
+# decoration, not a 1:1 rename. Ids matching nothing pass through as-is: still more informative
+# than a raw executor address, and a signal to extend this table.
```
louise-poole: This is very similar to `normalize_solver`'s docstring no? I think this can be split better - this toml one should have only the info needed for the person populating this config to understand what is needed here (not how it is applied).

louise-poole (commented):
Much needed improvements! Looking gooood! Just some nit-picks.

tools/hindsight/src/decoder/solvers/mod.rs
tamaralipows: This should ideally be easily expandable as we discover more veto cases

tamaralipows: Restructured so each new veto case is one more if-block (2ee081ef0) — keeping the generic name since more cases will land here. 🤖

tools/hindsight/src/decoder/solvers/mod.rs
tamaralipows: Rewrote it — generic now, lifi is just the first case. 2ee081ef0. 🤖

tools/hindsight/src/decoder/registry/ethereum.toml
tamaralipows: Took your wording (s/venue/settling venue/ for accuracy). 2ee081ef0. 🤖

tools/hindsight/src/decoder/registry/ethereum.toml
tamaralipows: Done — Permit2 labeled inline. 2ee081ef0. 🤖

tools/hindsight/src/resolve/monitor.rs
tamaralipows: Agreed the struct was misnamed — it was never really "RPC args". Renamed it to `ChainArgs` (chain name + how to reach it), still flattened into every subcommand, so any future non-RPC chain-specific config (tycho_url defaults etc.) has an obvious home that all subcommands inherit. 2ee081ef0. 🤖

tools/hindsight/src/decoder/registry/ethereum.toml
tamaralipows: Trimmed to config-author info; the matching mechanics stay on `normalize_solver`. 2ee081ef0. 🤖

louise-poole (approved):
Looks good! Only small gripe with the clauded commits is that docstring changes were tagged as 'refactor' instead of 'docs'. Would have preferred the latter.

tamaralipows:
@louise-poole not too late to change them ma'am! :)

PR #313, opened 2026-07-14
refactor(hindsight): Decoding flexibility and organization

tamaralipows:
Restructures the hindsight decoder so adding a new client, or a new way to read a swap, is small and local. This is organization — the trades that already decode still decode the same way.

The commits first put decoding behind two traits (`DecodeStrategy` and `VenueKnowledge`), then unify them into one `TradeDecoder` trait. Extending the two-trait design made it ambiguous which trait decoded a transaction and threatened a third, so the final commit collapses selection onto one axis: the matched entity — sender, maker, or venue — lists the decoders to try.

Please start with the [readme](https://github.com/propeller-heads/fynd/blob/tnl/hindsight-structural/tools/hindsight/README.md).
And look at this [PR](https://github.com/propeller-heads/fynd/pull/324) for reference for what it can look like to add a venue or a solver.


tools/hindsight/src/decoder/transfer_ledger.rs
```diff
@@ -239,41 +244,26 @@ fn net_positive(
...
-    drop_residue_legs(&mut net_sent, flows, received);
-    drop_residue_legs(&mut net_received, flows, sent);
+    // The one-directional residue condition looks at the tracked address's gross flow on the
+    // side opposite the leg: for a net-sent leg that is what it received, and vice versa.
+    drop_residue_legs(&mut net_sent, amounts_by_token, |amounts| amounts.received);
+    drop_residue_legs(&mut net_received, amounts_by_token, |amounts| amounts.sent);
 
     if net_sent.len() != 1 || net_received.len() != 1 {
-        // Flow on both sides but more than one significant token on one of them: a real batch
+        // TraderFlow on both sides but more than one significant token on one of them: a real batch
```
louise-poole: This looks like an accidental rename? Maybe we can just say "Value on both sides" instead?

tools/hindsight/src/decoder/transfer_ledger.rs
```diff
@@ -342,7 +330,7 @@ mod tests {
     }
 
     #[test]
-    fn multi_hop_nets_correctly() {
+    fn multi_hop_swap() {
```
louise-poole: Please prepend test names with `test_`

tools/hindsight/src/decoder/maker.rs
```diff
@@ -36,14 +36,14 @@ use crate::decoder::{
 ///   EIP-7702 delegation) is indistinguishable from a pool here, so its fills are dropped.
 pub(crate) async fn find_maker_trade<P: Provider>(
```
louise-poole: The word 'maker' here is a bit confusing (has heavy connotations with things like 'market maker' etc.) Maybe we can use the word 'principal' to refer to underlying swapper? `find_principle_trade`? Or something similar?

tools/hindsight/src/decoder/netting.rs
```diff
@@ -0,0 +1,145 @@
+//! Transfer-netting: recover a swap from what actually moved.
+//!
+//! The evidence is the ERC-20 `Transfer` events plus the native transfers recovered from the
+//! trace (see `transfer_ledger`) — what actually moved, not what any contract or calldata
+//! declared. It needs no knowledge of any router's format.
+//!
+//! This module is both a toolkit and two decoders. The toolkit — [`sender_flow`] and
```
louise-poole: I know claude loves to do this - but I'm not a fan of the square brackets around struct or fn references...

tools/hindsight/src/decoder/netting.rs
```diff
@@ -0,0 +1,145 @@
...
+//! [`venue_flow`] — is the shared netting engine the venue decoders build on. The decoders are
```
louise-poole: It looks like the main netting logic (or 'engine') is actually in `transfer_ledger.rs` (`net_swap`, `net_trade` etc). So this file just contains decoders no? Maybe we can rename the file `netting_decoders` so that this is more clear. I jumped to this file expecting the main netting calculation to be here and it isn't.

tools/hindsight/src/decoder/maker.rs
louise-poole: Actually I don't think Principal works too well with the other 'marker' related structs in this code. Maybe `Intent` was better? Or `Beneficiary`? IF we decide maker is indeed best then that's fine... I think none of these options are perfect 😅

louise-poole (approved):
Thanks! Just a couple smaller suggestions, otherwise I've approved!

tools/hindsight/src/decoder/maker.rs
tamaralipows: Renamed to Intent

PR #314, opened 2026-07-14
feat: add fynd-benchmark generate-requests subcommand

tvinagre:
Adds a `generate-requests` subcommand to `fynd-benchmark` that produces a per-chain synthetic swap-request dataset (JSON) for capacity load testing. The existing `download-trades` dataset is Ethereum-only; this covers other chains.

## What it does

- Fetches token and pool-count data from Tycho via the new `fynd_rpc::protocols::fetch_token_pool_stats`, which reuses the pool-count query behind `derive-connector-tokens` (counts pools per token, joins Tycho token metadata for symbols and decimals).
- Samples token pairs weighted by pool count, excluding same-token pairs, using a seeded `StdRng` so output is deterministic for a given `--seed` and token set.
- Log-spaces amounts across ~4 orders of magnitude around one whole token, scaled by each token's decimals.
- Writes a JSON array loadable by the existing `--requests-file` loaders (`load_request_templates` / `load_requests_from_file`).

## CLI

```
fynd-benchmark generate-requests --chain base [--tycho-url <url>] \
  --top-n-tokens 50 --num-requests 2000 --seed 42 --output trades.json
```

Accepted `--chain` values: `ethereum`, `base`, `unichain`, `bsc`, `arbitrum`, `polygon` (case-insensitive). `zksync` parses but has no default Tycho URL, so pass `--tycho-url`. The Tycho URL defaults to the chain's Fynd endpoint; auth uses `TYCHO_API_KEY`.

## Tests and validation

Sampling and serialization are covered with fixture token data (no network): amount scaling, weighted pair sampling with no self-pairs, seed determinism, minimum-token guard, and a round-trip that loads the generated dataset back through `load_request_templates`.

The Tycho fetch is a thin network layer. Against a hosted endpoint it was exercised only up to the authentication boundary (the RPC call is made and auth is enforced). Live end-to-end generation against ethereum and base is validated in-cluster, where a one-off Job in the `staging-fynd` namespace runs the command with `TYCHO_API_KEY` mounted from the existing secret, so the key never leaves the cluster.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3386,6 +3386,7 @@ dependencies = [
  "fynd-rpc",
  "fynd-tools-common",
  "num-bigint",
+ "rand 0.9.4",
  "reqwest 0.12.28",
  "serde",
  "serde_json",
--- a/fynd-rpc/CLAUDE.md
+++ b/fynd-rpc/CLAUDE.md
@@ -9,7 +9,7 @@ infrastructure.
 |---|---|
 | `builder.rs` | `FyndRPCBuilder` wraps `FyndBuilder`, adds HTTP server config. `FyndRPC` struct runs the server with graceful shutdown |
 | `config.rs` | `WorkerPoolsConfig` (TOML loader), `BlocklistConfig`, `defaults` module re-exporting `fynd-core` defaults + HTTP-specific ones |
-| `protocols.rs` | `fetch_protocol_systems()` — paginated Tycho RPC call to discover available protocols; `resolve_protocols()` — higher-level wrapper used by `serve` and `scale` that expands `all_onchain`/`native_onchain` tokens and applies min-TVL filtering |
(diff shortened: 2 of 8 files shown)


tvinagre:
Two robustness fixes from production use on 2026-07-15 (commit e782f598):

**Client-side fetch timeout.** Against a degraded indexer (tycho-base-beta), the per-protocol-system component fetches hung indefinitely (18+ min byte-flat stalls). Each Tycho component fetch and the token-metadata fetch now run under a client-side deadline. New `--fetch-timeout-secs` flag (default 300s per fetch). On timeout the run fails fast with an error naming the protocol system that stalled and suggesting `--protocols` to exclude it, rather than hanging.

**`--min-tvl` flag.** The dedicated tycho-fynd-* indexers plan-gate component queries ("tvl_gt parameter is required on this plan"), so generate-requests previously could not query them at all. New optional `--min-tvl` passes `tvl_gt` to the component query (reusing the same `with_tvl_gt` mechanism the `scale` subcommand's `--min-tvl` builds on). The flag help notes that tycho-fynd-* endpoints require it.

Both new knobs are carried on a new `TokenPoolStatsParams` struct threaded into `fynd_rpc::protocols::fetch_token_pool_stats`.

Verification (nightly-2026-06-28): `cargo fmt --check`, `clippy --all --all-features --all-targets -D warnings`, and rustdoc all clean; `cargo nextest run -p fynd-rpc -p fynd-benchmark` = 126 passed. New tests cover the timeout error path (deadline reported, inner error surfaced, success passthrough) and CLI parsing of both flags.

PR #323, opened 2026-07-17
feat(encoding): sign exclusive-swap quotes into Ekubo user_data

louise-poole:
Jira task: https://propeller-heads.atlassian.net/browse/ENG-6134 

Add a signed_quote module that builds the EIP-712-signed payload for exclusive Ekubo legs and stamps it into the tycho Swap.user_data. The Encoder derives the per-swap fee from the leg's committed_amount_out and signs with a controller key.

TODO: 
- Rebase on [this](https://github.com/propeller-heads/fynd/pull/304) PR and remove local committed_amount_out stub on Swap.

### DIFF
--- a/fynd-core/src/encoding/encoder.rs
+++ b/fynd-core/src/encoding/encoder.rs
@@ -23,7 +23,7 @@ use tycho_simulation::tycho_common::{models::Chain, Bytes};
 use crate::{
     encoding::{
         router_fees::{FeeRates, RouterFees, SharedRouterFees},
-        signed_quote::SignedSwapSigner,
+        exclusive_swap::ExclusiveSwapSigner,
     },
     EncodingOptions, FeeBreakdown, OrderQuote, QuoteStatus, SolveError, Transaction,
 };
@@ -45,8 +45,8 @@ pub struct Encoder {
     chain: Chain,
     router_address: Option<Bytes>,
     router_fees: SharedRouterFees,
-    /// Signs exclusive Ekubo legs. `None` disables signing (no controller key configured).
-    signed_swap_signer: Option<SignedSwapSigner>,
+    /// Signs exclusive legs. `None` disables signing (no controller key configured).
+    exclusive_swap_signer: Option<ExclusiveSwapSigner>,
 }
 
 impl TryFrom<&OrderQuote> for Solution {
@@ -143,20 +143,20 @@ impl Encoder {
                     .build()
             })
             .transpose()?;
-        let signed_swap_signer = SignedSwapSigner::from_env(chain.id())?;
+        let exclusive_swap_signer = ExclusiveSwapSigner::from_env(chain.id())?;
         Ok(Self {
             tycho_encoder,
             chain,
             router_address,
             router_fees: SharedRouterFees::default(),
-            signed_swap_signer,
+            exclusive_swap_signer,
         })
     }
 
-    /// Overrides the signed-swap signer, replacing whatever was read from the environment.
+    /// Overrides the exclusive-swap signer, replacing whatever was read from the environment.
     #[must_use]
-    pub fn with_signed_swap_signer(mut self, signer: SignedSwapSigner) -> Self {
-        self.signed_swap_signer = Some(signer);
+    pub fn with_exclusive_swap_signer(mut self, signer: ExclusiveSwapSigner) -> Self {
+        self.exclusive_swap_signer = Some(signer);
         self
     }
 
@@ -211,15 +211,15 @@ impl Encoder {
 
             let solution = Solution::try_from(quote)?
                 .with_user_transfer_type(encoding_options.transfer_type().clone());
-            let solution = match &self.signed_swap_signer {
-                Some(signer) => Self::stamp_signed_swaps(solution, quote, signer)?,
+            let solution = match &self.exclusive_swap_signer {
+                Some(signer) => Self::stamp_exclusive_swaps(solution, quote, signer)?,
                 None => {
                     // Fail fast rather than emit on-chain-invalid unsigned calldata for an
                     // exclusive leg: an exclusive route requires a signature.
                     if has_exclusive_leg(quote) {
                         return Err(SolveError::FailedEncoding(
                             "quote routes through an exclusive pool but no signing key is \
-                             configured (set SIGNED_SWAP_CONTROLLER_KEY)"
+                             configured (set EXCLUSIVE_SWAP_CONTROLLER_KEY)"
                                 .to_string(),
                         ));
                     }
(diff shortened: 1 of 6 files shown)


fynd-core/src/encoding/encoder.rs
```diff
@@ -41,6 +44,8 @@ pub struct Encoder {
     chain: Chain,
     router_address: Bytes,
     router_fees: SharedRouterFees,
+    /// Signs exclusive Ekubo legs. `None` disables signing (no controller key configured).
+    signed_swap_signer: Option<SignedSwapSigner>,
```
tamaralipows: `signed_swap_signer` sounds a bit funny haha. Let's maybe stick to the `exclusive` terminology we decided with Ksenia's PR?

fynd-core/src/encoding/encoder.rs
```diff
@@ -131,9 +137,17 @@ impl Encoder {
             chain,
             router_address,
             router_fees: SharedRouterFees::default(),
+            signed_swap_signer,
         })
     }
 
+    /// Overrides the signed-swap signer, replacing whatever was read from the environment.
+    #[must_use]
```
tamaralipows, line 157: Oh I've never seen this before? `must_use`? This means it can't be read from env?

fynd-core/src/encoding/encoder.rs
```diff
@@ -208,6 +235,52 @@ impl Encoder {
         Ok(quotes)
     }
 
+    /// Stamps controller-signed `user_data` onto each exclusive leg of `solution`.
+    ///
+    /// A leg is exclusive when its fynd swap carries a committed amount. Solution swaps are built
+    /// 1:1 with the route's swaps, so they are matched by index. Returns `solution` unchanged when
+    /// no leg is exclusive.
+    fn stamp_signed_swaps(
```
tamaralipows: Same here with the terminology

fynd-core/src/encoding/encoder.rs
```diff
@@ -208,6 +235,52 @@ impl Encoder {
...
+        if !has_exclusive_leg(quote) {
+            return Ok(solution);
+        }
+
+        let route = quote.route().ok_or_else(|| {
+            SolveError::FailedEncoding("successful quote must have a route".to_string())
+        })?;
+        let fynd_swaps = route.swaps();
+
+        if fynd_swaps.len() != solution.swaps().len() {
```
tamaralipows: I don't remember what's the difference between quote and solution actually? Was a bit confused by this (didn't go deeper into the code to check). Could we add a comment explaining why we add this check actually?

fynd-core/src/encoding/encoder.rs
```diff
@@ -208,6 +235,52 @@ impl Encoder {
...
+                "route and solution swap counts diverged".to_string(),
+            ));
+        }
+
+        let swaps = solution
+            .swaps()
+            .iter()
+            .cloned()
+            .zip(fynd_swaps.iter())
+            .map(|(tycho_swap, fynd_swap)| {
```
tamaralipows: Also super confused by this, what's the difference?

fynd-core/src/encoding/signed_quote.rs
```diff
@@ -0,0 +1,535 @@
+//! Builds the controller-signed `user_data` payload for exclusive Ekubo swaps.
```
tamaralipows: Not Ekubo specific, let's keep this general

fynd-core/src/encoding/signed_quote.rs
```diff
@@ -0,0 +1,535 @@
...
+/// The extension takes `ceil(grossOutput · (fee << 32) / 2⁶⁴)` from the output, so the largest
+/// fee that never charges more than the surplus (`gross − committed`) is
+/// `floor((gross − committed) · 2³² / gross)`, clamped to `u32`. Rounding down biases toward
+/// under-capture, so the taker always receives at least `committed`.
+fn derive_fee_q32(gross: &BigUint, committed: &BigUint) -> u32 {
+    if gross <= committed {
+        return 0;
+    }
+    let surplus = gross - committed;
+    let scaled = (surplus * BigUint::from(1u64 << 32)) / gross;
```
tamaralipows, line 175: I commented this on Ksenia's PR but I understand this PR might be the better place to solve it - should we add a buffer here? So that the user doesn't just get the best non-exclusive price, but also even slightly better?

fynd-core/src/encoding/signed_quote.rs
```diff
@@ -0,0 +1,535 @@
...
+) -> Result<Address, SolveError> {
+    let bytes = attribute(component, "extension")?;
+    Address::try_from(bytes).map_err(|_| {
+        SolveError::FailedEncoding("extension attribute is not a 20-byte address".to_string())
+    })
+}
+
+/// Rebuilds the 32-byte `PoolConfig` word: `extension(20) ‖ fee(u64, 8) ‖ pool_type_config(4)`.
+fn pool_config_word(
+    component: &tycho_simulation::tycho_common::models::protocol::ProtocolComponent,
```
tamaralipows: Claude pet peeve: when the `ProtocolComponent` isn't just imported at the top. These long lines make the code harder for me to read :[ not sure your opinion

tamaralipows (commented):
Thank you! Just a few questions/comments

fynd-core/src/encoding/encoder.rs
```diff
@@ -208,6 +235,52 @@ impl Encoder {
...
+        let route = quote.route().ok_or_else(|| {
+            SolveError::FailedEncoding("successful quote must have a route".to_string())
+        })?;
+        let fynd_swaps = route.swaps();
+
+        if fynd_swaps.len() != solution.swaps().len() {
+            return Err(SolveError::FailedEncoding(
+                "route and solution swap counts diverged".to_string(),
+            ));
+        }
```
Troshchk: We check quote.route() twice here: once inside has_exclusive_leg (which does not error on empty route) and again on line 252 (which errors on empty route). I think the error on line 252 is actually unreachable -- if has_exclusive_leg returns true, the route is guaranteed Some
                                                                                                                                                        
I think we should check for the route first and fail if it's missing and then do the exclusive leg check, may be something like:                                                                                                                 
```                                                                                                                                                         
let route = quote.route().ok_or_else(|| {
        })?;                                                                                               
let fynd_swaps = route.swaps();
if !fynd_swaps.iter().any(|s| s.committed_amount_out().is_some()) {                                                                                 
     return Ok(solution);                                                                                                                            
}
```

fynd-core/src/encoding/signed_quote.rs
Troshchk: Why do you think it is needed?
Relevant comment in the other PR [here](https://github.com/propeller-heads/fynd/pull/304#discussion_r3596752705)

Troshchk (commented):
Thank you! 
Left a couple of qs

fynd-core/src/encoding/encoder.rs
louise-poole: It is a linter hint stating that _when_ this fn is used, the returned values must be used and cannot be silently discarded (it is not enforcing that this fn itself must be used).

fynd-core/src/encoding/encoder.rs
```diff
@@ -237,24 +237,33 @@ impl Encoder {
...
+            return Ok(solution);
+        }
 
-        if fynd_swaps.len() != solution.swaps().len() {
+        // `route_swaps` carry `committed_amount_out` and the pool attributes; `solution_swaps` are
+        // built 1:1 from them by `Solution::try_from` and are what the router executes. We read the
+        // committed amount from the route swap but stamp `user_data` onto the matching solution
+        // swap, so the two lists must stay index-aligned. A length divergence means that 1:1
+        // mapping broke upstream; bail rather than risk signing the wrong leg.
+        if route_swaps.len() != solution.swaps().len() {
```
Troshchk: But can it actually happen?
Also, should this check be part of the `Solution::try_from` in this case?

tamaralipows (commented):
Thank you so much! Looks great! One last thing, could you have a look at this comment? https://github.com/propeller-heads/fynd/pull/304#discussion_r3633953409

If we go that way we will have to implement in this PR. Else we are all good :)

fynd-core/src/encoding/encoder.rs
louise-poole: True - it can't happen. I don't think the check needs to be in `Solution::try_from` (it's trivial to do it there as that's where we build it 1:1 anyway). I'll remove this.

tamaralipows (approved):
Thanks!

PR #324, opened 2026-07-17
feat(hindsight): New chain, venue, and solver decoding

tamaralipows:
Extends the decoder (#313) to a new chain, more clients, and a client-attribution layer that identifies the order-flow owner when the entry point alone cannot.

Client addresses and fingerprints come from the [Top Client Swap Addresses — Hindsight Targets](https://propeller-heads.atlassian.net/wiki/spaces/DEFI/pages/3834904577/Top+Client+Swap+Addresses+-+Hindsight+Targets#Confirmed-clients) doc (sources cited there, example txs verified on-chain 2026-07-17).

## Base chain

- Base address book (`registry/base.toml`): wrapped native, stablecoins, batch settlers, solvers, and venues, with fee collectors verified on-chain.
- OP-stack receipts: fetch as `AnyTransactionReceipt` so the system deposit transaction (type `0x7e`) does not fail the whole block.
- `base_pools.toml` worker-pools config for running the monitor on Base.
- Monitor chain-head lag threshold defaults to ~20 minutes of the chain's block time (2s on Base), not a fixed block count.

## Monitor

- Ctrl-C shuts the run down cleanly: the signal is raced against every long await (initial build, per-block session, feed-death rebuild), the solver's workers and background tasks are torn down, and the process exits.

## Client attribution

A single step overrides the venue after decode when a client owns the order flow without being `tx.to`. It is registry-driven; the provider-specific pieces (reading LiFi's event, reading a CoW order's appData) sit behind the decoder/solver modules. Four fingerprints:

- **owning trader** (`[client_owners]`) — kpk's Safes settling through CoW, surfaced from the CoW `Trade` event's owner.
- **CoW appData tag** (`[client_appdata]`) — LlamaSwap, from the static appData hash (appCode "DefiLlama") its CoW orders commit, read from the settle calldata.
- **fee wallet** (`[client_fees]`) — Phantom, Robinhood; the fee is grossed back so the settled output stays comparable. Checked only inside a matched solver trade, so a bare dust transfer to a fee wallet is not mistaken for flow.
- **provider integrator tag** (`[client_integrators]`) — LiFi frontends (Infinex on ETH and Base; Robinhood's LiFi leg), read from the LiFi swap event.

This closes the earlier gap where a client routing through a shared router (0x, 1inch) looked like that router's own trade.

## Venues and solvers

- **Rabby** — own router, fee removed.
- **Coinbase Wallet** — own proxies plus fee wallet.
- **Rainbow** — own router; input-side fee read from calldata.
- **Fly** (Magpie) — Base solver.
- **0x quote** recovery from positive slippage.

## CoW settlements

Single-order CoW settlements are read from the GPv2 `Trade` event: exact executed amounts, the fee backed out of the input, and the owner taken directly from the event (so kpk attribution is exact). Runs only when CoW is the entry point — a CoW settlement reached through a venue (Relay) decodes as that venue's trade with CoW as the underlying solver. The Intent-role decoders (CoW, plus the generic net-flow finder it falls back to) live under a new `intents/` module, mirroring `venues/`. This covers the large majority of CoW flow: in a recent ~700-block on-chain sample, 97% of settlements were single-order (456 of 471); multi-order batches fall back to net-flow decoding.

## Limitations

- **CoW multi-order batches** fall back to net-flow decoding (one trade per transaction), so batch volume is under-counted, as before.
- **Rainbow** decodes only its ETH→token entry; the token→ETH and token→token entries encode a basis-point fee and are declined (a coverage gap, not a mispriced trade) until verified on-chain.
- **LlamaSwap** is caught only on its CoW leg (the appData hash). Its 0x-surplus recipient, tried first, turned out dead — no on-chain inflows for months — and its Kyber leg is tagged only off-chain, so those are not decodable here. LlamaSwap's CoW flow is very sparse (0 orders in a 26h/3,773-settlement on-chain sample), so it rarely appears — expected, not a miss.
- **Coinbase Wallet**'s proxies have been quiet since late 2025, so it mainly decodes historical flow.
- **Relay on Base** decodes ~65% of its flow (670 of 1,027 in the Base run); the rest deliver the swap output to a receiver other than the sender, so net-flow decoding declines — a coverage miss, not a mispriced trade.

## Notes

- kpk is a low-frequency client by design — the targets doc characterizes it as very large, infrequent tickets (single trades in the $10Ms), so it is expected not to appear in a short sample window. Its attribution is exercised by the CoW Trade decoder (owner read from the event), which fired throughout the run; a longer window would surface it.
- Rainbow's fee mechanic was not what its "fee held in-router" description implied — it is an input-side fee argument in the call, retained by the router (verified on tx `0xe09cf895…`). The decoder reads that argument rather than a fee transfer.
- CoW's on-chain fee is zero on modern orders (priced into the order), so the input-fee back-out is exact for current flow and best-effort for legacy non-zero fees.
- Fee-identified clients are only recognized inside an already-matched solver trade, which sidesteps the dust-spray problem (labeled fee wallets are heavily spammed) that a naive "transfers to the fee wallet" scan would hit.

## Sample run

Exercised across three monitor runs — **Ethereum** (1000 + 300 blocks, `--min-tvl 100`) and **Base** (1000 blocks, `--min-tvl 10`) — 8,587 re-solved trades combined. Everything listed is new in this PR (Base support included).

| Category | Combined across the three runs |
|---|---|
| **Client attributions** (wallets/frontends) | phantom 171 · rabby 101 · robinhood 97 · rainbow 1 — all Ethereum. kpk, LlamaSwap, Coinbase, Infinex are wired but too rare to appear in these windows |
| **Solver** | fly 32 — Base |
| **Decoders** | sender-netting 3927 · intent-netting 1991 · relay-netting 1435 · metamask-netting 640 · **cow-trade 492** · rabby-netting 101 · **rainbow-calldata 1** |
| **Verdicts** (top-of-block) | win 1877 · loss 1292 · unsolvable 5388 · coverage-miss 20 · sandwiched 10 |

New decoders this PR are **cow-trade** and **rainbow-calldata**; the `*-netting` decoders are the existing pipeline, shown for volume context. `cow` and `fly` are solvers (settlement/routing), not client attributions — CoW settlement decoding shows up as the `cow-trade` decoder.

**Why kpk, LlamaSwap, Coinbase, and Infinex have no live example here.** Their fingerprints are wired and unit-tested; they just had no matching trade in these windows — each is low-frequency or off its former flow:

- **kpk** — very large, infrequent tickets by design (single trades in the $10Ms, per the targets doc). Attributed from the CoW `Trade` owner, which fired throughout — just never for a kpk Safe.
- **LlamaSwap** — its only on-chain fingerprint is the CoW appData hash, and its CoW flow is very sparse (0 orders in a 26h / 3,773-settlement on-chain sample).
- **Coinbase Wallet** — its swap proxies have been quiet since late 2025, so it mainly decodes historical flow.
- **Infinex** — read from the LiFi integrator tag; LiFi decoded 54× on Base but no Infinex-tagged order landed in the ~33-min Base window.

**Wins sanity — no real-money artifacts.** On trades settling >$1k (245 wins) bps tops out at 244 and 221 fall under 50 bps. Extreme bps is a dust effect only: all 63 wins above 1000 bps are sub-$10 trades, and the two >100% ("impossible") wins are sub-$1 dust ($0.30 and $0.10 settled, ~$0.005 total). No fee-backed client (phantom / rabby / robinhood / metamask) appears among the top wins — the fee back-out holds. No structural artifacts: no zero-amount or same-token trades.

### DIFF
--- a/tools/hindsight/src/decoder/clients.rs
+++ b/tools/hindsight/src/decoder/clients.rs
@@ -1,16 +1,20 @@
 //! Order-flow client attribution.
 //!
 //! A trade's venue is normally the contract the trader entered through (`tx.to`). Some clients own
-//! the order flow without being that contract — kpk's Safes settle through `CoW`, so `tx.to` is
-//! the solver and the client is only visible as the order's owner. This module recognizes those
-//! clients from the decoded flow and overrides the venue label accordingly.
+//! the order flow without being that contract, and are recognized here from the decoded flow —
+//! overriding the venue label. Every fingerprint is registry-driven; nothing here knows about a
+//! specific client or provider.
 //!
-//! Two fingerprints today:
-//! - **owning trader** — the swap's net flow was read from a known client address (kpk's Safes).
-//! - **fee wallet** — the trade routed through a shared router (0x) but a known client fee wallet
-//!   received the output-token fee (Phantom, Robinhood). The fee is backed out so the settled
-//!   output stays gross. This is only checked inside an already-matched solver trade, so the
-//!   dust-spray gotcha (bare transfers to a fee wallet) does not apply.
+//! Three fingerprints, tried in order:
+//! - **owning trader** — the swap's net flow was read from a known client address
+//!   (`[client_owners]`).
+//! - **fee wallet** — a known client fee wallet received the output-token fee (`[client_fees]`);
+//!   the fee is backed out so the settled output stays gross. Checked only inside an
+//!   already-matched solver trade, so a bare dust transfer to a fee wallet is not mistaken for
+//!   flow.
+//! - **provider integrator tag** — a provider's event carried an integrator string mapped to a
+//!   client (`[client_integrators]`). The tag is extracted by the caller (it is provider-specific;
+//!   see `crate::decoder::solvers::integrator`), so this module stays generic.
 
 use std::collections::HashSet;
 
@@ -19,22 +23,27 @@ use alloy::primitives::{Address, U256};
 use crate::decoder::{decode::TraderFlow, registry::Registry, transfer_ledger::TransferLedger};
 
 /// The order-flow client for a decoded flow, when a fingerprint matches. On a fee-wallet match the
-/// client's fee is backed out of `flow` (added to the gross output), unless a venue decoder already
+/// client's fee is backed out of `flow` (added to the gross output) unless a venue decoder already
 /// accounted a fee.
 pub(crate) fn attribute(
     registry: &Registry,
     flow: &mut TraderFlow,
     ledger: &TransferLedger,
+    integrator: Option<&str>,
 ) -> Option<String> {
     if let Some(client) = registry.client_for_owner(flow.tracked) {
         return Some(client.to_string());
     }
-    let (client, fee) = fee_client(registry, ledger, flow.swap.token_out)?;
-    if flow.venue_fee_out.is_none() {
-        flow.venue_fee_out = Some(fee);
-        flow.swap.amount_out = flow.swap.amount_out.saturating_add(fee);
+    if let Some((client, fee)) = fee_client(registry, ledger, flow.swap.token_out) {
+        if flow.venue_fee_out.is_none() {
+            flow.venue_fee_out = Some(fee);
+            flow.swap.amount_out = flow.swap.amount_out.saturating_add(fee);
+        }
+        return Some(client);
     }
-    Some(client)
+    integrator
+        .and_then(|tag| registry.client_for_integrator(tag))
+        .map(str::to_string)
 }
 
 /// The client whose fee wallet received the output token in this trade, with the fee amount.
@@ -71,7 +80,7 @@ mod tests {
         let kpk_safe = address!("0x4f2083f5fbede34c2714affb3105539775f7fe64");
         let ledger = TransferLedger::from_transaction(&[], &[]);
         let mut flow = TraderFlow::without_fees(kpk_safe, swap(addr(10), 1, addr(11), 2));
-        assert_eq!(attribute(&registry, &mut flow, &ledger).as_deref(), Some("kpk"));
+        assert_eq!(attribute(&registry, &mut flow, &ledger, None).as_deref(), Some("kpk"));
     }
 
     #[test]
@@ -79,7 +88,7 @@ mod tests {
         let registry = Registry::ethereum();
         let ledger = TransferLedger::from_transaction(&[], &[]);
         let mut flow = TraderFlow::without_fees(addr(9), swap(addr(10), 1, addr(11), 2));
-        assert_eq!(attribute(&registry, &mut flow, &ledger), None);
+        assert_eq!(attribute(&registry, &mut flow, &ledger, None), None);
     }
 
     #[test]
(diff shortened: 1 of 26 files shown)


louise-poole (approved):
Blind approval so we can merge and get this in the cluster over the weekend. Will do a real review later today.

tools/hindsight/src/decoder/venues/rainbow.rs
```diff
@@ -0,0 +1,148 @@
...
+
+/// The input-side fee of a `fillQuoteEthToToken` call — its 4th (static `uint256`) argument — or
+/// `None` for any other selector or a too-short input.
+fn eth_to_token_fee(input: &[u8]) -> Option<U256> {
+    let (selector, args) = input.split_at_checked(4)?;
+    if selector != FILL_QUOTE_ETH_TO_TOKEN {
+        return None;
+    }
+    // Args are `buyToken, to, data (offset), feeAmount`; `feeAmount` is the 4th 32-byte word.
+    let word = args.get(3 * 32..4 * 32)?;
```
louise-poole, line 57: Why not use `sol!` here like you do in other decoders?

tools/hindsight/src/decoder/venues/rabby.rs
```diff
@@ -0,0 +1,189 @@
...
+        if flow.swap.token_out == Address::ZERO && flow.venue_fee_out.is_none() {
+            let wrapped_fee = ctx
+                .transfer_ledger
+                .received_by(&addresses.fee_collectors)
+                .get(&ctx.registry.wrapped_native())
+                .copied()
+                .filter(|fee| !fee.is_zero());
+            if let Some(fee) = wrapped_fee {
+                flow.venue_fee_out = Some(fee);
+                flow.swap.amount_out = flow.swap.amount_out.saturating_add(fee);
```
louise-poole, line 53: Small improvement idea: maybe add a struct fn to `TraderFlow` object that does this saturating add of the venue fee to the amount out within the struct. I think I see that this logic is used at least twice and it's nice to centralise something like that (maybe saturating add is buggy then we only need to fix one place).

tools/hindsight/src/resolve/monitor.rs
```diff
@@ -314,10 +317,59 @@ async fn build_solver(
         .map_err(|e| anyhow::anyhow!("failed to build solver: {e}"))
 }
 
+/// Default chain-head lag threshold in blocks — about 20 minutes at the chain's block time, so the
+/// wall-clock tolerance stays roughly the same across chains. Used when `--max-lag-blocks` is not
+/// given; an unrecognised chain falls back to the 12-second-block assumption.
```
louise-poole, line 322: I'm confused by this comment... is the the block time (2 sec for base and 12 for ethereum) or num blocks we can lag?

tools/hindsight/src/resolve/monitor.rs
```diff
@@ -314,10 +317,59 @@ async fn build_solver(
...
+fn default_lag_blocks(chain: &str) -> u64 {
```
louise-poole, line 323: Why not use the Chain enum as input here? Why convert to a str first? Also I think the Chain enum from tycho already has block times associated with it (could be mistaken though).

tools/hindsight/src/decoder/clients.rs
louise-poole, line 1: Should this not be called 'venues' according to our chosen terminology?

tools/hindsight/src/decoder/mod.rs
```diff
@@ -251,6 +262,21 @@ impl<P: Provider> Decoder<P> {
             return None;
         }
 
+        // A client fingerprint (owning trader, CoW appData tag, fee wallet, or integrator tag — see
+        // `clients`) overrides the entry-point label, backing any client fee out before the quote
+        // check reads the grossed output. The appData tag is read from a batch settler's calldata;
+        // other transactions carry none.
+        let integrator = solvers::integrator(logs);
+        let app_data = intents::client_tag(registry, entry_point, &root.input);
```
louise-poole, line 270: We should rename this to venue_tag

louise-poole (commented):
Nice! Just a couple minor things I noticed:

tools/hindsight/src/decoder/clients.rs
tamaralipows: 100%

tools/hindsight/src/resolve/monitor.rs
tamaralipows: This is the total lag - so for any chain we try not to be 20 mins behind. Will clarify

PR #325, opened 2026-07-18
feat: add hosted-API auth and per-chain routing to clients

tvinagre:
## What

Adds hosted-API authentication and per-chain request routing to both typed Fynd clients (Rust + TypeScript), so they work out-of-the-box against the hosted gateway at `fynd-api.propellerheads.xyz` instead of forcing users onto raw `curl`/`fetch`.

- **Rust** (`fynd-client`): new `FyndClientBuilder::with_api_key` + `with_chain`, backed by a `HostedConfig` struct. The API key is sent as `Authorization: Bearer ***`; the chain slug rewrites request paths to `/v1/{chain}/quote|health|info` and derives the EVM `chain_id` used for signing.
- **TypeScript** (`@kayibal/fynd-client`): new `apiKey`/`chain`/`headers` on `FyndClientOptions`. Chain routing is implemented as an `openapi-fetch` middleware that rewrites `/v1/` → `/v1/{chain}/`, so call sites stay typed against the generated schema literals. `createFyndClient` and the `Middleware` type are now re-exported from the package root (resolves the long-standing #12 export gap).

## Why

The hosted gateway requires a bearer token and per-chain paths (`/v1/{chain}/quote`). Neither client could produce those requests.

Bundled fix: `build_quote_only` hardcoded `chain_id: 1`, so a quote-only client on a non-mainnet chain would sign transactions stamped for mainnet whenever `SigningHints` supplied nonce, fees, and gas. The chain ID is now derived from the chain slug, and `build()` cross-checks the slug against the RPC-reported chain ID.

## Backward compatibility

Both options are opt-in. Unset, the clients reproduce the previous unauthenticated requests to the unprefixed `/v1/…` paths — self-hosted Fynd users see zero behavior change.

## Supported chains

`ethereum` (1), `base` (8453), `arbitrum` (42161), `bsc` (56), `polygon` (137), `unichain` (130). Unknown slugs fail at build time with a `Config` error listing the accepted set.

## Validation

| Gate | Result |
|---|---|
| `cargo check --package fynd-client` | pass |
| `cargo clippy --workspace --all-targets --all-features` | pass, 0 warnings |
| `cargo +nightly fmt --all --check` | pass |
| `cargo test -p fynd-client` | 93 lib + 9 integration + 5 new = all pass |
| TS `typecheck` / `lint` / `test` | pass / 0 warnings / 157 pass |

New tests were mutation-tested (suppressing the bearer header, middleware, and chain_id derivation failed 6 Rust + 5 TS tests; the 2 TS tests that stayed green assert *absence* of the new behavior, which is correct).


tvinagre:
Heads-up for reconciliation when this merges: PR #327 (benchmark gateway auth, branch `tl/benchmark-gateway-auth`, based on `tl/release-0.90.0-rc`) independently adds an overlapping client-auth implementation to `clients/rust/src/client.rs` — `FyndClientBuilder::with_api_key` plus `with_quote_path`/`with_health_path` path overrides and a non-retryable `FyndError::RateLimited` (HTTP 429) variant. Both PRs touch `client.rs` and will conflict; they should be reconciled into one auth API. Note the two currently disagree on header format: this PR sends `Authorization: Bearer <key>` while #327 sends the raw key. The raw form is what the deployed gateway expects — verified against the live prod gateway with an admin key (raw key → 200, `Bearer <key>` → 401), since `ph-nginx-auth`'s `auth.lua` matches the entire raw `Authorization` value against `user_tokens.token` with no Bearer stripping.

clients/rust/src/client.rs
```diff
@@ -358,6 +358,52 @@ mod erc20 {
...
+// ============================================================================
+
+/// Chain slugs accepted by the hosted Fynd gateway, paired with their EVM chain ID.
+const SUPPORTED_CHAINS: [(&str, u64); 6] = [
+    ("ethereum", 1),
+    ("base", 8453),
+    ("arbitrum", 42161),
+    ("bsc", 56),
+    ("polygon", 137),
+    ("unichain", 130),
```
tamaralipows, line 372: plasma? 👀

tamaralipows (approved):
Thanks for this - just one question/request

clients/rust/src/client.rs
brunoguerios: Does it really make sense to add Plasma here since we don't have a hosted Tycho for Plasma? 🤔

PR #320, opened 2026-07-17
feat(config): centralize solver defaults into a layered config file

zizou0x:
## What

All solver-tuning defaults now live in one file,
`fynd-core/src/config/default_config.toml`, embedded in the binary. It
deserializes directly into a complete `config::Config` (every field
required except the chain-specific `min_tvl`), so adding a config field
without updating the file fails the build's tests.

Configuration resolves field by field through three layers, highest
priority first:

1. CLI flags (only flags the user actually set)
2. Local config file: `--config-file` / `CONFIG_FILE`, default `./fynd.toml`
3. Embedded default config

## Library API

- `config::embedded_default()` returns the cached, complete default `Config`
- `Config::apply(&PartialConfig)` overlays a layer (exhaustive
  destructuring: a new `PartialConfig` field without an overlay arm is a
  compile error)
- `Config::validate()` range-checks the resolved result;
  `FyndBuilder::apply_config(&Config)` validates automatically
- `FyndBuilder::new` seeds its defaults from the embedded config

## Backward compatibility

- `worker_pools.toml` (file or `--worker-pools-config`/`WORKER_POOLS_CONFIG`)
  keeps working; its pools override the config file's so existing
  deployments keep their pools. A deprecation warning points to `fynd.toml`.
- The repo's `worker_pools.toml` is renamed to `fynd.toml` and extended
  into a full config example; docker-compose and the record-market /
  integration-test includes follow.
- Breaking for lib consumers: the tuning constants in
  `fynd_core::solver::defaults` (and their fynd-rpc re-exports) are
  removed; read `config::embedded_default()` instead.

## Behavior

Defaults are unchanged — the embedded values match the previous
constants, and `--help` now renders them from the config file. Startup
logs the full resolved config.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #328, opened 2026-07-20
feat(config): add remote config layer fetched from S3

zizou0x:
Stacked on #320 (config centralization).

## What

Adds a remote config layer to the resolution stack. Per-chain tuned values are
pulled from S3 at startup and applied right above the embedded defaults:

```
CLI flags > local config file (fynd.toml) > remote config (S3) > embedded default
```

Default URL (per chain, overridable via `--remote-config-url` / `REMOTE_CONFIG_URL`;
`--no-remote-config` disables):

```
https://s3.eu-central-1.amazonaws.com/repo.propellerheads-propellerheads/fynd/presets/{chain}/latest.toml
```

The payload is a `PartialConfig` in the same schema as `fynd.toml`. Empty payloads
are published for all six supported chains (verified end to end against the live
bucket, including a tuned-value round trip).

## Fail-safe behavior

The remote layer can never take a solver down:

- Fetch errors and timeouts log a warning and resolve without the remote layer;
  the caller bounds the fetch with a timeout (binary: 2 s)
- Bounded retries with backoff for transient failures only (connect errors, 5xx)
- Response size capped at 256 KiB before parsing
- Non-2xx statuses outside 4xx/5xx are rejected explicitly (S3 answers an
  unfollowable 301 on region mismatches)
- Unknown fields in the payload are ignored (forward compatibility); a payload
  whose pools reference an unknown algorithm is rejected wholesale so outdated
  binaries fall back instead of failing at solver build
- No panic paths reachable from network input
- The final resolved config is validated as a whole, regardless of source

## Library API

- `Config::apply_remote(url, timeout)` — async layer application; returns `self`
  unchanged on any fetch problem
- `config::get_default(chain)` — embedded + remote with default URL and built-in
  2 s budget, in one call
- `config::remote::fetch_remote_config(url)` — raw fetch for custom error handling

Covered by mock-server tests: valid payload, 5xx retry, 4xx no-retry, unfollowable
redirect, oversized body, garbage TOML, unknown algorithm, and fallback-on-timeout.


brunoguerios (approved):
Nice one!
Fits in nicely on top of the previous PR 👏 
LGTM ✅

PR #329, opened 2026-07-21
feat(config): introduce tuning presets (balanced, deep)

zizou0x:
Stacked on #328 (remote config layer).

## What

Adds the preset concept: a `config::Preset` (`balanced` | `deep`) selects which
*defaults* apply. Both the embedded `default_config.toml` and the per-chain remote
document hold one section per preset (same layout in both files); every layer above
— local config file, CLI flags, legacy worker_pools.toml — behaves unchanged on top.
The two presets currently share identical values; `deep` gets tuned later.

- Selection: `--preset` / `PRESET` (default `balanced`); lib: `get_default(chain, preset)`,
  `Config::apply_remote(url, preset, timeout)`. A preset can never be set from a config file.
- Remote URL stays one document per chain (`presets/{chain}/latest.toml`); the fetch
  selects the preset's section. A missing section (or empty document) is a quiet
  "nothing tuned" state — the already-published empty payloads remain valid.
- The unknown-algorithm safety check runs only on the selected section, so a future
  preset section can use newer algorithms without breaking binaries on other presets.
- `Preset` is strum-derived; parsing, listing, and `--help` values are generated, and a
  unit test fails if the enum and the file sections drift. Adding a preset = one
  variant + one section per file.
- `FyndBuilder::new` and fynd-rpc's `builtin_default()` pin to `balanced`.

## From the design review (fresh-context agent)

Applied: unknown keys in hand-written config files now warn (remote stays fully
permissive for forward compatibility, guarded by a field-list sync test); the remote
layer logs exactly which fields it overrides; the standalone router-timeout exception
is documented at the constant; layer-count doc fix; `--help` text cleanup.

Not applied: `clap::ValueEnum` for `Preset` (fynd-core has no clap dependency — strum
is the boundary-respecting choice); trimming `get_default` (deliberate convenience
API); changing legacy worker_pools.toml pools precedence (explicit
backward-compatibility decision, already warns); value-range floors in `validate()`
(no tuning bounds established yet — the new override logging provides visibility
instead).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3436,6 +3436,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_with",
+ "strum 0.27.2",
  "tempfile",
  "thiserror 1.0.69",
  "tokio",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -143,6 +143,9 @@ tycho-execution = ">=0.331.0"
 tycho-simulation = ">=0.331.0"
 typetag = "0.2"
 
+# Enum derives (Display, FromStr, variant listing)
+strum = { version = "0.27", features = ["derive"] }
+
 # Compression
 zstd = "0.13"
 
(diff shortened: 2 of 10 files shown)


brunoguerios (approved):
LGTM ✅

PR #330, opened 2026-07-21
feat(hindsight): measure slippage by re-executing top routes at back-of-block

kayibal:
## What

The hindsight monitor now measures each trade twice at back-of-block. The fresh re-solve at state N stays (the `back` comparison keeps its original meaning: what routing at the block's end state would deliver), and the top-of-block route is additionally re-executed against state N — same pools, split fractions, and input amount — to record the slippage between quote time and execution time, in bps and USD.

## Changes

- **`fynd_core::replay_route` (new public API)**: re-executes an already-built `Route` against a given `MarketState`. Swaps run in the route's topological order; positive `split` fractions apply to the token's collected balance with the `0.0` remainder convention; post-swap states are threaded so shared pools see depleted reserves. Pool simulations run through `get_amount_out_guarded`, so pool-math panics become per-trade `ReplayError`s instead of unwinding the caller. Pools and tokens resolve from the market state, not the route's embedded quote-time states.
- **Monitor**: the top solve keeps its `Route` in memory; `SteppingSolver` gains `reexecute`. After `advance()`, each trade is re-executed (for slippage) and solved fresh (for `back`). An unsolvable top records "no top-of-block route to re-execute" as the slippage side; a failed replay (e.g. pool gone at N) records the error.
- **Route attribution derived, not stored**: `SolvedAmount` carries `algorithm` plus the solved `Route` (boxed, serde-skipped); the readable path in JSONL/logs is rendered from the route at record time via a new `fynd_core::Route::token_symbol` accessor. The stored `RouteSummary` type and the monitor's token-symbols plumbing are removed; JSONL keys (`algorithm`, `route`) are unchanged.
- **JSONL**: new top-level `"slippage": {"bps", "usd"}` field, signed in both directions, USD valued at the back-of-block price snapshot. Null when nothing was replayed.
- **Prometheus**: `hindsight_slippage_bps` (signed histogram, venue/solver/chain/outcome labels — outcome is the trade's headline verdict), `hindsight_slippage_usd` (signed histogram, same labels), and `hindsight_positive_slippage_usd` (positive-only; its `_sum` is the running revenue if positive slippage were charged). Positive outliers ≥ $1000 get the same mispricing-inspection warn log as savings.

## Notes

- Slippage is computed from the raw outcomes before the coverage-miss reclassification, and survives independently of the `back` solve.
- The re-executed side reuses the top quote's gas deduction (same route ⇒ same gas) instead of re-deriving gas prices at N.
- Sandwiched trades are not excluded from slippage metrics: the comparison is Fynd-quote vs Fynd-re-execution, so the settled trade's MEV does not enter it.

## Testing

- Unit tests for `replay_route` (sequential, split fractions, shared-pool depletion, market-state divergence, missing state/token errors) and `Route::token_symbol`
- Tests for slippage math, verdict-labeled metrics rendering, JSONL fields, the re-execution flow, the back solve running without a top route, and route-path derivation (245 hindsight tests pass)
- `cargo +nightly clippy --all-targets --all-features` clean; fmt and doc build clean

### DIFF
--- a/fynd-core/src/replay.rs
+++ b/fynd-core/src/replay.rs
@@ -1,0 +1,340 @@
+//! Re-execute an already-built [`Route`] against a (possibly newer)
+//! [`MarketState`](crate::feed::market_data::MarketState).
+//!
+//! A route emitted by a solving algorithm pins pools, token order, and split fractions. Replaying
+//! it against a later block's pool states answers "what would this exact route have produced at
+//! that state" — the in-process equivalent of submitting the already-encoded transaction at that
+//! block. Used by tooling (e.g. `hindsight`) to measure slippage between quote time and
+//! execution time.
+
+use std::collections::HashMap;
+
+use num_bigint::BigUint;
+use tycho_simulation::tycho_common::{models::Address, simulation::protocol_sim::ProtocolSim};
+
+use crate::{
+    algorithm::split_primitives::split_amount,
+    feed::market_data::MarketState,
+    types::{ComponentId, Route, Swap},
+};
+
+/// The outcome of replaying a route: final output and summed per-swap gas.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct RouteReplay {
+    /// Amount of the route's output token produced.
+    pub amount_out: BigUint,
+    /// Sum of the per-swap gas estimates reported by the simulations.
+    pub gas: BigUint,
+}
+
+/// Why a route could not be replayed against a market state.
+#[derive(Debug, thiserror::Error)]
+pub enum ReplayError {
+    /// The route carries no swaps (only reachable via deserialization — [`Route::new`] rejects
+    /// an empty swap list).
+    #[error("route has no swaps")]
+    EmptyRoute,
+    /// The market state carries no simulation state for a pool in the route (removed by the feed
+    /// or filtered out since the route was built).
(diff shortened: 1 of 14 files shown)


fynd-core/src/replay.rs
```diff
@@ -0,0 +1,340 @@
...
+        };
+        *remaining -= &amount_in;
+
+        let sim = post_swap
+            .get(swap.component_id())
+            .map(|state| state.as_ref())
+            .or_else(|| market.get_simulation_state(swap.component_id()))
+            .ok_or_else(|| ReplayError::MissingState(swap.component_id().to_string()))?;
+        let result = sim
+            .get_amount_out(amount_in, token_in, token_out)
```
tamaralipows: I see everywhere else is using `get_amount_out_guarded` - should we not be using it here instead?

tamaralipows (approved):
Thank you!

fynd-core/src/replay.rs
kayibal: Good catch — fixed in 73e289ae: the replay now calls `get_amount_out_guarded`, so a pool-math panic becomes a `ReplayError::Simulation` for that trade instead of unwinding through the monitor. (Branch was also rebased onto main in the same push, which is why the other hashes moved.)

PR #333, opened 2026-07-21
feat(hindsight): add offline HTML report subcommand

tamaralipows:
Adds a `report` subcommand that turns a `monitor` run's comparison JSONL into a single
self-contained HTML file. It reads the data the monitor already writes — no chain, Tycho, or
network access.

```
hindsight report --comparisons-dir ./out [--venue relay] [-o report.html]
```
Screenshot 2026-07-28 at 11 14 19

Note: The unsolvable was because I set TVL too high in that particular run.

### DIFF
--- a/tools/hindsight/src/report/html.rs
+++ b/tools/hindsight/src/report/html.rs
@@ -1,0 +1,386 @@
+//! Render an aggregated [`Report`] to a single self-contained HTML file.
+//!
+//! No external assets or network: the verdict pie is a CSS `conic-gradient`, styling is one inline
+//! stylesheet, so the file opens offline. The panels mirror the Grafana dashboard's value views —
+//! the headline Fynd savings (`hindsight_savings_usd`), the verdict split, coverage by notional,
+//! per-solver/venue breakdowns, and the top-saving trades — and skip the block/latency health
+//! panels the JSONL does not carry.
+
+use std::fmt::Write as _;
+
+use crate::report::aggregate::{
+    Count, GroupStats, Report, Savings, Summary, TradeRow, VerdictStat,
+};
+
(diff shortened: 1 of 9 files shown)


tools/hindsight/src/report/html.rs
```diff
@@ -0,0 +1,386 @@
...
+    bps.map_or_else(|| "—".to_string(), |b| format!("{b:+.1}"))
+}
+
+/// A signed USD amount with a thousands-separated integer part, e.g. `-$1,234.56`.
+fn fmt_usd(value: f64) -> String {
+    let sign = if value < 0.0 { "-$" } else { "$" };
+    let rounded = format!("{:.2}", value.abs());
+    let (int_part, frac) = rounded
+        .split_once('.')
+        .unwrap_or((rounded.as_str(), "00"));
```
Troshchk: I think `split_once` cannot fail here, so just unwrap or expect? Coz value "00" is never reached

Troshchk (commented):
Thank you! 🙌 
I have left one comment and also I think it will be better if we use stacked barcharts instead of pie charts to be in line with the hindsight dashboard

tools/hindsight/src/report/html.rs
tamaralipows: Turns out it can — `format!("{:.2}", f64::INFINITY)` gives `"inf"`, no dot, so it fell through to `"00"` and printed `$inf.00`. An unwrap there would panic.

Added an `is_finite` check up front instead, so non-finite renders as `—` and the split can't fail.

Troshchk: I did not expect this, thanks for checking it!

Troshchk (approved):
Thank you!

PR #334, opened 2026-07-21
feat(rpc): serve a hosted-gateway Swagger UI at /docs/hosted/

tvinagre:
## What

Serves a **second** Swagger UI from the fynd backend, describing the hosted-gateway API (per-chain routing + bearer auth) alongside the existing self-hosted UI.

| | `/docs/` (self-hosted, unchanged) | `/docs/hosted/` (new) |
|---|---|---|
| Spec | `/api-docs/openapi.json` | `/api-docs/hosted/openapi.json` |
| Paths | `/v1/quote`, `/v1/health`, `/v1/info` | `/v1/{chain}/quote`, `/v1/{chain}/health`, `/v1/{chain}/info` |
| Auth | None | `ApiKeyAuth` security scheme (API key as raw `Authorization` header value, no `Bearer ` prefix) — Swagger UI shows "Authorize" button |
| Server URL | Relative (same origin) | `FYND_HOSTED_SWAGGER_URL` env var (default `https://fynd-api.propellerheads.xyz`) |

## Why

The hosted gateway at `fynd-api.propellerheads.xyz` routes per chain (`/v1/{chain}/quote`) and authenticates every request. But the Swagger UI it proxies shows the wrong paths (`/v1/quote`), has no "Authorize" button, and "Try it out" only produces 401s/404s. Users have no usable interactive docs for the hosted API.

The hosted spec is built **programmatically** from the self-hosted one (`api/docs.rs`) — no duplicate `#[derive(OpenApi)]` annotations, so endpoint docs stay in one place and can't drift. Every `/v1/<endpoint>` path is rewritten to `/v1/{chain}/<endpoint>` with a required `chain` path parameter, a bearer security scheme is declared and required globally, and `servers[0].url` is set from the env var.

## Backward compatibility

`/docs/` and `/api-docs/openapi.json` are **unchanged** — self-hosted deployments see zero behavior change. The hosted UI is registered first (actix matches resources in registration order; `/docs/{_:.*}` would otherwise swallow `/docs/hosted/...`) — a test asserts both UIs stay reachable.

OpenAPI drift check: **no drift** — `ApiDoc` is untouched, so `clients/openapi.json` stays in sync.

## Validation

| Gate | Result |
|---|---|
| `cargo +nightly fmt --all --check` | pass |
| `cargo +nightly clippy --locked --all --all-features --all-targets -- -D warnings` | pass, 0 warnings |
| `cargo doc` (RUSTDOCFLAGS=-D warnings) | pass |
| `cargo test -p fynd-rpc --all-features --lib` | 10 passed (api::docs), 0 failed |
| `cargo nextest run --workspace` (non-integration) | 1031 pass, 0 fail |
| OpenAPI drift check (`cargo run -- openapi` vs `clients/openapi.json`) | NO DRIFT |

New tests cover: self-hosted paths unchanged, no security on self-hosted, hosted paths carry `{chain}`, experimental paths included, chain parameter on every operation, bearer auth scheme, server URL, default URL, both UIs reachable (registration-order regression test), and both spec endpoints serve the right paths.

## Follow-up (separate PR, ph-nginx-auth)

Follow-up PRs are open: [ph-nginx-auth#34](https://github.com/propeller-heads/ph-nginx-auth/pull/34) routes the gateway's `/docs/` → backend's `/docs/hosted/` and `/api-docs/openapi.json` → `/api-docs/hosted/openapi.json` — unauthenticated, not `access_by_lua_file` as originally planned: Swagger UI fetches assets and specs without an `Authorization` header, and `/api-docs/openapi.json` is the documented unauthenticated version-discovery endpoint. [helm-configuration#805](https://github.com/propeller-heads/helm-configuration/pull/805) (draft) rolls out the v14 gateway image and sets `FYND_HOSTED_SWAGGER_URL` on the staging backends. The self-hosted `/docs/` on the backend stays available for private deployments.

### DIFF
--- a/fynd-rpc/src/api/docs.rs
+++ b/fynd-rpc/src/api/docs.rs
@@ -1,0 +1,309 @@
+//! OpenAPI spec construction and Swagger UI registration.
+//!
+//! Two specs are served from the same binary:
+//!
+//! - `/docs/` + `/api-docs/openapi.json` — self-hosted deployments, describing the paths this
+//!   process routes (`/v1/quote`) on the origin it is reached at, with no authentication.
+//! - `/docs/hosted/` + `/api-docs/hosted/openapi.json` — the hosted gateway, which routes per chain
+//!   (`/v1/{chain}/quote`) and authenticates requests with a bearer token.
+
+use std::env;
+
+use actix_web::web;
+use serde_json::json;
+use utoipa::{
+    openapi::{
+        path::{Operation, Parameter, ParameterBuilder, ParameterIn},
+        security::{Http, HttpAuthScheme, SecurityScheme},
+        Components, ObjectBuilder, OpenApi, PathItem, Required, SecurityRequirement, Server, Type,
+    },
+    OpenApi as _,
+};
+use utoipa_swagger_ui::SwaggerUi;
+
+use crate::api::ApiDoc;
+#[cfg(feature = "experimental")]
+use crate::api::ExperimentalApiDoc;
+
+/// Environment variable overriding the server URL advertised by the hosted spec.
+const HOSTED_SERVER_URL_ENV: &str = "FYND_HOSTED_SWAGGER_URL";
+
+/// Server URL advertised by the hosted spec when [`HOSTED_SERVER_URL_ENV`] is unset.
+const DEFAULT_HOSTED_SERVER_URL: &str = "https://fynd-api.propellerheads.xyz";
+
+/// Name of the bearer security scheme declared by the hosted spec.
+const BEARER_SCHEME_NAME: &str = "BearerAuth";
+
+/// Registers both Swagger UIs and the specs they load.
+///
+/// The hosted UI is registered first because actix matches resources in registration order:
+/// `/docs/{_:.*}` also matches `/docs/hosted/index.html` and would serve a 404 for it.
+pub(crate) fn configure_docs(cfg: &mut web::ServiceConfig) {
+    let hosted = SwaggerUi::new("/docs/hosted/{_:.*}")
+        .url("/api-docs/hosted/openapi.json", hosted_spec(&hosted_server_url()));
+    let self_hosted =
+        SwaggerUi::new("/docs/{_:.*}").url("/api-docs/openapi.json", self_hosted_spec());
+    cfg.service(hosted).service(self_hosted);
+}
+
+/// Builds the spec describing the endpoints this process routes.
+fn self_hosted_spec() -> OpenApi {
+    #[allow(unused_mut)]
+    let mut openapi = ApiDoc::openapi();
+    #[cfg(feature = "experimental")]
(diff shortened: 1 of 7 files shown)


fynd-rpc/src/api/docs.rs
```diff
@@ -0,0 +1,320 @@
...
+
+use crate::api::ApiDoc;
+#[cfg(feature = "experimental")]
+use crate::api::ExperimentalApiDoc;
+
+/// Environment variable overriding the server URL advertised by the hosted spec.
+const HOSTED_SERVER_URL_ENV: &str = "FYND_HOSTED_SWAGGER_URL";
+
+/// Server URL advertised by the hosted spec when [`HOSTED_SERVER_URL_ENV`] is unset.
+const DEFAULT_HOSTED_SERVER_URL: &str = "https://fynd-api.propellerheads.xyz";
```
louise-poole: I'm not sure we want our hosted service specifics leaked here like this? 👀 Do we need a default? Can't we make the env var required? Ooooor just default to not registering a hosted endpoint if a url for it is not set?

fynd-rpc/src/api/docs.rs
```diff
@@ -0,0 +1,320 @@
...
+    OpenApi as _,
+};
+use utoipa_swagger_ui::SwaggerUi;
+
+use crate::api::ApiDoc;
+#[cfg(feature = "experimental")]
+use crate::api::ExperimentalApiDoc;
+
+/// Environment variable overriding the server URL advertised by the hosted spec.
+const HOSTED_SERVER_URL_ENV: &str = "FYND_HOSTED_SWAGGER_URL";
```
louise-poole: Why use an env var only? All other server configs are passed with `ServeArgs` defined on the cli and use clap's `env` attribute to allow that value to be set and an env var too if preferred. Would be nice if we remain consistent with the existing args.

louise-poole (commented):
Some small comments on the config set up, otherwise looks good to me!

fynd-rpc/src/api/docs.rs
tvinagre: Good catch — the default is gone. The hosted UI is now opt-in: only registered when a URL is provided via `--hosted-swagger-url` / `FYND_HOSTED_SWAGGER_URL`. Self-hosted deployments serve just `/docs/` and never see the gateway URL. See 5e1e865a.

fynd-rpc/src/api/docs.rs
tvinagre: Agreed — added `--hosted-swagger-url` to `ServeArgs` with clap's `env = "FYND_HOSTED_SWAGGER_URL"` attribute, consistent with the other server settings. It threads through `FyndRPCBuilder` → `configure_app` → `configure_docs` as an `Option<String>` instead of being read from the env inside `docs.rs`. See 5e1e865a.

louise-poole (approved):

PR #335, opened 2026-07-21
perf(rpc): defer quote failure-logging off the response path

brunoguerios:
## What

Moves the `/v1/quote` failure-capture logging off the HTTP response path.

- `ReplayRequest` is now an owned type with two methods: `capture()` extracts the routing-essential allowlist from the request (address clones, no JSON) and runs on the request path; `to_json()` performs the `serde_json` serialization.
- The handler builds the cheap `capture()` before the DTO is consumed, then on a failed quote spawns a detached `actix_web::rt::spawn` task that serializes and emits the log line, carrying the current tracing span so the line keeps its request context.
- `replay_json` is now a test-only wrapper over `capture().to_json()`.

Previously the serialization ran synchronously on every request. It now runs only for failed quotes and off the response path, so it does not add to response latency. The replay JSON output is unchanged, and the failure-only logging invariant is preserved.

This does not affect the reported `solve_time_ms` (measured inside the router). It positions a later follow-up to enrich the log with data such as request volume, which would run in the same deferred task.

## Testing

- `cargo nextest run -p fynd-rpc --all-features` — 56 passed
- `cargo check -p fynd-rpc --all-features` — clean
- `cargo clippy -p fynd-rpc --all-features --all-targets -- -D warnings` — clean
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --package fynd-rpc` — clean

### DIFF
--- a/fynd-rpc/src/api/handlers.rs
+++ b/fynd-rpc/src/api/handlers.rs
@@ -103,29 +103,41 @@ pub(crate) async fn quote(
         },
         Err(error) => RequestOutcome::Failed { code: solve_error_code(error) },
     };
-    // Only failed quotes are logged (successful quotes are the common case and
-    // would dominate log volume); see RequestOutcome::is_failure. Emitting is
-    // deferred to a detached task so serialization never adds latency to the
-    // response; the current tracing span is carried over so the line keeps its
-    // request context.
-    if outcome.is_failure() {
+    // Only failed quotes get a capture line (successful quotes are the common
+    // case and would dominate log volume; see RequestOutcome::is_failure), and
+    // successful solves above the slow threshold get a slow_solve line. Both
+    // are emitted from a detached task so serialization never adds latency to
+    // the response; the current tracing span is carried over so the lines keep
+    // their request context.
+    let slow_solve_time_ms = match &outcome {
+        RequestOutcome::Solved { solve_time_ms, .. }
+            if *solve_time_ms > request_capture::SLOW_SOLVE_THRESHOLD_MS =>
+        {
+            Some(*solve_time_ms)
+        }
+        RequestOutcome::Solved { .. } | RequestOutcome::Failed { .. } => None,
+    };
+    if outcome.is_failure() || slow_solve_time_ms.is_some() {
         let span = tracing::Span::current();
         actix_web::rt::spawn(async move {
-            span.in_scope(|| log_request_capture(num_orders, &replay_capture.to_json(), &outcome));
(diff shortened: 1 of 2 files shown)


fynd-rpc/src/api/handlers.rs
```diff
@@ -66,9 +66,10 @@ pub(crate) async fn quote(
     }
 
     // Capture a re-issuable, signature-free copy of the request BEFORE the core
-    // conversion consumes `dto_request`.
+    // conversion consumes `dto_request`. This is cheap (no serialization); the
+    // JSON encoding is deferred to the failure-only task below.
     let num_orders = dto_request.orders().len();
-    let replay_json = request_capture::replay_json(&dto_request);
+    let replay_capture = request_capture::ReplayRequest::capture(&dto_request);
```
tvinagre, line 73: Can we move this to inside 
`     if outcome.is_failure() {` ? Otherwise we're serializing for nothing

brunoguerios: This line isn't actually serializing anything. It's only capturing a copy before `dto_request` is consumed. Serialization happens only within `outcome.is_failure` by doing `&replay_capture.to_json()` in the detached task.
Let me know if I'm missing something.

fynd-rpc/src/api/handlers.rs
```diff
@@ -102,9 +103,16 @@ pub(crate) async fn quote(
...
-    // would dominate log volume); see RequestOutcome::is_failure.
+    // would dominate log volume); see RequestOutcome::is_failure. Emitting is
+    // deferred to a detached task so serialization never adds latency to the
+    // response; the current tracing span is carried over so the line keeps its
+    // request context.
     if outcome.is_failure() {
-        log_request_capture(num_orders, &replay_json, &outcome);
+        let span = tracing::Span::current();
+        actix_web::rt::spawn(async move {
+            let _guard = span.enter();
```
tvinagre: **Span guard held inside an async block**

This is safe today because the block has no `.await`, but an entered span guard held across a yield point is the classic tracing footgun (it corrupts span tracking on that worker thread). Since the PR description positions this task for future enrichment, that's exactly when an `.await` would land here. The body is fully synchronous, so this is immune to that future edit at no cost:

```rust
actix_web::rt::spawn(async move {
    span.in_scope(|| log_request_capture(num_orders, &replay_capture.to_json(), &outcome));
});
```

(or `.instrument(span)` on the async block)

tvinagre (approved):

PR #336, opened 2026-07-22
fix: correct hosted-api onboarding docs and improve serve-path observability

zizou0x:
## What

Fixes findings from two AI-generated reviews shared in Slack (PR #326 docs review + a serve-path observability report), verified against the code. This PR takes the **docs fixes** and the **safe, non-behavioral observability fixes**. The behavioral observability items (health-gating on derived freshness, `MarketDataStale` emission, `InsufficientLiquidity` surfacing) are intentionally left out — they need design sign-off and are tracked separately.

### Docs — `docs/get-started/hosted-api.md`
- **Broken quickstart links (×3):** `../quickstart/README.md` → `quickstart/README.md`. The page is at `docs/get-started/`, so `../quickstart/` 404'd.
- **Rust client version pin:** `fynd-client = "0.92"` → `"0.97"`. `with_api_key`/`with_chain` landed in 0.96+ (0.95 was never published to crates.io), so the pinned example didn't compile.
- **Typed-client slippage types:** TS `slippage: '0.005'` → `0.005` (typed field is `number`); Rust `EncodingOptions::new("0.005")` → `new(0.005)` (`f64`). The raw REST/curl body keeps the string form, which is correct and unchanged.
- **Nonexistent `/v1/ws`:** removed the "streaming quotes over WebSocket" claim — there is no client-facing WebSocket handler (self-hosted or otherwise). *If WS streaming is on the roadmap, re-add with accurate wording.*
- **Undefined `transition` tier:** removed from the pricing prose; the table row was already deleted, leaving prose referencing a missing row.
- **Telegram handle:** displayed `@FyndPortalBot` → `@fynd_portal_bot` to match the href (usernames are underscore-sensitive).
- **Undocumented fields:** documented `total_gas_estimate`, `amount_in`, and the `/info` `chain_id` and `version` fields. *(The `version` example value is illustrative — worth a sanity check against the live `/info` response.)*

### Observability — safe / non-behavioral
- **`TychoFeed::run` docstring** claimed "runs indefinitely, reconnecting on failures" — it doesn't; the feed is fail-stop (transport reconnection lives in tycho-client). Corrected, and noted that `reconnect_delay` is currently unused by the feed.
- **`HealthStatus.num_solver_pools`** docstring said "active" — clarified it's the *configured* count, not live worker threads (doesn't decrease when workers stop).
- **`--metrics-host` flag** added so the Prometheus exporter can bind loopback-only. Defaults to `0.0.0.0`, preserving current behavior.

## Not included (tracked for follow-up)
- Health `/v1/health` cannot detect a post-warmup derived-pipeline stall (returns 200 while quotes are dead) — the incident-class gap. Fix: surface `derived_last_success_age_ms` and gate `healthy` on it.
- Worker panics are unrecoverable/invisible (no `catch_unwind`, no respawn, no live-worker gauge).
- `SolveError::MarketDataStale` has no production call site — quotes never fail on staleness.
- `QuoteStatus::InsufficientLiquidity` unreachable; 0.97's `NoPathReason` work helps but has no IL variant and only reaches the capture log, not the response.
- Consuming Tycho `sync_states` (Delayed/Stale) instead of discarding them.

## Test plan
- [ ] `cargo check` / CI green (I could not run it in my environment — missing OpenSSL dev headers — so compilation is verified by inspection only; CI is the gate).
- [ ] `--metrics-host 127.0.0.1` binds loopback only; default still binds `0.0.0.0`.
- [ ] Rendered docs page: links resolve, code samples compile against `fynd-client 0.97`.


docs/get-started/hosted-api.md
```diff
@@ -102,12 +102,15 @@ curl -H "Authorization: $FYND_API_KEY" \
...
-  "permit2_address": "0x000000000022d473030f116ddee9f6b43ac78ba3"
+  "permit2_address": "0x000000000022d473030f116ddee9f6b43ac78ba3",
+  "version": "0.97.0"
 }
 ```
 
+* `chain_id` — the EVM chain ID this backend serves (e.g. `1` for Ethereum). Corresponds to the `chain` path segment in the request URL.
 * `router_address` — the on-chain contract that executes swaps. Submit your encoded swap transaction to this address (it's already set as `transaction.to` in an encoded quote — you don't set it yourself). It is **chain-specific**: each chain returns its own router address.
 * `permit2_address` — the [Permit2](https://uniswap.org/blog/permit2) contract address. Used for permit-based token approvals (see [Approvals](#approvals) below).
+* `version` — the Fynd binary version serving this chain (the example value is illustrative). The hosted backends track the latest release.
```
zizou0x: Hosted won't necessarily be the latest release, it will depend on when we do a prod release.

docs/get-started/hosted-api.md
```diff
@@ -371,7 +376,7 @@ Full example: [`clients/typescript/examples/tutorial/main.ts`](https://github.co
 ```toml
 # Cargo.toml
 [dependencies]
-fynd-client = "0.92"
+fynd-client = "0.97"
```
zizou0x, line 380: Would there be a way to not commit to any version somehow? Or add a comment to look for latest

fynd-core/src/feed/mod.rs
```diff
@@ -39,6 +39,10 @@ pub(crate) struct TychoFeedConfig {
     pub(crate) tvl_buffer_ratio: f64,
     /// Reconnect delay on connection failure.
     /// Default is 5 seconds.
+    ///
+    /// NOTE: currently unused by `TychoFeed` — the feed is fail-stop and relies on
+    /// tycho-client's internal transport reconnection rather than reconnecting itself.
+    /// Retained for config compatibility; see the tracking issue before wiring or removing.
```
zizou0x: Can we remove it then?

tvinagre (approved):

PR #337, opened 2026-07-22
feat: export per-protocol pool count and sync status metrics

zizou0x:
## What

- `MarketState` now maintains a per-protocol pool count, updated incrementally on component upsert/remove — no scan of the component map on read.
- New `MetricsSampler` background task samples the market state every 10s (`defaults::METRICS_SAMPLE_INTERVAL`) with a non-blocking read: if a block update holds the write lock, the tick is skipped. Exports:
  - `market_pools_per_protocol{protocol}` — pool count per protocol. Protocols with a sync status but no pools export 0 rather than no series.
  - `market_protocol_sync_state{protocol}` — numeric state code using the same mapping as the `tycho_integration_protocol_sync_state` dashboards: 1=started, 2=ready, 3=delayed, 4=stale, 5=advanced, 6=ended.
  - `market_protocol_block{protocol}` — block number from the protocol's synchronizer state; the gap to `market_current_block` gives per-protocol lag.
- The sampler task handle follows the router-fee handle pattern: stored on `Solver`/`SolverParts`/`FyndRPC`, aborted on shutdown. `SolverParts::into_components` returns one more `JoinHandle`.
- Grafana: adds a "Pools per Protocol" timeseries panel and a "Protocol Sync Status" status-history panel with the value mappings above.

Sampling on an interval instead of emitting per block keeps this work off the feed's write-lock path, and the sync-status gauges keep updating when the feed itself stalls. Prometheus scrapes at a coarser interval than blocks arrive, so no resolution is lost.

## Testing

- `pool_counts_by_protocol_tracks_upserts_and_removals` — incremental count maintenance, including re-upsert not double-counting and removal resetting to zero.
- `emit_market_metrics_records_pool_and_sync_gauges` — gauge emission verified with `DebuggingRecorder`.
- fmt, clippy, and doc build clean; 1023 workspace tests pass. The 8 fynd-core integration tests fail locally only because the Git LFS market-recording fixture is unavailable in my environment.


monitoring/grafana/dashboards/fynd.json
```diff
@@ -220,6 +220,75 @@
         "defaults": {},
         "overrides": []
       }
+    },
```
tvinagre, line 223: The hosted dashboards are on our terraform repo - can you add it there? This one is only for the self-hosting docker-compose

tvinagre (approved):

monitoring/grafana/dashboards/fynd.json
zizou0x: Yes, I will add it when this is deployed on hosted staging!

PR #338, opened 2026-07-22
fix(rpc): fail fast when the computation manager task exits

zizou0x:
## Problem

`FyndRPC::run()` supervises the HTTP server, Tycho feed, gas worker, and the derived-data **computation manager** with a single `tokio::select!` — with **no surrounding loop**. The feed and gas arms treat their task ending as fatal (stop the server gracefully, tear everything down). The computation-manager arm was the odd one out:

```rust
_ = &mut computation_manager_handle => {
    warn!("Computation manager stopped unexpectedly");
    // Continue running - derived data won't be updated but solver can still work
}
```

Because there is no loop, **"continue running" never happens.** The arm logs, then falls through to the shared shutdown tail (aborts the router-fee worker, shuts down all worker pools) and returns `Ok(())`.

Net effect when the derived pipeline task ends — event-channel close, a shutdown race with the feed arm, or a panic:

- the whole solver shuts down and the **process exits `0`** (looks like a clean, intentional stop, not a failure);
- the HTTP server is **not** gracefully stopped in this arm (unlike feed/gas), so in-flight requests can be cut mid-teardown;
- whether the instance comes back depends entirely on the orchestrator's restart policy — an `on-failure`-style policy treats exit-`0` as intentional and **leaves it down**.

The comment is also misleading: the manager is never respawned, so "the solver can still work" would mean serving quotes on **frozen derived data forever**.

## Change

Make the computation-manager arm behave like the feed/gas arms: log an **error**, stop the HTTP server gracefully, abort siblings, and set a `fatal_error` so `run()` returns `Err`. `run_solver` → `main` then propagate a **non-zero exit**, and the orchestrator restarts the instance (crash-only recovery).

```rust
_ = &mut computation_manager_handle => {
    error!("Computation manager stopped unexpectedly, shutting down solver");
    server_handle.stop(true).await;
    server_task.await.ok();
    feed_handle.abort();
    gas_price_worker_handle.abort();
    fatal_error = Some(std::io::Error::other("computation manager stopped unexpectedly"));
}
```

## Why fail-fast instead of actually continuing

Nothing respawns the computation manager, so "continuing" would serve quotes on stale/frozen derived data indefinitely with no path to recovery. A restart is the only real recovery, so we make it a loud, non-zero crash.

## Scope / non-goals

- **Only** the computation-manager arm changes; the server/feed/gas arms are untouched. (Note: those arms currently also return exit-`0` after logging an error — arguably worth revisiting, but out of scope here.)
- Does **not** cover the *manager-alive-but-computations-failing* stall (the task never exits, so this arm never fires). That needs a `/v1/health` derived-freshness gate — tracked separately.

## Testing

`FyndRPC::run()` owns live background tasks and has no existing unit coverage; this change is verified by inspection + CI (compile/clippy). Behavioral validation is best done in staging: kill/panic the computation manager and confirm the process now exits non-zero and is restarted (previously it exited `0`).


zizou0x:
Quick note: if possible we should have a retry logic directly in the code instead of relying on k8s restart.

tvinagre (approved):

PR #340, opened 2026-07-22
feat: distinguish amount_too_small from no_graph_path in route failures

brunoguerios:
## What

Adds `NoPathReason::AmountTooSmall` so dust quotes stop being reported as `no_graph_path`.

Bellman-Ford reported every zero routed-output as `NoGraphPath`, conflating "no route exists" with "the amount is too small to route." During SPFA, each successful hop simulation now checks whether the amount entering the hop covers that hop's own gas; if not, the failure is labeled `amount_too_small` instead. Input-side and single-hop, so healthy and too-large orders never trip it (too-large fails via simulation error). PathFrankWolfe inherits the behavior; the reason flows through aggregation into the `quote_failure` log.

Motivation: steady ETH↔USDC `no_graph_path` failures on Base turned out to be dust-sized requests (confirmed by replay), indistinguishable in the logs from real routing gaps.

Scope: Bellman-Ford + PathFrankWolfe. MostLiquid already reports dust as `insufficient_liquidity` and is unchanged.

## Compatibility

Reason label only: order status stays `NoRouteFound`, wire `QuoteStatus` unchanged, no OpenAPI drift.

## Performance

The check is gated on the already-computed `net_candidate <= 0`, keeping it off the hot path. A/B replay benchmark (recorded market, interleaved runs): ungated cost was ~4% per quote; gated is within measurement noise of main. Details in the PR comment.

## Caveats

- Gas-aware `amount_too_small` requires `token_in` to be priced; unpriced dust falls back to `no_graph_path` (under-report, never a mislabel).
- The flag is per-solve: a no-path solve that also contains a dust hop on some branch can be labeled `amount_too_small`. Rare, label-only.

## Testing

608 tests pass (clippy / fmt / doc / OpenAPI drift clean). Key cases: direct + mid-route dust → `amount_too_small`; too-large revert, low-rate pool with healthy input, connector-excluded, unreachable → `no_graph_path`; gas-aware tiny input → `amount_too_small`.

### DIFF
--- a/fynd-core/src/algorithm/bellman_ford.rs
+++ b/fynd-core/src/algorithm/bellman_ford.rs
@@ -462,20 +462,28 @@ impl BellmanFordAlgorithm {
                         // (input stays large) nor a too-large amount (huge input); those fail via
                         // other paths. Catches dust at token_in and amounts that decay below a
                         // hop's gas mid-route.
-                        let u_price = Self::resolve_token_price(
-                            ctx.node_address.get(&u),
-                            ctx.token_prices.as_ref(),
-                            spot_product[u_idx],
-                            ctx.node_address.get(&ctx.token_in_node),
-                        );
-                        if Self::gas_adjusted_amount(
-                            &amount[u_idx],
-                            &result.gas,
-                            ctx.gas_price_wei.as_ref().unwrap(),
-                            u_price.as_ref(),
-                        ) <= BigInt::ZERO
-                        {
-                            input_below_hop_gas = true;
+                        //
+                        // Probe the input-side dust signal only on uneconomic edges: an input
+                        // below this hop's gas implies net_candidate <= 0 (output value <=
+                        // input value < hop gas <= cumulative gas), so economic relaxing edges
+                        // can never trip it — gating keeps the extra price lookup and BigInt
+                        // math off the hot path (measured ~3-4% of quote time when unguarded).
+                        if !input_below_hop_gas && net_candidate <= BigInt::ZERO {
+                            let u_price = Self::resolve_token_price(
+                                ctx.node_address.get(&u),
+                                ctx.token_prices.as_ref(),
+                                spot_product[u_idx],
+                                ctx.node_address.get(&ctx.token_in_node),
+                            );
+                            if Self::gas_adjusted_amount(
+                                &amount[u_idx],
+                                &result.gas,
+                                ctx.gas_price_wei.as_ref().unwrap(),
+                                u_price.as_ref(),
+                            ) <= BigInt::ZERO
+                            {
+                                input_below_hop_gas = true;
+                            }
                         }
                         let net_existing = Self::gas_adjusted_amount(
                             &amount[v_idx],
--- a/fynd-core/src/algorithm/bellman_ford.rs
+++ b/fynd-core/src/algorithm/bellman_ford.rs
@@ -100,11 +100,8 @@ struct SPFAResult {
     edge_gas: Vec<BigUint>,
     /// Cumulative spot-price product from token_in to each node (for gas fallback).
     spot_product: Vec<f64>,
-    /// True if some hop's own input amount couldn't cover that single hop's gas cost
-    /// (gas-aware) or a literal zero output was produced (gas-unaware). Input-side and
-    /// single-hop, so it never trips for a healthy or too-large order (input stays
-    /// large); it catches dust at token_in and amounts that decay below a hop's gas
-    /// mid-route. Distinguishes a too-small amount from a genuinely missing route.
+    /// True if some hop's input couldn't cover that hop's own gas (gas-aware) or a sim
+    /// produced a literal zero output (gas-unaware) — i.e. the amount is dust, not unroutable.
     input_below_hop_gas: bool,
 }
 
@@ -272,13 +269,9 @@ impl BellmanFordAlgorithm {
 
         let out_idx = ctx.token_out_node.index();
         if spfa.amount[out_idx].is_zero() {
-            // amount[out] == 0. If some hop's own input amount couldn't cover that single
-            // hop's gas cost (gas-aware), or a successful simulation produced a literal zero
-            // output (gas-unaware), the requested amount is too small to route to a usable
-            // output at some hop -> AmountTooSmall (dust or gas-unprofitable, mid-route
-            // included). A too-large amount instead fails by simulation error, so it never
-            // sets the flag. Anything else (unreachable, connector/forbid-revisits excluded,
-            // sim error, missing state, timeout) -> NoGraphPath.
+            // Dust (a hop's input below its own gas) -> AmountTooSmall; everything else
+            // (unreachable, filtered, sim error incl. too-large, missing state, timeout)
+            // -> NoGraphPath.
             let reason = if spfa.input_below_hop_gas {
                 NoPathReason::AmountTooSmall
             } else {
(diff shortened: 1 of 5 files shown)


brunoguerios:
## Happy-path performance: measured and mitigated

**Method:** deterministic A/B replay — recorded ethereum market fixture through the full pipeline (`Solver::from_recording`, all worker pools, gas-aware), 44 scenarios × 30 passes (n=1,320 quotes/run, all successful = pure happy path), release build, wall-clock per quote, runs interleaved base↔branch to control machine drift. Baseline = merge-base `2830ff1`.

**Round 1 — ungated check (as first pushed):** consistent regression, mean **+4.0%** (~+215µs), p50 +3.4%, p95 +3.7%, vs a base-to-base noise floor of ~1–2%. Cause: the input-vs-hop-gas probe (`resolve_token_price` + BigInt `gas_adjusted_amount`) ran on every successful edge simulation.

**Fix (`2217210e`):** gate the probe on the already-computed `net_candidate <= 0` — an input below its hop's gas implies a non-positive net (output value ≤ input value < hop gas ≤ cumulative gas), so economic relaxing edges can never trip it and now skip it entirely. Semantics unchanged; all 60 algorithm tests pass as-is (both gas-aware tests have `net_candidate <= 0` and still exercise the gated path).

**Round 2 — gated:** branch indistinguishable from base — GATED-1 (5,472µs), BASE-4 (5,483µs), GATED-2 (5,455µs) within 0.5% of each other; residual delta below the round's own base-to-base drift (3.8%).

| | mean µs | p50 | p95 |
|---|---|---|---|
| base (avg of 2) | 5,382 | 5,978 | 8,177 |
| branch ungated (avg of 2) | 5,635 | 6,291 | 8,577 |
| branch gated (avg of 2) | 5,464 | 6,043 | 8,326 |

Reason-labeling behavior is failure-path only; the probe is the only hot-path touch and is now gated off the economic majority.

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -496,7 +496,9 @@ fn aggregate_no_route_reason(
                 NoPathReason::SourceTokenNotInGraph | NoPathReason::DestinationTokenNotInGraph => {
                     return Some(*reason)
                 }
-                NoPathReason::NoGraphPath | NoPathReason::NoScorablePaths => {
+                NoPathReason::NoGraphPath |
+                NoPathReason::NoScorablePaths |
+                NoPathReason::AmountTooSmall => {
```
tvinagre: This is the one thing I'd fix before merge: `failed_solvers` is populated in async completion order (`collect_responses` pushes as `pending.next()` resolves), so first-wins within this tier is nondeterministic when pools disagree. Two pools can legitimately diverge — e.g. a `max_hops=1` BF pool reports `NoGraphPath` (unreachable) while a `max_hops=3` pool reports `AmountTooSmall` (dust mid-route) — and the logged reason then flips run-to-run depending on which pool finishes first. The doc comment promises "the most informative reason", but that's only implemented for the token-not-in-graph tier.

Suggestion: rank explicitly within the tier, most specific wins — `AmountTooSmall > NoScorablePaths > NoGraphPath` (e.g. a small `fn specificity(reason) -> u8` and keep the max). Makes the telemetry this PR exists for deterministic.

Related but fine to keep out of scope: `InsufficientLiquidity` failures (MostLiquid/WaterFill pools) are a different `SolveError` variant and contribute nothing here, so a mixed fleet keeps two vocabularies for the same dust order.

fynd-core/src/algorithm/mod.rs
```diff
@@ -311,6 +311,9 @@ pub enum NoPathReason {
     NoGraphPath,
     /// Paths exist but none could be scored (e.g., missing edge weights).
     NoScorablePaths,
+    /// A path exists, but the requested amount is too small to produce a
+    /// non-zero output (dust).
+    AmountTooSmall,
```
tvinagre, line 320: Two small wording points on the variant docs:

1. "A path exists" isn't guaranteed — the flag is a global latch over all explored edges, so a solve where `token_out` is genuinely unreachable can still latch on a side branch (your PR caveat covers this shape).
2. The label's meaning shifts with scoring mode: gas-unaware is output-side (literal zero output), gas-aware is input-side (input below hop gas). The same market condition — healthy-sized input whose output floors to zero on an extreme-rate pool — reads `AmountTooSmall` gas-unaware but `NoGraphPath` gas-aware (which `test_no_graph_path_when_output_uneconomic_but_input_economic` pins deliberately). Defensible, but worth a sentence here so dashboard readers know the semantics depend on deployment config.

Also the `Display` text below says "output amount too small" while the gas-aware check is input-side.

fynd-core/src/algorithm/bellman_ford.rs
```diff
@@ -444,6 +450,27 @@ impl BellmanFordAlgorithm {
                             ctx.gas_price_wei.as_ref().unwrap(),
                             v_price.as_ref(),
                         );
+                        // Dust signal: this hop's input can't cover its own gas. Input-side, so
+                        // healthy and too-large orders (large inputs) never trip it. Gated on
+                        // net_candidate <= 0 (implied by input < hop gas) to keep the extra
+                        // price lookup off the hot path.
+                        if !input_below_hop_gas && net_candidate <= BigInt::ZERO {
```
tvinagre, line 457: Observation, no change needed: in gas-aware mode this is strictly "input uneconomic for some explored hop", not dust in the absolute sense. A $50 order relaxing past a pool whose single-hop gas costs $80 latches the flag, and since the latch covers branches irrelevant to `token_out`, a solve failing for an unrelated reason can then read `amount_too_small`. Rare on mainstream pools, but worth keeping in mind when reading Base dashboards during gas spikes.

One more known hole in the same direction as your unpriced-token caveat: a zero-gas hop can never latch (`input - 0 > 0`), so dust flooring to zero through it under-reports as `no_graph_path`. Consistent with the never-mislabel bias — just noting it.

fynd-rpc/src/api/request_capture.rs
```diff
@@ -392,6 +393,7 @@ mod tests {
         );
         assert_eq!(no_route_reason_code(Some(NoPathReason::NoGraphPath)), "no_graph_path");
         assert_eq!(no_route_reason_code(Some(NoPathReason::NoScorablePaths)), "no_scorable_paths");
+        assert_eq!(no_route_reason_code(Some(NoPathReason::AmountTooSmall)), "amount_too_small",);
```
tvinagre: Nit, stray trailing comma:

```suggestion
        assert_eq!(no_route_reason_code(Some(NoPathReason::AmountTooSmall)), "amount_too_small");
```

tvinagre (commented):
Reviewed with a focus on `AmountTooSmall` coverage/consistency across algorithms and on docs. Overall this looks good: detection is deliberately conservative (under-reports, never mislabels healthy orders), PathFrankWolfe inherits correctly via the initial full-amount BF solve (split-phase errors are only swallowed after a valid single-path exists), and zero-amount orders can't reach SPFA since order validation rejects them. I ran the BF, PFW, aggregation, and request-capture suites at head: 73/73 pass.

Two notes that don't anchor to a line:

- **Docs**: nothing missing in-repo — `no_route_reason` is log-only (not in `openapi.json`, `QuoteStatus` unchanged), no file under `docs/` or `.claude/` enumerates reason codes, and the CHANGELOG comes from release-plz. The variant rustdoc added here is the declaration point.
- **Ops heads-up for merge**: any Grafana/Loki panel that groups or filters on `no_route_reason` needs `amount_too_small` added, and the Base `no_graph_path` rate will drop when this ships — expected, but it will look like a regression fixed itself.

One suggested fix inline (aggregation determinism); the rest are minor observations.

tvinagre (approved):

PR #345, opened 2026-07-23
ci(release): restore workflow_dispatch with a required tag input

kayibal:
## What

Restores the manual trigger on `release.yaml` — this time functional: a **required `tag` input** is used for checkout and the npm version step in both jobs, so a dispatch publishes the tag's content while running the **current workflow from main**.

## Why (reverses #309's removal, with new information)

Release-event runs pin the workflow file **at the tag's commit**. That means the idempotent-publish fix (#343) can never reach the already-tagged v0.97.2: re-running the failed run *and* deleting/recreating the Release both execute the old workflow, which dies on the crates that already published. Verified on [run 30004711614](https://github.com/propeller-heads/fynd/actions/runs/30004711614).

So the earlier reasoning "re-run covers all recovery" holds for flakes, but not for the one case where the workflow itself is broken at the tagged commit — that's exactly what a dispatch-from-main covers (as suggested in [#309's review](https://github.com/propeller-heads/fynd/pull/309#issuecomment-thread)).

Publishing is idempotent since #343, so dispatching an already-published tag is a safe no-op.

## After merge

Dispatch with `tag=0.97.2` → skips `fynd-core` + npm (already published), publishes the four missing crates → crates.io reaches 0.97.2 → release-plz unblocks and opens the 0.98.0 release PR.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,6 +3,16 @@ name: Release
 on:
   release:
     types: [created]
+  # Manual escape hatch: (re)publish an existing tag with the CURRENT workflow from
+  # main. Release-event runs pin the workflow file at the tag's commit, so a workflow
+  # fix can never reach an already-tagged release any other way. Publishing is
+  # idempotent, so re-dispatching a fully published tag is a no-op.
+  workflow_dispatch:
+    inputs:
+      tag:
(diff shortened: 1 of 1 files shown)


kayibal:
Closing — going with a local publish of the missing crates instead; the dispatch escape hatch adds more workflow surface than the recovery is worth.

PR #347, opened 2026-07-23
fix: contain pool simulation panics in solver algorithms

zizou0x:
## What

Routes all solver-side `ProtocolSim::get_amount_out` calls (bellman_ford, path_frank_wolfe/split_primitives, most_liquid) through `get_amount_out_guarded`, which wraps the call in `catch_unwind` and converts a panic into a `SimulationError::FatalError`. The algorithms already treat simulation errors as "skip this pool", so a panicking pool no longer kills the worker thread. On a contained panic the guard logs the triggering amount and token pair (address + symbol) so the offending pool/quote can be tracked down.

The diagnostic `get_amount_out` probe in `derived/computations/pool_depth.rs` is routed through the same guard, so a panicking pool degrades to a diagnostic string instead of killing the computation-manager worker.

Also removes `SimResult.gas`, which was written but never read outside tests (gas totals flow through `SplitExecution.total_gas`); the zero-gas override test now sums per-hop gas from `hop_results`.

## Why

Prod incident 2026-07-23 (~10:22–12:02 UTC): a quote for ~7.9e28 USDC triggered `attempt to divide by zero` in ruint (U256 math inside tycho-simulation pool code). Each occurrence killed one solver worker thread; after all workers on both ethereum pods had panicked, every quote returned `no_route_found` (100% no_route on the Quote outcomes panel) until the pods were manually recreated ~1.5h later.

With this change the poison request fails with a simulation error for the affected pool/path and the workers stay alive.

## Tests

- `sim_guard`: panic → `FatalError` with the panic message; success passes through
- `split_primitives::test_simulate_path_contains_simulation_panic`: a `ProtocolSim` performing a real U256 division by zero surfaces as `AlgorithmError::SimulationFailed` instead of unwinding

### DIFF
--- a/fynd-core/src/algorithm/sim_guard.rs
+++ b/fynd-core/src/algorithm/sim_guard.rs
@@ -18,42 +18,57 @@ use tycho_simulation::tycho_common::{
     },
 };
 
-/// Calls `sim.get_amount_out`, converting a panic into a `SimulationError::FatalError`.
+/// Extension trait adding panic-guarded simulation calls to every [`ProtocolSim`].
 ///
-/// On a contained panic, logs the input that triggered it (amount and token pair) so the
-/// offending pool/quote can be tracked down from the logs.
-pub(crate) fn get_amount_out_guarded(
-    sim: &dyn ProtocolSim,
-    amount_in: BigUint,
-    token_in: &Token,
-    token_out: &Token,
-) -> Result<GetAmountOutResult, SimulationError> {
-    // `amount_in` is cloned into the call so the original stays available for the panic log.
-    let outcome = catch_unwind(AssertUnwindSafe(|| {
-        sim.get_amount_out(amount_in.clone(), token_in, token_out)
-    }));
+/// Crate-private: solver code calls the guarded variants; the raw `ProtocolSim` methods
+/// remain untouched for callers that manage panics themselves.
+pub(crate) trait GuardedProtocolSim {
+    /// Calls `get_amount_out`, converting a panic into a `SimulationError::FatalError`.
+    ///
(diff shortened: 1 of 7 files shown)


fynd-core/src/algorithm/sim_guard.rs
```diff
@@ -0,0 +1,77 @@
...
+        .unwrap_or_else(|panic_payload| {
+            let message = panic_payload
+                .downcast_ref::<&str>()
+                .copied()
+                .or_else(|| {
+                    panic_payload
+                        .downcast_ref::<String>()
+                        .map(String::as_str)
+                })
+                .unwrap_or("<non-string panic payload>");
```
zizou0x: We need a debug log here containing all relevant infos, so we can track these panics down.

zizou0x: Added in ee04f07. On a contained panic the guard now logs the triggering `amount_in`, both token addresses and symbols, and the panic message:

```
warn!(
    %amount_in,
    token_in = %token_in.address,
    token_out = %token_out.address,
    token_in_symbol = %token_in.symbol,
    token_out_symbol = %token_out.symbol,
    panic = message,
    "pool simulation panicked; skipping pool"
);
```

I used `warn!` rather than `debug!` since a contained third-party panic is an anomaly we want visible in prod (debug is filtered there) — happy to drop it to `debug!` if you'd prefer. `amount_in` is cloned into the guarded call so the original stays available for the log.

Also guarded the diagnostic `get_amount_out` probe in `pool_depth.rs` (b54c4ed), which the PR note had left unguarded.

kayibal (approved):

PR #348, opened 2026-07-23
fix: respawn panicked solver workers instead of losing them

zizou0x:
## Problem

Prod incident 2026-07-23. A quote request caused a divide-by-zero panic in pool math. Each occurrence killed
one solver worker thread. The panic was logged only at shutdown `join()`, about 100 minutes later. After all
workers on both ethereum pods died, every quote returned `no_route_found` for about 1.5 hours, until the pods
were recreated.

## Root cause

`spawn_workers_generic` ran `SolverWorker::run` directly in the thread body. A panic unwound the thread and
ended it. The pool kept the `JoinHandle`, so it reported the full worker count and stayed silent. Nothing
logged the loss until shutdown.

## Fix

Each worker thread now runs worker sessions in a loop, in `WorkerContext::run_sessions`. A session is wrapped
in `catch_unwind`. `spawn_workers_generic` only spawns threads.

- A panic ends the session, logs at ERROR (pool, algorithm, worker_id, message), increments
  `worker_pool_worker_panics_total{pool}`, and respawns the worker after a backoff.
- Backoff starts at 100 ms and doubles to a 2 s cap. The worker gives up on the 10th consecutive failure. A
  session that lives 600 s or more resets the budget.
- On give-up the process logs and exits 1, so the orchestrator restarts it. Workers are homogeneous, so a
  budget-exhausting failure is almost certainly shared. Exactly one thread runs the exit.
- A fresh worker seeds its `allow_stale` readiness from the shared derived-data store, so a respawned worker
  does not fail tasks `NotReady` until the next broadcast event.
- Session shutdown receivers come from `resubscribe()`, so workers exit when the pool drops its sender. A
  pre-check on the long-lived receiver catches a shutdown sent while no session was listening.

Complements #347, which stops that specific panic escaping the simulation call.

## Notes for reviewers

Read `supervisor.rs` first, then the exit policy in `pool.rs` — that is the one product decision here. Read
the last commit separately: it is a pure test refactor that folds three mock algorithms into one stub, and it
accounts for most of the diff size.

Two accepted trade-offs:

- Graceful shutdown can take up to 2 s longer. A worker sleeping in backoff cannot observe the signal.
- The 600 s stability window resets the budget rather than decaying it, so failures never expire on their
  own. A worker that panics about every 9 minutes still exits the process after about 90 minutes. Deliberate:
  the give-up path targets deterministic panics, which burn the budget in about 11 seconds.

`worker_pool_worker_panics_total` under-reports the give-up case, because ~11 s of crash loop rarely spans a
scrape. Alert on restart count and the give-up log line.

## Follow-ups

- Per-task `catch_unwind` inside `SolverWorker::run`, so a panicking task answers its requester instead of
  dropping the oneshot.
- Gate the task select arm on readiness.
- Share one panic-payload downcast helper between `sim_guard.rs` and `supervisor.rs`.
- Give the pool a health signal: `num_workers()` reports the static count and would over-report if a worker
  ever gives up without exiting the process.

### DIFF
--- a/fynd-core/src/worker_pool/registry.rs
+++ b/fynd-core/src/worker_pool/registry.rs
@@ -199,7 +199,6 @@ where
         // Subscribed before the thread starts so shutdown signals sent at any point,
         // including while the worker is recovering from a panic, are never missed.
         let mut shutdown_rx = params.shutdown_tx.subscribe();
-        let shutdown_tx = params.shutdown_tx.clone();
         let algorithm_name = params.algorithm.clone();
         let pool_name = params.pool_name.clone();
         let factory = factory.clone();
@@ -219,7 +218,7 @@ where
                     // consumed (or poisoned) when it panics.
                     let session_event_rx = event_rx.resubscribe();
                     let session_derived_event_rx = derived_event_rx.resubscribe();
-                    let session_shutdown_rx = shutdown_tx.subscribe();
+                    let session_shutdown_rx = shutdown_rx.resubscribe();
 
                     // A shutdown sent while no session was listening (e.g. mid-respawn) is
                     // buffered in the receiver created before the thread started.
@@ -603,6 +602,45 @@ mod tests {
             handle
                 .join()
                 .expect("worker thread should shut down cleanly");
+        }
+    }
+
+    #[tokio::test]
+    async fn workers_exit_when_pool_drops_shutdown_sender() {
+        let (market, _) = setup_market_weighted(vec![]);
+        let (_task_tx, task_rx) = async_channel::bounded::<SolveTask>(10);
+        let (event_tx, _) = broadcast::channel::<MarketEvent>(10);
+        let (derived_event_tx, _) = broadcast::channel(10);
+        let (shutdown_tx, _) = broadcast::channel(1);
+
+        let params = SpawnWorkersParams {
+            algorithm: "panic_on_poison".to_string(),
+            pool_name: "test_pool".to_string(),
+            num_workers: 1,
+            algorithm_config: AlgorithmConfig::default(),
+            task_rx,
+            market_data: market,
+            derived_data: DerivedData::new_shared(),
+            event_rx: event_tx.subscribe(),
+            derived_event_rx: derived_event_tx.subscribe(),
+            shutdown_tx,
+            liquidity_scope: LiquidityScope::default(),
+            exclude_protocols: Vec::new(),
+            fallback_fee_tiers: SharedFeeTiers::default(),
+        };
+        let factory = |_config: AlgorithmConfig| PanicOnPoisonAlgorithm;
+        let workers = spawn_workers_generic(params, &factory);
+        // `params` (holding the only shutdown sender) is consumed and dropped above.
+
+        for handle in workers {
+            tokio::time::timeout(
+                Duration::from_secs(5),
+                tokio::task::spawn_blocking(move || handle.join()),
+            )
+            .await
+            .expect("worker should exit when the shutdown sender drops")
+            .unwrap()
+            .expect("worker thread should exit cleanly");
         }
     }
 
--- a/fynd-core/src/worker_pool/registry.rs
+++ b/fynd-core/src/worker_pool/registry.rs
@@ -578,6 +578,123 @@ mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn worker_gives_up_after_repeated_rapid_panics() {
+        let (market, _) = setup_market_weighted(vec![]);
+        let (_task_tx, task_rx) = async_channel::bounded::<SolveTask>(10);
+        let (event_tx, _) = broadcast::channel::<MarketEvent>(10);
+        let (derived_event_tx, _) = broadcast::channel(10);
+        let (shutdown_tx, _) = broadcast::channel(1);
+        let gave_up = Arc::new(std::sync::atomic::AtomicBool::new(false));
+        let gave_up_flag = Arc::clone(&gave_up);
+
+        let params = SpawnWorkersParams {
+            algorithm: "panic_on_poison".to_string(),
+            pool_name: "test_pool".to_string(),
+            num_workers: 1,
+            algorithm_config: AlgorithmConfig::default(),
+            task_rx,
+            market_data: market,
+            derived_data: DerivedData::new_shared(),
+            event_rx: event_tx.subscribe(),
+            derived_event_rx: derived_event_tx.subscribe(),
+            shutdown_tx: shutdown_tx.clone(),
+            liquidity_scope: LiquidityScope::default(),
+            exclude_protocols: Vec::new(),
+            fallback_fee_tiers: SharedFeeTiers::default(),
+            respawn_policy: RespawnPolicy {
+                initial_backoff: Duration::from_millis(1),
+                max_backoff: Duration::from_millis(2),
+                max_attempts: 3,
+                stable_session: Duration::from_secs(60),
+            },
+            on_worker_gave_up: Arc::new(move || {
+                gave_up_flag.store(true, std::sync::atomic::Ordering::SeqCst)
+            }),
+        };
+        let factory = |_config: AlgorithmConfig| -> PanicOnPoisonAlgorithm {
+            panic!("deterministic init panic")
+        };
+        let workers = spawn_workers_generic(params, &factory);
+
+        for handle in workers {
+            tokio::time::timeout(
+                Duration::from_secs(5),
+                tokio::task::spawn_blocking(move || handle.join()),
+            )
+            .await
+            .expect("worker should give up instead of retrying forever")
+            .unwrap()
+            .expect("worker thread should exit cleanly after giving up");
+        }
+        assert!(gave_up.load(std::sync::atomic::Ordering::SeqCst));
+    }
+
+    #[tokio::test]
+    async fn shutdown_sent_mid_respawn_is_not_lost() {
+        let (market, _) = setup_market_weighted(vec![]);
+        let derived_data = DerivedData::new_shared();
+        let (task_tx, task_rx) = async_channel::bounded(10);
+        let (event_tx, _) = broadcast::channel::<MarketEvent>(10);
+        let (derived_event_tx, _) = broadcast::channel(10);
+        let (shutdown_tx, _) = broadcast::channel(1);
+
+        let params = SpawnWorkersParams {
+            algorithm: "panic_on_poison".to_string(),
+            pool_name: "test_pool".to_string(),
+            num_workers: 1,
+            algorithm_config: AlgorithmConfig::default(),
+            task_rx,
+            market_data: market,
+            derived_data,
+            event_rx: event_tx.subscribe(),
+            derived_event_rx: derived_event_tx.subscribe(),
+            shutdown_tx: shutdown_tx.clone(),
+            liquidity_scope: LiquidityScope::default(),
+            exclude_protocols: Vec::new(),
+            fallback_fee_tiers: SharedFeeTiers::default(),
+            respawn_policy: RespawnPolicy {
+                initial_backoff: Duration::from_millis(500),
+                ..RespawnPolicy::default()
+            },
+            on_worker_gave_up: Arc::new(|| {}),
+        };
+        let factory = |_config: AlgorithmConfig| PanicOnPoisonAlgorithm;
+        let workers = spawn_workers_generic(params, &factory);
+
+        let token_a = token(0x01, "A");
+        let token_b = token(0x02, "B");
+
+        // The poison task panics mid-solve; its response channel is dropped. The worker
+        // now sleeps through its 500ms backoff with no session listening for shutdown.
+        let (poison_tx, poison_rx) = oneshot::channel();
+        let poison_order = order(&token_a, &token_b, POISON_AMOUNT as u128, OrderSide::Sell);
+        task_tx
+            .send(SolveTask::new(Uuid::new_v4(), poison_order, poison_tx))
+            .await
+            .unwrap();
+        assert!(poison_rx.await.is_err());
+
+        // Sent while the worker is mid-respawn (no session receiver listening yet). The
+        // buffered `shutdown_rx.try_recv()` pre-check must still catch it.
+        shutdown_tx.send(()).unwrap();
+
+        for handle in workers {
+            tokio::time::timeout(
+                Duration::from_secs(5),
+                tokio::task::spawn_blocking(move || handle.join()),
+            )
+            .await
+            .expect("buffered shutdown sent mid-respawn should not be lost")
+            .unwrap()
+            .expect("worker thread should shut down cleanly");
+        }
+
+        // Keep the sender alive until after the join so the worker exits via the
+        // buffered shutdown signal, not because the task channel closed.
+        drop(task_tx);
+    }
+
     #[test]
     fn test_registry_spawns_path_frank_wolfe() {
         let (shutdown_tx, _) = broadcast::channel(1);
--- a/fynd-core/src/worker_pool/registry.rs
+++ b/fynd-core/src/worker_pool/registry.rs
@@ -268,14 +268,12 @@ mod tests {
     use super::*;
     use crate::{
         algorithm::{
-            most_liquid::DepthAndPrice,
-            test_utils::{order, setup_market_weighted, token},
-            Algorithm, AlgorithmError,
+            test_utils::{order, setup_market_weighted, token, StubAlgorithm},
+            AlgorithmError,
         },
-        derived::{computation::ComputationRequirements, DerivedData},
-        feed::market_data::{MarketData, StateLabel},
-        graph::petgraph::{PetgraphStableDiGraphManager, StableDiGraph},
-        types::{quote::OrderSide, Order, RouteResult, SolveError},
+        derived::DerivedData,
+        feed::market_data::MarketData,
+        types::{quote::OrderSide, SolveError},
     };
 
     fn make_params(algorithm: &str, num_workers: usize) -> SpawnWorkersParams {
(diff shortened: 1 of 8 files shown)


brunoguerios (commented):
The respawn direction is good and the shutdown buffering is race-free. Three things to address before merge:

1. Worker threads now hold a shutdown sender clone, so dropping the pool without an explicit shutdown no longer stops the workers. Use resubscribe instead.
2. The retry loop respawns forever. Add exponential backoff with a bounded attempt budget that resets on stable sessions, and kill the process when any worker gives up (workers are homogeneous, so a budget-exhausting failure is almost certainly shared; exiting at the first give-up avoids a permanently degraded pool).
3. A respawned worker starts not-ready behind a green health check. Seed its readiness tracker from the shared derived data.

Inline comments carry the details, plus test and doc fixes. Note the branch now conflicts with main: spawn_workers_generic gained liquidity_scope, exclude_protocols and fallback_fee_tiers, which must move into the session loop on rebase.

**15 comments** — 0 Blocker · 4 Important · 3 Suggestion · 6 Nit · 2 Context.

### Important
- **IMPL `#1`** Worker threads hold a shutdown sender clone, so the shutdown channel can never close — `registry.rs:170`
- **TESTS `#1`** Poison-task result is discarded, so the test can pass without a panic — `registry.rs:512`
- **TESTS `#2`** No test for the buffered-shutdown path (shutdown sent mid-respawn) — `registry.rs:190`
- **DOCS `#1`** `spawn_workers_generic` doc contradicts the new respawn behavior — `registry.rs:139`

### Suggestions
- **DESIGN `#2`** Session/supervision loop hidden in a 75-line inline closure — `registry.rs:178`
- **DESIGN `#3`** Fixed 100ms backoff allows a permanent 10 Hz error-log/rebuild loop — `registry.rs:42`
- **DOCS `#2`** "(or poisoned)" invents a tokio semantics that does not exist — `registry.rs:184`

### Nits
- **DESIGN `#4`** Third bespoke mock `Algorithm` impl in tests — `registry.rs:449`
- **TESTS `#4`** `POISON_AMOUNT as u128` cast — `registry.rs:514`
- **DOCS `#3`** Loop-top comment narrates the fixed bug and is verbose — `registry.rs:179`
- **DOCS `#4`** `session` binding holds the session's result, not the session — `registry.rs:198`
- **DOCS `#6`** Test doc says the panic "does not kill the worker" — `registry.rs:449`
- **DOCS `#7`** Non-parallel names for the two test tasks — `registry.rs:513`

### Context
- **BIZ `#2`** Respawned worker answers NotReady (after waiting the full algorithm timeout) until the next derived-data event — `registry.rs:198`
- **BIZ `#3`** Deterministic startup panics now retry forever instead of dying once — `registry.rs:183`

---

**[DOCS `#1` · Important]**

**Description**: The doc says "The `factory` closure is called once per worker to create the algorithm instance" (line 146), but the loop now calls `factory` once per session — again after every panic respawn (registry.rs:205). Factory implementors must now tolerate repeated calls; the doc states the opposite contract, and the bullet list omits the panic-containment behavior that is now the function's most important guarantee.

**Suggestion**: Update the doc: state that each worker thread runs sessions in a loop, a panic respawns the worker after a backoff, and the factory is called at every (re)start. Remove or correct "called once per worker".

fynd-core/src/worker_pool/registry.rs
```diff
@@ -159,35 +165,93 @@ where
...
+                            worker
+                                .run(
+                                    session_event_rx,
+                                    session_derived_event_rx,
+                                    task_rx.clone(),
+                                    session_shutdown_rx,
+                                )
+                                .await;
+                        });
+                    }));
```
brunoguerios: Please seed the fresh worker's readiness tracker from the shared DerivedData at session start. allow_stale requirements are satisfiable by data that is already computed, so the respawned worker becomes ready immediately — the same effect as the lag-recovery path this PR cites, which keeps its tracker. Note that /v1/health stays green through all of this (it checks shared derived data, not per-worker readiness), and a worker that pulls a task while cold cannot become ready mid-wait: the readiness notify fires from the same select loop that is blocked on the wait, so each cold task burns the full algorithm timeout.

---

**[BIZ `#2`]**

**What changed**: Each respawn constructs a brand-new `SolverWorker`, whose `ReadinessTracker` starts empty and only becomes ready via `DerivedDataEvent`s on a freshly resubscribed channel. All three built-in algorithms declare `allow_stale` requirements (`token_prices`, plus `spot_prices` for bellman_ford/path_frank_wolfe), and a fresh tracker with `allow_stale` requirements is not ready until it observes a `ComputationComplete` (see `new_tracker_not_ready_with_stale_requirements`). So after a panic, the respawned worker still pulls tasks off the shared queue, blocks each one in `wait_until_ready` for up to the algorithm timeout, then fails it with `SolveError::NotReady` — until the next block's computations re-emit events. This differs from the PR summary's "same as the existing lag-recovery path": lag recovery keeps the live worker and its tracker state; respawn is equivalent to a cold start.

**Where**: `fynd-core/src/worker_pool/registry.rs:198–225`; `fynd-core/src/worker_pool/worker.rs:544–553` (task fails NotReady), `fynd-core/src/worker_pool/worker.rs:357` (waits full timeout)

- Whether a not-yet-ready worker consuming tasks from the shared `async_channel` turns one panic into a window of user-visible NotReady quote errors (task stolen from healthy workers), and how long that window is at production block cadence.
- Whether derived-data events re-fire every block (so the window closes on the next block) or only on change.

fynd-core/src/worker_pool/registry.rs
```diff
@@ -159,35 +165,93 @@ where
...
+                            );
+                            metrics::counter!(
+                                "worker_pool_worker_panics_total",
+                                "pool" => pool_name.clone()
+                            )
+                            .increment(1);
+                            thread::sleep(RESPAWN_BACKOFF);
+                        }
+                    }
+                }
```
brunoguerios: Please make the retry loop give up instead of respawning forever: exponential backoff (100ms doubling, capped at ~2s), give up after ~10 attempts, and reset the budget when a session survives past a stability threshold (~10 minutes). A deterministic failure then stops retrying within seconds, while spaced transient panics (the incident shape) keep respawning indefinitely because each stable session resets the budget. When any worker gives up, kill the whole process: workers are homogeneous, so a failure that exhausts one worker's budget is almost certainly shared; exiting at the first give-up avoids a permanently degraded pool, and a crash loop is visible to operators where a green pod with zero solvers is not.

---

**[BIZ `#3`]**

**What changed**: The tokio runtime build `expect`, the algorithm factory (e.g. `MostLiquidAlgorithm::with_config(...).expect("invalid worker configuration...")`), and `initialize_graph` all now run inside `catch_unwind`. A panic that is deterministic (bad config, runtime build failure) previously killed each worker thread once, silently; it now produces an unbounded respawn loop per worker — an ERROR log line and a counter increment every ~100ms per worker, forever — until the pool is shut down. There is no retry cap or backoff escalation. Correspondingly, `WorkerPool::shutdown`'s `handle.join()` error arm (`pool.rs:163`) becomes unreachable for caught panics, so anything keyed on the shutdown-time "worker thread panicked" log must key on the new mid-run line or the new metric.

**Where**: `fynd-core/src/worker_pool/registry.rs:183–254` (loop with no attempt cap), `fynd-core/src/worker_pool/registry.rs:44` (fixed 100ms backoff)

- Log/metric volume math for a deterministic init panic across a full pool (num_workers × 10 attempts/sec) and whether alerting on `worker_pool_worker_panics_total` distinguishes a one-off solve panic from a crash loop.
- That a misconfigured pool failing loudly-forever (instead of running with zero workers) is the intended operational contract.

fynd-core/src/worker_pool/registry.rs
```diff
@@ -159,35 +165,93 @@ where
...
             })
```
brunoguerios: **[DESIGN `#2` · Suggestion]**

**Description**: The thread closure now nests four concerns — supervision loop, `catch_unwind`, runtime construction, worker construction — inside `spawn_workers_generic`, whose doc still frames it as "common parts of spawning workers" (`registry.rs:139–148`). The ten captured variables (factory, config, data handles, four channel ends, ids, names) travel together into every session: one type waiting to be named.

**Suggestion**: Extract the loop into a named unit, e.g. a `WorkerContext<A, F>` struct with a `run_sessions()` method; the spawn loop then reads as spawn-only and the respawn policy gets a name, a doc comment, and a place for its own tests.

fynd-core/src/worker_pool/registry.rs
```diff
@@ -37,6 +39,10 @@ pub(crate) const AVAILABLE_ALGORITHMS: &[&str] =
 /// Default algorithm to use if none specified.
 pub(crate) const DEFAULT_ALGORITHM: &str = "most_liquid";
 
+/// Pause before respawning a panicked worker, so a panic during worker
+/// initialization cannot turn the respawn loop into a hot spin.
+const RESPAWN_BACKOFF: Duration = Duration::from_millis(100);
```
brunoguerios: Use the doubling-with-cap pattern the Rust client's RetryConfig already uses (100ms initial, doubled, 2s cap) — duplicate it inline rather than extracting shared code. This folds into the give-up policy requested on the startup-panics comment, which also bounds the log volume this finding is about.

---

**[DESIGN `#3` · Suggestion]**

**Description**: A deterministic panic in `factory` (e.g. its `expect` on an invalid config, `registry.rs:272–275`) or `initialize_graph` respawns each worker ~10 times per second forever, each iteration logging at ERROR, incrementing the panics counter, and rebuilding the graph — a 16-worker pool produces ~160 error lines/sec indefinitely. Across all workers of all pools this is the log-flood/churn failure mode this codebase was bitten by before (see the busy-spin rationale in `fynd-core/src/worker_pool/worker.rs:1080–1085`). Pros of capped exponential backoff: bounds log volume and rebuild churn; standard supervisor pattern. Cons: extra state; slower recovery after a transient burst; the current constant is simple and documented.

**Suggestion**: Double the backoff up to a cap of a few seconds (e.g. 100ms doubling to 10–30s), resetting when a session survives noticeably longer than the current backoff or serves a task. Keeps the fast recovery for the incident's per-quote panic while bounding log/metric volume for a poisoned worker.

fynd-core/src/worker_pool/registry.rs
```diff
@@ -365,6 +443,108 @@ mod tests {
...
+        }
+
+        fn computation_requirements(&self) -> ComputationRequirements {
+            ComputationRequirements::none()
+        }
+
+        fn timeout(&self) -> Duration {
+            Duration::from_secs(1)
+        }
+    }
```
brunoguerios: **[DESIGN `#4` · Nit]**

`PanicOnPoisonAlgorithm` repeats the ~30-line trait boilerplate already written twice in `fynd-core/src/worker_pool/worker.rs` (`MockAlgorithm`, `InvalidRouteAlgorithm`); a configurable mock in `algorithm::test_utils` would absorb all three.

fynd-core/src/worker_pool/registry.rs
```diff
@@ -159,35 +165,93 @@ where
         let event_rx = params.event_rx.resubscribe();
         let derived_event_rx = params.derived_event_rx.resubscribe();
         let algorithm_config = params.algorithm_config.clone();
-        let shutdown_rx = params.shutdown_tx.subscribe();
+        // Subscribed before the thread starts so shutdown signals sent at any point,
+        // including while the worker is recovering from a panic, are never missed.
+        let mut shutdown_rx = params.shutdown_tx.subscribe();
+        let shutdown_tx = params.shutdown_tx.clone();
```
brunoguerios: **[IMPL `#1` · Important]**

**Category**: bug

**Plain English**: Stopping the solver pool through the normal explicit stop still works. But when a program simply discards the pool object without that call, the worker threads used to notice the pool was gone and exit on their own. After this change they keep running in the background indefinitely, holding their threads, unless the work queue also happens to be torn down. Discarding the pool should still make its workers exit.

**Root cause**: Each worker thread captures `shutdown_tx = params.shutdown_tx.clone()` for its lifetime, solely to mint per-session receivers — so the shutdown broadcast channel keeps at least one live sender per worker and can never report `Closed` while any worker runs.

- **Behavior**: dropping the `WorkerPool` without calling shutdown no longer terminates workers — `SolverWorker::run`'s closed-channel arm (`fynd-core/src/worker_pool/worker.rs:453–456`) and the `try_recv` `Closed` case are now unreachable, `fynd-core/src/worker_pool/registry.rs:188–196`
- **Complexity**: three shutdown handles per thread where two suffice, and the session receiver is minted from a sender clone while the sibling event receivers use `resubscribe()`, `fynd-core/src/worker_pool/registry.rs:171, 186–188`
- **Documentation**: the comment "or the pool dropped the sender" describes a case the captured clone makes unreachable, `fynd-core/src/worker_pool/registry.rs:194`

**Proposed fix**: Delete the `shutdown_tx` clone (line 171) and mint the session receiver as `let session_shutdown_rx = shutdown_rx.resubscribe();` (line 188). This single change resolves every symptom above: it restores the closed-channel exit path (Behavior), removes the extra captured binding and makes all three session receivers uniform (Complexity), and makes the existing comment accurate again (Documentation). The no-lost-shutdown guarantee is unchanged: a value sent before the resubscribe is still caught by the `try_recv` check that follows.

fynd-core/src/worker_pool/registry.rs
```diff
@@ -365,6 +443,108 @@ mod tests {
...
+        let token_b = token(0x02, "B");
+
+        // The poison task panics mid-solve; its response channel is dropped.
+        let (poison_tx, poison_rx) = oneshot::channel();
+        let poison_order = order(&token_a, &token_b, POISON_AMOUNT as u128, OrderSide::Sell);
+        task_tx
+            .send(SolveTask::new(Uuid::new_v4(), poison_order, poison_tx))
+            .await
+            .unwrap();
+        let _ = poison_rx.await;
```
brunoguerios: **[TESTS `#1` · Important]**

**Description**: `let _ = poison_rx.await` resolves with `Err(RecvError)` when the sender is dropped mid-unwind (a real panic) but also with `Ok(...)` if the worker responds without panicking (e.g. a future refactor runs solves in `tokio::spawn` and converts the panic into a `SolveError`). In that case the respawn path is never exercised, yet the second task still succeeds and the whole test stays green — a vacuous pass, against the project's "verify tests catch failures" rule.

**Suggestion**: Assert the receive error: `poison_rx.await.expect_err("poison task must panic, not respond")` (the dropped sender is the proof the panic unwound the task). Wrap the await in `tokio::time::timeout` like the second task so a stuck worker fails instead of hanging.

fynd-core/src/worker_pool/registry.rs
```diff
@@ -159,35 +165,93 @@ where
...
+                    let session_derived_event_rx = derived_event_rx.resubscribe();
+                    let session_shutdown_rx = shutdown_tx.subscribe();
+
+                    // A shutdown sent while no session was listening (e.g. mid-respawn) is
+                    // buffered in the receiver created before the thread started.
+                    match shutdown_rx.try_recv() {
+                        Err(broadcast::error::TryRecvError::Empty) => {}
+                        // Received a shutdown, or the pool dropped the sender.
+                        _ => break,
+                    }
```
brunoguerios: **[TESTS `#2` · Important]**

**Description**: The pre-subscribed `shutdown_rx` plus the per-iteration `try_recv()` is dedicated machinery for "a shutdown sent while the worker is between sessions is not lost" — a behavior the PR body explicitly claims — but no test exercises it: the existing test's `shutdown_tx.send(())` at line 539 is consumed by a live session's `session_shutdown_rx`, and `try_recv` only ever sees `Empty`. Removing the buffered receiver would break no test.

**Suggestion**: Add a test that sends the poison task, awaits `poison_rx` (worker is now in the 100 ms backoff), sends shutdown, keeps `task_tx` alive, and joins the thread under a timeout (e.g. `tokio::time::timeout(_, tokio::task::spawn_blocking(|| handle.join()))`). With the buffered receiver the join completes; without it the signal is lost and the join times out.

fynd-core/src/worker_pool/registry.rs
```diff
@@ -365,6 +443,108 @@ mod tests {
...
+        };
+        let factory = |_config: AlgorithmConfig| PanicOnPoisonAlgorithm;
+        let workers = spawn_workers_generic(params, &factory);
+
+        let token_a = token(0x01, "A");
+        let token_b = token(0x02, "B");
+
+        // The poison task panics mid-solve; its response channel is dropped.
+        let (poison_tx, poison_rx) = oneshot::channel();
+        let poison_order = order(&token_a, &token_b, POISON_AMOUNT as u128, OrderSide::Sell);
```
brunoguerios: **[TESTS `#4` · Nit]**

Declare `POISON_AMOUNT` as `u128` (`BigUint: From<u128>` covers the comparison at line 470) and drop the cast at line 514.

fynd-core/src/worker_pool/registry.rs
```diff
@@ -159,35 +165,93 @@ where
...
-                        .run(event_rx, derived_event_rx, task_rx, shutdown_rx)
-                        .await;
-                });
+                // Run worker sessions until clean shutdown. A panic while solving (e.g. pool
+                // math dividing by zero) only ends the current session: it is reported loudly
+                // and the worker is respawned, instead of silently losing the thread until
+                // every worker in the pool is dead.
+                loop {
+                    // Fresh receivers for this session; the previous session's receivers are
+                    // consumed (or poisoned) when it panics.
```
brunoguerios: **[DOCS `#2` · Suggestion]**

**Description**: Tokio broadcast receivers are never "poisoned" — poisoning is a `std::sync::Mutex` concept. The receivers are simply moved into the session closure and dropped during unwind, and that happens when the session starts, not "when it panics". The comment gives a wrong reason for a correct action and could make a maintainer think receiver reuse after a panic is unsafe for channel-integrity reasons.

**Suggestion**: Replace with the accurate reason, e.g.: "Fresh receivers for this session; the previous session's receivers were moved into it and dropped when it ended."

fynd-core/src/worker_pool/registry.rs
```diff
@@ -159,35 +165,93 @@ where
...
-
-                    worker.initialize_graph().await;
-                    worker
-                        .run(event_rx, derived_event_rx, task_rx, shutdown_rx)
-                        .await;
-                });
+                // Run worker sessions until clean shutdown. A panic while solving (e.g. pool
+                // math dividing by zero) only ends the current session: it is reported loudly
+                // and the worker is respawned, instead of silently losing the thread until
+                // every worker in the pool is dead.
```
brunoguerios: **[DOCS `#3` · Nit]**

"Run worker sessions until clean shutdown" restates the loop, "reported loudly" is vague (the log and metric are right below), and "instead of silently losing the thread until every worker in the pool is dead" narrates the old incident. Trim to the WHY, e.g.: "Contain panics (e.g. pool math dividing by zero) to the current session so one bad task cannot permanently kill the worker thread."

fynd-core/src/worker_pool/registry.rs
```diff
@@ -159,35 +165,93 @@ where
...
+                                    session_event_rx,
+                                    session_derived_event_rx,
+                                    task_rx.clone(),
+                                    session_shutdown_rx,
+                                )
+                                .await;
+                        });
+                    }));
+
+                    match session {
```
brunoguerios: **[DOCS `#4` · Nit]**

`let session = catch_unwind(...)` binds a `Result`, so `match session { Ok(()) => ... }` reads as matching a session object. Rename to `session_result`.

fynd-core/src/worker_pool/registry.rs
```diff
@@ -365,6 +443,108 @@ mod tests {
         let _ = shutdown_tx.send(());
     }
 
+    /// Amount marking the order whose solve panics in [`PanicOnPoisonAlgorithm`].
+    const POISON_AMOUNT: u64 = 666;
+
+    /// Algorithm that panics while solving the poison order and returns an error
+    /// otherwise. Used to verify that a panicking task does not kill the worker.
```
brunoguerios: **[DOCS `#6` · Nit]**

The mechanism is kill-and-respawn — the log at line 244 itself says "respawning worker" — so "does not kill the worker" understates what happens. Say "does not permanently lose the worker" or "the worker respawns".

fynd-core/src/worker_pool/registry.rs
brunoguerios: **[DOCS `#7` · Nit]**

`poison_tx`/`poison_rx` are named after the order but `ok_tx`/`ok_rx` after the outcome, while the order itself is `normal_order`. Rename to `normal_tx`/`normal_rx` to match the `poison_*`/`normal_*` pairing.

PR #351, opened 2026-07-24
feat(rpc): break down quote-failure reasons into namespaced causes

brunoguerios:
## What

Every failed order in the `quote_failure` log now carries a namespaced `group/reason` slug (`graph/no_graph_path`, `request/max_gas_exceeded`, `infra/queue_full`, …) instead of an empty `""` for anything that wasn't a graph-level path reason. Whole-request errors log `http/<code>` into the same column. This kills the misleading single empty legend on the Grafana dashboards.

Groups: `graph/*` (topology + depth, incl. `insufficient_liquidity`), `data/*` (missing/stale inputs), `infra/*` (timeout, queue, internal), `guard/price_check_failed` (found quote rejected vs external price), `request/max_gas_exceeded` (caller's gas cap), `http/*` (whole-request solver errors).

## How

- New `SolveError` variants: `MaxGasExceeded`, `MissingData`, `SimulationFailed`.
- Worker classifies algorithm errors into specific `SolveError`s instead of flattening to `AlgorithmError(String)`.
- `OrderQuote` carries `no_route_cause: Option<SolveError>`; the router aggregates failed pools by informativeness tier. `no_route_reason()` kept as a compat shim.
- RPC log column renamed `no_route_reasons` → `failure_reasons` with the slug vocabulary.

## Notes

- **Log-only** — no change to HTTP responses or `QuoteStatus`. CI green (`nextest --workspace` 1093/1093).
- `worker_router_solver_failures_total{error_type}` shifts some counts from `other` to `no_route`.
- Four `SolveError` variants that can't reach the per-order log today map to `"unknown"` by design (deferred, not speculatively slugged).

### DIFF
--- a/fynd-core/src/types/quote.rs
+++ b/fynd-core/src/types/quote.rs
@@ -28,7 +28,7 @@ use tycho_simulation::tycho_common::{
 };
 use uuid::Uuid;
 
-use super::primitives::ComponentId;
+use super::{internal::SolveError, primitives::ComponentId};
 use crate::{
     algorithm::NoPathReason, feed::market_data::StateLabel, price_guard::config::PriceGuardConfig,
     AlgorithmError,
@@ -791,9 +791,10 @@ pub struct OrderQuote {
     /// The state overlay this quote was computed against.
     /// When no overlay was requested this is the block number of the base state at solve time.
     solved_against: StateLabel,
-    /// Why no route was found (internal use only; only set for `NoRouteFound`).
+    /// Why this order failed (internal use only; set by the router fallback,
+    /// skipped during serialization).
     #[serde(skip)]
-    no_route_reason: Option<NoPathReason>,
+    no_route_cause: Option<SolveError>,
     /// Order-level surplus summary, populated when this quote executes through an exclusive
     /// pool.
     ///
@@ -836,7 +837,7 @@ impl OrderQuote {
             sender,
             receiver,
             solved_against,
-            no_route_reason: None,
+            no_route_cause: None,
             surplus: None,
         }
     }
@@ -1004,14 +1005,23 @@ impl OrderQuote {
         &self.solved_against
     }
 
-    /// Records why no route was found. Internal; skipped during serialization.
-    pub(crate) fn set_no_route_reason(&mut self, reason: Option<NoPathReason>) {
-        self.no_route_reason = reason;
+    /// Records why this order failed. Internal; skipped during serialization.
+    pub(crate) fn set_no_route_cause(&mut self, cause: Option<SolveError>) {
+        self.no_route_cause = cause;
     }
 
-    /// Returns the recorded no-route reason, if any.
+    /// Returns the recorded failure cause, if any.
+    pub fn no_route_cause(&self) -> Option<&SolveError> {
+        self.no_route_cause.as_ref()
+    }
+
+    /// Returns the no-route path reason, when the failure cause was a
+    /// route-finding failure that reported one.
     pub fn no_route_reason(&self) -> Option<NoPathReason> {
-        self.no_route_reason
+        match self.no_route_cause.as_ref() {
+            Some(SolveError::NoRouteFound { reason, .. }) => *reason,
+            _ => None,
+        }
     }
 }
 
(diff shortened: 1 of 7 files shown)


fynd-rpc/src/api/request_capture.rs
```diff
@@ -156,62 +172,35 @@ impl RequestOutcome {
...
-        RequestOutcome::Failed { code } => info!(
-            event = "quote_failure",
-            num_orders,
-            outcome = *code,
+            failure_reasons = ?failure_reasons,
             request = %request_json,
             "quote failure captured"
         ),
+        RequestOutcome::Failed { code } => {
+            let failure_reasons = [format!("http/{}", code.to_lowercase())];
```
tvinagre: I'd fix this one before merge: `failure_reasons` now means two incompatible things under the same key in the same `event="quote_failure"` stream. In the `Solved` arm it's documented as one-slug-per-order, index-aligned with `order_statuses` (length == num_orders). Here in the `Failed` arm it's a fixed one-element array with no `order_statuses` sibling — and this PR is what introduces the key into this branch for the first time. A Loki query written against the Solved-side contract (unpack per order, count weighted by order) silently gets wrong numbers when a Failed event flows through it — the exact class of quiet dashboard corruption this PR is eliminating.

Suggestion: name this field in the singular (`failure_reason = %slug`, not an array) so the two shapes can't be conflated by queries, or failing that, document on `log_request_capture` that the key's cardinality differs by outcome variant.

tvinagre (commented):
Reviewed at 4df050b9. The mechanics check out under a full trace: the log-only claim holds (the per-order worker mapping never reaches `solve_error_code` — worker failures terminate in the `rank_quotes` fallback, and the top-level `Err` path only carries empty-pool-list, price-guard-internal, and encoding errors), the four unslugged variants are genuinely unreachable per-order today, both new match sites are exhaustive without wildcards, and the touched suites plus clippy `-D warnings` are clean at head.

Three comments inline: one log-schema issue I'd fix before merge, one hardening suggestion, one cleanup. One more that doesn't anchor to the diff:

**The PR description's metrics note looks stale.** It says `worker_router_solver_failures_total{error_type}` "shifts some counts from `other` to `no_route`", but `NoPath → NoRouteFound → "no_route"` was already the mapping before this PR, and the newly classified variants (`InsufficientLiquidity`, `MissingData`, `SimulationFailed`) all still fall through `_ => "other"` in the `error_type` match. As far as I can tell no counts shift at all — worth correcting the description. Separately, that wildcard means the new variants are invisible in the metric; explicit arms would let each new classification show up under its own label (fine as a follow-up).

Feedback on the broader cause-carrier pattern (the `SolveError`-on-`OrderQuote` design and the slug vocabulary layout) will come separately — deliberately keeping this review to the mechanical items.

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -413,7 +413,11 @@ impl WorkerPoolRouter {
                 any_q.sender().clone(),
                 any_q.receiver().clone(),
                 any_q.solved_against().clone(),
-            )
+            );
+            // Workers only return Success quotes, so responses here mean every
+            // quote was filtered out by the request's max_gas.
+            fallback.set_no_route_cause(Some(SolveError::MaxGasExceeded));
```
tvinagre: The inference is correct today — I verified the comment's invariant: workers only construct `Success` quotes (`worker.rs:297`), the price guard runs after ranking, so max_gas really is the only filter that can empty `valid_quotes` while `responses.quotes` is non-empty.

But the cause is deduced by elimination two modules away from the filter that produces it, and nothing enforces the invariant. The day a worker returns a non-`Success` status inside `Ok(...)` (or any new filter lands in ranking), this branch starts stamping `request/max_gas_exceeded` onto unrelated failures — no compile error, no test failure, just a misleading dashboard sending someone to tune gas caps.

Suggestion: observe instead of deduce — derive the label from evidence in hand:

```rust
let max_gas_filtered = responses.quotes.iter()
    .any(|(_, q)| options.max_gas().is_some_and(|max| q.gas_estimate() > max));
fallback.set_no_route_cause(max_gas_filtered.then_some(SolveError::MaxGasExceeded));
```

Same behavior now, but future changes can't silently invalidate the label. A `debug_assert!` that all quotes here are `Success` would also pin the cross-module invariant.

fynd-core/src/types/quote.rs
```diff
@@ -923,14 +924,23 @@ impl OrderQuote {
...
 
-    /// Returns the recorded no-route reason, if any.
+    /// Returns the recorded failure cause, if any.
+    pub fn no_route_cause(&self) -> Option<&SolveError> {
+        self.no_route_cause.as_ref()
+    }
+
+    /// Returns the no-route path reason, when the failure cause was a
+    /// route-finding failure that reported one.
     pub fn no_route_reason(&self) -> Option<NoPathReason> {
```
tvinagre, line 1020: This shim has zero production callers at head — `handlers.rs` was migrated to `no_route_cause()`, and the only remaining call sites are four tests. Per the repo's replace-don't-deprecate convention I'd port those tests and delete it (it also drags the `NoPathReason` import into this module for no production purpose). One caveat: it's `pub` on fynd-core, so worth a quick check that no external consumer of the crate uses it before removing.

brunoguerios: Since we're planning on taking a deeper look into these logs next week, I'd avoid this breaking change of removing `no_route_reasons` for now., as there's a risk next week's refactor might break it again. What do you think?

fynd-rpc/src/api/request_capture.rs
brunoguerios: I'd like to keep the amounts of quote failure reasons relative to the amounts of orders, because each order can either succeed or fail for a different reason.
My suggested solution then is to replicate the request failure reason `num_orders` times.
What do you think?

fynd-core/src/worker_pool_router/mod.rs
brunoguerios: Agreed and applied here ✅

fynd-core/src/types/quote.rs
tvinagre: fine for me

tvinagre (approved):

PR #354, opened 2026-07-27
chore: update tycho crates to 0.341.6

zizou0x:
## What

Updates the tycho crates (`tycho-client`, `tycho-common`, `tycho-ethereum`, `tycho-execution`, `tycho-simulation`) from 0.340.0 to 0.341.6 in `Cargo.lock`.

## Notes on the socket2 diff

The re-resolution also switched `quinn`, `quinn-udp` and `hyper-util` from `socket2 0.6.4` to `socket2 0.5.10`. These crates declare a version range spanning both socket2 majors (`>=0.5.7, <0.7`), so the resolver is free to pick either side; a partial `cargo update` re-answered the choice. Both socket2 versions remain in the lockfile (tokio still uses 0.6.4) and the resolved configuration is within quinn-udp's supported range. No crate versions changed other than the tycho ones.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4055,7 +4055,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.4",
+ "socket2 0.5.10",
  "system-configuration",
  "tokio",
  "tower-service",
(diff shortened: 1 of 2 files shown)


tvinagre (approved):
shouldn't we also bump the `toml`?

PR #355, opened 2026-07-27
fix(hindsight): log terminal errors so stdout log pipelines show them

louise-poole:
The fmt subscriber writes to stdout, while a main that returns Err prints only to stderr. Collectors that follow stdout showed a run that stopped mid-startup with no reason.

Log the error through tracing before returning it. Returning it too keeps the exit code and the stderr fallback for the subcommands that install no subscriber.


tamaralipows (approved):
Thank you!

Troshchk (approved):
Thank you!

PR #359, opened 2026-07-27
fix(hindsight): New client follow-ups 

tamaralipows:
Two things: a min-notional floor on the Grafana metrics, and the review comments left on #324.

## Metrics floor

Sub-$100-notional trades were swamping the unweighted win-rate and bps quantiles — a few wei of rounding on a dust trade is tens to thousands of bps, so the win median and win-rate reflected noise, not real trades. A `MIN_NOTIONAL_USD = 100.0` floor now gates `hindsight_trades_total` and `hindsight_savings_bps`. The USD-weighted histograms (`volume_usd`/`savings_usd`/`improvement_usd`) still count every trade, so total savings and volume stay complete. A trade whose notional cannot be priced is kept — we do not drop what we cannot measure.

Applies uniformly to ethereum and base; no config, no helm change.

## PR #324 review comments

- **rainbow.rs** — decode the fee with a `sol!` function definition + `abi_decode` instead of a hand-rolled selector constant and word slicing.
- **rabby.rs / venue attribution** — fold the duplicated "record an output-token fee and gross it back into `amount_out`" into `TraderFlow::gross_output_fee`, so the double-count guard lives in one place.
- **monitor.rs** — `default_lag_blocks` now takes the `Chain` enum instead of a chain-name string, and reads the block time from `Chain::try_block_time_secs` instead of a local per-chain table; the doc comment states the threshold is a ~20-minute wall-clock budget expressed as a block count. Chains other than ethereum and base now get their real block time, so their default threshold changes (unichain: 100 → 1200 blocks).
- **clients → venue** — order-flow front-ends are the "venue" tier in the decoder's terminology, not a separate "client" concept. Renamed the module (`clients.rs` → `venue_attribution.rs`), its registry maps/methods/address-book sections (`client_*` → `venue_*`), and `intents::client_tag` → `venue_tag`.

### DIFF
--- a/tools/hindsight/src/resolve/monitor.rs
+++ b/tools/hindsight/src/resolve/monitor.rs
@@ -56,6 +56,9 @@ const POLL_INTERVAL: Duration = Duration::from_millis(50);
 const DECODE_RPC_LAG_RETRIES: usize = 5;
 const DECODE_RPC_LAG_BACKOFF: Duration = Duration::from_millis(1500);
 
+/// Wall-clock budget behind chain head that `default_lag_blocks` converts into a block count.
+const LAG_BUDGET_SECS: u64 = 20 * 60;
+
 /// Inputs for the live monitor — the `monitor` subcommand's CLI arguments, used directly as its
 /// configuration.
 #[derive(clap::Args)]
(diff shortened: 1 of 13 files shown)


Troshchk (approved):
Thank you @tamaralipows!
Looks good to me ✨

tools/hindsight/src/resolve/monitor.rs
```diff
@@ -317,12 +317,13 @@ async fn build_solver(
...
 
-/// Default chain-head lag threshold in blocks — about 20 minutes at the chain's block time, so the
-/// wall-clock tolerance stays roughly the same across chains. Used when `--max-lag-blocks` is not
-/// given; an unrecognised chain falls back to the 12-second-block assumption.
-fn default_lag_blocks(chain: &str) -> u64 {
-    let block_secs = match chain.to_lowercase().as_str() {
-        "base" => 2,
+/// The default `--max-lag-blocks`: a ~20-minute wall-clock budget for how far behind chain head the
+/// monitor may fall before rebuilding, expressed as a block count at the chain's block time so the
+/// budget stays about the same wall-clock length on every chain. `Chain` carries no block time, so
```
louise-poole: Chain does though? See [here](https://github.com/propeller-heads/tycho/blob/1e1e4cdec24ded088248eb0f075814992d109b39/crates/tycho-common/src/models/mod.rs#L483).

louise-poole (commented):
Just one last thing, otherwise looks great! Thanks for going back to address these for me 🙏

PR #360, opened 2026-07-27
fix(hindsight): cut Base block lag and keep partially-traced blocks

tamaralipows:
The staging Base monitor runs 4-6s behind chain head. Not solve cost — block
processing is 0.49s p95 and the pod uses 323m CPU with no limit. The lag is in
fetch-and-decode, which the existing metric starts after.

- The receipts-RPC wait backed off in fixed 1500ms steps, sized for Ethereum's
  12s blocks; on Base it fired on 71% of blocks. It now polls every 250ms against
  a budget in blocks, shared with `--max-lag-blocks` via one `block_time` helper.

- A single transaction the RPC could not trace failed its entire block, skipping
  every trade in it — 254 blocks on staging, ~5% of the run, which biases the
  win-rate and savings aggregates. Now only that transaction is dropped, counted
  by `hindsight_untraced_transactions_total`.

  To be clear about what did **not** change: Blocks are still dropped when the 
  monitor is behind — `--max-lag-blocks` still ends the session and rebuilds, and a
  block whose receipts never arrive still fails into `hindsight_skipped_blocks_total`. 
  The surviving trades in a partly-traced block are unaffected by the missing one:
  they have their own receipts and traces, sandwich detection reads the receipts
  slice rather than traces, and the solver is stepped so block N is still solved
  against state N-1.

- Adds `hindsight_chain_head_lag_blocks` and `hindsight_rpc_index_wait_seconds`,
  bucketed so they render as histograms.

The Ethereum lag budget grows from 7.5s to 36s: waiting beats skipping, since a
skipped block is a permanent hole in the aggregates. Set
`DECODE_RPC_LAG_BUDGET_BLOCKS` to 1 to keep Ethereum near its old behaviour.

### DIFF
--- a/tools/hindsight/src/resolve/monitor.rs
+++ b/tools/hindsight/src/resolve/monitor.rs
@@ -286,6 +286,7 @@ async fn decode_block_when_available<P: Provider>(
         }
         tokio::time::sleep(DECODE_RPC_LAG_POLL).await;
     }
+    telemetry::record_rpc_index_wait(started.elapsed());
     decoder.decode_block(block).await
 }
 
@@ -566,6 +567,7 @@ async fn run_session<P: Provider>(
             .get_block_number()
             .await
         {
+            telemetry::record_head_lag_blocks(head.saturating_sub(target));
             if head.saturating_sub(target) > pacing.max_lag_blocks {
                 return SessionEnd::Unhealthy(format!(
                     "monitor is {} blocks behind head {head}; presuming an unhealthy session",
--- a/tools/hindsight/src/resolve/monitor.rs
+++ b/tools/hindsight/src/resolve/monitor.rs
@@ -26,7 +26,7 @@ use fynd_core::{
     BlockStepController, FyndBuilder, Solver,
 };
 use num_bigint::BigUint;
-use tracing::{info, warn};
+use tracing::{debug, info, warn};
 use tycho_simulation::tycho_common::models::{Address as CoreAddress, Chain};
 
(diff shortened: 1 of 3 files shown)


tools/hindsight/src/resolve/monitor.rs
```diff
@@ -241,31 +242,48 @@ fn biguint_to_u256(value: &BigUint) -> U256 {
...
-/// Tycho stream that drives `block`, so poll the RPC head until it reaches `block` (bounded retries
-/// with backoff) — that distinguishes a transient race from a real failure. A block still
-/// undecodable once the RPC has indexed it is a genuine error and surfaces to the caller.
+/// Tycho stream that drives `block`, so poll the RPC head until it reaches `block` or `budget`
+/// expires — that distinguishes a transient race from a real failure. A block still undecodable
+/// once the RPC has indexed it is a genuine error and surfaces to the caller.
 async fn decode_block_when_available<P: Provider>(
     decoder: &mut Decoder<P>,
     block: u64,
+    budget: Duration,
```
Troshchk, line 254: Should it be the number of blocks? Is it related to DECODE_RPC_LAG_BUDGET_BLOCKS?

Troshchk (approved):
Thank you!
I have only one question

tools/hindsight/src/resolve/monitor.rs
Troshchk: I saw it is recalced with the block_time

PR #362, opened 2026-07-28
feat(hindsight): attribute each route to its algorithm and protocols

tamaralipows:
Adds the two things the dashboard was missing: the route Fynd took, and which algorithm won it.

A solved re-solve state now carries the winning algorithm plus the route rendered as a readable path:

```
USDT -[uniswap_v2]-> DAI -[vm:balancer]-> WETH
```

Token symbols come from the solver's token table; a token with no entry falls back to a shortened address (`0xababab…`). Protocol ids are Tycho's own, so a new DEX integration reads correctly with no lookup table.

A **split** fans legs out of one token and so cannot share a single chain — each leg becomes its own path joined by ` + `, carrying its share of the input:

```
USDC -[uniswap_v3 60%]-> WETH + USDC -[vm:curve 40%]-> WETH
USDT -[uniswap_v3 25%]-> WETH + USDT -[vm:curve 75%]-> DAI -[uniswap_v2]-> USDC
```

Surfaced three ways:

- **Prometheus**: `algorithm` label on `hindsight_trades_total`, `savings_bps`, `savings_usd`, `improvement_usd` (`"none"` when unsolved). Split by venue for the per-client comparison. The path is not a label — it is per-trade and would explode cardinality.
- **Loki**: `algorithm` and `route` on the existing `trade comparison` line. `route` is **last** on purpose: its value contains spaces, so a LogQL regexp can only bound it by end-of-line.
- **JSONL**: `algorithm` and `route` as top-level fields on both the `top` and `back` state objects, alongside the nested per-hop route that keeps the pools and amounts.

Dashboard panels: propeller-heads/terraform-infrastructure#218 (merge this first).

`monitor` attributes the winning pool, so the algorithm comparison needs a `--worker-pools-config` with more than one pool. The default `worker_pools.toml` has one, so every trade would read `algorithm=bellman_ford`.

### DIFF
--- a/tools/hindsight/CLAUDE.md
+++ b/tools/hindsight/CLAUDE.md
@@ -115,6 +115,23 @@ savings aggregates; unsolved states keep their coverage verdicts).
   like-for-like. Carries `sandwich` evidence when a bracket pair was found.
 - `RangeComparison` — a trade re-solved at both block states, including gas-netted settled output.
 - `Outcome` — `Solved`, `Partial`, or `Unsolvable`.
+- `RouteSummary` — which algorithm won a solved state and which protocols its route traded
+  through. Carried on `SolvedAmount` as typed fields (not dug out of the serialized quote) so the
+  metrics and the per-trade log line read it directly.
+
+### Route attribution
+
+A solved state records the algorithm whose route won the quote — the worker pool that beat the
+others on that order — and the route's protocol mix. It surfaces three ways:
+
+- **Prometheus**: an `algorithm` label on `hindsight_trades_total`, `hindsight_savings_bps`,
+  `hindsight_savings_usd`, and `hindsight_improvement_usd`. Split any of them by venue to see
+  which algorithm serves that venue's flow best. Unsolved states carry `algorithm="none"`.
+  Protocols are deliberately *not* a label: a route's protocol set is combinatorial and would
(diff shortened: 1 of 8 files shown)


tools/hindsight/CLAUDE.md
```diff
@@ -95,6 +95,44 @@ savings aggregates; unsolved states keep their coverage verdicts).
...
+  `SolvedAmount` as typed fields (not dug out of the serialized quote) so the metrics and the
+  per-trade log line read it directly.
+
+### Route attribution
+
+A solved state records the algorithm whose route won the quote — the worker pool that beat the
+others on that order — and that route rendered as a readable path:
+
+```
+USDT -(uniswap_v2)-> DAI -(vm:balancer)-> WETH
```
louise-poole: Lol - I'm going insane. I was getting very confused as to why some hops had connecting symbols with `-(` and others had with `)->`. Took me literally shaking my head and looking again to realise the protocol is 'written' inside the arrow.

I wonder if maybe square brackets would help with that? Not sure - I leave it to you to decide!

`USDT -[uniswap_v2]-> DAI -[vm:balancer]-> WETH`

louise-poole (approved):
Looks good!!! Thanks! 🙏

PR #366, opened 2026-07-29
refactor!: rename liquidity pools to components

Troshchk:
Rename all liquidity-pool naming to "component" (matching Tycho's ProtocolComponent and the component_id wire field); "pool" alone now always means a worker pool -- docs, comments, tests, and internal identifiers only, with no behavior change.

 Breaking changes:

  - fynd-core public API: DerivedData pool-depth accessors are now component_depths() / component_depths_block() / set_component_depths() / clear_component_depths(); PoolDepths / PoolDepthKey are now ComponentDepths / ComponentDepthKey; PoolDepthComputation is now ComponentDepthComputation; MarketState::pool_counts_by_protocol() is now component_counts_by_protocol().
  - GET /v1/prices (experimental feature): response fields pool_depths → component_depths, both the top-level array and the entry in blocks. Request parameters (include=depths) are unchanged.
  - TypeScript client: Swap.poolId → Swap.componentId. The quote wire format is unchanged.
- PoolConfig -> WorkerPoolConfig
- WorkerPoolConfig (spawn-level) -> WorkerPoolSpawnConfig
- FyndBuilder::add_pool -> add_worker_pool
- SolverBuildError::NoPools -> NoWorkerPools
- defaults::{POOL_TASK_QUEUE_CAPACITY, POOL_MIN_HOPS, POOL_MAX_HOPS, POOL_TIMEOUT_MS} -> WORKER_POOL_*

Not changed (intentionally, to avoid breaking consumers): the quote API wire format, Prometheus metric names (market_pools, market_pools_per_protocol), and the pool_depths computation ID used as the computation metric label.

### DIFF
--- a/.claude/CODEBASE.md
+++ b/.claude/CODEBASE.md
@@ -65,7 +65,7 @@ See `docs/ARCHITECTURE.md` for the full architecture diagram and detailed compon
    `WaterFillAlgorithm`
 5. **MarketState** (`fynd-core/src/feed/market_data.rs`) — `Arc<RwLock<>>` of all component/token/gas state; accessed via `MarketData` handle
 6. **TychoFeed** (`fynd-core/src/feed/tycho_feed.rs`) — Background task: Tycho WebSocket → MarketState → broadcast events
-7. **Derived Data** (`fynd-core/src/derived/`) — Pre-computed spot prices, component depths, token gas prices
+7. **Derived Data** (`fynd-core/src/derived/`) — Pre-computed spot prices, component (pool) depths, token gas prices
 8. **Encoding** (`fynd-core/src/encoding/`) — Encodes solved routes into on-chain transactions via `TychoEncoder`
 9. **Graph** (`fynd-core/src/graph/`) — `GraphManager` trait + `PetgraphStableDiGraphManager` implementation
 
@@ -76,7 +76,7 @@ See `docs/ARCHITECTURE.md` for the full architecture diagram and detailed compon
 2. Writes new component/token/state data into `MarketState` (write lock)
 3. Broadcasts `MarketEvent` → each `SolverWorker` updates its local graph via `GraphManager`
 4. `GasPriceFetcher` runs independently on a timer → fetches gas price from RPC node → writes to `MarketState`
-5. Triggers `ComputationManager` → runs spot prices → component depths → token gas prices (in dependency order) → broadcasts `DerivedDataEvent` → workers update edge weights
+5. Triggers `ComputationManager` → runs spot prices → component (pool) depths → token gas prices (in dependency order) → broadcasts `DerivedDataEvent` → workers update edge weights
 
 **Quote request path** (`POST /v1/quote`):
 1. `RouterApi` validates the request
--- a/.claude/CODEBASE.md
+++ b/.claude/CODEBASE.md
@@ -7,13 +7,13 @@ multiple DeFi protocols in real-time.
 ## What is Fynd
 
 Fynd is a solver that indexes live DEX liquidity via Tycho's streaming API, maintains an in-memory
-graph of token pairs and pools, and runs pluggable routing algorithms on dedicated OS threads to
+graph of token pairs and components (liquidity pools), and runs pluggable routing algorithms on dedicated OS threads to
 find optimal swap paths. It exposes an HTTP RPC for quote requests and returns the best
 gas-aware solution with optional on-chain transaction encoding.
 
 Key properties:
 - **Multi-protocol**: Routes through any on-chain protocol supported by Tycho, plus RFQ protocols
-- **Real-time**: Tycho Stream keeps all pool states synchronized every block
+- **Real-time**: Tycho Stream keeps all component states synchronized every block
 - **Multi-algorithm competition**: Multiple worker pools compete in parallel; best result wins
 - **Gas-aware**: Best solution selected by net output after gas costs
 - **Extensible**: Implement the `Algorithm` trait to add new routing strategies
@@ -63,9 +63,9 @@ See `docs/ARCHITECTURE.md` for the full architecture diagram and detailed compon
 4. **Algorithm trait** (`fynd-core/src/algorithm/`) — Pluggable route-finding; built-in:
    `MostLiquidAlgorithm`, `BellmanFordAlgorithm`, `PathFrankWolfeAlgorithm`, and
    `WaterFillAlgorithm`
-5. **MarketState** (`fynd-core/src/feed/market_data.rs`) — `Arc<RwLock<>>` of all pool/token/gas state; accessed via `MarketData` handle
+5. **MarketState** (`fynd-core/src/feed/market_data.rs`) — `Arc<RwLock<>>` of all component/token/gas state; accessed via `MarketData` handle
 6. **TychoFeed** (`fynd-core/src/feed/tycho_feed.rs`) — Background task: Tycho WebSocket → MarketState → broadcast events
-7. **Derived Data** (`fynd-core/src/derived/`) — Pre-computed spot prices, pool depths, token gas prices
+7. **Derived Data** (`fynd-core/src/derived/`) — Pre-computed spot prices, component depths, token gas prices
 8. **Encoding** (`fynd-core/src/encoding/`) — Encodes solved routes into on-chain transactions via `TychoEncoder`
 9. **Graph** (`fynd-core/src/graph/`) — `GraphManager` trait + `PetgraphStableDiGraphManager` implementation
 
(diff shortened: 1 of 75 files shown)


fynd-rpc/src/api/prices.rs
```diff
@@ -71,9 +71,9 @@ pub struct ComputationBlocks {
     /// Block at which spot prices were computed. `None` if not yet available.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub spot_prices: Option<u64>,
-    /// Block at which pool depths were computed. `None` if not yet available.
+    /// Block at which component depths were computed. `None` if not yet available.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub pool_depths: Option<u64>,
+    pub component_depths: Option<u64>,
```
louise-poole, line 76: Since the API is now hosted and public facing we instantly break users systems if we deploy a change like this. If we want to do a breaking change on the API we then need to introduce a new version of the API (v2) and host both for a while until everyone has switched over to the new one.
That said, I don't think this change is important enough to warrant a v2. Can you revert these API interface changes and rather add a note somewhere recommending the name change when/if we ever make a v2?

.claude/CODEBASE.md
```diff
@@ -63,9 +63,9 @@ See `docs/ARCHITECTURE.md` for the full architecture diagram and detailed compon
 4. **Algorithm trait** (`fynd-core/src/algorithm/`) — Pluggable route-finding; built-in:
    `MostLiquidAlgorithm`, `BellmanFordAlgorithm`, `PathFrankWolfeAlgorithm`, and
    `WaterFillAlgorithm`
-5. **MarketState** (`fynd-core/src/feed/market_data.rs`) — `Arc<RwLock<>>` of all pool/token/gas state; accessed via `MarketData` handle
+5. **MarketState** (`fynd-core/src/feed/market_data.rs`) — `Arc<RwLock<>>` of all component/token/gas state; accessed via `MarketData` handle
 6. **TychoFeed** (`fynd-core/src/feed/tycho_feed.rs`) — Background task: Tycho WebSocket → MarketState → broadcast events
-7. **Derived Data** (`fynd-core/src/derived/`) — Pre-computed spot prices, pool depths, token gas prices
+7. **Derived Data** (`fynd-core/src/derived/`) — Pre-computed spot prices, component depths, token gas prices
```
louise-poole: This one sounds a little odd just simply because we usually refer to depth with the term 'pool'. Maybe for ease of reading:
```suggestion
7. **Derived Data** (`fynd-core/src/derived/`) — Pre-computed spot prices, component (pool) depths, token gas prices
```

.claude/CODEBASE.md
```diff
@@ -76,7 +76,7 @@ See `docs/ARCHITECTURE.md` for the full architecture diagram and detailed compon
 2. Writes new component/token/state data into `MarketState` (write lock)
 3. Broadcasts `MarketEvent` → each `SolverWorker` updates its local graph via `GraphManager`
 4. `GasPriceFetcher` runs independently on a timer → fetches gas price from RPC node → writes to `MarketState`
-5. Triggers `ComputationManager` → runs spot prices → pool depths → token gas prices (in dependency order) → broadcasts `DerivedDataEvent` → workers update edge weights
+5. Triggers `ComputationManager` → runs spot prices → component depths → token gas prices (in dependency order) → broadcasts `DerivedDataEvent` → workers update edge weights
```
louise-poole: Same here

.claude/CODEBASE.md
```diff
@@ -91,7 +91,7 @@ See `docs/ARCHITECTURE.md` for the full architecture diagram and detailed compon
 
 - **Actix/Tokio runtime** (async I/O): HTTP server, TychoFeed, WorkerPoolRouter, gas fetcher, ComputationManager
 - **Worker pools** (dedicated OS threads): Each `SolverWorker` has a local graph and single-thread tokio runtime
-- **Communication**: `async_channel` (pool queues), `oneshot` (responses), `broadcast` (events), `Arc<RwLock<>>` (shared data)
+- **Communication**: `async_channel` (worker pool queues), `oneshot` (responses), `broadcast` (events), `Arc<RwLock<>>` (shared data)
```
louise-poole, line 94: ❤️

clients/rust/src/types.rs
```diff
@@ -613,7 +613,7 @@ pub enum BackendKind {
 pub enum QuoteStatus {
     /// A valid route was found and `route`, `amount_out`, and `gas_estimate` are populated.
     Success,
-    /// No swap path exists between the requested token pair on any available pool.
+    /// No swap path exists between the requested token pair on any available component.
```
louise-poole: This is a public interface so needs to make sense to people not used to our internal terminology too:
```suggestion
    /// No swap path exists between the requested token pair on available swap components (pools).
```

docs/guides/custom-algorithm.md
```diff
@@ -123,9 +123,10 @@ impl Algorithm for DirectPoolAlgorithm {
             // Validate every candidate route before returning it. The solver worker rejects
             // invalid routes (disconnected swaps, repeated tokens, malformed splits) and that
             // failure drops the whole solution for this worker pool. Skipping invalid candidates
-            // here lets a later pool be chosen instead. Any custom algorithm should validate the
-            // routes it might return, and — when it ranks multiple candidates — fall through to
-            // the next-best valid one rather than returning the invalid route.
+            // here lets a later component be chosen instead. Any custom algorithm should validate
```
louise-poole, line 126:
```suggestion
            // here lets a later worker pool be chosen instead. Any custom algorithm should validate
```

fynd-core/examples/custom_algorithm.rs
```diff
@@ -137,9 +137,10 @@ impl Algorithm for DirectPoolAlgorithm {
             // Validate every candidate route before returning it. The solver worker rejects
             // invalid routes (disconnected swaps, repeated tokens, malformed splits) and that
             // failure drops the whole solution for this worker pool. Skipping invalid candidates
-            // here lets a later pool be chosen instead. Any custom algorithm should validate the
-            // routes it might return, and — when it ranks multiple candidates — fall through to
-            // the next-best valid one rather than returning the invalid route.
+            // here lets a later component be chosen instead. Any custom algorithm should validate
```
louise-poole, line 140:
```suggestion
            // here lets a later worker pool be chosen instead. Any custom algorithm should validate
```

fynd-core/src/derived/mod.rs
```diff
@@ -22,11 +22,11 @@
 //!                 SpotPriceComputation
 //!                    /           \
 //!                   v             v
-//!    PoolDepthComputation    TokenGasPriceComputation
+//!    ComponentDepthComputation    TokenGasPriceComputation
```
louise-poole: Can you shift `SpotPriceComputation` and the arrows to line up better with the longer name here?

fynd-rpc/src/api/prices.rs
louise-poole: Thinking on this further - this is gated behind a `experimental` flag and I don't think it's exposed on our hosted API right now. We could get away with changing this.

docs/guides/custom-algorithm.md
Troshchk: Here, component is intended. Skipping lets the loop continue and return a route on a later component

fynd-core/examples/custom_algorithm.rs
Troshchk: same as https://github.com/propeller-heads/fynd/pull/366#discussion_r3673487761

fynd-rpc/src/api/prices.rs
Troshchk: Ok, should I also double check with cairoline?

louise-poole: I checked - we're not using it on the hosted API. But yes, maybe check with them if they know of any users using this endpoint on self-hosted Fynd?

fynd-core/src/worker_pool/pool.rs
```diff
@@ -34,7 +34,7 @@ use crate::{
 
 /// Configuration for the worker pool.
 #[derive(Debug)]
-pub struct WorkerPoolConfig {
+pub struct WorkerPoolSpawnConfig {
```
louise-poole: This is a public interface that I know people actively use. Let's rather not rename it?

fynd-core/src/solver.rs
```diff
@@ -132,10 +132,10 @@ fn parse_connector_tokens(
     Ok(Some(set))
 }
 
-/// Configuration for one worker pool, used by [`FyndBuilder::add_pool`].
+/// Configuration for one worker pool, used by [`FyndBuilder::add_worker_pool`].
 #[must_use]
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct PoolConfig {
+pub struct WorkerPoolConfig {
```
louise-poole: Same here - let's keep this renaming to internals only? I don't want to add friction to someone updating their Fynd version over something as small as renaming.

fynd-core/src/solver.rs
```diff
@@ -672,38 +673,39 @@ impl FyndBuilder {
...
 
-    /// Adds a named worker pool using the given [`PoolConfig`].
+    /// Adds a named worker pool using the given [`WorkerPoolConfig`].
     ///
     /// # Errors
     ///
     /// Returns [`SolverBuildError::AlgorithmConfig`] if any address in `connector_tokens` is not
     /// valid hex.
-    pub fn add_pool(
+    pub fn add_worker_pool(
```
louise-poole: Here too

fynd-rpc/src/api/prices.rs
Troshchk: Checked, it's ok to rename

louise-poole (approved):
Looks good!

PR #370, opened 2026-07-29
fix(water-fill): route simulation calls through the panic guard

pistomat:
## Summary

`water_fill` was the only algorithm still calling raw `ProtocolSim::get_amount_out`, at three sites: path simulation (`simulate_step`), `allocation_commit`, and discovery's `simulate_edge`. A pool-math panic in its bulk fill loops (~10^4 sim calls per solve) unwound through the solver worker thread and killed it permanently — workers are plain OS threads whose join handles are only reaped at shutdown, so each panic silently shrank pool capacity until restart. This is exactly the failure mode #7d3a070f's `sim_guard` was introduced to contain; that commit converted `bellman_ford`, `most_liquid`, and `split_primitives` but missed `water_fill`.

## Fix

Route all three sites through `get_amount_out_guarded`. A contained panic now surfaces as `SimulationError` and takes the same `.ok()?` skip-the-pool path as any ordinary simulation failure.

## Testing

- New end-to-end regression test: two-pool market where one pool's math panics (`DivByZeroSim`); the solve must succeed via the healthy pool. Verified red/green: with the guard reverted the test dies on a real `ruint` divide-by-zero; with it, passes (the guard fires 12x per run, covering all three sites).
- Full `check.sh` battery green (fmt, clippy, doc, 1168 tests).

## Known follow-up (out of scope)

A panicking pool now logs one `warn!` per contained call — water_fill's chunked re-simulation can emit hundreds per order for a single bad pool. Worth a per-solve dead-path memo or per-pool rate limit in a follow-up.

### DIFF
--- a/fynd-core/src/algorithm/water_fill.rs
+++ b/fynd-core/src/algorithm/water_fill.rs
@@ -44,6 +44,7 @@ use tycho_simulation::{
 
 use super::{
     most_liquid::DepthAndPrice,
+    sim_guard::GuardedProtocolSim,
     split_primitives::{build_split_route, HopDescriptor, PathAllocation, SimulatedHop},
     Algorithm, AlgorithmConfig, MostLiquidAlgorithm, NoPathReason,
 };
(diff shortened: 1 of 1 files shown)


louise-poole (approved):
Thank you!

PR #371, opened 2026-07-29
feat(hindsight): decode and resolve reverted Relay swaps

kayibal:
## What

Stacked on #388 (base branch `feat/hindsight-calldata-decoding`). Now that the calldata-extraction
layer (`SwapIntent`, per-solver parsers, `RelayCalldata`) has been split out into #388, this PR's
scope is the revert measurement pipeline only: the hindsight monitor also captures Relay swaps
that reverted on-chain and measures whether a Fynd quote would have cleared the trader's floor.
For each reverted Relay swap it decodes the intended trade from the transaction's trace —
reverted transactions emit no logs, but their call frames retain full calldata, reusing the same
`SwapIntent` recovery the settled path uses — then solves it at top-of-block (N-1) and fresh at
back-of-block (N) and records whether each quote is at or above `min_amount_out`. Purpose:
quantify how many reverts a sequencer that fills against interim block state would avoid, and by
what margin.

A reverted trade is modeled as the same `DecodedTrade` a settled one is, distinguished by a
`status` field (`settled` or `reverted` with a cause) rather than a parallel set of types — a
revert is still a trade, just one that did not fill.

## Changes

- **Revert matching** (`decoder/matching.rs`): `matching::select` handles both settled and
  reverted receipts in one function, returning one `MatchedSolverTrade` with a `reverted` flag.
  Status-0 receipts whose `tx.to` is a Relay entry point are traced in the same bounded wave as
  settled trades (venue allowlist, currently `relay` only).
- **Unified trace walk and cause classification** (`decoder/trace.rs`): one
  `trace::find_solver_frame` — prefer the frame that settled the swap, and when nothing settled,
  fall back to the frame that tried. This also protects a settled trade whose router tried one
  solver (reverted) before succeeding via another: it is never mis-attributed to the abandoned
  attempt. `classify_revert_cause` asks each registered `SolverKnowledge` impl whether it
  recognizes a frame as a slippage floor (`is_slippage_floor` — Fly's `InsufficientAmountOut()`
  selector, KyberSwap's "Return amount is not enough" reason) rather than checking solver-specific
  constants itself; `trace.rs` keeps only the generic subtree walk, out-of-gas, and
  deepest-revert-reason logic. `RevertCause` is `slippage_floor`, `out_of_gas`, or `other`.
  Undecodable reverts are kept with unknown terms so parser coverage is measurable.
  `decode_block` returns one `Vec<DecodedTrade>` with mixed status.
- **Unified re-solve** (`resolve/`): `resolve_block_range` takes the one trade list. Trades with
  unknown terms are recorded but not solved; settled and decodable-reverted trades go through the
  same top-of-block/back-of-block wave. `RangeComparison` gains `status` (+ cause) and a
  `fillable`/`margin_bps` judgment per state, computed against `min_amount_out` for both settled
  and reverted trades whenever a floor is known — a settled trade cleared it by construction, but
  the margin is still informative.
- **JSONL**: everything writes to the one `comparisons-YYYY-MM-DD.jsonl` stream
  (`RotatingWriter` is no longer prefix-parameterized); a record's `status` field tells a settled
  trade from a reverted one. The `report` subcommand filters to `status == "settled"` explicitly
  before aggregating savings/win-rate, rather than relying on absent fields.
- **Prometheus**: `hindsight_reverted_swaps_total{venue,chain,cause,decoded}`,
  `hindsight_revert_fillable_total{venue,chain,state,fillable}`, and
  `hindsight_revert_margin_bps{state,…}` (signed histogram) — unchanged metric names, now driven
  off `status` instead of a separate revert type.

## Notes

- Revert causes are recorded, not filtered: only `slippage_floor` reverts are avoidable by
  fresher state, but the avoidance rate stays computable against either denominator offline.
- Cause classification scans the whole reverted subtree for slippage markers before falling back
  to the deepest frame's error — the informative reason can sit on an outer frame.

## Testing

- 311 tests pass in the crate: revert matching, the unified trace walk and cause classification
  against four real Base revert traces (Fly slippage, Kyber slippage, out-of-gas, transfer
  failure), the unified resolve wave, JSONL records (settled and reverted), report status
  filtering, telemetry rendering.
- `cargo check --workspace --all-features`, `cargo clippy --locked --workspace --all-targets
  --all-features -- -D warnings`, `cargo +nightly fmt --check`, and `RUSTDOCFLAGS="-D warnings"
  cargo doc --no-deps -p hindsight -p fynd-core -p fynd-rpc-types -p fynd-rpc -p fynd-client` all
  clean.

Pre-commit hook not used for the commits on the rebuilt branch: `cargo nextest run --workspace
--bin fynd` SIGABRTs on this machine on an unrelated `fynd-core` test (a `conda`/`cc` PATH issue
breaking Rust panic unwinding, environmental). Verified manually instead; see above.


tools/hindsight/src/decoder/matching.rs
```diff
@@ -19,6 +19,39 @@ pub(crate) struct MatchedSolverTrade<'a> {
...
+
+/// A transaction that reverted on-chain, matched only by its entry point being a
+/// revert-decoded venue. Every field is `Copy` (a reference, an `Address`, a `&'static str`), so
+/// callers that only read it can dereference a borrowed match instead of taking ownership.
+#[derive(Clone, Copy)]
+pub(crate) struct MatchedRevertedTrade<'a> {
+    pub receipt: &'a AnyTransactionReceipt,
+    pub entry_point: Address,
+    pub venue: &'static str,
+}
```
kayibal: Why do we need this? Couldn't we just add a reverted flag to the struct above? This struct basically is a duplication of the above struct.

tools/hindsight/src/decoder/mod.rs
```diff
@@ -161,6 +163,99 @@ fn intent_fields(intent: Option<&SwapIntent>) -> (Option<U256>, Option<U256>, Op
...
+    /// The registry name of the solver frame's `to`, or the entry-point label when no solver
+    /// frame was found in the reverted trace at all.
+    pub solver: String,
+    /// The trader's swap terms recovered from the solver frame's calldata. `None` when no solver
+    /// frame was found, its `to` has no `swap_intent` support, or the calldata did not parse —
+    /// kept (not filtered out) so parser coverage is measurable against every reverted candidate.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub intent: Option<SwapIntent>,
+    pub cause: RevertCause,
+}
```
kayibal: Can't we model this with DecodedTrade? Just because the trade did not fill, it doesn't mean it wasn't a trade.

tools/hindsight/src/decoder/trace.rs
```diff
@@ -90,7 +90,31 @@ pub(crate) fn find_solver_frame<'a>(
...
+/// solver frame either reverted itself (the router bubbled the failure up) or sits inside a
+/// reverted ancestor while having itself run clean before a later sibling exhausted gas — both
+/// shapes are seen live (see the trace fixtures). Since a revert emits no logs, this is the only
+/// way to find which router the trader was routed through and recover its calldata.
+pub(crate) fn find_reverted_solver_frame<'a>(
+    frame: &'a CallFrame,
+    registry: &Registry,
+) -> Option<&'a CallFrame> {
+    find_solver_frame_impl(frame, registry, true)
+}
```
kayibal: Why do we need a separate path here? The function could simply ignore wether the transaction failed or not.

tools/hindsight/src/decoder/trace.rs
```diff
@@ -101,7 +125,98 @@ pub(crate) fn find_solver_frame<'a>(
...
+        .find_map(|child| find_solver_frame_impl(child, registry, tolerate_reverts))
+}
+
+/// Fly's `InsufficientAmountOut()` custom error selector (`DexAggregator.sol`) — the marker for a
+/// slippage-floor revert on Fly.
+const FLY_INSUFFICIENT_AMOUNT_OUT: [u8; 4] = [0xe5, 0x29, 0x70, 0xaa];
+/// `KyberSwap`'s slippage-floor revert reason.
+const KYBERSWAP_INSUFFICIENT_RETURN: &str = "Return amount is not enough";
+/// The geth call tracer's exact error string for a frame that ran out of gas.
+const OUT_OF_GAS: &str = "out of gas";
```
kayibal: Solver specific data in general purpose module, this is a red flag.

PR #372, opened 2026-07-29
feat: split the exclusive route improvement with the user

tamaralipows:
An exclusive-access route is quoted when it at least matches the public reference net of gas. Whatever it produces on top is split: the user keeps a share, the protocol captures the rest.

```
improvement  = exclusive_net - public_net     (must be >= 0)
margin       = improvement * user_share_bps / 10_000   (rounded up)
required_net = public_net + margin
committed    = max(public_amount_out, required_net + exclusive_gas)
surplus      = exclusive_amount_out - committed
```

The share comes from `EXCLUSIVE_ROUTE_USER_SHARE_BPS`, default 1_000 (a tenth). A share of 0 leaves the user at the public net; 10_000 hands over the whole improvement and captures nothing. A route that ties the public reference is quoted at the public price, with nothing to capture.

Because the margin scales with the improvement rather than the trade, the effective mark-up drops below a basis point when the improvement is small.

### DIFF
--- a/fynd-core/src/worker_pool_router/mod.rs
+++ b/fynd-core/src/worker_pool_router/mod.rs
@@ -25,6 +25,7 @@ pub mod config;
 
 use std::{
     collections::{HashMap, HashSet},
+    sync::LazyLock,
     time::{Duration, Instant},
 };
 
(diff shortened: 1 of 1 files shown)


fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -47,6 +47,14 @@ use crate::{
     SolveParams, SurplusInfo,
 };
 
+/// Minimum improvement over the public route, in basis points of the public route's output net of
+/// gas, that an exclusive-access route must deliver before it is quoted.
+///
+/// The margin goes to the user: the commitment is raised by it, so the protocol only captures what
+/// the route produces on top. A candidate that beats the public route by exactly this margin is
+/// still quoted, with zero surplus.
+const EXCLUSIVE_ROUTE_IMPROVEMENT_BPS: u64 = 1;
```
louise-poole: Just a thought - do you think play with sub-bps values here would be useful? Might help us with fine-tuning edging out the competition.

louise-poole: Side thought: maybe the mark-up can be a fraction of the public vs exclusive amount out diff (the surplus diff)? That way we're only marking up the amount when we can afford to do so and still win. And we mark it up relative to how much surplus margin we have: so more surplus potential = better user price we offer = more likely to land onchain (everyone wins).

louise-poole: Maybe we can make this configurable via an env var? To make it easier to play with/update?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -712,8 +726,9 @@ fn combine_with_surplus(
         return public_ranked;
     };
 
-    // The candidate route must beat the committed reference net-of-gas.
-    if exclusive_candidate.amount_out_net_gas() <= committed.amount_out_net_gas() {
+    // The candidate route must beat the public reference net-of-gas by the improvement margin.
+    let required_net_amount_out = with_improvement_margin(committed.amount_out_net_gas());
+    if exclusive_candidate.amount_out_net_gas() < &required_net_amount_out {
```
louise-poole: What about the case where we set the improvement margin to 0? Then we need the `<=` check back no?

fynd-core/src/worker_pool_router/mod.rs
Troshchk: I like your idea with it being a fraction of surplus

Troshchk (commented):
Thank you @tamaralipows!
I agree with Loiuse's comments, nothing more to add

fynd-core/src/worker_pool_router/mod.rs
tamaralipows: A share of 0 doesn't need the `<=` check any more. The gate is now `exclusive_net < public_net → skip`: a route that ties the public reference is quoted at the public price with zero surplus, and a route that beats it is quoted with the user keeping a share of the difference (default 10%, set by `EXCLUSIVE_ROUTE_USER_SHARE_BPS`). At a share of 0 the user is left at the public net and we capture the whole difference.

fynd-core/src/worker_pool_router/mod.rs
tamaralipows: The sub bps i think inherently comes then i make it a fraction of the surplus since this amount will already be tiny relative to the trade. So I implemented just the second point :) lmk what you think

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -47,6 +48,59 @@ use crate::{
...
+/// Basis-point denominator: `10_000` bps is the whole improvement.
+const BPS_DENOMINATOR: u32 = 10_000;
+
+/// The configured user share, parsed from [`ENV_USER_IMPROVEMENT_SHARE_BPS`].
+static USER_IMPROVEMENT_SHARE_BPS: LazyLock<u32> = LazyLock::new(user_improvement_share_bps_env);
+
+/// Reads the user share from the environment, falling back to
+/// [`DEFAULT_USER_IMPROVEMENT_SHARE_BPS`] when the variable is unset or unusable.
+fn user_improvement_share_bps_env() -> u32 {
+    let Ok(raw) = std::env::var(ENV_USER_IMPROVEMENT_SHARE_BPS) else {
```
Troshchk, line 68: Why do we allow user to set it?

tamaralipows: This is for us to make it easier to change/update (we are the user here remember)

louise-poole (approved):
Nice! Looks good! 👌

PR #373, opened 2026-07-29
feat: quote exclusive routes when no public route exists

tamaralipows:
[ENG-6279](https://propeller-heads.atlassian.net/browse/ENG-6279).

When no public route is found, the exclusive candidate is quoted instead of dropped. With no reference price to match, the commitment is the route's output minus a default fee:

```
fee       = exclusive_leg_amount_out * NO_PUBLIC_ROUTE_FEE_BPS / 10_000   (rounded down)
committed = route_amount_out - fee
```

The fee is 5 bps, the rate Ekubo charges on its most traded ETH/USDC pool, so the quote stays competitive with the other solvers our API clients compare us against.

The fee base is the exclusive leg, not the whole quote. On a split route the public branches are already priced by the market, so there is nothing to take from them; charging the whole route could also make the fee exceed what the exclusive leg is able to withhold in `pin_commitment`.

The candidate is skipped if the commitment would not cover gas; the `NoRouteFound` placeholder stays on as the price-guard fallback.

`combine_with_surplus` is split into `best_exclusive_candidate`, `matched_commitment`, `default_fee_commitment` and `pin_commitment` so both commitment modes share the per-leg attribution.

[ENG-6279]: https://propeller-heads.atlassian.net/browse/ENG-6279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/fynd-core/src/worker_pool_router/mod.rs
+++ b/fynd-core/src/worker_pool_router/mod.rs
@@ -881,16 +881,16 @@ fn default_fee_commitment(exclusive_candidate: &OrderQuote) -> Option<BigUint> {
     Some(committed_amount_out)
 }
 
-/// Pins `candidate` to `committed_amount_out`: stamps the per-leg committed amounts the encoder
-/// consumes, pins `amount_out` and `amount_out_net_gas`, and attaches the [`SurplusInfo`].
+/// Pins `exclusive_candidate` to `committed_amount_out`: stamps the per-leg committed amounts the
+/// encoder consumes, pins `amount_out` and `amount_out_net_gas`, and attaches the [`SurplusInfo`].
 ///
 /// Everything the route produces above the commitment is what the exclusive components capture.
-fn pin_commitment(candidate: &OrderQuote, committed_amount_out: BigUint) -> OrderQuote {
-    let exclusive_route_amount_out = candidate.amount_out();
-    let exclusive_gas_cost = exclusive_route_amount_out - candidate.amount_out_net_gas();
+fn pin_commitment(exclusive_candidate: &OrderQuote, committed_amount_out: BigUint) -> OrderQuote {
+    let exclusive_route_amount_out = exclusive_candidate.amount_out();
+    let exclusive_gas_cost = exclusive_route_amount_out - exclusive_candidate.amount_out_net_gas();
     let surplus_amount = exclusive_route_amount_out - &committed_amount_out;
 
-    let mut surplus_quote = candidate.clone();
+    let mut surplus_quote = exclusive_candidate.clone();
 
     // Final output of each swap's path, walked backwards (a path's terminal output propagates
     // to its chained predecessors). Converting captured surplus into a leg's token needs the
--- a/fynd-core/src/worker_pool_router/mod.rs
+++ b/fynd-core/src/worker_pool_router/mod.rs
@@ -100,14 +100,12 @@ fn user_margin(improvement: &BigUint, user_share_bps: u32) -> BigUint {
     (improvement * BigUint::from(user_share_bps) + (&denominator - 1u32)) / denominator
 }
 
-/// Fee taken from an exclusive leg's output, in basis points, when the public worker pools returned
-/// no route within the solve timeout.
+/// The fee in basis points that the protocol takes from an exclusive leg. It applies when the
+/// public worker pools find no route in the solve timeout.
 ///
-/// Without a public quote there is no reference price to pin the commitment to, so the protocol
-/// takes this cut of the exclusive leg's output and the user receives the rest — a price instead
-/// of no price at all. Only the exclusive leg is charged: the public branches of a split route
-/// pay out in full. Set to the fee Ekubo charges on its most traded ETH/USDC pool, so a quote
-/// priced this way stays competitive with the other solvers our API clients compare us against.
+/// Without a public quote there is no reference price. The protocol takes this fee and the user
+/// gets the rest. The value is the fee of Ekubo's most used ETH/USDC pool. This keeps the quote
+/// competitive with the other solvers.
 const NO_PUBLIC_ROUTE_FEE_BPS: u32 = 5;
 
 /// Which liquidity a solver pool (a group of workers) routes through, and therefore what its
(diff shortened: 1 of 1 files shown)


fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -55,6 +55,15 @@ use crate::{
...
 const EXCLUSIVE_ROUTE_IMPROVEMENT_BPS: u64 = 1;
 
+/// Fee taken from an exclusive route's output, in basis points, when no public route exists to
+/// match against.
+///
+/// Without a public quote there is no reference price to pin the commitment to, so the protocol
+/// takes this cut of the route's output and the user receives the rest — a price instead of no
+/// price at all. Set high on purpose: it bounds what the LPs give away while the pool's price
+/// cannot be checked against the public market.
+const NO_PUBLIC_ROUTE_FEE_BPS: u64 = 100;
```
tamaralipows: I know we said we'd make this high - let me know if this is too high though, taking suggestions

tamaralipows: Actually claude says this is not high enough xD

louise-poole: Considering we're often still competing with other solvers here (users of our API like Relay compare out output to what other solvers give) we do not want to make this excessively high. Let's look at what ekubo charges on their normal pools and maybe set it to that to start?

louise-poole: Looks like Ekubo's most popular ETH/USDC pool uses a 5bps fee.

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -55,6 +55,15 @@ use crate::{
 /// still quoted, with zero surplus.
 const EXCLUSIVE_ROUTE_IMPROVEMENT_BPS: u64 = 1;
 
+/// Fee taken from an exclusive route's output, in basis points, when no public route exists to
```
louise-poole: Let's clear up the wording here a little: it's not that no public route exists... it's just that none was found within the timeout.

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -720,47 +766,63 @@ fn combine_with_surplus(
...
-    let required_net_amount_out = with_improvement_margin(committed.amount_out_net_gas());
-    if exclusive_candidate.amount_out_net_gas() < &required_net_amount_out {
-        return public_ranked;
+/// Returns the amount to commit against a successful public quote:
+/// `max(public_amount_out, required_net + gas)`.
+///
+/// Returns `None` when the candidate fails a gate — it must beat the public output net-of-gas by
+/// the improvement margin and produce at least the public output. `combine_with_surplus` documents
+/// why each bound is there.
+fn matched_commitment(public_reference: &OrderQuote, candidate: &OrderQuote) -> Option<BigUint> {
```
Troshchk: can we rename to exclusive_candidate for clarity?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -720,47 +766,63 @@ fn combine_with_surplus(
...
-    // Together with the net check above, these gates guarantee the route covers the committed
-    // amount.
-    let committed_amount_out =
-        (&required_net_amount_out + &exclusive_gas_cost).max(public_amount_out.clone());
+/// Returns the amount to commit when no public quote succeeded: the candidate's output minus
+/// `NO_PUBLIC_ROUTE_FEE_BPS`, rounded in the user's favour.
+///
+/// Returns `None` when the commitment would not cover the route's gas — quoting a route the user
+/// nets nothing on is worse than reporting no route.
+fn default_fee_commitment(candidate: &OrderQuote) -> Option<BigUint> {
```
Troshchk: Can we also change to exclusive_candidate?

fynd-core/src/worker_pool_router/mod.rs
Troshchk: I am thinking -- if we make liquidity_scope `all` by default, we might encounter case where no public route is found more often.
So may be we can measure/test an impact of this number somehow?

Troshchk (commented):
Thank you @tamaralipows!
Just a minor thing, otherwise looks good ✨

fynd-core/src/worker_pool_router/mod.rs
Troshchk: Saw you message about now running a test -- may be we can have a separate task to calibrate this value?

fynd-core/src/worker_pool_router/mod.rs
louise-poole: I think the fee should be calculated on the exclusive leg only, not on the whole quote's amount out (unless I misunderstood what 'candidate' means here?

louise-poole: This is important for split routes. Maybe we should have a split route case in one of the tests?

fynd-core/src/worker_pool_router/mod.rs
louise-poole: I'm noticing some naming inconsistencies I didn't notice before. Like sometimes we say 'surplus' and sometimes we say 'excess'. I'll make a note to do a clean up at the end (no action required on this PR).

fynd-core/src/worker_pool_router/mod.rs
tamaralipows: Agreed. Set to 5 bps for now to match Ekubo's ETH/USDC pool. We can remember to calibrate it against real traffic — hindsight can replay with the public pool disabled to measure how often we hit this path and what it costs in win rate.

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -775,34 +776,53 @@ fn best_exclusive_candidate<'a>(
...
+        .route()?
+        .swaps()
+        .iter()
+        .find(|swap| policy.is_exclusive(swap.protocol_component()))?
+        .amount_out();
+
+    let realized_amount_out = exclusive_candidate.amount_out();
+    let fee = exclusive_leg_amount_out * BigUint::from(NO_PUBLIC_ROUTE_FEE_BPS) /
+        BigUint::from(10_000u64);
     let committed_amount_out = realized_amount_out - fee;
```
Troshchk, line 870: Just a note: now we only allow having one exclusive leg and the exclusive leg to only be the last one, so it works fine now
In the future we will probably allow any leg/multiple legs to be exclusive, so we will have to change this logic

Troshchk: The `pin_commitment` is already written for the other cases -- you may refer to this, but we can also leave it as is and implement in v2

Troshchk (approved):

PR #374, opened 2026-07-29
feat(hindsight): decode the remaining Router V3 chains, and fix input-side venue fees

tamaralipows:
Adds embedded address books for **unichain, arbitrum, bsc and polygon**, resolved from one `BUILTIN_CHAINS` table instead of a per-chain constructor — `--chain` now takes all six. The second commit fixes a fee bug this surfaced, which also corrects ethereum and base.

Deploys via **propeller-heads/helm-configuration#840**, which needs a `versions.yml` bump to a `hindsight` tag containing this PR.

## Decoded on each chain

`decode` against live RPC on this branch. The window is ~5 minutes of **chain time**, not a fixed
block count: block times range from 0.25s to 1.5s here, so 50 blocks would be 12s on arbitrum and
75s on polygon and the counts would not be comparable.

| Chain | Block time | Blocks | Trades | Venues decoded |
|---|---|---|---|---|
| bsc | 0.45s | 666 | 1572 | pancakeswap 1116, okx 269, relay 69, uniswap 46, cow 18, robinhood 13, kyberswap 10, 1inch 4, fly 4, lifi 2, paraswap 1 |
| polygon | 1.50s | 200 | 184 | quickswap 128, relay 28, kyberswap 6, uniswap 6, okx 3, paraswap 1, fly 1, robinhood 1, lifi 1, phantom 1 |
| arbitrum | 0.25s | 1000 | 93 | uniswap 51, cow 22, camelot 10, okx 4, kyberswap 3, 1inch 1, relay 1 |
| unichain | 1.00s | 300 | 1 | fly 1 |

Relay, the one client live on all four, decodes on each at a rate that tracks its volume there. Its
absence from a short arbitrum window is sampling, not coverage: Relay took 106 fee legs per 20k
arbitrum blocks, so a few hundred blocks is expected to contain none.

Two counts worth reading as findings rather than throughput figures:

- **bsc settles 2.36 decodable trades per block**, the highest of the fleet, which is why
  propeller-heads/helm-configuration#840 does not deploy a bsc monitor yet — the monitor cannot
  re-solve that many inside a 0.45s block. The address book still ships, so `decode --chain bsc`
  works.
- **unichain is very thin.** This window found one trade; an earlier 50-block window found four, so
  the rate is somewhere between roughly 300 and 7000 trades a day and this sample cannot pin it
  down. Its monitor will be mostly idle.

## Clients live per chain

Established on-chain, not copied from the mainnet book. Fee legs counted only when paid by a settlement contract, so dust sprays are excluded.

| Client | unichain | arbitrum | bsc | polygon |
|---|---|---|---|---|
| Relay, Rabby, Robinhood | yes | yes | yes | yes |
| MetaMask, Rainbow, Jumper, Binance Wallet, Coinbase, LlamaSwap, Infinex | no | yes | yes | yes |
| Phantom | no | no | no | yes |

Each book's header records how every entry was verified and what was found absent — CoW emits no trades on unichain and LiFi is not deployed there, so that book has no batch settlers and no integrator tier.

Three clients came out of decoding LiFi's `integrator` string per chain: **Binance Wallet** (`_binancewallet`, 1043 swaps per 20k blocks on bsc — the largest client here), **Jumper** and **Coinbase's Base App**. All three use the existing integrator tier, no code.

Each chain's largest native DEX is registered as a solver — PancakeSwap, QuickSwap, Camelot — which also needs no code and is where most retail flow enters. Ranking unknown entry points found PancakeSwap's Universal Router 2 at 4% of all bsc swaps.

## Second commit: input-side venue fees

`fee_venue` only checked the output token, because the venues it was written for (Phantom, Robinhood) take the buy token. Coinbase's Base App skims 0.95% off the **sell** token before routing, so its fee was never backed out: `amount_in` stayed above what reached the pools, and Fynd re-solved on that inflated size looked like it won by the fee.

Both sides are now checked; output wins when a wallet received both tokens. Coinbase's fee wallet is registered on ethereum and base too — 2942 fee legs per 8k blocks on base, 943 on ethereum — plus the 0x Settler each chain routes that flow through.

Effect on the existing chains: Coinbase appears as a venue, and savings on that flow drop by about the fee. Nothing that decodes today decodes differently — the output side is still checked first, and every pre-existing test passes unchanged.

## Live monitor sessions

Short live sessions per chain (`monitor` against the beta indexer, with each chain's real worker
pools and `--min-tvl 10`) to check the whole path end to end, and to look for wins too large to be
real. Deliberately short — the cluster run is what will surface rare suspicious wins.

| Chain | Blocks | Records | Verdicts (top) | Largest win | USD accuracy |
|---|---|---|---|---|---|
| bsc | 60 | 151 | 108 unsolvable, 29 loss, 14 win | 526 bps | fixed, see below |
| polygon | 40 | 32 | 24 unsolvable, 7 loss, 1 win | ~0 bps | 1.0x |
| arbitrum | 40 | 3 | 2 unsolvable, 1 win | 90 bps | 1.0x |
| unichain | 40 | 0 | — | — | — |

USD accuracy is measured, not assumed: a stablecoin-out trade's true value follows from its amount,
so comparing the reported valuation against it gives an exact error factor.

**This found a real bug.** One bsc record claimed a 353 bps win worth **$177,223** on a trade whose
actual output was 613.77 USDT. The bps was right — it is a ratio, so decimals cancel — but the
dollar figure was inflated ~8000x. `Prices::usd_per_native_wei` averages every priced anchor
stablecoin, so an anchor the solver misprices scales every USD figure on the chain instead of being
outvoted:

| bsc anchors | USD error factor |
|---|---|
| USDT only | **1.0x** (exact, n=15) |
| USDT + USDC | 6,123x |
| USDT + USDC + BUSD | 8,174x |

Binance-Peg USDC and BUSD are thin on the DEXs indexed here next to USDT, which dominates BSC
stablecoin volume, so bsc now anchors on USDT alone. Arbitrum and polygon measure 1.0x with their
full anchor sets, so this is BSC token liquidity rather than the valuation code — but the averaging
has no plausibility guard on any chain, which is left as a follow-up.

Two things to keep an eye on in the cluster data rather than fix blind:

- **bsc had one win above 500 bps** (526), on an illiquid PancakeSwap token. Plausible on a thin
  pool, but it is the shape a decode fault takes.
- **`Unsolvable` dominates** — 108 of 151 on bsc, 94 of them `NoRouteFound`. Expected, since most
  bsc flow is long-tail PancakeSwap tokens Fynd does not index. 14 were `Timeout`, a direct
  consequence of re-solving on production's 100ms budget rather than 10s.

Throughput was measured in the same sessions: bsc re-solves a block in 41ms median (p90 136ms)
against a 450ms block, and ran end to end at 0.95x real time from a laptop over public RPC with no
skipped blocks. propeller-heads/helm-configuration#840 deploys all four.

## Verification

Every contract confirmed deployed on its chain; every token by its own `symbol()` and `decimals()`; OKX's concurrent router versions by the OKX proxies embedded in their bytecode; Coinbase's fee wallet binds only to `base-app` swaps. Relay, the one client on all four chains, decoded 21 trades over 200 polygon blocks — all through `relay-netting`, all stablecoin pairs within 11 bps of 1.0, no anomalies.

Local CI green (fmt, clippy, doc, 1171 tests). The 7 `fynd-core::integration` failures are the pre-existing Git-LFS fixture issue on this machine.

### DIFF
--- a/tools/hindsight/src/decoder/registry.rs
+++ b/tools/hindsight/src/decoder/registry.rs
@@ -15,6 +15,7 @@ use std::{
 use alloy::primitives::{Address, B256};
 use anyhow::Context;
 use serde::Deserialize;
+use tycho_simulation::tycho_common::models::Chain;
 
 /// The built-in address books, embedded at compile time (validated by tests, so they cannot fail
 /// to parse at runtime).
@@ -25,15 +26,29 @@ const ARBITRUM_TOML: &str = include_str!("registry/arbitrum.toml");
 const BSC_TOML: &str = include_str!("registry/bsc.toml");
 const POLYGON_TOML: &str = include_str!("registry/polygon.toml");
 
-/// The chains with a built-in address book, in the order `Registry::load` reports them.
-const BUILTIN_CHAINS: [(&str, &str); 6] = [
-    ("ethereum", ETHEREUM_TOML),
-    ("base", BASE_TOML),
-    ("unichain", UNICHAIN_TOML),
-    ("arbitrum", ARBITRUM_TOML),
-    ("bsc", BSC_TOML),
-    ("polygon", POLYGON_TOML),
-];
+/// The chains that have a built-in address book, only for enumerating them: the tests cover every
+/// book through this, and `Registry::builtin` names them when asked for a chain that has none.
+/// The lookup itself is `builtin_book`, so no list is walked to resolve a chain.
+const BUILTIN_CHAINS: [Chain; 6] =
+    [Chain::Ethereum, Chain::Base, Chain::Unichain, Chain::Arbitrum, Chain::Bsc, Chain::Polygon];
+
+/// The address book embedded for `chain`, or `None` for a chain Hindsight has none for.
+///
+/// Keyed on `Chain` rather than a chain name so the caller's string is parsed once, by
+/// `fynd_core`'s parser, instead of being matched against spellings here. The wildcard arm is
+/// forced: `Chain` is `#[non_exhaustive]`, so this cannot be an exhaustive match, which means a
+/// chain Tycho adds later reaches it as "no built-in book" rather than as a compile error.
+fn builtin_book(chain: Chain) -> Option<&'static str> {
+    match chain {
+        Chain::Ethereum => Some(ETHEREUM_TOML),
+        Chain::Base => Some(BASE_TOML),
+        Chain::Unichain => Some(UNICHAIN_TOML),
+        Chain::Arbitrum => Some(ARBITRUM_TOML),
+        Chain::Bsc => Some(BSC_TOML),
+        Chain::Polygon => Some(POLYGON_TOML),
+        _ => None,
+    }
+}
 
 /// On-disk shape of a chain's address book. Field meanings are documented in
 /// `registry/ethereum.toml`.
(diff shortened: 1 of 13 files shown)


tools/hindsight/src/decoder/registry.rs
```diff
@@ -138,26 +152,29 @@ impl Registry {
             return Self::from_toml(&text)
                 .with_context(|| format!("invalid registry file {}", path.display()));
         }
-        match chain.to_lowercase().as_str() {
-            "ethereum" => Ok(Self::ethereum()),
-            "base" => Ok(Self::base()),
-            other => anyhow::bail!(
-                "no built-in decoder address registry for chain '{other}' (only ethereum, base); \
+        let name = chain.to_lowercase();
```
louise-poole: Why not use `Chain` directly in `BUILTIN_CHAINS`? We leave room for error if we have these extra steps of converting to str and then matching on those. Also maybe `BUILTIN_CHAINS` can be a small fn that maps from Chain to toml path. Makes this look-up logic way less complex and you do not need to iterate through a whole list on each look-up.

louise-poole (approved):
LFG 🚀

tools/hindsight/src/decoder/registry.rs
tamaralipows: Done — the lookup now takes a `Chain` and matches on it, so no chain name is compared as a string and nothing is iterated to resolve one.

I did keep the list of six chains, but it is no longer used to look anything up. Two things still need to know which chains we support: the error message that tells you which ones exist, and the tests that check every address book parses. Neither can be derived from a match, and Rust will not let me enumerate the `Chain` variants automatically.

So deleting the list would not remove it — the same six names would reappear in the test file, and the error message would quietly go stale when we add a chain. I left it in one place instead and added a test that fails if it ever disagrees with the lookup.

PR #375, opened 2026-07-29
feat(fynd-rpc): gate exclusive liquidity on a per-request entitlement

tamaralipows:
Exclusive pools are reserved for selected Fynd clients, but nothing stopped an arbitrary caller of `/v1/quote` from routing through them: eligibility was a property of the configured worker pools, not of the request.

## What changed

`ExclusiveAccess` (`Granted` / `Denied`, defaulting to `Denied`) is now a required argument to `WorkerPoolRouter::quote`. Every call site states which it wants.

`fynd-rpc` resolves it per request from the `x-exclusive-access` header:

```
x-exclusive-access: true    -> Granted
anything else, or absent    -> Denied
```

Only the exact value `true` (case-insensitive) grants access. The header is set by the authenticating proxy that already injects `user-identity` and `x-user-plan`, so entitlement lives with API-key resolution rather than in the solver. **This is only meaningful while the server is unreachable except through that proxy** — otherwise a caller can set the header themselves.

A `Denied` request is ranked as if no exclusive-access pool were configured:

- exclusive-access candidates are dropped before ranking, and the surplus overlay is skipped;
- the request no longer waits for the exclusive pool's response before early-returning.

`Solver::quote` passes `Granted`. A library embedder configures its own pools and exclusivity policy, so there is no untrusted caller to gate at that layer.

## Behaviour

`Denied` is not an error. The caller gets the public quote — the same answer a deployment with no exclusive pools would give — so a missing entitlement costs price, not availability. No new response state and nothing for clients to handle.

## Tests

`test_router_exclusive_access_entitlement` runs a public pool (net 800) against a strictly better exclusive pool (net 1100) and asserts whether the winning route contains the exclusive leg. It sets `min_responses(2)` so the denied case has an exclusive candidate in hand to discard; without that it passed even with the gate inverted, because the early-return path hid the leak.

`test_exclusive_access_from_headers` covers absent, `true`, `TRUE`, `false`, empty, `1`, and `" true "`.

Both were verified to fail with the gate inverted.

## Not in this PR

The header only decides who may *route* through exclusive liquidity. Excluding the pool from self-hosted instances that hold no controller key, and the encoder's behaviour when a route has an exclusive leg but no signer, are tracked separately under [ENG-6153](https://propeller-heads.atlassian.net/browse/ENG-6153).

[ENG-6153]: https://propeller-heads.atlassian.net/browse/ENG-6153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/fynd-rpc/CLAUDE.md
+++ b/fynd-rpc/CLAUDE.md
@@ -22,7 +22,7 @@ infrastructure.
 
 | Endpoint | Handler | Description |
 |---|---|---|
-| `POST /v1/quote` | `handlers::quote` | Submit orders, receive optimal routes |
+| `POST /v1/quote` | `handlers::quote` | Submit orders, receive optimal routes. The `x-exclusive-access: true` request header (set by the authenticating proxy, never by the caller) lets the request route through exclusive liquidity; any other value or none restricts it to public pools. Only meaningful when the server is unreachable except through that proxy |
 | `GET /v1/health` | `handlers::health` | Health check (data freshness, derived data readiness, gas-price staleness, pool count). Returns 503 when market data is stale, derived data is not ready, or the gas price is stale |
 | `GET /v1/info` | `handlers::info` | Static metadata about this Fynd instance (version, chain, spender address) |
 | `GET /v1/prices` | `handlers::get_prices` | Token prices, spot prices, pool depths (experimental feature only) |
--- a/fynd-core/src/lib.rs
+++ b/fynd-core/src/lib.rs
@@ -86,5 +86,6 @@ pub use worker_pool::{
     TaskQueueHandle,
 };
 pub use worker_pool_router::{
-    config::WorkerPoolRouterConfig, LiquidityScope, SolverPoolHandle, WorkerPoolRouter,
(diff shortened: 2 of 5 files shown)


fynd-rpc/CLAUDE.md
```diff
@@ -22,7 +22,7 @@ infrastructure.
 
 | Endpoint | Handler | Description |
 |---|---|---|
-| `POST /v1/quote` | `handlers::quote` | Submit orders, receive optimal routes |
+| `POST /v1/quote` | `handlers::quote` | Submit orders, receive optimal routes. The `x-exclusive-access: true` request header (set by the authenticating proxy, never by the caller) lets the request route through exclusive liquidity; any other value or none restricts it to public pools. Only meaningful when the server is unreachable except through that proxy |
```
Troshchk, line 25: So the exclusive access will be available only via hosted Fynd?

PR #376, opened 2026-07-29
fix(derived): recover from broadcast lag incrementally

brunoguerios:
`ComputationManager` handled `broadcast::RecvError::Lagged` by treating the whole topology as changed and recomputing every pool. On busy chains that recompute took long enough to lag again on the next `recv()`, leaving the manager in back-to-back full recomputes.

Recovery now drains the buffered events with `try_recv`, coalesces them into one incremental `ChangedComponents`, and recomputes just that union. Removes the now-unused `ChangedComponents::all`. Adds `derived_manager_lag_recoveries_total` and `derived_manager_lagged_events_total`.

Components removed only within the dropped window keep stale `spot_prices`/`pool_depths` entries until restart — documented on `recover_from_lag`, with a topology-reconciliation follow-up tracked. Routing is unaffected: derived lookups are driven by existing graph edges, and worker graphs reinitialize from live `MarketState` on lag.

### DIFF
--- a/fynd-core/src/derived/manager.rs
+++ b/fynd-core/src/derived/manager.rs
@@ -38,18 +38,6 @@ pub struct ChangedComponents {
 }
 
 impl ChangedComponents {
-    /// Creates a marker for full recompute where all components are considered changed.
-    ///
-    /// Used for startup and lag recovery scenarios.
-    pub fn all(market: MarketDataView) -> Self {
-        Self {
-            added: market.component_topology().clone(),
-            removed: vec![],
-            updated: vec![],
-            is_full_recompute: true,
-        }
-    }
-
     /// Returns true if this update changes the graph topology (adds or removes components).
     pub fn is_topology_change(&self) -> bool {
         !self.added.is_empty() || !self.removed.is_empty()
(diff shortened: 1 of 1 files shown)


fynd-core/src/derived/manager.rs
```diff
@@ -432,6 +471,40 @@ impl ComputationManager {
...
+    ///
+    /// Trade-off: this only recomputes components named in the *drained* events, not
+    /// the ones lost in the *dropped* (skipped) window. Added/updated components missed
+    /// there self-correct on their next `MarketUpdated`. Components REMOVED only in the
+    /// dropped window are different: they are already gone from market topology, so they
+    /// never reappear in a future event, and this path never prunes them either — their
+    /// stale `spot_prices`/`pool_depths` entries persist for the life of the process
+    /// (production runs no full recompute anymore). This is bounded and does not affect
+    /// routing correctness (the routing graph is gated separately and self-resyncs), but
+    /// it is a known limitation; a topology-reconciliation follow-up is tracked to prune
```
tvinagre: I don't understand the last sentence

tvinagre (approved):
LGTM! Elegant :)

fynd-core/src/derived/manager.rs
brunoguerios: It was saying that stale entries from the dropped window are harmless for routing, because derived data is only read per graph edge and a removed component has no edges left. Reworded the comment to say that directly and dropped the follow-up reference which didn't really exist

PR #377, opened 2026-07-30
feat!: replace ExclusivityPolicy with scope based filtering

Troshchk:
Some liquidity is exclusive to Fynd: components that can only be swapped with off-chain authorization. Quotes must keep them out of the committed public reference, while still routing through them to capture surplus above the best  public rate.

 Previously, deciding what counts as "exclusive" was the deployer's job -- a user-supplied policy predicate wired through the builder, with a build error when scoped pools lacked one. This PR removes that configuration surface entirely. Components are now classified from their own data, using the same marker list tycho-simulation's stream filters use, so what a deployment ingests and what it classifies as exclusive can never disagree, and new exclusive pool kinds arrive with a dependency bump instead of a code change.

What a deployment configures is reduced to one per-pool setting: its liquidity scope. A `public_only` pool establishes the committed public reference; an `all` pool (the default) routes through everything the stream delivers.
Deployments that never opt into exclusive components configure nothing and behave exactly as before.

### DIFF
--- a/fynd-core/src/feed/exclusivity.rs
+++ b/fynd-core/src/feed/exclusivity.rs
@@ -15,8 +15,8 @@
 use std::collections::HashMap;
 
 use tycho_simulation::{
+    evm::protocol::ekubo_v3::EXCLUSIVE_EXTENSIONS,
     tycho_common::models::{protocol::ProtocolComponent, Address},
-    EXCLUSIVE_EXTENSIONS,
 };
 
 use crate::{
--- a/fynd-core/src/feed/exclusivity.rs
+++ b/fynd-core/src/feed/exclusivity.rs
@@ -14,10 +14,7 @@
 
 use std::collections::HashMap;
 
-use tycho_simulation::{
-    evm::protocol::ekubo_v3::EXCLUSIVE_EXTENSIONS,
-    tycho_common::models::{protocol::ProtocolComponent, Address},
-};
+use tycho_simulation::tycho_common::models::{protocol::ProtocolComponent, Address};
 
 use crate::{
     feed::{events::MarketEvent, market_data::MarketState},
@@ -25,16 +22,12 @@ use crate::{
 };
 
 /// Returns `true` when the component offers exclusive liquidity, i.e. is swappable only with
-/// off-chain authorization.
+/// off-chain authorization. The `is_exclusive` static attribute is stamped by tycho-simulation's
+/// decoder; absence means the component is public.
 pub(crate) fn is_exclusive(component: &ProtocolComponent) -> bool {
     component
         .static_attributes
-        .get("extension")
-        .is_some_and(|extension| {
-            EXCLUSIVE_EXTENSIONS
-                .iter()
-                .any(|addr| addr.as_slice() == &extension[..])
-        })
+        .contains_key("is_exclusive")
 }
 
 /// Removes exclusive components from a full topology map.
@@ -77,22 +70,18 @@ fn filter_component_ids(market: &MarketState, ids: &[ComponentId]) -> Vec<Compon
         .collect()
 }
 
-/// Stamps the signed-exclusive-swap extension onto a component's static attributes so tests can
-/// build exclusive components without protocol-specific fixtures.
+/// Stamps the `is_exclusive` attribute onto a component's static attributes, mirroring
+/// tycho-simulation's decoder tagging, so tests can build exclusive components without
+/// protocol-specific fixtures.
 #[cfg(test)]
 pub(crate) fn mark_exclusive(component: &mut ProtocolComponent) {
-    component.static_attributes.insert(
-        "extension".to_string(),
-        EXCLUSIVE_EXTENSIONS[0]
-            .as_slice()
-            .into(),
-    );
+    component
+        .static_attributes
+        .insert("is_exclusive".to_string(), vec![1u8].into());
 }
 
 #[cfg(test)]
 mod tests {
-    use alloy::primitives::address;
-
     use super::*;
     use crate::{
         algorithm::test_utils::{component, token},
@@ -115,23 +104,17 @@ mod tests {
         market
     }
 
-    fn component_with_extension(extension: Vec<u8>) -> ProtocolComponent {
+    fn component_with_extension() -> ProtocolComponent {
         let mut c = public_component("pub-1");
         c.static_attributes
-            .insert("extension".to_string(), extension.into());
+            .insert("extension".to_string(), vec![0x55, 0x19].into());
         c
     }
 
     #[rstest::rstest]
-    #[case::exclusive_extension(exclusive_component("excl-1"), true)]
+    #[case::tagged(exclusive_component("excl-1"), true)]
     #[case::missing_attribute(public_component("pub-1"), false)]
-    #[case::other_extension(
-        component_with_extension(
-            address!("0x517E506700271AEa091b02f42756F5E174Af5230").as_slice().to_vec()
-        ),
-        false
-    )]
-    #[case::garbage_attribute(component_with_extension(vec![0x55, 0x19]), false)]
+    #[case::untagged_extension(component_with_extension(), false)]
     fn test_is_exclusive(#[case] component: ProtocolComponent, #[case] expected: bool) {
         assert_eq!(is_exclusive(&component), expected);
     }
--- a/fynd-core/src/feed/exclusivity.rs
+++ b/fynd-core/src/feed/exclusivity.rs
@@ -22,8 +22,8 @@ use crate::{
 };
 
 /// Returns `true` when the component offers exclusive liquidity, i.e. is swappable only with
-/// off-chain authorization. The `is_exclusive` static attribute is stamped by tycho-simulation's
-/// decoder; absence means the component is public.
+/// off-chain authorization. Signaled by the `is_exclusive` static attribute; absence means the
+/// component is public.
 pub(crate) fn is_exclusive(component: &ProtocolComponent) -> bool {
     component
         .static_attributes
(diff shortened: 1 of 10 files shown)


fynd-core/src/feed/exclusivity.rs
```diff
@@ -1,115 +1,107 @@
...
-/// The predicate is supplied by the caller rather than hard-coded against an id or protocol name,
-/// so the notion of "exclusive" can evolve (e.g. a hook address allowlist) without touching the
-/// routing core.
-#[derive(Clone)]
-pub struct ExclusivityPolicy {
-    /// Returns `true` when the component is exclusive and therefore excluded from public
-    /// quotes.
-    is_exclusive: Arc<dyn Fn(&ProtocolComponent) -> bool + Send + Sync>,
+/// Returns `true` when the component's `extension` static attribute names a known exclusive
+/// extension contract.
```
louise-poole: This is specific to Ekubo pools. When we support more exclusive protocols we'd add more identifying logic here. I would update this doc string to be more generic and leave the implementation details (like how we identify the pool) as inline comments.

fynd-core/src/worker_pool/pool.rs
```diff
@@ -46,15 +46,9 @@ pub struct WorkerPoolConfig {
...
-    /// (default: `None`, no filtering).
-    ///
-    /// `All` is safe as the default because it only applies when no `ExclusivityPolicy` is
-    /// configured — meaning no exclusive components exist to exclude. When a policy is set,
-    /// `FyndBuilder::assemble_components` always constructs
-    /// `Some(policy)` for `Public`-scoped pools, so this default is never relied on in
-    /// that path.
-    exclusivity_policy: Option<ExclusivityPolicy>,
+    /// Which liquidity this pool's workers ingest: `PublicOnly` filters exclusive components
+    /// out of each worker's graph; `All` (the default) applies no filtering.
```
louise-poole: I think the specifics of the enum options should live in `LiquidityScope`'s doc comments instead of here. It's unexpected to have to update this doc if we later go extend/change the LiquidityScope options.

fynd-core/src/worker_pool/registry.rs
```diff
@@ -59,12 +60,9 @@ pub(crate) struct SpawnWorkersParams {
...
     /// Sender for shutdown signals.
     pub shutdown_tx: broadcast::Sender<()>,
-    /// Exclusivity policy applied to every worker in this pool, cloned per worker.
-    ///
-    /// `Some(policy)` filters exclusive components out of each worker's graph (public pools);
-    /// `None` keeps every component (exclusive-access pools, or no exclusive components
-    /// configured).
-    pub exclusivity_policy: Option<ExclusivityPolicy>,
+    /// Liquidity scope applied to every worker in this pool: `PublicOnly` filters exclusive
+    /// components out of each worker's graph; `All` applies no filtering.
```
louise-poole: Same here

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -354,15 +343,18 @@ impl WorkerPoolRouter {
...
+        // surplus overlay then needs one response from each; with a single scope (e.g. every
+        // pool on the `All` default), plain count-based gating applies.
         let exclusive_access_pool_names: HashSet<String> = self
             .solver_pools
             .iter()
             .filter(|p| p.liquidity_scope() == LiquidityScope::All)
             .map(|p| p.name().to_string())
             .collect();
-        let has_exclusive_access_pool = !exclusive_access_pool_names.is_empty();
+        let surplus_routing_active = !exclusive_access_pool_names.is_empty() &&
```
louise-poole: Looks like you use 2 different logic approaches to derive `surplus_routing_active`. Maybe you can extract this to a helper fn so we have 1 central place that identifies active surplus routing?

fynd-core/src/solver.rs
louise-poole: This still has the old scope names and default. Please remove specifics from the comments at callsites so that we don't need to update them when the underlying struct changes?

fynd-core/src/feed/exclusivity.rs
```diff
@@ -1,115 +1,107 @@
...
-//! `ExclusivityPolicy` — the shared `MarketState` is never duplicated. Workers of public pools
-//! hold `Some(policy)` and filter; workers of exclusive-access pools hold `None` and ingest
-//! everything.
+//! achieved by filtering each worker's local graph topology/events through
+//! `filter_topology`/`scope_event` when the pool's `LiquidityScope` is `PublicOnly` — the shared
+//! `MarketState` is never duplicated. Workers of `All`-scoped pools ingest everything.
+//!
+//! A component is classified from its own data, whatever its protocol: its `extension` static
+//! attribute is matched against tycho-simulation's exported `EXCLUSIVE_EXTENSIONS` — the same
+//! knowledge the stream filters use, applied generically to every ingested component.
```
louise-poole: This is protocol specific and should not be included here in the main file docstring.

fynd-core/src/feed/exclusivity.rs
```diff
@@ -1,115 +1,107 @@
...
-            .into_iter()
-            .filter(|(id, _)| {
-                market
-                    .get_component(id)
-                    .is_none_or(|c| !self.is_exclusive(c))
-            })
-            .collect()
-    }
+/// Removes exclusive components from a full topology map.
+pub(crate) fn filter_topology(
```
tamaralipows: Maybe we should name this better, like `remove_exclusive_components`

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -41,20 +41,20 @@ use tycho_execution::encoding::{
...
-    encoding::encoder::Encoder, feed::exclusivity::ExclusivityPolicy,
-    price_guard::guard::PriceGuard, worker_pool::task_queue::TaskQueueHandle, BlockInfo,
-    EncodingOptions, Order, OrderQuote, Quote, QuoteOptions, QuoteRequest, QuoteStatus, SolveError,
-    SolveParams, SurplusInfo,
+    encoding::encoder::Encoder, feed::exclusivity::is_exclusive, price_guard::guard::PriceGuard,
+    worker_pool::task_queue::TaskQueueHandle, BlockInfo, EncodingOptions, Order, OrderQuote, Quote,
+    QuoteOptions, QuoteRequest, QuoteStatus, SolveError, SolveParams, SurplusInfo,
 };
 
 /// Which liquidity a solver pool (a group of workers) routes through, and therefore what its
```
tamaralipows, line 103: here we are using solver pool instead of worker pool and this isn't user-facing

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -83,14 +84,13 @@ pub struct SolverPoolHandle {
 }
 
 impl SolverPoolHandle {
-    /// Creates a new solver pool handle with the default [`LiquidityScope::PublicOnly`] scope.
+    /// Creates a new solver pool handle with the default [`LiquidityScope::All`] scope.
```
tamaralipows: Why is LiquidityScope::All the default? This means that by default the user will have access to exclusive pools? 🧐

fynd-core/src/worker_pool/worker.rs
```diff
@@ -113,15 +109,13 @@ where
...
+            liquidity_scope: LiquidityScope::default(),
         }
     }
 
-    /// Configures the policy that filters exclusive components out of this worker's graph.
-    ///
-    /// Public workers use `PublicOnly(policy)`; exclusive-access workers use `All`.
-    pub(crate) fn with_exclusivity_policy(mut self, policy: Option<ExclusivityPolicy>) -> Self {
-        self.exclusivity_policy = policy;
+    /// Sets which liquidity this worker ingests; see [`LiquidityScope`].
```
tamaralipows: Square brackets snuck in lol

tamaralipows (commented):
Thanks! My main question is about the default here

fynd-core/src/worker_pool/worker.rs
louise-poole: Also I feel it's implied that you would check `LiquidityScope` to see the options you can set here. I don't think the `see ['LiquidityScope'].` part is necessary

fynd-core/src/worker_pool_router/mod.rs
louise-poole: Only if they opted-in to exclusive pools on their stream (see [this](https://github.com/propeller-heads/fynd/pull/385) PR). 

I feel like we've gone back and forth on this defaulting a lot (sorry @Troshchk !), but things have been improved and made a lot cleaner now since the last discussion here... I think since we have a built-in exclusive filter and the opt-in nature of streaming exclusive pools I think it's safe for us to now default to `PublicOnly` and someone has to explicitly opt in to first stream exclusive pools and then on a worker-level opt in for `All` to get exclusive access on that worker only. That way even when exclusive pools are streamed the default is to not include them.
On that note should we rename `All` so that it's obvious it's the option coupled with exclusive access? `IncludeExclusive` maybe?

fynd-core/src/worker_pool_router/mod.rs
Troshchk: I thiiiiink that the distinction is nor purely user-facing.
It looks like, solver pool are used in `worker_pool` an worker_pool in `worker_pool_router`
I will have a look and address in the rename PR

fynd-core/src/worker_pool/worker.rs
Troshchk: I think the brackets here were for doc linking, but I have removed than part entirely

fynd-core/src/worker_pool_router/mod.rs
Troshchk: @tamaralipows The initial idea was, that by default we do not filter anything, so if the user opts in for the exclusive components, the user will get them
But as @louise-poole mentioned, to make it safer, we will now go back to `PublicOnly` being default

tamaralipows (approved):
Thank you! A lot clearer now :)

louise-poole (approved):
Thank you! Looks very nice now! 🙏

PR #378, opened 2026-07-30
docs(encoding): keep exclusive-swap docs protocol-agnostic

louise-poole:


Troshchk (approved):
Thank you @louise-poole! 😍

PR #383, opened 2026-07-30
chore: update tycho dependencies to 0.346.0

tamaralipows:
Bumps `tycho-simulation` and `tycho-execution` from 0.345.1 to 0.346.0.

The only upstream change in that range is Balancer V3 reClamm pool support ([tycho#1049](https://github.com/propeller-heads/tycho/pull/1049)). Fynd does not enumerate Balancer V3 pool types, so nothing needs wiring here.

**On the Angstrom attestation cache:** it is already on `main`. It landed upstream in 0.346.0's predecessor 0.344.0 and arrived in Fynd with the 0.345.1 bump in #380, so it is in v0.97.9 onwards. This PR does not add it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4056,7 +4056,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.5.10",
+ "socket2 0.6.4",
  "system-configuration",
  "tokio",
  "tower-service",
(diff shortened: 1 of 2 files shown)


tamaralipows:
Closing — the Angstrom attestation cache is already on main (landed upstream in 0.344.0, arrived via the 0.345.1 bump in #380). 0.346.0 only adds Balancer V3 reClamm, which we don't wire up, so there's nothing to gain here.

PR #385, opened 2026-07-30
feat(feed): stream exclusive pools via the exclusive: protocol prefix

louise-poole:
Allows Fynd users to opt-in to streaming exclusive pools. Currently only supported on Ekubo v3 and controlled by switching out which tycho filter is applied. 

Also synced the claude docs to reflect the current state of the exclusive liquidity feature. Do not worry about reviewing the doc changes too closely, I will update them with our final version of the feature once we're done.

### DIFF
--- a/.claude/CODEBASE.md
+++ b/.claude/CODEBASE.md
@@ -106,6 +106,7 @@ See `docs/ARCHITECTURE.md` for the full architecture diagram and detailed compon
 | `HTTP_PORT` | API port (default: `3000`) |
 | `WORKER_POOLS_CONFIG` | Worker pools config file (default: `worker_pools.toml`) |
 | `BLOCKLIST_CONFIG` | Blocklist config file |
+| `EXCLUSIVE_SWAP_CONTROLLER_KEY` | Restricted exclusive-liquidity deployments only — see [Exclusive liquidity](#exclusive-liquidity-restricted). Unset in ordinary deployments |
 | `RUST_LOG` | Tracing filter (e.g. `info,fynd=debug`) |
 | `METRICS_PORT` | Prometheus metrics server port (default: `9898`, requires `metrics` feature) |
 | `FYND_HOSTED_SWAGGER_URL` | Server URL advertised by the hosted OpenAPI spec. When unset, the hosted Swagger UI (`/docs/hosted/`) is not served — only the self-hosted `/docs/` |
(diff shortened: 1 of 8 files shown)


.claude/CODEBASE.md
```diff
@@ -135,6 +136,35 @@ See `docs/ARCHITECTURE.md` for the full architecture diagram and detailed compon
...
+## Exclusive Liquidity (restricted)
+
+A limited, opt-in service offered to specific deployments — **not** part of the normal routing path.
+Ordinary deployments configure no exclusivity policy, every pool stays `LiquidityScope::PublicOnly`,
+and nothing below applies. Treat it as out of scope unless a task names it.
+
+Worker pools are partitioned by `LiquidityScope` (`fynd-core/src/worker_pool_router/`):
+
+- `PublicOnly` (default) — public liquidity only. Its best candidate is the **committed amount out**,
+  the reference output a quote must at least deliver
```
tamaralipows, line 148: This isn't entirely true if there is no reference, since we are sometimes enabling exclusive routes if no public route found

fynd-core/src/feed/protocol_registry.rs
```diff
@@ -38,6 +44,75 @@ use tycho_simulation::{
 
 use super::DataFeedError;
 
+/// Marker prefix opting a protocol system into the stream variant that also admits its
+/// exclusive-liquidity pools, e.g. `exclusive:ekubo_v3`.
```
tamaralipows: This sentence is a bit funny lol. Can we be a bit more concise

tamaralipows (approved):
Thank you! I like the logic itself! Two minor docs things (and see slack)

docs/guides/server-configuration.md
```diff
@@ -115,7 +131,7 @@ Run `fynd serve --help` for the full list.
...
 | `--chains-config`                  | `TYCHO_CHAINS_CONFIG` | _(none)_                   | Path to the custom-chains `chains.yaml`. Required for a chain Tycho does not know as a built-in.                                                                                                              |
-| `-p, --protocols`                  | —                     | _(all on-chain)_           | Protocols to index (comma-separated). If omitted, all on-chain protocols available on your configured Tycho endpoint are fetched. Use `all_onchain` to combine auto-fetched protocols with explicit entries (e.g. `all_onchain,rfq:bebop`). |
+| `-p, --protocols`                  | —                     | _(all on-chain)_           | Protocols to index (comma-separated). If omitted, all on-chain protocols available on your configured Tycho endpoint are fetched. Use `all_onchain` to combine auto-fetched protocols with explicit entries (e.g. `all_onchain,rfq:bebop`). Prefix an entry with `exclusive:` to also stream that protocol's exclusive pools (e.g. `exclusive:ekubo_v3`). |
```
Troshchk: so if we have multiple exclusive pools -- will the user have to name them all one by one?
Weren't we discussing having a single separate toggle for adding exclusive pools?

docs/ARCHITECTURE.md
```diff
@@ -363,3 +365,29 @@ Worker Pool B (dedicated OS threads)
 * TychoFeed -> Workers: `broadcast` channel (MarketEvent)
 * ComputationManager -> Workers: `broadcast` channel (DerivedDataEvent)
 * All -> MarketState (via MarketData): `Arc<RwLock<>>` (read-heavy)
+
+***
+
+## Exclusive Liquidity (restricted)
+
+A limited, opt-in service offered to specific deployments. It is **not** part of the normal routing
+path: with no `ExclusivityPolicy` configured -- the default -- every pool is
```
Troshchk: But we configure this inside Fynd now?

fynd-core/src/feed/protocol_registry.rs
```diff
@@ -38,6 +44,75 @@ use tycho_simulation::{
...
+///
+/// Like `rfq:`, this is a Fynd-side marker: the prefix is stripped before the protocol system is
+/// registered on the stream, so Tycho only ever sees the bare name.
+const EXCLUSIVE_PREFIX: &str = "exclusive:";
+
+/// Protocol systems that offer an exclusive-liquidity stream variant, i.e. the ones that may be
+/// requested with the `exclusive:` prefix.
+const EXCLUSIVE_CAPABLE_PROTOCOLS: &[&str] = &["ekubo_v3"];
```
Troshchk, line 54: Should we import this from tycho, so there is only one source of truth?

Troshchk (commented):
Thank you @louise-poole
Left a couple of questions

docs/guides/server-configuration.md
louise-poole: The user has to individually set up their env vars to cater for each protocol's exclusivity requirements (like have signer keys etc) so individually opting in for the exclusivity here too isn't much effort.

fynd-core/src/feed/protocol_registry.rs
louise-poole: Since Fynd needs to be updated anyway to support signing and encoding on a new exclusive protocol I would prefer we maintain a list here. We might merge support on tycho before we're ready on Fynd so I would like Fynd to have final say on which it supports or not.

louise-poole: Though I'll add references to this list in the appropriate code locations to make sure it's obvious that this needs to be updated when you add new exclusive protocol support in Fynd.

docs/ARCHITECTURE.md
louise-poole: Yes - but on this version of the code that this PR is based on we don't do that yet. 
This is why I added this part to the PR description:
> Also synced the claude docs to reflect the current state of the exclusive liquidity feature. Do not worry about reviewing the doc changes too closely, I will update them with our final version of the feature once we're done.

.claude/CODEBASE.md
louise-poole: In the current version of the code this is true (your PR for this is still open). I'll do a fresh run through of all these claude docs after all our open PRs are merged.

PR #393, opened 2026-08-03
feat(core)!: allocate worker pools per request entitlement

louise-poole:
The router now picks which worker pools serve an order before dispatching it, instead of fanning out to all of them and filtering the results. An unentitled request is never enqueued to the exclusive-access pool — no CPU spent there, no waiting on it.

Entitlement comes from the `x-exclusive-access` header, set by the authenticating proxy. Only meaningful while Fynd is unreachable except through that proxy. A denied request isn't an error — it gets the public quote. Nothing changes on the wire.

Startup now rejects a config where every worker pool is `liquidity_scope = "all"`.

### DIFF
--- a/.claude/CODEBASE.md
+++ b/.claude/CODEBASE.md
@@ -58,7 +58,7 @@ See `docs/ARCHITECTURE.md` for the full architecture diagram and detailed compon
 ### Core Components
 
 1. **RouterApi** (`fynd-rpc/src/api/`) — Actix Web HTTP handlers: `POST /v1/quote`, `GET /v1/health`, `GET /v1/info`
-2. **WorkerPoolRouter** (`fynd-core/src/worker_pool_router/`) — Fans out orders to all pools, selects best by `amount_out_net_gas`
+2. **WorkerPoolRouter** (`fynd-core/src/worker_pool_router/`) — Allocates the pools that serve each order, fans out to those, selects best by `amount_out_net_gas`
 3. **WorkerPool** (`fynd-core/src/worker_pool/`) — N `SolverWorker` instances on dedicated OS threads per pool
 4. **Algorithm trait** (`fynd-core/src/algorithm/`) — Pluggable route-finding; built-in:
    `MostLiquidAlgorithm`, `BellmanFordAlgorithm`, `PathFrankWolfeAlgorithm`, and
@@ -80,7 +80,7 @@ See `docs/ARCHITECTURE.md` for the full architecture diagram and detailed compon
 
 **Quote request path** (`POST /v1/quote`):
 1. `RouterApi` validates the request
-2. `WorkerPoolRouter` fans out each order to all worker pools in parallel
+2. `WorkerPoolRouter` allocates the worker pools serving each order (an exclusive-access pool only for a request carrying the `x-exclusive-access` entitlement) and fans out to them in parallel
 3. Each pool's `TaskQueue` dispatches to a `SolverWorker` on a dedicated OS thread
 4. Worker calls `Algorithm::find_best_route` with its local graph + shared market/derived data
 5. `WorkerPoolRouter` collects results, ranks candidates by `amount_out_net_gas` descending; if price guard is enabled it validates in rank order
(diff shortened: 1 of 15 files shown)


fynd-core/src/worker_pool_router/allocation.rs
```diff
@@ -0,0 +1,166 @@
...
+    /// Whether this order may be served by exclusive-access worker pools.
+    exclusive_access: ExclusiveAccess,
+}
+
+impl OrderClass {
+    /// Classifies an order whose only distinguishing property is the caller's access.
+    pub(crate) fn new(exclusive_access: ExclusiveAccess) -> Self {
+        Self { exclusive_access }
+    }
+}
```
tamaralipows, line 55: I'm a bit confused by we need a struct with just one attr?

fynd-core/src/worker_pool_router/allocation.rs
```diff
@@ -0,0 +1,166 @@
...
+/// them from the full worker pool list, so a worker pool that was not selected cannot reappear in
+/// the ranking.
+pub(crate) struct Allocation<'a> {
+    /// Selected worker pools, in configuration order.
+    worker_pools: Vec<&'a SolverPoolHandle>,
+    /// Liquidity scope of each selected worker pool, keyed by worker pool name.
+    scopes: HashMap<String, LiquidityScope>,
+    /// Whether both scopes are selected — the single fact that early-return gating and the
+    /// ranking split branch on. See [`Allocation::surplus_routing_active`].
+    surplus_routing_active: bool,
```
tamaralipows: Let's stick to the "exclusive" terminology?

fynd-core/src/worker_pool_router/allocation.rs
```diff
@@ -0,0 +1,166 @@
...
+    let scopes: HashMap<String, LiquidityScope> = worker_pools
+        .iter()
+        .map(|worker_pool| (worker_pool.name().to_string(), worker_pool.liquidity_scope()))
+        .collect();
+    let surplus_routing_active = scopes
+        .values()
+        .any(|scope| *scope == LiquidityScope::IncludeExclusive) &&
+        scopes
+            .values()
+            .any(|scope| *scope == LiquidityScope::PublicOnly);
```
tamaralipows: Do we need to check the `PublicOnly` scope here? Doesn't `IncludeExclusive` automatically imply that surplus routing is active?

tamaralipows (commented):
Thanks, just some minor things - I would like @louise-poole's expertise here and then we can get this merged and hopefully live :)

fynd-core/src/worker_pool_router/allocation.rs
louise-poole: The idea was to make this easy to extend to other allocation rules later (see [this](https://propellerswap.slack.com/archives/C05LQALE9L1/p1785417776049769?thread_ts=1785355901.978609&cid=C05LQALE9L1) message/thread for my motivation there)

fynd-core/src/worker_pool_router/allocation.rs
louise-poole: I would say no, we don't need this public check. Technically for a correct exclusive routing set up there should be at least 1 public worker pool and 1 exclusive worker pool. But there is nothing preventing us from having an exclusive worker pool only...

.claude/CODEBASE.md
louise-poole, line 149: Recently noticed that this is out of date with our latest names. We should do a full sweep of the docs again and update them to reflect our final version (out of scope of this PR - should be done separately)

fynd-rpc/src/api/handlers.rs
```diff
@@ -54,11 +55,13 @@ pub(crate) fn configure_routes(cfg: &mut web::ServiceConfig) {
...
     )
 )]
-#[instrument(skip(state, request), fields(num_orders = request.orders().len()))]
+#[instrument(skip(state, request, http_request), fields(num_orders = request.orders().len()))]
 pub(crate) async fn quote(
     state: web::Data<AppState>,
     request: web::Json<dto::QuoteRequest>,
+    http_request: HttpRequest,
 ) -> Result<HttpResponse, ApiError> {
+    let access = exclusive_access::from_headers(http_request.headers());
```
louise-poole, line 68: I wonder if we should add `access` to the request_capture? Probably would be good no?

tamaralipows (approved):
Lets goo

PR #395, opened 2026-08-03
refactor!: rename exclusive-routing surplus to excess

Troshchk:
Rename exclusive-routing "excess" to "surplus", so there is a single name for the surplus


louise-poole (approved):
Thank you!!

PR #396, opened 2026-08-03
fix: fix terminal-leg check in has_valid_exclusive_route

Troshchk:
Fix has_valid_exclusive_route to detect terminal legs by the route's output token.

The old check just looked at the next swap in the list to guess if a leg was terminal, which broke for split routes where a mid-path exclusive leg happened to sit next to a sibling branch instead of its real continuation.
Now it checks whether the leg's output token matches the route's actual output token, which works no matter how the swaps are ordered.

### DIFF
--- a/fynd-core/src/worker_pool_router/mod.rs
+++ b/fynd-core/src/worker_pool_router/mod.rs
@@ -888,8 +888,8 @@ fn combine_with_surplus(
     result
 }
 
-/// Returns `true` only for routes carrying exactly one exclusive leg, positioned as the terminal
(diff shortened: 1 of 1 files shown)


tamaralipows (approved):
Looks good, but I see in the PR description, title, and commit message that it is a fix, but not what it is actually fixing (i.e. description of the bug). Pls write that somewhere for us to reference in the future. That would have also made it easier to review

louise-poole (approved):
Looks good! Thanks for spotting and fixing this 👍

PR #398, opened 2026-08-03
fix(token-prices): price a token from the median of its deepest paths

tamaralipows:
## Bug

Token prices come from the candidate path with the tightest round-trip spread. Spread only measures whether a probe-sized round trip returns symmetric. It says nothing about whether the rate is right, and a pool holding one side and almost none of the other quotes a wrong rate with a tight spread and deep liquidity.

On BSC such a WBNB-USDT pool priced USDT 12191x too low. No threshold on that pool alone can catch it: its reserves are internally consistent, so it looks like a legitimate pool for a token that really trades at that rate. What catches it is that no other path agrees.

The price feeds solving, not just reporting. `water_fill.rs` converts gas through it, which sets `amount_out_net_gas` — the router's ranking key. `most_liquid.rs` normalizes edge depth through it. hindsight's `value_usd` divides by it.

## Fix

Rank candidate paths by the depth of their thinnest hop and simulate deepest first, until `PRICED_PATHS_PER_TOKEN` (5) have quoted.

From three quotes up, the token's price is their median, so a lone disagreeing pool is outvoted. Below three it is the deepest path's, since two quotes cannot outvote each other.

Depth replaces spread as the ranking key because the wrong pool wins on spread, so ranking by spread would fill the sample with it.

Depth ranks paths rather than admitting them. A hop with no depth recorded reads as zero and sorts last instead of being dropped, so a token whose pools are all thin keeps a rough price rather than losing it.

Token prices now require `pool_depths`, which reorders the derived schedule to spot prices → depths → token prices.

## Cost

The cap counts quotes, not attempts, so a path that fails to simulate does not use up a token's budget. Worst case matches today — a token whose paths all fail tries them all, as the round-robin already did — and the common case is lower, since a token stops at five successes.

None of this is on the solve path. It runs per block in `ComputationManager`, and all three algorithms declare `allow_stale("token_prices")`, so no worker blocks on it. The one real cost is scheduling: prices now run after depths instead of alongside, so per-block freshness lags by the depth computation.

## Worth a look

- Selection changes for every token on every chain, not only mispriced ones. A token with one usable path is unaffected; a token with two gets the deeper of the two.
- 5 is the sample size, so it is also how many independent quotes one wrong pool has to outvote.
- Ranking by depth means a token's price can now come from a different pool than before even where nothing was wrong.

### DIFF
--- a/fynd-core/src/derived/computations/token_gas_price.rs
+++ b/fynd-core/src/derived/computations/token_gas_price.rs
@@ -1,18 +1,19 @@
-//! Computes the `mid_price` of tokens relative to a gas token (e.g., ETH), selecting paths
-//! by the lowest spread (the most reliable price) derived from full simulation of both buy and sell
-//! directions.
+//! Computes the `mid_price` of tokens relative to a gas token (e.g., ETH) as the median of what a
+//! token's deepest paths quote, each from a full simulation of both buy and sell directions.
 //!
 //! # Algorithm
 //!
-//! 1. **Path Discovery (DFS)**: Enumerate all paths from gas_token to each reachable token, scoring
-//!    by spot-price spread: `|forward_spot - 1/reverse_spot|`. Lower spread = better score.
+//! 1. **Path Discovery (DFS)**: Enumerate all paths from gas_token to each reachable token, keeping
+//!    the depth of each path's thinnest hop.
 //!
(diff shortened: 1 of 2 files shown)


tamaralipows:
Closing in favour of #401. This extended the bespoke path search rather than replacing it, and #401 fixes the same BSC mispricing by solving with the router instead — see Markus's point on this PR.

PR #399, opened 2026-08-03
feat(metrics): add per-pool solve duration histogram

brunoguerios:
Adds `worker_pool_solve_duration_seconds{pool}` plus a panel on the local dev dashboard. The Grafana panel for the hosted API dashboard follows in `terraform-infrastructure`, gated on this reaching prod.

## Why a new metric instead of a `pool` label on the existing one

`worker_router_solve_duration_seconds` is recorded once per order *after* the router has raced every pool, so it measures the fan-out and belongs to no single pool — labelling it with the winner would attribute the whole race to whoever won. Reusing the name would also silently change what 21 existing queries measure (12 VMRule exprs, 9 panels, all aggregating with `sum by (le)`), including the 6 prod `FyndQuoteLatencyHighP95` rules.

## Why the worker, not the router

On early return the router drops the remaining `pending` futures, so pools that hadn't answered yet never report. Emitting there would systematically omit the slowest pools — the ones this metric exists to find.

## Successes only

A pool that exhausts its timeout returns before the timing site, so it is absent here rather than slow; those land in `worker_router_solver_failures_total{error_type="timeout"}`. Same for solves failing in `wait_until_ready`. Worth knowing when a pool looks suspiciously fast.

## Review notes

- No exporter change needed — `src/main.rs` already maps the `_seconds` suffix onto `LATENCY_BUCKETS_SECONDS`, so this renders as a bucketed histogram, not a summary.
- Bucket edges near the pool timeouts are `0.25, 0.5, 1.0`, so p95 for a pool pressed against a 500ms budget is coarsely interpolated.
- The local dashboard panel is appended full-width at the bottom; inserting it beside the existing solve panel would have reflowed nine panels. Shows one series until a second pool is configured in `worker_pools.toml`.

Verified with the CI-pinned `nightly-2026-06-28`: `fmt --all --check` and `clippy -p fynd-core --all-targets --all-features` clean, `test -p fynd-core --lib worker_pool::worker` 20/20 including the new `solve_duration_metric_recorded`. Full-workspace clippy/tests left to CI.

Draft until the dashboard PR is ready.

### DIFF
--- a/fynd-core/src/worker_pool/worker.rs
+++ b/fynd-core/src/worker_pool/worker.rs
@@ -44,6 +44,17 @@ fn record_task_pickup_metrics(pool_name: &str, queue_wait: Duration, queue_depth
         .set(queue_depth as f64);
 }
 
+/// Records per-pool solve latency: one algorithm's own working time for one order, excluding
+/// queue wait. Unlike `worker_router_solve_duration_seconds`, which times the router racing every
+/// pool and so belongs to no single pool, this is attributable per pool.
+///
+/// Successful solves only — a pool that exhausts its timeout returns before this point and is
+/// counted in `worker_router_solver_failures_total{error_type="timeout"}` instead.
+fn record_solve_duration(pool_name: &str, solve_time: Duration) {
+    metrics::histogram!("worker_pool_solve_duration_seconds", "pool" => pool_name.to_string())
+        .record(solve_time.as_secs_f64());
+}
+
 /// A solver worker instance that maintains a market graph and processes solve requests.
 pub(crate) struct SolverWorker<A>
 where
@@ -341,9 +352,10 @@ where
             }
         };
 
-        let solve_time_ms = start_time.elapsed().as_millis() as u64;
+        let solve_time = start_time.elapsed();
+        record_solve_duration(&self.pool_name, solve_time);
 
-        Ok(SingleOrderQuote::new(order_quote, solve_time_ms))
+        Ok(SingleOrderQuote::new(order_quote, solve_time.as_millis() as u64))
     }
 
     /// Waits for required derived data to become ready, or until timeout.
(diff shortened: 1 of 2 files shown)


fynd-core/src/worker_pool/worker.rs
```diff
@@ -43,6 +43,17 @@ fn record_task_pickup_metrics(pool_name: &str, queue_wait: Duration, queue_depth
...
 
+/// Records per-pool solve latency: one algorithm's own working time for one order, excluding
+/// queue wait. Unlike `worker_router_solve_duration_seconds`, which times the router racing every
+/// pool and so belongs to no single pool, this is attributable per pool.
+///
+/// Successful solves only — a pool that exhausts its timeout returns before this point and is
+/// counted in `worker_router_solver_failures_total{error_type="timeout"}` instead.
+fn record_solve_duration(pool_name: &str, solve_time: Duration) {
+    metrics::histogram!("worker_pool_solve_duration_seconds", "pool" => pool_name.to_string())
+        .record(solve_time.as_secs_f64());
```
tvinagre, line 55: The solve time is targeted to always be below a second, so I guess it'd make more sense to use `milis` instead of seconds

brunoguerios: If I understand correctly, the standard is to always register the data in seconds and then handle precision within the bucket sizes.
Right now, having a `_seconds` suffix makes it inherit `LATENCY_BUCKETS_SECONDS`, which seems to be good enough to represent solve duration as well.
On top of that, Grafana dashboards should be able to auto-scale the display - so instead of presenting `0.001s` it will show `1ms` automatically.
Let me know if that makes sense or if I'm missing something.

tvinagre: you're correct - I did some research and learned that this is Grafana's convention. Thanks!

tvinagre (approved):

zizou0x (approved):

PR #400, opened 2026-08-03
fix(hindsight): Decoding issues

tamaralipows:
Three decoding faults from prod Base hindsight over 2026-08-01/02, plus two changes from review.

### 1. KyberSwap: a venue's fee was counted against Fynd

KyberSwap's router lets the venue that sent the order keep a share of what the user buys, listing
the addresses to pay in the transaction's calldata. The venue here is not in the address book, so
nothing else identifies it.

- **Before:** hindsight never read that list. It recorded only the part that reached the trader,
  while Fynd's re-solve was measured before any fee came out, so every trade from that venue
  looked like a Fynd win of exactly the fee — 20 on 2026-08-02, all at 1109–1112 bps, which is
  the 10% it charged.
- **Added:** `SolverKnowledge::fee_recipients`, implemented for KyberSwap by decoding its `swap`
  and `swapSimpleMode` calldata, plus `declared_output_fee`, which adds back what those addresses
  were actually paid. Taking the amount from the transfers means no router's fee format has to be
  decoded — KyberSwap writes "1000 bps", not a number of tokens.
- **Unchanged:** trades where a venue already accounted for the fee, and KyberSwap calldata
  nested inside another venue's transaction.

### 2. Relay: a fee payment was recorded as the whole trade

On a Relay fill, `decode_rebalance` finds the address that receives the bought token without
sending anything, and treats what it received as the trade's output.

- **Before:** if the person being paid had also passed the input token along earlier in the same
  transaction, they both sent and received, so they were skipped — leaving whoever took a fee as
  the only candidate. On tx `0x3dc3f83e` the trader was paid 0.12076 WETH and a fee address got
  0.000121 WETH (0.1%). The 0.1% became the trade, so Fynd — which routed the same pair to the
  same amount — appeared to win by 9,990,000 bps, or $225. That was the largest single saving
  reported across those two days on Base.
- **Added:** one check in that loop — a candidate holding under 1% of the token's movement through
  the transaction is skipped. The 1% is netting's existing cut-off, not a new number.
- **Unchanged:** every other reason `decode_rebalance` accepts or declines a shape.

### 3. The report counted trades the dashboards leave out

- **Before:** Prometheus dropped trades below the notional floor from the win rate and the bps
  percentiles, while the offline report counted all of them, so the two disagreed about the same
  run.
- **Added:** the report applies the same floor to its win rate and medians, and the constant moved
  from `telemetry.rs` to `resolve/compare.rs` beside the existing fill-ratio threshold, so both
  surfaces read it from one place.
- **Unchanged:** dollar totals, verdict counts and the trade tables still cover every record.

### 4. The floor drops from $100 to $10 (from review)

Every implausible reading sits below $1, so $100 was well past what the floor needs and excluded
63% of scored trades. Over 2026-08-02's 22,167 scored trades:

| floor | scored | kept | median win bps | max win bps |
|---|---|---|---|---|
| $0 | 22,167 | 100% | 10.3 | 4.4e15 |
| $1 | 18,217 | 82% | 7.0 | 1,220 |
| **$10** | **14,157** | **64%** | **4.4** | **1,111** |
| $100 | 8,225 | 37% | 2.8 | 1,111 |

Past a dollar the spread is real rather than noise — small trades are genuinely easier to beat —
so raising the floor further trades coverage for a lower median, not for accuracy. This moves the
existing Prometheus win rate and bps quantiles, not just the report.

### 5. Residue-ratio comment restored (from review)

Making the constant `pub(crate)` came with a rewritten comment that read worse than the original.
Reverted to the original sentence.

### DIFF
--- a/tools/hindsight/src/decoder/transfer_ledger.rs
+++ b/tools/hindsight/src/decoder/transfer_ledger.rs
@@ -259,10 +259,8 @@ fn net_positive(
     net
 }
 
-/// A leg is residue when it is under this fraction of its token's gross transaction flow:
-/// `leg * RESIDUE_GROSS_RATIO < gross` (1%). A trade moves the bulk of its token's flow; a fee or
-/// a refund moves a sliver. Netting adds two more conditions before pruning a leg (see
-/// `drop_residue_legs`); decoders that only need the size test use the ratio directly.
+/// A net leg is residue when its token routed between third parties and the leg is under this
+/// fraction of the token's gross transaction flow: `net * RESIDUE_GROSS_RATIO < gross` (1%).
 pub(crate) const RESIDUE_GROSS_RATIO: u64 = 100;
 
 /// Net the per-token amounts into a single swap (see `TransferLedger::net_swap`).
--- a/tools/hindsight/src/decoder/transfer_ledger.rs
+++ b/tools/hindsight/src/decoder/transfer_ledger.rs
@@ -210,6 +210,18 @@ impl TransferLedger {
             .collect()
     }
 
+    /// Gross flow of one token through the transaction, counting every transfer of it in either
+    /// direction. The denominator of the residue test.
+    pub(crate) fn token_gross(&self, token: Address) -> U256 {
+        let mut gross = U256::ZERO;
+        for &(transferred, _, _, value) in &self.transfers {
+            if transferred == token {
+                gross = gross.saturating_add(value);
+            }
+        }
+        gross
+    }
+
     /// Gross sent and received per token, summed across the group.
     fn group_totals(
         &self,
@@ -247,9 +259,11 @@ fn net_positive(
     net
 }
 
-/// A net leg is residue when its token routed between third parties and the leg is under this
-/// fraction of the token's gross transaction flow: `net * RESIDUE_GROSS_RATIO < gross` (1%).
-const RESIDUE_GROSS_RATIO: u64 = 100;
+/// A leg is residue when it is under this fraction of its token's gross transaction flow:
+/// `leg * RESIDUE_GROSS_RATIO < gross` (1%). A trade moves the bulk of its token's flow; a fee or
+/// a refund moves a sliver. Netting adds two more conditions before pruning a leg (see
+/// `drop_residue_legs`); decoders that only need the size test use the ratio directly.
+pub(crate) const RESIDUE_GROSS_RATIO: u64 = 100;
 
 /// Net the per-token amounts into a single swap (see `TransferLedger::net_swap`).
 fn net_trade(amounts_by_token: &HashMap<Address, TokenAmounts>) -> Option<NetSwap> {
--- a/tools/hindsight/src/resolve/compare.rs
+++ b/tools/hindsight/src/resolve/compare.rs
@@ -83,17 +83,13 @@ pub(crate) enum Verdict {
 /// loss. Without this cut, a single un-fillable whale trade dominates the USD aggregate.
 const MIN_FILL_RATIO: f64 = 0.5;
 
-/// Settled USD notional below which a trade is left out of the unweighted views — the win rate and
-/// the bps quantiles, in both the Prometheus metrics and the offline report.
+/// Settled USD notional below which a trade is left out of the unweighted views: the win rate and
+/// the bps quantiles. USD-weighted views keep every trade, as does a trade too small to price.
 ///
-/// On a dust trade a few wei of rounding is tens to thousands of bps, so dust dominates an
-/// unweighted median or win count while contributing nothing real. USD-weighted views keep every
-/// trade — a $5 win adds $5 — so total savings and volume stay complete, and a trade whose
-/// notional cannot be priced is kept: we do not drop what we cannot measure.
-///
-/// It lives here, beside `MIN_FILL_RATIO`, because both are scoring policy: every surface that
-/// reports a verdict reads them from one place rather than each choosing its own cut-off.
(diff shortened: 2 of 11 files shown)


tools/hindsight/src/decoder/transfer_ledger.rs
```diff
@@ -247,9 +259,11 @@ fn net_positive(
...
 }
 
-/// A net leg is residue when its token routed between third parties and the leg is under this
-/// fraction of the token's gross transaction flow: `net * RESIDUE_GROSS_RATIO < gross` (1%).
-const RESIDUE_GROSS_RATIO: u64 = 100;
+/// A leg is residue when it is under this fraction of its token's gross transaction flow:
+/// `leg * RESIDUE_GROSS_RATIO < gross` (1%). A trade moves the bulk of its token's flow; a fee or
+/// a refund moves a sliver. Netting adds two more conditions before pruning a leg (see
+/// `drop_residue_legs`); decoders that only need the size test use the ratio directly.
+pub(crate) const RESIDUE_GROSS_RATIO: u64 = 100;
```
louise-poole: I find this hard to follow/understand. The comment says this is a fraction, but it's a u64. The old comment made more sense to me than the additions of the new comment 👀

tools/hindsight/src/resolve/compare.rs
```diff
@@ -83,6 +83,18 @@ pub(crate) enum Verdict {
 /// loss. Without this cut, a single un-fillable whale trade dominates the USD aggregate.
 const MIN_FILL_RATIO: f64 = 0.5;
 
+/// Settled USD notional below which a trade is left out of the unweighted views — the win rate and
+/// the bps quantiles, in both the Prometheus metrics and the offline report.
+///
+/// On a dust trade a few wei of rounding is tens to thousands of bps, so dust dominates an
+/// unweighted median or win count while contributing nothing real. USD-weighted views keep every
+/// trade — a $5 win adds $5 — so total savings and volume stay complete, and a trade whose
+/// notional cannot be priced is kept: we do not drop what we cannot measure.
```
louise-poole: I think we shouldn't have call-site specifics in the global var's description.

tools/hindsight/src/resolve/compare.rs
```diff
@@ -83,6 +83,18 @@ pub(crate) enum Verdict {
...
+///
+/// It lives here, beside `MIN_FILL_RATIO`, because both are scoring policy: every surface that
+/// reports a verdict reads them from one place rather than each choosing its own cut-off.
+pub(crate) const MIN_NOTIONAL_USD: f64 = 100.0;
```
louise-poole: Why $100 cut off when we're just trying to filter out a few wei of rounding? Can't we use a lower cut-off?

tools/hindsight/src/resolve/compare.rs
```diff
@@ -83,6 +83,18 @@ pub(crate) enum Verdict {
...
+/// Settled USD notional below which a trade is left out of the unweighted views — the win rate and
+/// the bps quantiles, in both the Prometheus metrics and the offline report.
+///
+/// On a dust trade a few wei of rounding is tens to thousands of bps, so dust dominates an
+/// unweighted median or win count while contributing nothing real. USD-weighted views keep every
+/// trade — a $5 win adds $5 — so total savings and volume stay complete, and a trade whose
+/// notional cannot be priced is kept: we do not drop what we cannot measure.
+///
+/// It lives here, beside `MIN_FILL_RATIO`, because both are scoring policy: every surface that
+/// reports a verdict reads them from one place rather than each choosing its own cut-off.
```
louise-poole: I don't think we need the comment to motivate why it was moved - only need the comment to describe what it is.

tools/hindsight/src/decoder/transfer_ledger.rs
tamaralipows: Agreed — reverted to the original sentence verbatim in 3cd92f04a. Making it `pub(crate)` didn't need a rewritten comment.

tools/hindsight/src/resolve/compare.rs
tamaralipows: Fixed in 51f45ae2a — the doc now says what the value is and stops there.

tools/hindsight/src/resolve/compare.rs
tamaralipows: Dropped in 51f45ae2a.

tools/hindsight/src/resolve/compare.rs
tamaralipows: You're right, it was too high. Lowered to $10 in 51f45ae2a.

For context, the $100 isn't new — it came in with the original dust guard on 2026-07-27 (b06810b93); this PR only made the offline report use the same number, which is what surfaced it.

Over 2026-08-02's 22,167 scored trades on Base:

| floor | scored | kept | median win bps | max win bps |
|---|---|---|---|---|
| $0 | 22,167 | 100% | 10.3 | 4.4e15 |
| $1 | 18,217 | 82% | 7.0 | 1,220 |
| **$10** | **14,157** | **64%** | **4.4** | **1,111** |
| $100 | 8,225 | 37% | 2.8 | 1,111 |

Every implausible reading sits below $1 — a $1 floor already removes the 4.4e15 tail. Past that the spread is real rather than rounding: small trades are genuinely easier to beat, so a higher floor buys a lower median, not more accuracy. $10 clears the noise with a decade of margin and keeps 64% of the book instead of 37%.

Worth flagging that this moves the existing Prometheus win rate and bps quantiles too, not just the report, so those dashboards will step when it deploys.

louise-poole (approved):

PR #402, opened 2026-08-04
feat(core): drop a pAMM quote whose fallback misses min_amount_out

tamaralipows:
Fynd side of the pAMM simulation fix. Contract side: propeller-heads/tycho#1278.

## Problem

propeller-heads/tycho#1278 makes the PropAMMRouter fall back to a Uniswap V3 pool when a stale pAMM quote reverts the venue. That is not enough, because Fynd's minimum amount out describes the pAMM quote, not the fallback.

The pAMM quotes 2000 USDC. The user accepts 1% slippage.

```
1. Fynd quotes the pAMM leg          -> 2000 USDC
2. Fynd writes the minimum amount    -> 1980 USDC   (2000 less 1%)
3. Integrator simulates. pAMM stale. Router falls back to Uniswap V3.
4. Uniswap V3 pays                   -> 1950 USDC
5. TychoRouter compares 1950 < 1980  -> revert
```

Step 3 works. Step 5 fails, so the revert moves from `StaleUpdate` to `NegativeSlippage`.

The fallback rescues the trade only when it pays at least what the user accepted. Fynd cannot tell which case it is without pricing the fallback.

## How it works

`min_amount_out` keeps its present value: the pAMM quote, less fees, less the slippage the user accepted. The fallback amount is only used to reject routes.

**Learn which pool the router would use.** The router picks it with `resolvedFee(tokenIn, tokenOut)`, a public view function. A background task reads that on a timer and caches it. Governance changes it rarely, and only about 12 pairs across Fermi and Kipseli need one, so the quote path never waits on RPC.

**Keep the candidate pools indexed.** Each solver worker holds an index of Uniswap V3 pools by token pair and fee tier, built once and updated from market events. A pool qualifies on its component alone, so the index only moves when the market adds or removes one. Rebuilding it per block would cost 97 ms.

**Price the fallback while solving.** When a worker produces a route with a pAMM leg, it replays that route with each pAMM leg swapped for its fallback pool, and stamps the result on the route. Replay is the existing `replay_route`, so split fractions and shared-pool depletion behave as they do everywhere else.

**Reject the route when the fallback misses the floor.** The encoder compares the fallback fill against `min_amount_out`. A quote whose fallback cannot clear that floor returns `NoRouteFound`, because it can only revert. Lowering the floor to fit would pay the user less than they accepted, and would widen the window for a sandwich attack.

A route whose fallback cannot be priced at all is dropped in the same spirit, while solving.

## No client impact

Clients change nothing.

The wire format is untouched — `clients/openapi.json` is identical to main, so there is no field to set and no client to regenerate. The quoted amount out and the minimum amount out both keep their present meaning.

Fynd builds no price level stream, so nothing in this repo emits `propammfallback:`. Every route reports no pAMM leg and behaviour is unchanged until the service that builds the stream emits the family.

Once on, the only visible change is that a pAMM route with a weak fallback returns no route, instead of returning a quote that reverts on simulation.

## Next steps

  1. Merge Fynd #402. PropAMMFallbackExecutor is deployed, Tycho #1278 is merged, and this PR pins tycho 0.363.0, which carries the executor address, the executor model, and the `PriceLevelStreamBuilder` venue selection.
  2. Run a price level stream with the fallback router on. `PriceLevelStreamBuilder` reads the router's on-chain whitelist and emits `propammfallback:{venue}` for whitelisted venues. That stream is the switch, and it is built outside this repo.

## Limits

**bopAMM does not work.** The PropAMMRouter whitelist does not contain its venue, so bopAMM quotes continue to fail simulation.

**A venue outside the whitelist is worse than not doing this.** Every swap then executes on Uniswap V3 at a worse price than the pAMM gives.

**Large trades lose the pAMM route.** One pool in one fee tier moves 60 bps on 1000 WETH to USDC (about $1.87M), and 7 bps on $2M USDC to USDT. A trade whose fallback moves further than the user's slippage is not quoted at all, so the largest orders fall back to public liquidity rather than reverting.


fynd-core/src/propamm_fallback/fee_fetcher.rs
```diff
@@ -0,0 +1,347 @@
...
+                    continue;
+                }
+            };
+
+            for pair in pairs {
+                if !seen.insert((pair.token0, pair.token1)) {
+                    continue;
+                }
+                let fee = self
+                    .eth_call::<IPropAMMRouter::getPairFeeCall>(
```
tvinagre: Instead of multiple calls, can you make one multicall here to make it more efficient?

fynd-core/src/propamm_fallback/fee_fetcher.rs
```diff
@@ -0,0 +1,347 @@
...
+            .map_err(|e| FallbackFeeFetchError::Call { method, contract, reason: e.to_string() })?;
+        C::abi_decode_returns(&response).map_err(|e| FallbackFeeFetchError::Call {
+            method,
+            contract,
+            reason: format!("failed to decode response: {e}"),
+        })
+    }
+}
+
+fn to_address(raw: &Bytes, what: &str) -> Result<Address, FallbackFeeFetchError> {
```
tvinagre: dup code, we should already have this as a aux method in the codebase

fynd-core/src/propamm_fallback/mod.rs
```diff
@@ -0,0 +1,610 @@
...
+    for swap in route.swaps() {
+        if !swap
+            .protocol()
+            .starts_with(PROPAMM_FALLBACK_PREFIX)
+        {
+            substituted.push(swap.clone());
+            continue;
+        }
+
+        let fee = fees.resolved_fee(swap.token_in(), swap.token_out());
```
tvinagre: I think this whole fee naming is very conflicting with Router fees - I'd probably change this but it's fine to do this in another PR

tvinagre:
Overall: the design is sound and well-executed — the fallback pricing reuses `replay_route` so split/depletion semantics stay consistent, the pool index is maintained correctly (including the rebuild on broadcast lag, since it lives in `initialize_graph`), the wire format is genuinely untouched, and the tests cover the right edges (zero-fee override clearing, one failing venue not discarding others, token-order independence, split routes). Two correctness issues worth fixing before merge, one resource leak, and a few smaller points.

## Correctness

**1. The fallback floor isn't clamped to the venue floor — a fallback pool that out-prices the pAMM quote makes the happy path revert** (`fynd-core/src/encoding/encoder.rs:353`)

`min_amount_out` is replaced wholesale with the fallback-derived floor. If at solve time the router's fallback pool pays *more* than the pAMM quote (the algorithm picks routes net of gas, and the pool it evaluated isn't necessarily the one at the router's resolved fee tier, so this isn't impossible), the new floor exceeds what the pAMM path pays post-fee — and now the *venue* fill reverts, inverting the bug this PR fixes. A one-line guard closes it:

```rust
fee_breakdown.with_min_amount_received(
    fallback_floor.min_amount_received().min(fee_breakdown.min_amount_received()).clone()
)
```

**2. Labeled-state solves price the fallback against the wrong state** (`fynd-core/src/worker_pool/worker.rs:300`)

When `params.state_label()` is set, the algorithm solves against the overlay (`read_labeled`), but the new pAMM block does `self.market_data.read().await` (base view) and `fallback_amount_out` then replays against `market.base_market_state()` — which explicitly bypasses overlays. `replay_route` also ignores the states embedded in the swaps and resolves everything from the passed `MarketState`, so the entire substituted route (not just the fallback leg) is priced on a different state than the route was solved on. If labeled solves are only used by tooling this may be acceptable, but it should at least be a documented decision; the stamped floor silently disagreeing with the quote's `solved_against` state is surprising.

## Resource / operational

**3. The fallback fee fetcher task can't be shut down** (`fynd-core/src/solver.rs:940`, `1007`, `1084`)

`router_fee_fetcher` follows the pattern of keeping its `JoinHandle` and aborting it in `shutdown()`; the new fetcher's handle is dropped on the floor in all three build variants, so the task (with its RPC client) outlives `shutdown()` and `into_parts()` callers can't manage it either. Store it alongside `router_fee_handle`.

**4. Startup window runs on the compiled-in default fee tier** (`DEFAULT_FALLBACK_FEE = 3000`)

Until the first successful fetch — or indefinitely if the RPC is down at startup, since failures keep "previous" values that are just the default — per-pair overrides are missing. A pair that actually resolves to the 500 tier gets priced against the 3000-tier pool (wrong floor, potentially reintroducing the revert) or dropped if no 3000-tier pool exists. Given ~12 pairs, consider doing one blocking fetch at build time, or gating pAMM stamping until the first successful refresh.

## Minor

- `fynd-core/src/solver.rs:854`: `PROPAMM_VENUES.iter().filter_map(|v| Bytes::from_str(v).ok())` silently drops a malformed venue constant — a typo would silently lose that venue's fee overrides. These are compile-time constants, so failing loudly is cheap. Same block reuses `SolverBuildError::RouterFeeFetcher` for the *fallback* fetcher, which will produce a misleading error message.
- `fee_fetcher.rs`: the `seen` set keys on `(token0, token1)` unsorted, so two venues listing the same pair in opposite order trigger a duplicate `getPairFee` call. Harmless, but sorting the key matches the semantics everywhere else.
- Router address and venue list are hardcoded in `fynd-core` rather than config — fine for now given venue selection is the real switch, but a governance change or new venue means a redeploy.
- Design note (acknowledged in the PR): a route whose fallback can't be priced returns `no_route_found` for the whole order, even when plain Uniswap liquidity existed. Until venue selection lands this path is dead, but the eventual improvement is re-solving with pAMM legs excluded rather than returning nothing.

The PR's own open decision — capping the floor gap on large trades — should be resolved before venue selection turns this on, but it doesn't block this merge since the code path is inert until then.

fynd-core/src/propamm_fallback/fee_fetcher.rs
tamaralipows: Done. A refresh is now two calls: one Multicall3 batch for `fallbackFee` plus every venue's `getPairs`, and one for the `getPairFee` of every pair. It used to be one call per venue and per pair.

fynd-core/src/propamm_fallback/fee_fetcher.rs
tamaralipows: Done. The address conversion and the `eth_call` step both moved to a new private `fynd-core::rpc`, and the router fee fetcher uses them too, so there is one copy of each.

fynd-core/src/propamm_fallback/mod.rs
tamaralipows: Renamed in this PR, since I was in the file anyway: `FeeTiers`, `SharedFeeTiers`, `FeeTierFetcher`. These are Uniswap V3 fee tiers, so nothing shares a name with `RouterFees` any more.

tamaralipows:
Thanks — all of it is in. The branch is now **rebased onto `main` (`4f238653d`)**, so every hash changed; the review fixes are the last three commits, `8d5d080c4..80933092b`.

One history note: the three commits `5d291d18b`, `83350cb43` and `910825dd8` are squashed into `f5541b97e`. The first of them added an `EncodingOptions::cover_propamm_fallback` flag and the last one deleted it, so replaying them separately meant resolving conflicts inside code that does not survive. No commit carries that flag now, and `clients/openapi.json` has zero diff against `main`. The other four feature commits are unchanged in content.

**Fixed**

- **Labeled solves priced the fallback on the wrong state.** Real, and the worst of the set. The replay now runs on the same view the algorithm solved against, through `extract_subset_with_overlay`. A test registers an overlay that halves the fallback pool's price and asserts the stamped amount follows it; it fails against the old code.
- **The fetcher task could not be shut down.** Its handle now sits next to `router_fee_handle` everywhere that one does, `SolverParts` and `FyndRPC` included. No public signature changes: `SolverParts` hands the task out as an `AbortHandle` from a new `fee_tier_abort_handle`, so `into_components` keeps the ten elements it has on `main`. The abort handle stays valid after that call drops the `JoinHandle`, and `fynd-rpc` takes it beforehand and aborts it with the other background tasks.
- **The startup window ran on the compiled-in tier.** Gone, along with `DEFAULT_FALLBACK_FEE`. `SharedFeeTiers` is empty until the first successful read, and a worker that finds it empty drops the pAMM route. The wrong tier prices the wrong pool, and a floor from the wrong pool is the revert this path exists to prevent — so dropping is the safe direction, and the window is one RPC round trip, not a refresh interval.
- **A malformed venue constant was silently dropped.** Both the router address and the venue list now fail the build, under their own `SolverBuildError::FeeTierFetcher`.
- **Unsorted `seen` key.** Sorted, so two venues listing a pair in opposite order cost one `getPairFee`.

**Not a problem**

- **Clamping the floor to the venue floor.** That was true of an earlier revision. `910825dd8` stopped moving `min_amount_out` at all: it stays the venue quote less fees less the user's slippage, and the encoder only decides whether to keep the quote. A fallback that out-prices the pAMM quote now just passes the check.

**Left alone, on purpose**

- **Router address and venue list in code, not config.** Agreed, and a governance change means a redeploy. Venue selection is the real switch and it lands in the next PR; I would rather move both into config there than add a config surface for a path that is inert today.
- **`no_route_found` for the whole order when a fallback cannot be priced.** Also agreed, and re-solving with pAMM legs excluded is the right answer. It needs a second solve pass, so it is its own change.
- **Capping the floor gap on large trades.** Still open, and still blocking venue selection rather than this merge.

fynd-core/src/encoding/encoder.rs
```diff
@@ -295,6 +295,36 @@ impl Encoder {
...
+            if let Some(fallback) = &fallback_amount_out {
+                if !Self::fallback_clears_floor(
+                    fee_breakdown.min_amount_received(),
+                    fallback,
+                    encoding_options
+                        .client_fee_params()
+                        .map_or(0, |f| f.bps()),
+                    fee_rates,
+                )? {
+                    tracing::debug!(
```
tvinagre, line 319: Can we also emit a metric here? We should keep track of this on the hosted API to know how much this is hurting us.

tvinagre (approved):
Lgtm, can you already update tycho-client to include the relase of the fallback executor + sim?

fynd-core/src/encoding/encoder.rs
tamaralipows: Done. `propamm_fallback_quotes_total` counts every route with a pAMM leg, split by `outcome="dropped"` and `outcome="kept"`, so the hosted API shows what share of pAMM routes the floor removes.

tamaralipows:
Done — tycho is now pinned to `>=0.363.0`. That release has the deployed PropAMMFallbackExecutor address, the executor model, and the `PriceLevelStreamBuilder` venue selection that emits `propammfallback:{venue}`.

Fynd builds no price level stream, so nothing here emits that family yet. I updated the Next steps in the description to say so.

PR #406, opened 2026-08-06
feat(hindsight): measure route decay over the blocks after a quote

kayibal:
Adds a `decay` subcommand that quotes a sample of trade shapes at one block, then replays those
exact routes at each of the next `--offsets` blocks (default 5). Each offset records the total
move, the market drift a fresh solve at the same state also saw, and the difference — what holding
a stale route cost.

## Why a separate measurement

`monitor`'s existing `slippage` field solves a settled trade at N-1 and replays it at N. That state
already contains the trade's own price impact, so the number measures the route eating its own
shadow. The trades `decay` samples never execute: shapes are drawn from a `monitor` run's
comparison JSONL (pair and input amount only) and re-quoted at unrelated live blocks, which removes
that contamination.

## Cost

Rounds do not overlap, so per-block work is `--sample-size` replays plus `--sample-size` solves
regardless of offset depth. A run warns when a block's work exceeds half the chain's block time.

## Refactors carried along

- `SteppingSolver` moves to `resolve/step.rs` and the solver build/rebuild scaffolding to
  `resolve/session.rs`; both are now shared with `monitor`.
- `StepAdapter` gains an encoding opt-out. `decay` reads only `amount_out`, and requesting calldata
  would both cost work and drop routes an encoder rejects.
- `RotatingWriter` takes an explicit filename prefix, so decay output cannot be re-read as sampler
  input.
- The sampler tolerates reverted-transaction records, which carry null amounts by design, without
  counting them as malformed.

## Note on the open hindsight stack

This branch sits directly on `main`, independent of the open decoding stack (#388 → #371 → #391).
Those PRs also touch `resolve/jsonl.rs` and `resolve/monitor.rs`, so whichever merges second will
need a rebase; there is no functional overlap.

### DIFF
--- a/tools/hindsight/src/decay/mod.rs
+++ b/tools/hindsight/src/decay/mod.rs
@@ -1,0 +1,763 @@
+//! Route decay measured in non-overlapping rounds: quote a sample of trades at one block, then
+//! replay those exact routes at each of the following `--offsets` blocks.
+//!
+//! Each offset yields three numbers. The quoted route replayed at the later state gives the **total
+//! move**; a freshly solved quote at that same state gives the part of it that was **unavoidable
+//! market drift**; the difference is what holding a **stale route** cost. Only the last is
+//! something better routing or faster submission can recover.
+//!
+//! The trades are synthetic on purpose. `monitor`'s existing `slippage` field solves a settled
+//! trade at N-1 and replays it at N — a state that already contains that trade's own price impact,
+//! so it measures the route eating its own shadow. Here the sampled shapes never execute: they are
+//! drawn from historical flow (see [`sample`]) and re-quoted at unrelated live blocks, so nothing
+//! but genuine market movement and route staleness can move the measurement.
+//!
+//! Rounds do not overlap — a new sample is drawn only once the previous round's last offset is
+//! measured — so per-block work is `sample_size` replays plus `sample_size` solves regardless of
+//! how many offsets are configured.
+
+pub(crate) mod record;
+pub(crate) mod sample;
+mod summary;
+
+use std::{
+    path::PathBuf,
+    time::{Duration, Instant},
+};
+
+use tracing::{info, warn};
+
+use crate::{
+    decay::{
+        record::{DecayRecord, Measurement, QuotedTrade, ReplayFailure},
+        sample::{Sampler, TradeShape},
+        summary::Summary,
+    },
+    resolve::{
+        jsonl::RotatingWriter,
+        session::{self, Session, SolverArgs},
+        step::{Encoding, StepAdapter},
+        Outcome, SteppingSolver,
+    },
+    telemetry,
+};
+
+/// Filename prefix for decay output. Deliberately not `comparisons`, which [`sample`] globs as its
+/// input — sharing it would let a run read its own output back as trade shapes.
+const DECAY_PREFIX: &str = "decay";
+
+/// Fraction of a block's time the per-block measurement may occupy before it is warned about. Above
+/// this there is no margin left for feed jitter, and the run is one slow solve away from sliding
+/// behind head.
+const PACING_WARN_FRACTION: f64 = 0.5;
+
+/// Inputs for the decay measurement — the `decay` subcommand's CLI arguments.
+#[derive(clap::Args)]
+pub(crate) struct DecayArgs {
+    #[command(flatten)]
+    pub solver: SolverArgs,
+
+    /// Directory of a `monitor` run's `comparisons-*.jsonl` to draw trade shapes from. Only each
+    /// record's token pair and input amount are read; nothing about how it settled is used
+    #[arg(long)]
+    pub comparisons_dir: PathBuf,
+
+    /// Venue whose flow to sample, matched case-insensitively against each record's `venue`
+    #[arg(long, default_value = sample::RELAY_VENUE)]
+    pub venue: String,
+
+    /// Trades quoted per round. Every one is re-solved at every offset, so this sets the per-block
+    /// solver load: budget roughly 30ms per trade and keep the total well inside one block (on
+    /// Base's 2s blocks, 25 trades ≈ 750ms worst case)
+    #[arg(long, default_value_t = 25)]
+    pub sample_size: usize,
+
+    /// How many blocks after the quote to measure. Decay is front-loaded, so the first few offsets
+    /// carry most of the signal
+    #[arg(long, default_value_t = 5)]
+    pub offsets: u32,
+
+    /// Seed for the trade draw, so a run's sample is reproducible
+    #[arg(long, default_value_t = 0)]
+    pub seed: u64,
+
+    /// Write one JSON line per (trade, offset) into this directory as `decay-YYYY-MM-DD.jsonl`,
+    /// rotated at each UTC day boundary
+    #[arg(long)]
+    pub output_dir: Option<PathBuf>,
+}
+
+/// One session's terminal condition: the run completed (`--max-blocks` reached) or the session is
+/// unhealthy — the feed died — and the caller should rebuild the solver.
+enum SessionEnd {
+    Complete,
+    Unhealthy(String),
+}
+
+/// Counters that survive feed rebuilds, so `--max-blocks` and the final tally span the whole run.
+#[derive(Default)]
+struct Totals {
+    /// Blocks whose measurement work completed, the unit `--max-blocks` counts.
+    blocks: u64,
+    rounds: u64,
+    /// Rounds abandoned because the feed died before their last offset.
+    abandoned_rounds: u64,
+    records: u64,
+    /// Sampled shapes the solver could not quote at all, so they never entered a round.
+    unquotable: u64,
+    pool_gone: u64,
+    simulation_failed: u64,
+    no_market_reference: u64,
+}
+
+impl Totals {
+    fn count_failure(&mut self, failure: ReplayFailure) {
+        match failure {
+            ReplayFailure::PoolGone => self.pool_gone += 1,
+            ReplayFailure::SimulationFailed => self.simulation_failed += 1,
+            ReplayFailure::NoMarketReference => self.no_market_reference += 1,
+        }
+    }
+}
+
+/// Build the in-process stepped solver and measure route decay in rounds until `--max-blocks` or
+/// Ctrl-C. When the tycho feed dies the solver is torn down and rebuilt in place, and the round in
+/// flight is abandoned rather than finished against a different solver's states.
+pub(crate) async fn run(cfg: DecayArgs) -> anyhow::Result<()> {
+    anyhow::ensure!(cfg.offsets >= 1, "--offsets must be at least 1");
+    anyhow::ensure!(cfg.sample_size >= 1, "--sample-size must be at least 1");
+
+    let shapes = sample::load_shapes(&cfg.comparisons_dir, &cfg.venue)?;
+    // Taken before the solver args are moved into the session, which owns them from here on.
+    let rounds = RoundConfig {
+        sample_size: cfg.sample_size,
+        offsets: cfg.offsets,
+        max_blocks: cfg.solver.max_blocks,
+    };
+    let session = Session::prepare(cfg.solver).await?;
+    let block_budget = session::block_time(session.chain()).mul_f64(PACING_WARN_FRACTION);
+    info!(
+        sample_size = cfg.sample_size,
+        offsets = cfg.offsets,
+        pool = shapes.len(),
+        seed = cfg.seed,
+        block_budget_ms = block_budget.as_millis(),
+        "measuring route decay in rounds"
+    );
+
+    if let Some(port) = session.args().metrics_port {
+        telemetry::install_exporter(port)?;
+        info!(port, "serving Prometheus metrics at /metrics");
+    }
+
+    let output = match cfg.output_dir.as_ref() {
+        Some(dir) => {
+            let writer = RotatingWriter::open(dir, DECAY_PREFIX)?;
+            info!(path = %writer.current_path().display(), "appending decay records to JSONL");
+            Some(writer)
+        }
+        None => None,
+    };
+
+    let mut state = RunState {
+        sampler: Sampler::new(cfg.seed, shapes.len()),
+        summary: Summary::new(cfg.offsets),
+        totals: Totals::default(),
+        output,
+        block_budget,
+    };
+
+    // Resolves on Ctrl-C, so a long run stops cleanly at any await below — including the
+    // multi-minute solver builds. On shutdown the round in flight is abandoned and the output
+    // writer flushes as it drops.
+    let shutdown = session::shutdown_signal();
+    tokio::pin!(shutdown);
+
+    // The first build fails fast — an error here is a configuration problem. Rebuilds after a feed
+    // death retry forever, since the config is known good and failures are transient.
+    let (mut solver, mut controller) = tokio::select! {
+        biased;
+        () = &mut shutdown => return Ok(()),
+        built = session.build() => built?,
+    };
+    loop {
+        // Encoding off: decay reads only `amount_out`, so calldata is work we would throw away —
+        // and an encode failure would drop a route that measures perfectly well.
+        let adapter =
+            StepAdapter::new(&solver, &controller, session.args().timeout_ms, Encoding::Skipped);
+        let reason = tokio::select! {
+            biased;
+            () = &mut shutdown => {
+                info!("received Ctrl-C; shutting down");
+                break;
+            }
+            end = run_session(&adapter, &rounds, &shapes, &mut state) => match end {
+                SessionEnd::Complete => break,
+                SessionEnd::Unhealthy(reason) => reason,
+            },
+        };
+        warn!(reason, "session unhealthy; rebuilding the solver to resubscribe");
+        telemetry::record_feed_rebuild();
+        solver.shutdown();
+        let Some(built) = session.rebuild(shutdown.as_mut()).await else {
+            info!("received Ctrl-C during rebuild; shutting down");
+            report(&state);
+            return Ok(());
+        };
+        (solver, controller) = built;
+    }
+    solver.shutdown();
+    report(&state);
+    Ok(())
+}
+
+/// The round-shaping knobs, lifted out of [`DecayArgs`] before its solver args are moved into the
+/// [`Session`] that owns them.
+struct RoundConfig {
+    sample_size: usize,
+    offsets: u32,
+    max_blocks: Option<u64>,
+}
+
+/// Run state that outlives any one solver session, so a feed rebuild loses only the round in
+/// flight.
+struct RunState {
+    sampler: Sampler,
+    summary: Summary,
+    totals: Totals,
+    output: Option<RotatingWriter>,
+    block_budget: Duration,
+}
+
+/// Drive one solver session: draw a sample, quote it, measure it across the configured offsets, and
+/// repeat until the run completes or the feed dies.
+async fn run_session<S: SteppingSolver + ?Sized>(
+    adapter: &S,
+    cfg: &RoundConfig,
+    shapes: &[TradeShape],
+    state: &mut RunState,
+) -> SessionEnd {
+    // A quote needs an applied state to be quoted against.
+    if adapter.current_block().await.is_none() {
+        info!("waiting for solver to apply its first block…");
+        if let Err(e) = adapter.advance().await {
+            return SessionEnd::Unhealthy(e.to_string());
+        }
+    }
+
+    loop {
+        let Some(quote_block) = adapter.current_block().await else {
+            warn!("no applied block yet; advancing");
+            if let Err(e) = adapter.advance().await {
+                return SessionEnd::Unhealthy(e.to_string());
+            }
+            continue;
+        };
+
+        let quoted = quote_round(adapter, cfg, shapes, state, quote_block).await;
+        state.totals.rounds += 1;
+        let round_id = state.totals.rounds;
+        if quoted.is_empty() {
+            // Nothing to replay, but the offsets must still elapse — otherwise the next round would
+            // quote against the very state this one did.
+            warn!(
+                block = quote_block,
+                round_id, "no sampled trade could be quoted; skipping the round"
+            );
+            for _ in 0..cfg.offsets {
+                if let Err(e) = adapter.advance().await {
+                    return SessionEnd::Unhealthy(e.to_string());
+                }
+            }
+            continue;
+        }
+
+        for offset in 1..=cfg.offsets {
+            if let Err(e) = adapter.advance().await {
+                // The round dies with the session: its remaining offsets would be measured against
+                // a rebuilt solver's states, which are not continuous with this
+                // round's quote.
+                state.totals.abandoned_rounds += 1;
+                return SessionEnd::Unhealthy(e.to_string());
+            }
+            let started = Instant::now();
+            let Some(measured_block) = adapter.current_block().await else {
+                warn!(round_id, offset, "no applied block after advancing; abandoning the round");
+                state.totals.abandoned_rounds += 1;
+                break;
+            };
+            let at = RoundAt { round_id, quote_block, measured_block, offset };
+            measure_offset(adapter, state, &quoted, at).await;
+
+            let elapsed = started.elapsed();
+            telemetry::record_block_seconds(elapsed.as_secs_f64());
+            if elapsed > state.block_budget {
+                warn!(
+                    round_id,
+                    offset,
+                    elapsed_ms = elapsed.as_millis(),
+                    budget_ms = state.block_budget.as_millis(),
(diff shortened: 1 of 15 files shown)


tools/hindsight/src/decay/record.rs
```diff
@@ -0,0 +1,335 @@
...
+            (Err(failure), _) => Some(failure),
+            (Ok(_), None) => Some(ReplayFailure::NoMarketReference),
+            (Ok(_), Some(_)) => None,
+        };
+        let bps = match (replayed, measurement.fresh) {
+            (Some(replayed), Some(fresh)) => {
+                DecayBps::new(trade.quoted_amount_out(), replayed, fresh)
+            }
+            _ => None,
+        };
```
kayibal: :robot: **A zero amount in any leg produces a record with neither `bps` nor `failure`, and that silently removes the worst measurements from the tail statistics.**

`DecayBps::new` calls `bps_against` twice, which calls `raw_bps_diff`. That function returns `None` when **either** argument is non-positive (`tools/common/src/bps.rs:13-17` filters both with `v > 0.0`), not just when the baseline is zero. So `DecayBps::new` returns `None` if `quoted`, `replayed`, **or** `fresh` is zero.

Now trace `replayed = Ok(U256::ZERO)`, `fresh = Some(non-zero)`:

- line 165-169: `failure` is `None`, because `replayed` is `Ok(_)` and `fresh` is `Some(_)`.
- line 170-175: `bps` is `None`, because `DecayBps::new` rejected the zero.

The record now has neither, which contradicts this type's own doc comment at line 94-95 ("Every record carries either `bps` ... or `failure` ... so a missing measurement is always attributable rather than silently absent"). `Totals::count_failure` is never called for it, so the run's tally shows no failure, and `Summary::record` is never called for it, so it does not appear in any per-offset statistic. `Totals::records` still counts it, so `records` and the failure counters silently stop reconciling.

Why this matters beyond bookkeeping: a replayed output of zero is *the* extreme decay case — the held route became worthless at the later block. Those are precisely the observations `p01_bps` and `tail_share` exist to capture, and they are the ones being dropped. The reported tail is therefore optimistic by an unknown amount.

Reachable paths to a zero: `fynd_core::replay_route` returning a zero `amount_out` for a drained or degenerate pool, and `step.rs:230-233`'s `biguint_to_u256` overflow branch, which logs a warning and returns `U256::ZERO`.

A fix could either add a fourth `ReplayFailure` variant for "zero output" so it is counted, or treat a zero replayed amount as -10000 bps (total loss), which is what it economically is. Either way the invariant in the doc comment should hold. Worth a test: `DecayRecord::build(&trade(10_000), measurement(Ok(U256::ZERO), Some(U256::from(9_950))))` currently yields `bps: None, failure: None`.

Separately, line 66's doc ("Returns `None` when the original quote is zero") is incomplete — it also returns `None` when either of the other two is zero.

kayibal (commented):
:robot: **Review of the `decay` subcommand**

I read the diff against `origin/main` rather than the commit message, and checked each of the PR's claims against the code.

**Overall:** the structure is good and the refactor is clean. `SteppingSolver::current_block` on the trait, `StepAdapter`/`Session` in their own modules, and the `Encoding` opt-out are all the right shape, and the extraction out of `monitor.rs` looks behaviour-preserving to me (details below). The decomposition math itself is correct and well tested. My concerns are concentrated in two places: **measurements that silently disappear from the summary**, and **the run having no guard against sliding behind chain head**.

I have left 8 inline comments. Ranked:

1. `decay/record.rs:170` — a zero amount in any of the three legs produces a record with neither `bps` nor `failure`. This silently drops exactly the worst outcomes from `p01_bps` / `tail_share`.
2. `decay/mod.rs:283` — the per-block budget check excludes `quote_round`, and one block in every `--offsets` does 50% more solver work than the PR's cost model says. The warning cannot fire on the busiest block.
3. `decay/mod.rs:58` — `--max-lag-blocks` is accepted but never enforced by `decay`. Nothing stops the run drifting arbitrarily far behind head.
4. `decay/mod.rs:289` — nothing checks that `measured_block == quote_block + offset`, so a gap or reorg puts a longer measurement into a shorter offset's bucket.
5. `decay/record.rs:30` — `PoolGone` classification depends on the exact wording of a `fynd-core` error string, with no test guarding the wording.
6. `decay/sample.rs:36` — the "`monitor` records reverted transactions" claim does not describe the code on this branch.
7. `decay/mod.rs:267` — an all-unquotable round does not count toward `--max-blocks`.
8. `decay/summary.rs:3` — the summary grows without bound for the life of the process.

---

### Methodology: the market/execution split has no noise control

This is not a bug, but it affects the headline number the tool exists to produce, so I want to flag it.

`execution_slippage_bps = (replayed - fresh) / quoted`. `replayed` is deterministic given the state — same route, same pools. `fresh` is not: it comes from a time-boxed worker-pool solve, so two solves of the same order at the same state can return different routes. All of that variance lands in the split.

The direction matters. A fresh solve is never *better* than the true optimum, so when it underperforms, `market_movement_bps` reads more negative than it should and `execution_slippage_bps` reads less negative — which **inflates `market_share` and deflates `execution_share`**, the exact ratio the summary reports and the exact ratio PR #297 is being compared against.

There is a cheap control available: measure offset 0 as well. At offset 0 the replay must equal the quote exactly (same state, same route), so any nonzero `market_movement_bps` at offset 0 is pure solver noise, and you get a per-run number to subtract or at least to report alongside the split. It costs one extra measurement per round.

A related diagnostic that is currently invisible: how often `execution_slippage_bps` is *positive* — the stale route beating a fresh solve. That is physically impossible in a static market, so its frequency is a direct read on how noisy the reference is.

### Monitor refactor

I diffed `run` and `run_session` line by line against `origin/main`. Everything moved verbatim: `advance`'s barrier, timeouts, warn intervals, `Pacing`, the head-lag check, `decode_block_when_available`, the price snapshots, the JSONL write. Two things I specifically checked and they are fine:

- `Encoding::Requested` is passed at `monitor.rs:187`, so `EncodingOptions::new(0.005)` is still applied — `ENCODING_SLIPPAGE` is the same 0.005.
- `load_pools_config` reproduces the old inline logic exactly, including the default-path-only fallback.

The `RotatingWriter` prefix claim holds: `is_comparisons_file` requires the literal `comparisons-` prefix, and `decay` writes `decay-`, so a decay run cannot re-read its own output. (It could not anyway within a single run, since `load_shapes` runs before the writer is opened — but across runs the guard is real.)

### Minor, take or leave

- `--registry` / `HINDSIGHT_REGISTRY` reach `decay` through `ChainArgs`, but `decay` never calls `load_registry()`. `tools/hindsight/CLAUDE.md` now says `decay` takes `--registry` "to load a custom address book", which is not true. Either drop the flag from the docs for this subcommand or note that it is unused.
- `order_quote_to_outcome` still builds `quote_json` via `slim_quote` for every solve, including all `2 * sample_size` solves per block in `decay`, which never reads it. Since the point of `Encoding::Skipped` was to stop paying for output nobody consumes, the same reasoning applies here.
- `decay` reuses `telemetry::record_block_seconds`, but a decay "block" measures one offset (excluding the quote round) while a monitor "block" measures a full top/back pass. Two different quantities under one Prometheus series name.
- `FEED_DEAD_TIMEOUT` in `resolve/step.rs:33` is `pub(crate)` but only used inside `step.rs`.
- `decay/mod.rs::run_session` is ~90 lines with roughly a dozen branch points. Under the 100-line limit but over the complexity-8 guidance; splitting the offset loop into its own function would bring both down.

### Tests

Good coverage overall — the round-shape tests in `decay/mod.rs` are testing behaviour, not implementation, and the mock is well built. Of the edge cases worth having: empty sample ✅, malformed JSONL ✅, feed death mid-round ✅, offsets outside the summary's range ✅. Not covered: a block gap between offsets (there is no logic for it — see comment 4), and a zero amount in any leg (see comment 1).

tools/hindsight/src/decay/mod.rs
```diff
@@ -0,0 +1,763 @@
...
+            if elapsed > state.block_budget {
+                warn!(
+                    round_id,
+                    offset,
+                    elapsed_ms = elapsed.as_millis(),
+                    budget_ms = state.block_budget.as_millis(),
+                    trades = quoted.len(),
+                    "measurement exceeded its per-block budget; lower --sample-size"
+                );
+            }
```
kayibal: 🤖  **The per-block budget warning cannot fire on the busiest block, and the PR's cost model undercounts that block by 50%.**

Two related things.

*The cost claim.* The PR says per-block work is "`--sample-size` replays plus `--sample-size` solves regardless of offset depth". That holds for offsets `1..offsets-1`. It does not hold for the block where a round ends and the next begins. Trace the loop: the first round quotes at B0 and measures B1..B5; control then returns to the top of the outer `loop` and line 249 reads `current_block()`, which is **still B5**. So B5 gets the offset-5 measurement (25 replays + 25 solves) *and* the next round's quote (25 more solves) — 75 solver operations, not 50. With the default `--offsets 5` that is one block in five running 50% hot.

*The warning.* `started` is taken at line 283, after `advance()`, and `elapsed` at line 292 covers only `measure_offset`. `quote_round` is called at line 257, outside that window, and is never budget-checked — it has its own `Instant` at line 344 used purely for a log line. So on B5, the block that actually does the most work, the budget check sees only two thirds of it.

Concrete failure on Base (2s blocks, so `block_budget` is 1s): `measure_offset` takes 0.9s and `quote_round` takes 0.9s. B5 consumes 1.8s of a 2s block. No warning fires, because the check only ever sees 0.9s. Repeat that and the run drifts behind head with the operator seeing a clean log.

The same omission affects `telemetry::record_block_seconds` at line 293 — the metric under-reports decay's real per-block cost for the same reason.

Simplest fix: start the timer before `quote_round` for the block where both happen, or check the quote round against the same budget separately. Either way the docs' "warns when a block's work exceeds half the chain's block time" should be true of the worst block, not just the typical one.

tools/hindsight/src/decay/mod.rs
```diff
@@ -0,0 +1,763 @@
...
+/// Fraction of a block's time the per-block measurement may occupy before it is warned about. Above
+/// this there is no margin left for feed jitter, and the run is one slow solve away from sliding
+/// behind head.
+const PACING_WARN_FRACTION: f64 = 0.5;
+
+/// Inputs for the decay measurement — the `decay` subcommand's CLI arguments.
+#[derive(clap::Args)]
+pub(crate) struct DecayArgs {
+    #[command(flatten)]
+    pub solver: SolverArgs,
```
kayibal, line 58: :robot: **`--max-lag-blocks` is exposed on `decay` but never enforced, so nothing stops the run sliding arbitrarily far behind chain head.**

`SolverArgs` is flattened in whole, so `hindsight decay --max-lag-blocks 50` parses and is accepted. `decay` then never reads it: the field appears nowhere in `decay/mod.rs`. Its own help text (`resolve/session.rs:73-78`) promises "beyond which the session is considered unhealthy and the solver is rebuilt", which `decay` does not do.

This is not only a dead flag. `monitor` uses that budget as its only real defence against a specific failure: the solver falls behind, Tycho's buffer keeps handing it blocks instantly, so `advance()` never errors and `FEED_DEAD_TIMEOUT` never fires, while the run quietly measures state from hours ago. `resolve/step.rs:28-33` documents exactly where that ends — buffer backpressure kills the subscriptions and nothing resubscribes.

`decay` has no equivalent guard. It never calls `peek_next_block()` at the top of its loop (as `monitor` does at `monitor.rs:238-245`), never polls the RPC head, and never compares itself to it. Its only exit is `advance()` erroring after 15 minutes of no block at all.

Concrete scenario: `--sample-size 100` on Base. Each offset block takes ~5s against a 2s block time, so the run loses 3s per block. After an hour it is ~900 blocks behind, still logging "measured offset" happily. The measurements are not *wrong* (they are block-relative), but the run is no longer sampling live market conditions, which is the premise of the whole exercise, and it is heading for the subscription death described above.

Either wire `max_lag_blocks` up (the `decay` loop already has an RPC URL in `session.args().chain.rpc_url`), or remove the flag from `decay`'s surface so it does not promise something it does not do.

tools/hindsight/src/decay/mod.rs
```diff
@@ -0,0 +1,763 @@
...
+                return SessionEnd::Unhealthy(e.to_string());
+            }
+            let started = Instant::now();
+            let Some(measured_block) = adapter.current_block().await else {
+                warn!(round_id, offset, "no applied block after advancing; abandoning the round");
+                state.totals.abandoned_rounds += 1;
+                break;
+            };
+            let at = RoundAt { round_id, quote_block, measured_block, offset };
+            measure_offset(adapter, state, &quoted, at).await;
```
kayibal: :robot: **Nothing checks that `measured_block == quote_block + offset`, so a gap or reorg files a longer measurement under a shorter offset.**

`advance()` (`resolve/step.rs:159-200`) waits for a block *strictly newer* than the previous one, not for exactly the next one. If the Tycho stream skips a block on a reorg or a resync, one `advance()` can move the applied state forward by more than one block. `measured_block` at line 284 will reflect that, but line 289 files the measurement under `offset` regardless, and `Summary::record` buckets purely on `offset`.

That this happens in practice is already acknowledged in this codebase — `monitor.rs:301-311` warns explicitly when the applied block is not the expected target, for the same reason.

Concrete failure: the stream applies N, then N+2. The measurement taken after the first `advance()` is a *two-block* decay, but it lands in the offset-1 bucket. Since the whole thesis is that decay is front-loaded, and offset 1 is the number that carries it, contaminating offset 1 with two-block moves overstates exactly the figure the tool is built to report.

The JSONL keeps both `quote_block` and `measured_block`, so the raw data is recoverable offline — but the per-offset summary logged at exit, which is the primary output, is not.

`measured_block - quote_block` is right there at line 289. Comparing it against `offset` and either warning or skipping the measurement would be a couple of lines.

tools/hindsight/src/decay/record.rs
```diff
@@ -0,0 +1,335 @@
...
+impl ReplayFailure {
+    /// Classify a `reexecute` failure string. `fynd_core::ReplayError::MissingState` renders as
+    /// "no simulation state for component …", which is the one case we separate out.
+    pub(crate) fn classify(reason: &str) -> Self {
+        if reason.contains("no simulation state for component") {
+            Self::PoolGone
+        } else {
+            Self::SimulationFailed
+        }
+    }
```
kayibal, line 44: :robot: **`PoolGone` depends on the exact wording of a `fynd-core` error message, with nothing guarding that wording.**

The substring matched here comes from `#[error("no simulation state for component {0}")]` on `ReplayError::MissingState` (`fynd-core/src/replay.rs:39`). It matches today. But nothing connects the two: someone rewording that `thiserror` attribute — a change that looks entirely cosmetic and that no test in `fynd-core` covers — would silently reclassify every `PoolGone` as `SimulationFailed`.

That is a quiet failure. The counters still add up, the run still finishes, and the operator sees `pool_gone = 0` and concludes routes are never being invalidated by pools leaving the feed — which, per this enum's own doc comment at line 11-13, is the signal being looked for.

`ReplayError` is `pub`, so the cheap guard is a test in this module that constructs the real error and asserts the classifier handles it:

```rust
let rendered = format!(
    "re-execution failed: {}",
    fynd_core::ReplayError::MissingState("0xabc".to_string())
);
assert_eq!(ReplayFailure::classify(&rendered), ReplayFailure::PoolGone);
```

The existing test at line 280-290 asserts against a hand-written string, which passes whether or not `fynd-core` still produces it.

Better still would be threading the typed error through `SteppingSolver::reexecute` instead of stringifying it at `step.rs:155`, but that is a bigger change than this PR needs.

tools/hindsight/src/decay/sample.rs
```diff
@@ -0,0 +1,354 @@
...
+/// (`"status": "reverted"`), where they are explicitly `null` — a revert settles no amounts, so
+/// there is nothing to decode. Those rows are well-formed and expected, not corrupt, so they must
+/// parse and then be dropped for having no shape rather than counted as malformed.
+#[derive(Deserialize)]
+struct ComparisonRow {
+    venue: Option<String>,
+    token_in: Option<String>,
+    token_out: Option<String>,
+    amount_in: Option<String>,
+}
```
kayibal, line 45: :robot: **The reverted-record rationale does not describe the code on this branch.**

This comment states that "`monitor` also records **reverted** transactions (`\"status\": \"reverted\"`), where they are explicitly `null`". I could not find that anywhere in this branch. `comparison_record` (`resolve/jsonl.rs:155-196`) is the only producer of `comparisons-*.jsonl`, and it emits no `status` field at all; `token_in`, `token_out` and `amount_in` are non-optional there and always written. The fixture at line 262 also carries `"cause"` and `"decoder":"reverted"`, neither of which `comparison_record` produces.

So the shape being defended against comes from the decoding stack (#388 / #371 / #391) that the PR description says is not merged. Against this codebase, the test at line 259 asserts the handling of a record no writer produces, and the PR description's claim that "the sampler tolerates reverted-transaction records" describes future behaviour as present behaviour.

The `Option<String>` fields themselves are fine and I would keep them — parsing JSONL from a long-running appender defensively is correct regardless. It is the justification that is wrong, and it is the kind of comment that sends the next reader looking for a writer that does not exist. Either reword to "records may carry null trade fields" without attributing it to `monitor`, or land this after the decoding stack so the claim becomes true.

tools/hindsight/src/decay/mod.rs
```diff
@@ -0,0 +1,763 @@
...
+                block = quote_block,
+                round_id, "no sampled trade could be quoted; skipping the round"
+            );
+            for _ in 0..cfg.offsets {
+                if let Err(e) = adapter.advance().await {
+                    return SessionEnd::Unhealthy(e.to_string());
+                }
+            }
+            continue;
+        }
```
kayibal, line 318: :robot: **A round where nothing quotes does not count toward `--max-blocks`, so a run can advance forever without terminating.**

`state.totals.blocks` is only incremented at line 305, inside the offset loop. This branch advances `cfg.offsets` blocks and then `continue`s without touching it, so the `--max-blocks` check at line 314 never sees them.

If the solver can never quote the sampled shapes — a `--venue` whose pairs the market does not cover, a token set the solver has no pools for, a `--comparisons-dir` from a different chain — every round lands here and `hindsight decay --max-blocks 100` runs until it is killed, burning blocks and logging the same warning. `--max-blocks` reads as a hard bound on the run; here it is not one.

Counting these blocks in `totals.blocks` would fix the termination. It would also be worth failing the run outright after N consecutive fully-unquotable rounds, since that state is a configuration error rather than something that recovers.

Minor, related: even in the normal case `--max-blocks N` consumes N+1 blocks (the first quote block is never counted), and one extra per round after that. Worth a word in the flag's help text if the count is meant to be exact.

tools/hindsight/src/decay/summary.rs
```diff
@@ -0,0 +1,324 @@
+//! Per-offset aggregation of a decay run's measurements.
+//!
+//! Accumulates every measurement in memory, which a run of any realistic length affords: at the
+//! default sample size a day of Base produces ~1.1M measurements, or ~1.7 MB of `f64` per offset.
+
```
kayibal, line 10: :robot: **The summary accumulates for the life of the process, which is a long time for this command.**

Every complete measurement pushes three `f64`s into `OffsetSamples` and nothing ever drops them. The note here says "a run of any realistic length affords" it, but the arithmetic is heavier than the comment suggests: the "~1.7 MB per offset" figure counts one of the three vectors. At the stated ~1.1M measurements per day on Base that is closer to **26 MB/day** across all offsets and vectors.

That is genuinely fine for a day. It is less fine given that this command shares its whole session and rebuild-forever design with `monitor`, which is built to run unattended for weeks — three weeks lands around 550 MB, all of it live because `stats()` needs the full sorted slice for percentiles.

Not a blocker, and nothing here needs redesigning for the first runs. But the comment should reflect the real number, and if runs are expected to go past a few days a streaming quantile sketch (t-digest / P²) or a periodic flush-and-reset would remove the ceiling entirely.

kayibal:
:robot: **Correction to my earlier review comment on the market/execution split, plus two concrete proposals.**

My summary comment argued that fresh-solve noise biases `execution_slippage_bps` *toward zero* and inflates `market_share`. After checking this against the two collected decay datasets (450k Ethereum records, 705k Base), **that directional claim is wrong**, and one of my supporting arguments was invalid. Walking both back.

### What I got wrong

I treated the fact that a fresh solve usually returns *more* than the original quote as evidence of an artifact. It isn't. The fresh solve returns the best of many candidate routes, and the original route is one of the candidates, so random movement in the *other* routes can only push that maximum up. `fresh >= replayed` is the structural expectation, not a symptom.

I also framed cases where a better fresh route exists as contamination. They are not — if a better route genuinely exists at the later block, then holding the stale route did cost that difference, which is exactly what this subcommand sets out to measure.

### What does hold up

Only the wrong-sign cases: `execution_slippage_bps > 0`, meaning the fresh solve came back with **less** than replaying the original route. The fresh solve was free to pick that route, so this is the solver failing to find something it had found a few blocks earlier.

| | Ethereum | Base |
|---|---|---|
| wrong-sign share of all measured records | 3.5% | 5.9% |
| wrong-sign share of records where the number is non-zero | 20.0% | 22.4% |
| median magnitude of those | 1.68 bps | 1.21 bps |

### Why this matters more than I first thought — the sign, not the size

These records carry positive values, and positives cancel negatives inside an average. They were wiping out **49%** of the genuine measured cost on Ethereum and **74%** on Base. Removing them roughly doubles the reported decay on Ethereum and quadruples it on Base:

| mean `execution_slippage_bps` | as computed today | wrong-sign cases removed |
|---|---|---|
| Ethereum | -0.2209 | **-0.4328** (1.96x) |
| Base | -0.0487 | **-0.1889** (3.88x) |

So the current output *understates* the cost of holding a stale route.

### Proposal 1 — clamp `execution_slippage_bps` at zero

A real in-block solver holds the original route and would take the better of the two, so its outcome is `max(replayed, fresh)` and a worse fresh solve gains nothing. Under that model the correct value is `min(0, (replayed - fresh) / quoted)`.

Clamping is preferable to dropping the record (clamped -0.4328 vs dropped -0.4486 on Ethereum, -0.1889 vs -0.2007 on Base — nearly identical): clamping keeps the trade in the sample contributing exactly zero, which is what actually happens, whereas dropping shrinks the population and makes per-offset counts non-comparable.

Worth noting the published summary table is already effectively clamped — computing "average gain per quote" as total gains over all quotes is the same operation. It is `DecayBps` / `Summary` that are not, so the JSONL and the logged summary currently disagree with the table.

### Proposal 2 — record the fresh solve's route identity

`DecayRecord::algorithm` stores only `trade.quote.algorithm` (record.rs:125, 189). Because the fresh solve's route is not recorded, two very different situations are indistinguishable in the data:

- the fresh solve found a genuinely different, better route (real, measurable gain), versus
- the fresh solve returned the same route with a different number (solver inconsistency).

Adding the fresh solve's algorithm — or better, its route — makes the remaining noise measurable instead of inferred. This is cheap now and unanswerable retroactively for data already collected.

### One residual bias, in the opposite direction

Because the fresh solve takes a maximum over many route valuations, an alternative that merely *looks* better through estimation error wins that maximum and is reported as achievable gain. That inflates the measured cost. It is smaller than the sign flips, it cannot be sized without proposal 2, and it partly offsets the understatement above — so the corrected figures should be read as better, not exact.

My original inline comment about a zero amount silently dropping both `bps` and `failure` (record.rs:163-175) is unaffected by any of this and still stands.

tools/hindsight/src/decay/record.rs
kayibal: Worth fixing, if that ever happens I think it is an edge case and I would prefer it to be marked as a failure instead of distorting statistics.

tools/hindsight/src/decay/mod.rs
kayibal: The next round should simply start at B6, not at B5.

tools/hindsight/src/decay/mod.rs
kayibal: Wire it up.

tools/hindsight/src/decay/mod.rs
kayibal: Skip reorged blocks

tools/hindsight/src/decay/record.rs
kayibal: Yeah add the cheap test guard.

tools/hindsight/src/decay/sample.rs
kayibal: Reword the comment

tools/hindsight/src/decay/mod.rs
kayibal: Make it respect max-blocks regardless of how many quotes we produced

tools/hindsight/src/decay/summary.rs
kayibal: Update the comment accordingly.

PR #409, opened 2026-08-06
fix: log and exit non-zero when the tycho feed stops

brunoguerios:
40 of 43 fynd shutdowns in the last 48h ended the Tycho stream *cleanly*: `TychoFeed::run` broke its
loop, returned `Ok(())`, the spawn wrapper logged nothing, `tycho_feed_failures_total` did not
increment, and the process exited 0 (`reason: Completed`). The dominant prod failure mode was
invisible to both logs and metrics.

Now any end of the stream — clean or error — returns
`StreamError("<stream> stream ended (last block: N)")`. The existing spawn wrappers in `solver.rs`
log it at ERROR and increment `tycho_feed_failures_total` with no change. The feed arm of
`FyndRPC::run` sets `fatal_error`, so the process exits 1 and k8s records `reason: Error`. The gas
arm had the same defect and gets the same one-line fix.

Not in scope: reconnect-with-backoff inside `TychoFeed`, and the upstream tycho-client
`RevertPositionNotFound` fix (tycho repo) that causes most of these stream ends.
`run_with_step_controller` keeps its clean end — that end is deliberate test behavior.

Before prod rollout: dashboards or alerts that key on `last_terminated_reason="Completed"` for fynd
pods now see `Error`, and Loki queries that count the INFO line `Protocol stream ended` must switch
to the `tycho feed error` ERROR line or to the metric.

Local CI green (fmt, clippy, doc, 1270 tests). Not yet verified on staging — that needs a forced
stream end on one chain to confirm the exit code and the log line.

### DIFF
--- a/fynd-core/src/feed/tycho_feed.rs
+++ b/fynd-core/src/feed/tycho_feed.rs
@@ -105,12 +105,28 @@ impl TychoFeed {
         self.event_tx.clone()
     }
 
+    /// Builds the terminal error for a stream that ended without an error. The feed's job is
+    /// to never end, so a clean end is still a failure. The last block makes the log line
+    /// self-sufficient during triage.
+    async fn stream_ended_error(&self, stream_name: &str) -> DataFeedError {
+        let last_block = match self
+            .market_data
+            .read()
+            .await
+            .last_updated()
+        {
+            Some(block) => block.number().to_string(),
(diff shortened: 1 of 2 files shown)


azf20:
bumped into this issue when tycho stopped yesterday afternoon

PR #412, opened 2026-08-07
feat: Add an offline benchmark and a profiling tool

carloszanella:
This runs over a selected dataset of orders that needs to be downloaded (its not committed).

Two entry points over the same setup:

  ./scripts/bench.sh --name my-run          compare configs, write a report
  ./scripts/profile.sh --config X --order N one config, one thread, no files

It includes the logic to generate benchmark results, a set of algorithm configurations, and a visualizer for the data. Results are saved per run and gitignored.

I included a readme with run instructions


fynd-core/benches/common/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+/// # Errors
+///
+/// Returns the reason the config could not be used — missing, unparseable, or without an
+/// `algorithm` key — so the caller can print it and record it in the report rather than inventing
+/// its own wording.
+pub fn load_bench_config(label: &str) -> Result<BenchConfig, String> {
+    let path = configs_dir().join(format!("{label}.toml"));
+    let contents = std::fs::read_to_string(&path)
+        .map_err(|error| format!("{} could not be read: {error}", path.display()))?;
+    let pool_fields: toml::Table = toml::from_str(&contents)
```
louise-poole: Small nit-pick: we are trying to be hyperconscious about the fact that 'pool' can be ambiguous here in a solver's codebase. Whenever we refer to worker pools like this we try to always say `worker_pool` and not rely on the code context for someone to figure out what pool this is. It has saved our sanity in the long run.
```suggestion
    let worker_pool_fields: toml::Table = toml::from_str(&contents)
```

fynd-core/benches/common/mod.rs
```diff
@@ -0,0 +1,433 @@
...
+
+    let algorithm = pool_fields
+        .get("algorithm")
+        .and_then(|value| value.as_str())
+        .ok_or_else(|| format!("{} has no `algorithm` key", path.display()))?
+        .to_string();
+    let max_hops = pool_fields
+        .get("max_hops")
+        .and_then(|value| value.as_integer())
+        .unwrap_or(0) as usize;
```
louise-poole: 0 hops seems like a bad fallback value no? In the readme it says defaults are provided for omitted configs. In the rest of the code we use 3 hops as a default.

fynd-core/benches/algorithm_bench.rs
```diff
@@ -0,0 +1,1184 @@
...
+        // Dropped before the next is built, so no two configs ever solve at the same time.
+        drop(solver);
+    }
+
+    assert!(
+        results
+            .first()
+            .is_some_and(|(config, _)| config.label == BASELINE),
+        "the baseline {BASELINE} could not be built, so nothing can be compared"
+    );
```
louise-poole, line 1392: Could we not assert this earlier? Maybe in the `Err` arm above? Then we abort as soon as this errors and we don't waste time waiting for all solvers to start up etc. 
And maybe add the failure reason to the message?

fynd-core/benches/configs/decomposition_d2.toml
louise-poole, line 1: Should this be pushed for an algorithm that isn't being pushed?

louise-poole (approved):
Minor comments, otherwise looks good!

fynd-core/benches/configs/decomposition_d2.toml
carloszanella: prob added by mistake

PR #413, opened 2026-08-07
feat(rpc): serialize /v1/prices token prices as decimal strings

tvinagre:
Serializes `/v1/prices` token prices as exact decimal strings (BigUint long-division, up to 17 significant digits, plain notation, magnitude always preserved) instead of `f64`, so the wire format has no cross-language float-rendering ambiguity. Price entries are sorted by token address for a deterministic response.

Marks the endpoint `x-experimental` in the OpenAPI spec, generates the committed spec and TypeScript schema with the experimental feature, and enables the feature in the Dockerfile build so the hosted binary serves the route.

Tokens whose price cannot be serialized (zero or oversized numerator/denominator) are skipped and counted in a single warn per request.


fynd-rpc/src/api/prices.rs
```diff
@@ -7,6 +7,12 @@ use serde::{Deserialize, Serialize};
 use tycho_simulation::tycho_common::models::Address;
 use utoipa::{IntoParams, ToSchema};
 
+/// Version of the unit contract used by [`TokenPriceEntry::price`].
+pub const PRICE_UNIT_CONTRACT_V1: &str = "PRICE_UNIT_CONTRACT_V1";
```
tvinagre: These two consts are only referenced from test code (and the openapi test in `mod.rs` asserts the literal strings instead of using them). If the contract identifier never appears in the response, they are public API with no runtime purpose — either serialize them in `PricesResponse` (my preference, see review summary) or move them into the test module.

fynd-rpc/src/api/prices.rs
```diff
@@ -100,7 +106,11 @@ pub struct TokenPriceEntry {
     /// Token address.
     #[schema(value_type = String, example = "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48")]
     pub token: Address,
-    /// Price relative to gas token as a float.
+    /// `rawTokenUnitsPerRawGasUnit`: raw target-token units divided by raw gas-token units.
+    ///
+    /// This approximate `f64` follows `PRICE_UNIT_CONTRACT_V1` and is intended only for display
+    /// and analytics. Consumers must normalize both tokens' decimals before using it.
+    #[schema(example = 0.000000003)]
     pub price: f64,
```
tvinagre: Freezing `f64` as the V1 wire type bakes in the weakest part of the contract:

- `ComponentDepthEntry.depth` in the same response is already a decimal string, so the response is split between two representations.
- The server holds the exact rational (`numerator`/`denominator`) and discards precision here; tokens whose ratio under/overflows f64 are dropped entirely rather than represented.
- The fixture's `rawPrice` byte-equality assertions pin serde_json's float rendering (`1500.0`), which other languages do not share — JS renders the same double as `"1500"`, and `1e21` as `"1e+21"` vs serde_json's `"1e21"`. So "byte-identical" only holds for Rust consumers.

A decimal-string field (like `depth`) would remove the formatting coupling, keep precision, and make the unit contract trivial to state. If f64 must stay for compatibility, consider at least comparing parsed values in the fixture tests instead of strings (see comment in `handlers.rs`).

fynd-rpc/src/api/prices.rs
```diff
@@ -148,17 +159,181 @@ pub fn price_to_f64(
     if denominator.is_zero() {
         return None;
     }
-    let n = numerator.to_f64()?;
-    let d = denominator.to_f64()?;
-    Some(n / d)
+    let price = numerator.to_f64()? / denominator.to_f64()?;
+    if price.is_finite() && price > 0.0 {
```
tvinagre: Two behavior changes worth surfacing in the PR body:

- Previously a zero numerator produced `Some(0.0)` and the token appeared in the response with `"price": 0.0`; now it is dropped with a warn. Fine for display purposes, but it is a wire change for anyone already consuming the experimental endpoint.
- The old code could return `Some(inf)` when `BigUint::to_f64` saturates to infinity instead of returning `None` — the `is_finite()` check fixes that regardless of num-bigint version behavior. Good catch.

fynd-rpc/src/api/prices.rs
```diff
@@ -148,17 +159,181 @@ pub fn price_to_f64(
...
+                        .expect("rejected case needs an error")),
+                    "{}",
+                    case.name
+                ),
+                other => panic!("unknown fixture expectation {other}"),
+            }
+        }
+    }
+
+    fn validate_fixture_case<'a>(
```
tvinagre: This function re-implements the *frontend's* validation rules in a Rust test: `missing_decimals`, `decimal_scale_out_of_range`, `duplicate_address`, and `gas_token_mismatch` are checks the server never performs (the response does not even carry decimals). For those cases the test validates the fixture against rules written a few lines above — it exercises no production code and will silently diverge from the real validator in fynd-frontend.

Suggest trimming the fixture (and this test) to server-observable behavior: the accepted serialization cases plus the non-finite/non-positive rejections that `price_to_f64` actually enforces. The frontend-only cases can live in the frontend repo, next to the code they describe.

fynd-rpc/src/api/handlers.rs
```diff
@@ -435,6 +440,143 @@ mod tests {
...
+            let response_entry = response_prices
+                .iter()
+                .find(|entry| {
+                    entry["token"]
+                        .as_str()
+                        .unwrap()
+                        .eq_ignore_ascii_case(expected_address)
+                })
+                .unwrap_or_else(|| panic!("missing handler response for {}", case["name"]));
+            assert_eq!(
```
tvinagre: `response_entry["price"].to_string()` compared against the fixture string pins serde_json's exact float formatting (`3e-9`, `1500.0`). That formatting is an implementation detail of serde_json's ryu integration, not part of any spec — a serde_json bump could change it and this test would fail with no behavioral change on the wire. Parsing both sides to `f64` and comparing for equality tests the same thing without the coupling.

.github/workflows/ci.yaml
```diff
@@ -87,7 +87,8 @@ jobs:
           # to main, which changes the generated spec's version field. Regenerating and committing
           # openapi.json as part of the release would require a second post-release commit, so we
           # accept version drift here and only check that the API surface itself hasn't changed.
-          cargo run --locked -- openapi | jq 'del(.info.version)' > /tmp/openapi.json
+          cargo run --locked --features fynd-rpc/experimental -- openapi \
```
tvinagre, line 90: Generating the committed spec with `--features fynd-rpc/experimental` means `clients/openapi.json` and the generated TS client now advertise `/v1/prices` to every consumer of `@kayibal/fynd-client`, while a default (non-experimental) build 404s it — and does the hosted gateway route it at all? If the spec should describe the maximal build, consider marking experimental operations in the spec (e.g. an `x-experimental` extension or a note in the endpoint description) so client users know the endpoint may not exist on the server they are pointed at.

tvinagre (commented):
Reviewed together with the frontend consumer (propeller-heads/fynd-frontend#24). The validation hardening in `price_to_f64` is a real fix (the old code could return `Some(inf)` when `BigUint::to_f64` saturates), and consolidating spec generation behind `openapi_spec()` removes duplicated merge logic. My main concerns are with the shape of the contract itself, not the code.

### 1. The contract never appears on the wire — should it?

`PRICE_UNIT_CONTRACT_V1` exists as doc prose, two `pub const`s referenced only by tests, and a fixture file the frontend pins byte-for-byte (SHA-256 + commit hash). Nothing in the actual `/v1/prices` response identifies the unit or contract version. Two consequences:

- The frontend cannot verify the contract at runtime — fynd-frontend#24 has to *stamp* `unitContract: "PRICE_UNIT_CONTRACT_V1"` onto responses in its proxy rather than assert it. If semantics ever change (V2), consumers keep displaying wrong numbers with no signal.
- Every fixture change becomes a two-repo lockstep dance (JSON + `.sha256` + pinned commit constant on the frontend side), and the pinned commit goes stale the moment this branch is squash-merged.

Suggestion: put the unit in the response itself, e.g. a `price_unit: "raw_token_units_per_raw_gas_unit"` (or a contract-version string) field on `PricesResponse`. Then the OpenAPI spec — already drift-checked in CI and used to generate the TS client — is the single contract artifact, consumers assert instead of trust, and the shared fixture can shrink to serialization cases or go away.

### 2. Is f64 the right thing to freeze as a "stable" contract?

See inline comment on `TokenPriceEntry.price`. Short version: the same response already uses decimal strings for `depth`; the frontend immediately re-stringifies the float into BigNumber; and the byte-identical fixture actually pins serde_json's float formatting, which JS does not share (`String(1500.0)` → `"1500"`, not `"1500.0"`). If we are declaring V1 now, this is the moment to pick the wire type we want to live with.

### 3. Fixture scope

About half the fixture cases (`missing_decimals`, `overflow_scale`, `duplicate_address`, `gas_token_mismatch`, `anchorUsdAssumption`, the `expectedTokenUsd` fields) describe frontend normalization the server neither produces nor enforces — see inline comment on `validate_fixture_case`.

Smaller notes inline. One behavior note worth adding to the PR body: tokens with a zero price were previously serialized as `0.0` and are now dropped from the response entirely.

tvinagre:
## Review Summary

Deep re-review of the branch (three parallel analysis passes + manual verification; codex and gemini CLIs not installed on this machine, so the external passes were skipped). Fixes pushed as d63d1d8f.

### Findings

| # | Severity | Finding | Resolution |
|---|----------|---------|------------|
| 1 | P1 | `truncate_sig_digits` cut digits off instead of zeroing them, shrinking any price with >17 significant integer digits by 10x–10^84x (10^17 → `"10000000000000000"` = 10^16; 10^100 → 10^16). Four tests pinned the corrupted values, so CI stayed green. | Fixed: truncation now zero-pads to preserve magnitude; the four expectations corrected; boundary tests added (17/18-digit integer parts with fractions, 10^100, 10^399). |
| 2 | P1→removed | The `PRICE_UNIT_CONTRACT_V1` machinery (consts, `contract_version`/`price_unit` response fields, cross-repo byte-pinned fixture, fixture-driven tests) added coupling without a consumer that needs it. | Removed per review decision: unit semantics live in the OpenAPI schema description; if semantics ever change, version the path (`/v2/prices`) instead of a wire marker. Fixture deleted; handler test now uses inline cases. |
| 3 | P2 | Fractional zeros after a non-zero integer part didn't consume significant-digit budget: `(10^30+1)/10^30` serialized with 31 significant digits (up to 401-char strings), violating the documented 17-digit cap. | Fixed: zeros are significant once any non-zero digit precedes them; added an invariant test asserting ≤17 significant digits across both paths. |
| 4 | P2 | Per-token `warn!` in `get_prices` re-fires for the same tokens on every request (log flood at steady traffic). | Fixed: one `warn!` per request with a `skipped_tokens` count; per-token detail demoted to `debug!`. |
| 5 | P3 | `prices` array had non-deterministic HashMap order (varies per process/block), unlike the sorted `spot_prices`/`component_depths`. | Fixed: sorted by token address. |
| 6 | P3 | Operand-size guard rendered both BigUints to strings just to count digits (two allocations per token per request). | Fixed: bit-length bound (`1328 bits ≈ 400 digits`), constant time. |
| 7 | P3 | `unwrap_or(0)`/`unwrap_or('0')` in the long-division digit loop would silently corrupt a digit if the loop invariant ever broke. | Fixed: single `expect` stating the invariant — fails fast instead of corrupting output. |
| 8 | P3 | Warn message named impossible causes (non-finite/non-positive are f64 concepts; the only reachable cause is oversized operands, since `Price::new` panics on zeroes). | Fixed: message reworded; defensive-guard rationale documented in tests. |
| 9 | P4 | `price_to_f64` left dead by the decimal-string switch (no production callers), plus its 7 tests. | Fixed: removed; `fynd-rpc/CLAUDE.md` updated. |
| 10 | P3 | `limit` doesn't apply to `prices`. | Dismissed: documented behavior — `limit` is specified for `spot_prices`/`component_depths` only, and the frontend consumer needs the full price set. |
| 11 | P3 | `x-experimental` doesn't survive into `schema.d.ts`. | Dismissed: openapi-typescript drops vendor extensions; the marker is present in `clients/openapi.json`, which is the artifact spec consumers read. |
| 12 | P3 | Hosted spec advertises `/v1/{chain}/prices`; Dockerfile now builds the hosted binary with `experimental` on (compile-time, no runtime flag). | Deliberate — the goal is serving prices to the frontend. Open item: confirm the gateway actually routes `/v1/{chain}/prices` before relying on the hosted docs. |

Note for fynd-frontend#24: the fixture and its sha256/commit pinning were removed here, and the response no longer carries `unitContract`-style fields — the frontend's byte-identical fixture test and stamped `unitContract` field should be simplified accordingly. Its proxy parser must now read `price` as a JSON string rather than a number.

### Verification

- **Tests**: fynd-rpc 101/101 passed (nextest, `--all-features`), fynd bin 6/6 passed
- **Lint**: `cargo +nightly clippy --locked --all --all-features --all-targets -- -D warnings` clean
- **Format**: `cargo +nightly fmt` clean
- **Docs**: `RUSTDOCFLAGS="-D warnings" cargo doc` clean
- **OpenAPI drift**: regenerate + diff against committed spec — clean
- **TS client**: `tsc --noEmit` clean

### Commit

`d63d1d8f` feat(rpc): drop price unit contract and fix decimal truncation

tvinagre:
Follow-up from the review's re-verification pass, pushed as 665d675f: the public doc comment on `price_to_decimal_string` used intra-doc links to private consts, which fails `RUSTDOCFLAGS="-D warnings" cargo doc` when the `experimental` feature is enabled. Neither check.sh nor CI documents with that feature, so the pipeline never caught it. Links replaced with plain code spans; `cargo doc --all-features -D warnings` now passes.

One open item from the findings table remains a deploy-owner decision, restated for visibility: the Dockerfile now builds the production image with `fynd-rpc/experimental` unconditionally, permanently shipping `/v1/prices` — the CI spec-drift fix alone did not require this. Intentional for the frontend-prices rollout, but it deserves an explicit sign-off before merge.

fynd-rpc/src/api/prices.rs
tvinagre: Resolved by d63d1d8f — the consts and the whole contract machinery were removed.

fynd-rpc/src/api/prices.rs
tvinagre: Resolved by 2bbb17f2 + d63d1d8f — price is now an exact decimal string (BigUint long division, 17 significant digits, magnitude-preserving), matching the string representation depth already uses.

fynd-rpc/src/api/prices.rs
tvinagre: Addressed — the skip behavior is documented in the PR body, aggregated to a single warn per request in d63d1d8f, and the zero guards are kept as defensive (Price::new panics on zeroes).

fynd-rpc/src/api/prices.rs
tvinagre: Resolved by d63d1d8f — the fixture and its circular tests were removed; the handler test now uses inline, server-observable cases.

fynd-rpc/src/api/handlers.rs
tvinagre: Moot as of 2bbb17f2 + d63d1d8f — prices are exact decimal strings, so no serde_json float rendering is involved on the wire or in tests.

.github/workflows/ci.yaml
tvinagre: Addressed — /v1/prices is marked x-experimental in the committed spec (2bbb17f2), and the production image now builds with the feature, so deployed servers actually serve the advertised endpoint. Gateway routing of /v1/{chain}/prices is tracked in the PR comments.

tvinagre:
Closing the deploy sign-off item (confirmed by Thales): the unconditional `fynd-rpc/experimental` build stays. The endpoint is a read-only view over derived data that every build already computes (`HealthTracker` consumes it unconditionally), so the feature only registers the route — no added runtime cost. Hosted deployments sit behind the gateway, matching the tokens-endpoint rollout (#416). Remaining infra check before the frontend flips its flag: gateway routing of `/v1/{chain}/prices`.

fynd-rpc/src/api/prices.rs
```diff
@@ -134,98 +149,417 @@ pub struct ComponentDepthEntry {
...
+/// with at most `max_sig_digits` significant digits (trailing zeros removed from the
+/// fractional part, no scientific notation).
+///
+/// For values below one, leading fractional zeros do not consume significant-digit budget:
+/// for a value like 1/700 (0.00142...), the two leading zeros after the decimal point are
+/// skipped, and the full 17 significant digits are produced. Once any non-zero digit
+/// precedes them, fractional zeros are significant and do consume budget.
+///
+/// Example: `(3, 1_000_000_000, 17)` -> `"0.000000003"`, `(1500, 1, 17)` -> `"1500"`.
+fn biguint_division_to_decimal_string(
```
brunoguerios, line 193: This feels like the kind of thing you'd delegate to some math lib or similar source. Feels a bit error prone to do this "by hand".

brunoguerios (approved):
LGTM ✅ 

Added a small comment on the math to calculate the price, but if tests are passing, I wouldn't say it's a blocker.

PR #414, opened 2026-08-10
feat: Graph performance improvements

carloszanella:
I added a comment going over the changes: https://propeller-heads.atlassian.net/wiki/x/AQA15w

There is one larger commit: [feat: Implement MostLiquid V2, designed with token sequences](https://github.com/propeller-heads/fynd/pull/414/commits/cb6e963c799602d490ae844bc4c1a7eea2d494f1)

The others should be easy to review in isolation, so I suggest reviewing one commit at a time

## Benchmark results:

Baseline:
image

New solve times
image


fynd-core/src/algorithm/most_liquid.rs
```diff
@@ -587,130 +748,71 @@ impl Algorithm for MostLiquidAlgorithm {
...
+                let route = Self::build_route(&ctx, token_path, &solved)?;
+                if let Err(e) = route.validate() {
+                    trace!(error = %e, "best route failed validation");
+                    report.validation_failures += 1;
+                    None
+                } else {
+                    let amount_out = swap_on_route(&route, token_prices.as_deref(), &gas_price);
+                    Some(RouteResult::new(route, amount_out, gas_price.clone()))
+                }
+            }
```
louise-poole, line 655: And if the winner fails to build/validate? We should fallback to trying the next best route no?

fynd-core/src/algorithm/most_liquid.rs
```diff
@@ -188,219 +371,231 @@ impl MostLiquidAlgorithm {
...
         let mut current_amount = amount_in.clone();
-        let mut swaps = Vec::with_capacity(path.len());
+        let mut legs: SmallVec<[SwapOutcome; INLINE_EDGES]> = SmallVec::new();
+        let mut gas = BigUint::ZERO;
+        // What a unit of gas costs in the token the route pays out, which is the last leg's own
+        // output token. `None` when that token has no price.
+        let mut route_gas_price = None;
 
-        // Track state overrides for components we've already swapped through.
-        let mut state_overrides: HashMap<&ComponentId, Box<dyn ProtocolSim>> = HashMap::new();
```
louise-poole, line 361: Why did you drop the state overrides? Sometimes a pool is re-used at different hops (thinking mainly for 3 token pools etc) and without these overrides we'll simulate the second hop on that pool as if it wasn't swapped on already which is incorrect.

fynd-core/src/algorithm/most_liquid.rs
```diff
@@ -188,219 +371,231 @@ impl MostLiquidAlgorithm {
...
 
-        Ok(paths)
+    /// Sets whether a pair's simulation results are reused across one order's routes.
+    ///
+    /// On, a pair is asked once and the answer serves every route crossing it. Off, every route
+    /// asks every pool on the pair at the amount reaching it, which is exact but much slower.
+    pub fn with_cache_pair_swaps(mut self, enabled: bool) -> Self {
+        self.cache_pair_swaps = enabled;
+        self
     }
```
louise-poole: This looks related to the simulations overrides I mentioned already, but looks unused?

louise-poole (commented):
Did a quick look through this morning - these are the first things I've noticed. Nothing else jumped out at me beyond this though. Will do a proper review next week

carloszanella:
I force pushed after your comments just to rebase with latest main, didn't change any of the commits of this PR in the force push. all new changes you haven't reviewed are outside the force-push changes

fynd-core/src/algorithm/most_liquid.rs
carloszanella: that's a valid question. I looked into this and the potential validation reasons. They are:
- swaps are empty: if the "best" is an empty list of swaps then we have no graph. it would've failed earlier
- split validation: does not apply
- The other two are: DisconnectedSwaps and UnsupportedCycle: from what I checked they are not possible by design on the algorithm and they would also imply a bug in graph logic

I can add a loop around this if you prefer. I made it like this given the above and the extra code to "solve the next best" which would be unreachable if I understood it correctly

fynd-core/src/algorithm/most_liquid.rs
carloszanella: I judged it is not worth the extra code plus performance impact of cloning things and cache misses. I added some extra code to filter paths where the same pool shows up twice

fynd-core/src/algorithm/most_liquid.rs
carloszanella: atm unused. Id like to expose this at the config level at some point. removing this code for now

fynd-core/src/algorithm/mod.rs
```diff
@@ -141,13 +143,13 @@ impl AlgorithmConfig {
     /// When set, only these tokens may appear between `token_in` and `token_out`
     /// in a multi-hop route. The order endpoints are always allowed regardless.
     /// Pass an empty set to disallow all intermediate hops (only 1-hop routes possible).
-    pub fn with_connector_tokens(mut self, tokens: HashSet<Address>) -> Self {
+    pub fn with_connector_tokens(mut self, tokens: FxHashSet<Address>) -> Self {
```
louise-poole: Hmmm... is changing the public API here super necessary? This repo is actively used by many external users so I'm hesitant to alter any public APIs unless the change can be motivated as strongly necessary.

fynd-core/src/solver.rs
```diff
@@ -521,7 +522,7 @@ impl FyndBuilder {
     }
 
     /// Sets component IDs to exclude from the Tycho stream.
-    pub fn blocklisted_components(mut self, components: HashSet<String>) -> Self {
+    pub fn blocklisted_components(mut self, components: FxHashSet<String>) -> Self {
```
louise-poole: More public interface changes...

fynd-core/src/graph/token_graph.rs
```diff
@@ -0,0 +1,1364 @@
...
+            .ok_or_else(|| GraphError::TokenNotFound(from.clone()))?;
+        let to_ix = self
+            .get_token_ix(to)
+            .ok_or_else(|| GraphError::TokenNotFound(to.clone()))?;
+        Ok(self.paths_between_ix(from_ix, to_ix, filter))
+    }
+
+    /// Every token path from `from` to `to` within the filter's hop bounds.
+    ///
+    /// Empty when there is no route. Check [`TopologyGraph::expand_path`] for expanded pool paths.
```
louise-poole, line 170: Personally I'm not such a fan of the `[` `]` in our doc strings. It breaks things up a bit for me as a human reader. The '`' characters are enough to indicate code is being referenced.

fynd-core/src/graph/token_graph.rs
```diff
@@ -0,0 +1,1364 @@
...
+    /// to its left when it wraps.
+    ///
+    /// That product is unbounded — four legs of twenty pools is 160,000 routes — so `max_paths`
+    /// caps how many are written. The ones past the cap are dropped in counting order, which no
+    /// ranking has seen yet, so a cap trades routes the caller might have wanted for a bound on
+    /// what one sequence can allocate. `None` writes them all.
+    pub fn expand_path(
+        &self,
+        token_path: &[NodeIndex],
+        max_paths: Option<usize>,
```
louise-poole, line 202: Many of this places this is called use `None` here. I might have missed it, bit I didn't actually spot any callsites that set this value. Is it necessary to keep this logic here that no-one uses?

fynd-core/src/graph/token_graph.rs
```diff
@@ -0,0 +1,1364 @@
...
+    /// Routes come back shortest first. Within a length they are in no particular order.
+    pub fn paths_between_ix(
+        &self,
+        from: NodeIndex,
+        to: NodeIndex,
+        filter: &GraphQueryFilter,
+    ) -> Vec<TokenPath> {
+        if filter.min_hops == 0 || filter.min_hops > filter.max_hops {
+            return Vec::new();
+        }
```
louise-poole, line 181: Would it not be better to error on a misconfiguration like this? Returning an empty vec just silently swallows a configuration error.

fynd-core/src/algorithm/most_liquid.rs
```diff
@@ -739,11 +910,126 @@ impl Algorithm for MostLiquidAlgorithm {
...
+        pools: &[EdgeData<D>],
+        mut simulate: impl FnMut(&ComponentId) -> Option<(GetAmountOutResult, BigInt)>,
+    ) -> Option<HopResult> {
+        if let Some(choice) = self.pairs.get(&pair) {
+            if let Some(outcome) = choice.outcomes_by_amount.get(amount_in) {
+                return Some(outcome.clone());
+            }
+
+            // The pool was simulated but not at this amount. We assume that amounts dont move
+            // so much so we assume the same winner wins again at a slightly different amount.
```
louise-poole, line 992: I feel this might be a risky assumption. Maybe we should buffer by how much it can differ? The top pool for a $0.1 trade can differ greatly to the top pool for a $100k trade.

louise-poole: Oh wait - we're caching the simulations per-order no? Then my comment above doesn't apply. Ignore this 😅

louise-poole (approved):

fynd-core/src/graph/token_graph.rs
carloszanella: In the IDE it allows you to jump to the code definition or even read the docs of the function as a tool tip. Not sure if you knew that?

fynd-core/src/graph/token_graph.rs
carloszanella: I will review this but I think I am using it in the waterfill PR

PR #416, opened 2026-08-10
feat(rpc): add experimental GET /v1/tokens graph token endpoint

tvinagre:
Serves token metadata for exactly the tokens present in the routing graph, so the frontend can stop querying hosted Tycho's ~350K-token universe.

## What it does

- `GET /v1/tokens` (behind the `experimental` feature, like `/v1/prices`) returns every graph token with `address`, `symbol`, `decimals`, `tax`, `gas`, `quality`, plus two ranking signals:
  - `component_count` — number of components (pools) containing the token
  - `liquidity` — approximate routable liquidity in raw gas-token units: the sum of the token's directional component depths, each converted via its token gas price. Documented as sort/display-only; omitted for tokens without a computed gas price, which rank after priced ones by component count.
- The response is pre-sorted (liquidity desc, component count desc, address) and paginated with `offset`/`limit` (default limit 1000): `?limit=100&offset=1000` returns tokens ranked #1001–#1100, and an offset past the end yields an empty page. `total` carries the full list size and `block` the token-prices computation block; pages are consistent while `block` is unchanged, so clients restart from offset 0 when it advances mid-pagination.
- The list is computed lazily on first request per derived-data update — one pass over the component topology and depths — and cached in `AppState`. Nothing runs on the quote path; nothing runs when the endpoint is unused. Returns 503 until derived data is ready.
- `fynd-core` now re-exports the `ComponentDepths`/`TokenGasPrices` aliases its store getters already expose in public signatures.

The contract lives in the OpenAPI annotations and on the wire; the committed spec is unchanged because it excludes experimental endpoints (same as `/v1/prices`).

## Validation

- 7 new tests: ranking-fold unit tests (degree counting, depth×price summation, unpriced ordering, registry/topology mismatches) and handler tests (ranked response, cache + limit + offset paging + past-the-end offset, 503 before readiness)
- `cargo nextest`: 85 fynd-rpc tests pass, 747 across fynd-rpc + fynd-core (the recorded-market integration test fails locally due to the known Git-LFS fixture decode issue, unrelated)
- clippy `-D warnings`, fmt, doc build, and OpenAPI drift check all clean

The frontend consumer PR is propeller-heads/fynd-frontend#27.

### DIFF
--- a/fynd-rpc/src/api/handlers.rs
+++ b/fynd-rpc/src/api/handlers.rs
@@ -389,9 +389,14 @@ const DEFAULT_TOKENS_LIMIT: usize = 1000;
 /// (descending), then `component_count`, then address. The list is recomputed lazily at
 /// most once per derived-data update and cached; nothing runs on the quote path.
 ///
+/// Paginate with `offset`/`limit` (e.g. `?limit=100&offset=1000` returns tokens ranked
+/// #1001-#1100). Pages are consistent while the response `block` is unchanged; restart
+/// from offset 0 when it advances mid-pagination.
+///
 /// # Query Parameters
 ///
 /// - `limit` - Maximum number of tokens returned (default: 1000)
+/// - `offset` - Number of tokens to skip from the start of the ranked list (default: 0)
 #[utoipa::path(
     get,
     path = "/v1/tokens",
@@ -410,6 +415,7 @@ pub async fn get_tokens(
     let limit = query
         .limit
         .unwrap_or(DEFAULT_TOKENS_LIMIT);
+    let offset = query.offset.unwrap_or(0);
 
     let cache_key = {
         let store = state.derived_data.read().await;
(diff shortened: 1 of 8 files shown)


fynd-rpc/src/api/handlers.rs
```diff
@@ -361,6 +365,110 @@ pub async fn get_prices(
...
+
+    let entries = {
+        let market = state.market_data.read().await;
+        build_token_entries(
+            &market.component_topology(),
+            market.token_registry_ref(),
+            depths.as_ref(),
+            token_prices.as_ref(),
+        )
+    };
```
brunoguerios: Claude suggests releasing the lock before building entries so it doesn't hold the lock for too long.
```suggestion
    let (topology, token_registry) = {
        let market = state.market_data.read().await;
        (market.component_topology(), market.token_registry_ref().clone())
    };
    let entries =
        build_token_entries(&topology, &token_registry, depths.as_ref(), token_prices.as_ref());
```

brunoguerios (approved):
Added a small improvement suggestion, but other than that it LGTM ✅

fynd-rpc/src/api/handlers.rs
tvinagre: Applied in c365e343 — topology and registry are snapshotted under one read guard and the ranking fold runs outside it, so the per-block feed writer is never blocked by the recompute. The two maps stay mutually consistent since they come from the same guard; depths/prices already came from the separate derived-data lock.

PR #420, opened 2026-08-11
feat: Emit more detailed metrics on solver quotes

carloszanella:
Have a per order quote metric emission for each worker pool.

Do some refactoring in the module to make this possible:
- add struct with a new field to replace tuple
- abstract some functions

I will probably have to add more observability into Fynd solvers in the near future - e.g. protocol usage


brunoguerios (commented):
LGTM ✅

brunoguerios (approved):

PR #421, opened 2026-08-12
fix(rpc): invert liquidity unit conversion in GET /v1/tokens

tvinagre:
Fixes the token ranking reported as "not ordered by usefulness" on mainnet.

## The bug

`TokenGasPrices` stores **token raw units per gas-token raw unit**, so converting a raw depth to gas units divides by that ratio (`depth * denominator / numerator`) — the convention `most_liquid.rs` documents and follows. `build_token_entries` multiplied instead, so the reported `liquidity` was `correct × price²`: the ranking was dominated by decimals and inverse token value, floating cheap 18-decimal tokens to the top and burying USDC/USDT/WBTC. The endpoint tests had encoded the inverted arithmetic as their expectations, which is why they passed.

## The fix

- `price_ratio_f64` reworked into `gas_units_per_token_unit` (returns `denominator / numerator`), so the unit direction is stated once where the arithmetic lives, with the same finite/positive skip guard via the reciprocal.
- Corrected the two tests that pinned the inverted values (USDC fixture: 400 raw at 2-per-gas is 200 gas units, not 800) and the doc comments on `GraphTokenEntry::liquidity` / `build_token_entries`.
- New regression test: two tokens with equal true gas-unit liquidity but prices 18 orders of magnitude apart must score equal (within 1e-9 relative) and fall through to the component-count tie-break. Under the inverted code their scores differ by 36 orders of magnitude. Verified the test fails on the reverted code.
- OpenAPI artifacts regenerated (the field description is embedded in the spec); the pre-existing `info.version` drift on main is left as-is since CI strips it.

Known approximation, unchanged here: multi-asset pools contribute one directional depth per direction, inflating absolute values for Curve-style pool tokens. Independent of this bug.

## Validation

`cargo nextest` 109 passed, clippy `-D warnings` clean, fmt clean, OpenAPI drift check clean.

### DIFF
--- a/clients/openapi.json
+++ b/clients/openapi.json
@@ -564,7 +564,7 @@
               "null"
             ],
             "format": "double",
-            "description": "Approximate routable liquidity in raw gas-token units: the sum of this token's\ndirectional component depths converted via its gas price. Approximate `f64`,\nintended for sorting and display only. Absent when the token has no computed\ngas price."
+            "description": "Approximate routable liquidity in raw gas-token units: the sum of this token's\ndirectional component depths, each divided by the token's gas price (token raw units\nper gas-token raw unit). Approximate `f64`, intended for sorting and display only.\nAbsent when the token has no computed gas price."
           },
           "quality": {
             "type": "integer",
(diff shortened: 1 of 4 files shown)


brunoguerios (approved):
LGTM ✅

PR #422, opened 2026-08-12
fix(hindsight): veto fee-on-transfer trades

tamaralipows:
A fee-on-transfer token's contract splits transfers, sending the fee to its own wallet. Netting cannot see this: the fee lands inside the settled amounts, and Tycho's simulation does not support fee-on-transfer tokens, so the re-solve quotes as if the full amounts moved fee-free and every such trade records a fake win. On Polygon, a token with a 5% transfer fee read as a constant ~525 bps improvement on all of its trades. Tycho scores that token `quality: 100, tax: 0`, so the solver's `min_token_quality` filter cannot catch it upstream.

Such trades are now vetoed, on both the sell and the buy side. Registered venue fee collectors stay exempt (their fees are backed out downstream), and router-paid partner fees never match the shape — the settled price keeps the solver's own fee, per policy.


louise-poole (approved):
Good catch!

PR #430, opened 2026-08-17
refactor: improve exclusive liquidity request latency

louise-poole:
**Optimisation**: Quote ranking narrows candidates by borrow instead of cloning, removing one deep copy of every candidate's route from the request path.

**Metrics**: Makes the time a request spends waiting on each worker pool visible in metrics:
  - `worker_router_pool_response_seconds` (new) — per-pool latency as the router sees it: queue wait, readiness wait and the solve, recorded for failures as well as successes.
  - `worker_pool_readiness_wait_seconds` (new) — time spent waiting for derived data before solving.
  - `worker_pool_solve_duration_seconds` — now labelled by `outcome`, and recorded for every outcome rather than successes only.

### DIFF
--- a/fynd-core/src/worker_pool/worker.rs
+++ b/fynd-core/src/worker_pool/worker.rs
@@ -45,14 +45,41 @@ fn record_task_pickup_metrics(pool_name: &str, queue_wait: Duration, queue_depth
 }
 
 /// Records per-pool solve latency: one algorithm's own working time for one order, excluding
-/// queue wait. Unlike `worker_router_solve_duration_seconds`, which times the router racing every
-/// pool and so belongs to no single pool, this is attributable per pool.
+/// queue wait and readiness wait. Unlike `worker_router_solve_duration_seconds`, which times the
+/// router racing every pool and so belongs to no single pool, this is attributable per pool.
 ///
-/// Successful solves only — a pool that exhausts its timeout returns before this point and is
-/// counted in `worker_router_solver_failures_total{error_type="timeout"}` instead.
-fn record_solve_duration(pool_name: &str, solve_time: Duration) {
-    metrics::histogram!("worker_pool_solve_duration_seconds", "pool" => pool_name.to_string())
-        .record(solve_time.as_secs_f64());
+/// Recorded for every outcome, labelled by `outcome`. A histogram of successes alone hides the
+/// cost of the failures the request still waited for, which is how a pool can show a healthy p95
+/// while setting the request's.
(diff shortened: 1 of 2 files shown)


fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -523,8 +538,7 @@ impl WorkerPoolRouter {
                 // Timeout reached
                 _ = tokio::time::sleep_until(deadline_instant) => {
                     // Mark all remaining worker pools as timed out
-                    let elapsed_ms = deadline.saturating_duration_since(Instant::now())
-                        .as_millis() as u64;
+                    let elapsed_ms = start_time.elapsed().as_millis() as u64;
```
Troshchk, line 541: Nice fix!

fynd-core/src/worker_pool/worker.rs
```diff
@@ -45,14 +45,41 @@ fn record_task_pickup_metrics(pool_name: &str, queue_wait: Duration, queue_depth
...
+
+/// Classifies a solve outcome for the `outcome` label on `worker_pool_solve_duration_seconds`.
+fn solve_outcome(result: &Result<OrderQuote, SolveError>) -> &'static str {
+    match result {
+        Ok(_) => "success",
+        Err(SolveError::NotReady(_)) => "not_ready",
+        Err(SolveError::Timeout { .. }) => "timeout",
+        Err(SolveError::NoRouteFound { .. }) => "no_route",
+        Err(SolveError::InsufficientLiquidity { .. }) => "insufficient_liquidity",
+        Err(_) => "error",
```
Troshchk, line 81: why not to use `solver_error_label` for the error matching?
Is it ok, that we wrap some errors into simple "error" string?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -209,29 +209,61 @@ pub(crate) struct OrderResponses {
...
+    quotes: Vec<&'a WorkerPoolQuote>,
+    failed_solvers: &'a [(String, SolveError)],
+}
+
+/// Records how long a worker pool took to answer, as observed by the router.
+///
+/// Unlike `worker_pool_solve_duration_seconds`, which a worker reports for its own successful
+/// solves, this covers everything the request actually waits for — queue wait, readiness wait,
+/// and the solve — and is recorded for failures too (`outcome`). The gap between the two is
+/// what a request pays for a worker pool beyond its algorithm's own working time.
```
Troshchk, line 257: Would it make sense to add that the pools are worker pools?

Troshchk (commented):
Thank you!
Left a couple of qs

PR #431, opened 2026-08-18
feat(hindsight): solver-first decoding

tamaralipows:
Read me instead: https://propeller-heads.atlassian.net/wiki/spaces/GEN/pages/3897393161/Hindsight+Refactor


tools/hindsight/src/decoder/declared.rs
```diff
@@ -0,0 +1,450 @@
...
+/// Only the settling solver is asked — the outermost known solver in the trace — so a record's
+/// amounts and its solver label always come from the same solver. A transaction that merely
+/// touches another solver's router somewhere is not read by that solver.
+///
+/// `Ok(None)` means no solver was read at all — no known solver frame, or the settling solver's
+/// own data carried nothing it could parse — so the caller falls back to netting.
+///
+/// `Err(veto)` means the transaction is dropped, and netting is not tried. Either the solver said
+/// it is not a swap, or it named an output that the transfers do not show: once a solver has told
+/// us the trade, netting would answer a different question, so its answer is not substituted.
```
carloszanella: sounds either claude docs need an update or the code

---

**[BIZ `#4`]**

**What changed**: When the settling solver declares a swap but the recovered output is zero or below the declared `min_amount_out`, the transaction is vetoed (`Veto::OutputNotFound`) and netting is never tried — the trade vanishes from the output entirely rather than appearing as a netted record. Transactions that produced a record before this PR can now produce nothing.

**Where**: `tools/hindsight/src/decoder/declared.rs:32–37,64,94–112`, `tools/hindsight/src/decoder/mod.rs:253`

- `tools/hindsight/CLAUDE.md:149` (post-merge) and the design doc both still say both guards "fall through to the netting fallback" — the code does the opposite for the floor guard, and the quote guard now drops only the quote. One of the two needs correcting.
- Whether the veto is silent: an off-basis or misparsed `min_amount_out` in any solver decoder now deletes real trades with no counter, unlike the old fallback path.

tools/hindsight/src/decoder/attribution.rs
```diff
@@ -0,0 +1,485 @@
...
+    if let Some((venue, fee)) = fee_venue(registry, ledger, flow.token_in, flow.token_out) {
+        if let VenueFee::Output(amount) = fee {
+            flow.amount_out = flow.amount_out.saturating_add(amount);
+        }
+        return Some(venue);
+    }
+    integrator
+        .and_then(|tag| registry.venue_for_integrator(tag))
+        .map(str::to_string)
+}
```
carloszanella, line 127: worth checking whether we are actually not considering fees from amount in and whether it is the right thing to do

---

**[BIZ `#5`]**

**What changed**: `fee_collectors` is gone from every venue section. A venue fee is now corrected only when the venue was identified *by* its fee wallet, and only on the output side; an input-side fee is never subtracted from `amount_in`. Because `venue()` returns early on an owner or CoW `appData` match, a trade attributed by those fingerprints keeps an output fee inside its amounts even when the fee wallet is in `[venue_fees]`.

**Where**: `tools/hindsight/src/decoder/attribution.rs:91–113,131–153`, `tools/hindsight/src/decoder/registry/ethereum.toml:166–170`

- Coinbase Base App: the registry comment at `ethereum.toml:168–169` says its 0.95% sell-token cut "is netted out of the input", which the code no longer does. For its netted records `amount_in` is now ~95 bps larger than what reached the pools, which biases the comparison in Fynd's favour.
- `DecodedTrade`'s field docs at `tools/hindsight/src/decoder/mod.rs:84–88` still describe `amount_in`/`amount_out` in terms of `venue_fee_in`/`venue_fee_out`, fields that no longer exist.

tools/hindsight/src/decoder/mod.rs
```diff
@@ -353,13 +284,12 @@ impl<P: Provider> Decoder<P> {
...
-        // the quote check reads the grossed output. The appData tag is read from a batch settler's
-        // calldata; other transactions carry none.
-        let integrator = solvers::integrator(logs);
-        let app_data = intents::venue_tag(registry, entry_point, &root.input);
-        let venue = venue_attribution::attribute(
+        // A venue fingerprint (owning trader, CoW appData tag, fee wallet, or integrator tag —
+        // see `attribution`) overrides the entry-point label. The appData tag is read from a
+        // batch settler's calldata; other transactions carry none.
+        let integrator = solvers::lifi::integrator(logs);
+        let app_data = solvers::cow::venue_tag(registry, entry_point, &root.input);
```
carloszanella: sounds reasonable design redflag

---

**[DESIGN `#1`]**

**Description**: `decode_transaction` calls `solvers::lifi::integrator(logs)` and `solvers::cow::venue_tag(registry, entry_point, &root.input)` by name, so the top-level orchestrator depends on two leaf solver modules directly and the docs' claim that a solver needs "one row in `solvers::IMPLEMENTATIONS`" is only true for solvers that contribute nothing but a `declared` read. The dependency also points the wrong way: `mod.rs` now has to know that LiFi emits integrator tags and that CoW is the only settler with an `appData` field, which is exactly the venue-specific knowledge the refactor set out to push down. Maps to the `placement` directive.

**Suggestion**: Give the fingerprint reads the same registry-joined dispatch the declared read got — either a second defaulted method (`fn venue_fingerprint(&self, input: &[u8], logs: &[Log]) -> Option<VenueTag>`) resolved through `Registry::solver(...)`, or a small `const FINGERPRINTS: &[(&str, fn(...) -> Option<VenueTag>)]` table next to `IMPLEMENTATIONS`, so `mod.rs` names no solver.

tools/hindsight/src/decoder/attribution.rs
carloszanella, line 127: looks like this can distort amounts and also hints at some better design

---

**[DESIGN `#2`]**

**Root cause**: `venue` takes `flow: &mut SettledSwap` and grosses a matched fee wallet's receipt back into `flow.amount_out` (line 106) as a side effect of deciding a label. Because the correction lives inside the fingerprint chain, it is reached only when the fee-wallet fingerprint is the one that wins, and it is applied without regard to whether the fee already sits inside the recorded output. Maps to the `encapsulation` directive.

- **Architecture** *(was DESIGN `#2`)*: the module doc says "Attribution runs after decoding and only labels the record… Nothing here affects whether a trade decodes", which the `&mut` parameter contradicts, `tools/hindsight/src/decoder/attribution.rs:91–96`
- **Behavior — ordering** *(was DESIGN `#2`)*: an owner match (line 98) or a CoW `appData` match (line 101) returns before `fee_venue` runs, so a venue identified by one of those fingerprints never gets its output-side fee grossed back; the amount depends on which label won a race that has nothing to do with fees, `tools/hindsight/src/decoder/attribution.rs:97–103`
- **Behavior — double count** *(was IMPL `#4`)*: when the declared `output_recipient` is the router that received the gross output and then paid the fee wallet out of it, `recover_output`'s receipt already contains the fee and adding it again inflates `amount_out`, `tools/hindsight/src/decoder/attribution.rs:104–109`
- **Documentation** *(was IMPL `#4`)*: `README.md:158` states "on a netted flow the fee is backed out of the amounts; on a declared flow it is recorded only", but the add-back is unconditional on tier — and `README.md:138–141` in the same file says the opposite, `tools/hindsight/README.md:138–141, 158`

**Plain English**: For trades where a known wallet took the platform's cut, the recorded settled output can be inflated by that cut twice — once because the recipient's receipt already included it, and once because it is added back afterwards. For other trades the same cut is never added back at all, purely because the tool happened to recognise the venue by a different clue first. The comparison then shows the competing router delivering more or less than it did, and the dashboard reports a win or a loss that is not real. The correction should either apply to every trade of a given tier or to none, and never twice.

**Proposed fix**: Split the two jobs — keep `venue(...) -> Option<String>` pure over `&SettledSwap`, and expose the fee-wallet lookup as its own `fn output_fee(registry, ledger, token_out) -> Option<U256>` that `decode_transaction` runs unconditionally and applies itself, gated on the one documented rule (declared vs netted) once code and README agree on it. Making the correction the caller's step removes the `&mut`, decouples it from fingerprint ordering, and puts it at the one place that knows which tier produced `amount_out` and therefore whether the fee is already inside it.

tools/hindsight/src/report/mod.rs
```diff
@@ -54,6 +61,18 @@ pub(crate) fn run(args: ReportArgs) -> anyhow::Result<()> {
...
+/// `decode` column predates the marker and is kept either way.
+fn filter_netted(records: Vec<Comparison>, include_netted: bool) -> Vec<Comparison> {
+    if include_netted {
+        return records;
+    }
+    records
+        .into_iter()
+        .filter(|record| record.decode.as_deref() != Some("netted"))
+        .collect()
+}
```
carloszanella, line 74: always in favour of enums

---

**[DESIGN `#3`]**

**Root cause**: The PR's headline concept — the two decode tiers — is carried as string literals rather than a type, and it is stored twice: `decode` at `mod.rs:304` (`if declared.is_some() { "declared" } else { "netted" }`) is fully derivable from the sibling `decoder` field (`"solver-calldata"`, `"solver-logs"`, `"venue-netting"`, `"sender-netting"`, `"intent-netting"`). Maps to the `naming` and `yagni` directives.

- **Architecture** *(was DESIGN `#3`)*: the tier is copied through `RangeComparison::decode` and consumed by string comparison at `report/mod.rs:72` (`record.decode.as_deref() != Some("netted")`); a typo in any of the four spellings compiles and silently changes what the report includes, against the repo's own "enums for state machines, not boolean flags" guidance, `tools/hindsight/src/report/mod.rs:64–74`
- **Complexity** *(was SIMPLIFY `#6`)*: the same literals reappear at `resolve/mod.rs:624` and `report/mod.rs:166`, so one closed set is spelled out in six places, `tools/hindsight/src/resolve/mod.rs:624`
- **Naming** *(was DOCS `#10`)*: `decoder` and `decode` are adjacent JSONL columns one letter apart whose names do not distinguish them, and `decoder`'s doc says "see `decode`", which reads as a typo for itself, `tools/hindsight/src/decoder/mod.rs:74–80`

**Proposed fix**: Make `decoder` an enum (`DeclaredFromCalldata`, `DeclaredFromLogs`, `VenueNetting`, `SenderNetting`, `IntentNetting`) with `#[serde(rename = …)]` preserving the current wire strings, give it a `tier()` method that yields the declared/netted value, and have `report::filter_netted` match the variant. One type then owns the closed set, the second field stops being independently writable, and the six scattered literals collapse to the rename attributes. Rename the columns while you are there (`decode_source` / `decode_tier`) so the JSONL is readable without the source. Flag for the human: `report/record.rs`'s `decode: Option<String>` reads historical JSONL, so typing *that* field would turn an unrecognised value into a deserialize error instead of a silently kept record — keep the reader side a `String` or decide deliberately.

tools/hindsight/src/decoder/registry.rs
```diff
@@ -91,44 +92,37 @@ struct AddressBook {
...
+    pub(crate) decoder: &'static dyn crate::decoder::solvers::SolverDecoder,
+}
+
+impl std::fmt::Debug for Solver {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Solver")
+            .field("name", &self.name)
+            .finish_non_exhaustive()
     }
 }
```
carloszanella, line 119: also sounds like unnecessary trait complexity

---

**[DESIGN `#4`]**

**Description**: All nine impls override `declared`, so the default body exists solely to make `NoDecoder` compile, and `NoDecoder` exists solely to give `decoder_for` something non-`Option` to return. The trait is one stateless method on nine unit structs, and it also costs a hand-written `Debug` impl on `Registry::Solver` (lines 113–119) because `dyn SolverDecoder` isn't `Debug`. Maps to the `yagni` directive.

**Suggestion**: Make it `type DeclaredRead = fn(&[u8], &[Log]) -> Result<Option<DeclaredSwap>, Veto>;` with `IMPLEMENTATIONS: &[(&str, DeclaredRead)]` and `Solver { name, declared: Option<DeclaredRead> }`. Pros: deletes the trait, the default body, `NoDecoder`, and the manual `Debug`; `Option` states "book-only solver" honestly; `#[derive(Debug)]` works again. Cons: a future second method would need a struct-of-fn-pointers or the trait back; `fn` items are less discoverable than an `impl` block; loses the `Send + Sync` bound as documentation (fn pointers are both anyway).

tools/hindsight/src/decoder/attribution.rs
```diff
@@ -0,0 +1,485 @@
...
+                .copied()
+                .filter(|amount| !amount.is_zero())
+        };
+        if let Some(fee) = non_zero(&token_out) {
+            return Some((venue.clone(), VenueFee::Output(fee)));
+        }
+        if let Some(fee) = non_zero(&token_in) {
+            return Some((venue.clone(), VenueFee::Input(fee)));
+        }
+    }
```
carloszanella, line 177: yeah seems liek very similar functions that could be grouped

---

**[DESIGN `#6`]**

**Root cause**: `received_by(&HashSet<Address>) -> HashMap<Address, U256>` (many addresses, all tokens) has exactly one production caller, `fee_venue`, which needs one address and two tokens. The sibling `received_by_address(recipient, token)` (line 180) is already that query; the two differ in granularity on both axes for no reason.

- **Architecture** *(was DESIGN `#6`)*: the caller passes `HashSet::from([*wallet])` and then reads back exactly two tokens — a signature the only user has to work around, `tools/hindsight/src/decoder/attribution.rs:138–152`
- **Performance** *(was IMPL `#9`)*: per trade the whole transfer list is walked once per wallet in `[venue_fees]`, with a `HashSet` and a `HashMap` allocated each time, to answer a two-token question, `tools/hindsight/src/decoder/attribution.rs:132–154`
- **Complexity** *(was SIMPLIFY `#4`)*: deleting the method and its test removes about 20 lines and one `HashSet` import, `tools/hindsight/src/decoder/transfer_ledger.rs:223–237`

**Proposed fix**: Delete `received_by` and have `fee_venue` call `received_by_address(*wallet, token_out)` and `received_by_address(*wallet, token_in)` directly. The narrower query is the one the caller wanted, so the workaround, both allocations, and the whole-list rescan all go with it. Note the deleted method uses `+=` while `received_by_address` saturates; that only differs on a `U256` overflow that cannot occur here.

tools/hindsight/src/decoder/netting.rs
```diff
@@ -0,0 +1,354 @@
...
+    // Batch settlements and log-matched intent fills: the sender acts for the trader. A
+    // frame-matched transaction through an unknown wrapper has no other trader to find, so it
+    // falls through to the sender's own flow.
+    if let Some(flow) =
+        find_intent_trade(provider, transfer_ledger, &[entry_point, sender], registry, code_cache)
+            .await
+    {
+        return Some(("intent-netting", flow));
+    }
+    sender_flow(transfer_ledger, sender, entry_point).map(|flow| ("sender-netting", flow))
```
carloszanella: up to you to validate whether these redisual amounts are parsed into swaps or not, which si what claude is suggesting\

---

**[IMPL `#1`]**

**Category**: bug

**Description**: `fallback_flow` ends with an unconditional `sender_flow(transfer_ledger, sender, entry_point)`, which is reached for batch settlers and log-matched intent fills whenever `find_intent_trade` finds no clean-net EOA. On the base branch the Intent role's decoder list was `[cow::CowSettlement, intents::netting::IntentNetting]` with no sender fallback, precisely because — as `Registry::is_batch_settler`'s own doc still says — a settler's sender is a solver whose net flow is not a single swap. For a multi-order CoW batch (which `cow.rs` declines) this now nets the solver's, or the settlement contract's, residual buffer flow and records it as a trade. Maps to the `reviewability` directive: the refactor drops a deliberate guard and no commit body mentions it.

**Plain English**: When a batch settles several signed orders in one transaction, the tool can no longer read any single one of them, so it falls back to measuring what moved in and out of the settler's own account. That leftover balance change is inventory and rounding, not anybody's trade, but it is recorded as a settled swap with real token amounts and then compared against a fresh quote — producing wins or losses for trades that never existed. Before this change those transactions were simply skipped. A transaction entered through a batch settler should stay skipped when no individual order can be identified.

**Suggestion**: Restrict the trailing `sender_flow` fallback to the frame-matched-unknown-wrapper case it was written for — return `None` when `registry.is_batch_settler(entry_point)` (and for the log-matched filler case) instead of netting the sender.

tools/hindsight/src/decoder/declared.rs
```diff
@@ -0,0 +1,450 @@
...
+    transfer_ledger::{SettledSwap, TransferLedger},
+    veto::Veto,
+};
+
+/// Decode a transaction from the settling solver's own declaration: the decoder label recorded on
+/// the trade, the settled swap, and the terms the solver declared alongside it.
+///
+/// Only the settling solver is asked — the outermost known solver in the trace — so a record's
+/// amounts and its solver label always come from the same solver. A transaction that merely
+/// touches another solver's router somewhere is not read by that solver.
```
carloszanella, line 31: **[IMPL `#5`]**

**Category**: bug

**Description**: The walk checks a node then recurses into `calls` in order, so a solver sitting deep inside the first child's subtree is returned ahead of a solver that is a direct child later in the list. `declared_flow`'s contract ("the outermost known solver in the trace") and its guarantee that a record's amounts and solver label come from the same settling solver therefore do not hold for a transaction that touches two solvers on different branches.

**Plain English**: When one transaction passes through two known routers on separate call branches, the tool asks whichever one it reaches first while walking down, which can be the deeper, incidental one rather than the router that actually settled the trade. The record then carries one router's name next to another router's amounts, and a rejection rule belonging to the wrong router can throw the trade away. The router that settled the trade is the outer one, and that is the one that should be read.

**Suggestion**: Make the walk breadth-first (or track depth and keep the shallowest match) so the outermost solver frame wins, and keep the doc comment as the specification.

tools/hindsight/src/decoder/solvers/okx.rs
```diff
@@ -0,0 +1,194 @@
...
+/// OKX's sentinel for native ETH, normalized to the zero address like every other flow.
+const OKX_NATIVE_ETH: Address = address!("0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE");
+
+fn normalize_native(token: Address) -> Address {
+    if token == OKX_NATIVE_ETH {
+        Address::ZERO
+    } else {
+        token
+    }
+}
```
carloszanella: **[SIMPLIFY `#1`]**

**Root cause**: `0xEeee…EEeE` is an ecosystem-wide convention, not a per-solver fact, but each solver module declares its own constant for it and its own three-line normalizer. This PR took the duplication from three copies to six. Maps to the `dry` and `naming` directives.

- **Complexity** *(was SIMPLIFY `#1`)*: identical `normalize_native` bodies in `okx.rs:40–48`, `oneinch.rs:53–61`, `paraswap.rs:51–59`, `zeroex.rs:65–74`, `kyberswap.rs:26–27, 67–73`, `cow.rs:84, 124–130`, `tools/hindsight/src/decoder/solvers/paraswap.rs:51–59`
- **Naming** *(was DOCS `#16`)*: `OKX_NATIVE_ETH`, `ONEINCH_NATIVE_ETH`, `PARASWAP_NATIVE_ETH`, `COW_NATIVE_ETH` and `ZEROEX_NATIVE` all hold the same value; the per-solver prefixes claim a distinction that does not exist, `ZEROEX_NATIVE` breaks the suffix pattern its four siblings share, and the `_ETH` suffix is wrong on the BSC and Polygon books, `tools/hindsight/src/decoder/solvers/zeroex.rs:65–74`

**Proposed fix**: Put one `pub(crate) const NATIVE_TOKEN_SENTINEL` and one `pub(crate) fn normalize_native` in `solvers/mod.rs` and delete the five copies — about 40 lines, six constants, and five misleading names in one change, since a single shared constant has no per-solver prefix to get wrong. Nothing is lost: the shared function has the same body each copy already has.

tools/hindsight/src/decoder/attribution.rs
```diff
@@ -0,0 +1,485 @@
...
+}
+
+/// Which side of the swap a venue took its fee from, with the amount.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum VenueFee {
+    /// Skimmed off the input before the swap, so the settled input is smaller than the user spent.
+    Input(U256),
+    /// Taken out of the output after the swap, so the settled output is larger than the user kept.
+    Output(U256),
+}
```
carloszanella: not sure wyh we are not reading venue fee input? maybe related to the other commenbt about input fees

---

**[SIMPLIFY `#2`]**

**Root cause**: `fee_venue` computes an input-side fee amount and returns it in `VenueFee::Input(fee)`, but `venue` only matches `VenueFee::Output(amount)` — the payload is built and discarded, and only the variant's existence (venue matched) is used. Dropping the input-side correction is deliberate and `attribution.rs:86–90` documents it, but the address-book comments were not updated to match. Maps to the `reviewability` directive.

- **Complexity** *(was SIMPLIFY `#3`)*: an enum with a payload that has no reader, plus the doc block explaining why, `tools/hindsight/src/decoder/attribution.rs:115–122`
- **Design** *(was DESIGN `#12`)*: the type states a distinction the code no longer acts on, `tools/hindsight/src/decoder/attribution.rs:104–109`
- **Documentation** *(was IMPL `#10`, PERSONAL `#2`)*: `registry/ethereum.toml:167–169` and `registry/base.toml:94–96` still tell the reader Coinbase's 0.95% sell-token cut "is netted out of the input rather than added to the output", which nothing does any more; the base branch had `test_fee_wallet_input_side_fee_nets_the_input_down` covering exactly that, `tools/hindsight/src/decoder/registry/ethereum.toml:166–170`

**Proposed fix**: Return `Option<(String, Option<U256>)>` where the `Option<U256>` is the output-side fee only, delete the enum and its doc block, and correct the Coinbase comment in every chain's book to say the input-side cut is left inside `amount_in`. Removing the payload forces the address-book comments to be revisited in the same change, which is where the stale promise lives. Nothing is lost — the input-side branch still has to run so an input-side fee wallet still identifies the venue.

carloszanella (commented):
tried to make sense of the review items and posted the ones that looked relevant. up to you to judge.

---

Four further accepted findings (DESIGN `#9`, IMPL `#2`, SIMPLIFY `#3`, SIMPLIFY `#8`) are inline in a follow-up review — their lines had moved since this review was taken.

---

**[IMPL `#7`]**

`swapSimpleMode` is declared in the `sol!` block but never decoded — `declared` only tries `swapCall`. Either add the branch (simple-mode swaps currently fall to netting) or delete the declaration.

tools/hindsight/src/decoder/netting.rs
```diff
@@ -0,0 +1,354 @@
...
+}
+
+/// Net the sender's flow. When the sender nets nothing, fall back to the contract the transaction
+/// entered through (`tx.to`), for the rare shape where the swap output is delivered to that
+/// contract rather than back to the sender.
+pub(crate) fn sender_flow(
```
carloszanella: **[SIMPLIFY `#9`]**

Both are only called from within `netting.rs` and its tests; plain `fn` narrows the surface.

tools/hindsight/README.md
```diff
@@ -14,233 +14,179 @@ measurable value of adding Fynd to a venue.
...
-- **owning trader** (`[venue_owners]`) — the flow was read from a known venue address (kpk's Safes,
-  surfaced from the CoW decoder's owner).
-- **CoW appData tag** (`[venue_appdata]`) — the settled order committed a frontend tag (`appCode`)
-  whose appData hash maps to a venue (LlamaSwap). The hash is read from the settle calldata by
-  `intents::venue_tag`, so `venue_attribution.rs` stays protocol-agnostic.
-- **fee wallet** (`[venue_fees]`) — a known venue fee wallet took the output-token fee (Phantom,
-  Robinhood); the fee is grossed back. Only inside an already-matched trade, so a dust spray to a
-  fee wallet is not mistaken for flow.
+One method, because one question is being asked: what does this solver say about this
+transaction? A solver's parse fills the whole `SwapIntent` in one pass, including the recipient.
```
carloszanella: **[DOCS `#1`]**

**Description**: The block shows `fn declared(&self, input, logs, amount_in_hint: Option<U256>) -> Result<Option<Declaration>, Veto>` and an `enum Declaration { Terms(SwapIntent), Settled(SettledSwap) }`. `amount_in_hint` was dropped in `eecf1af6`, the two-variant `Declaration` in `0cf6a6e8`, and `SwapIntent` is now `DeclaredSwap`; the real signature is `fn declared(&self, input: &[u8], logs: &[Log]) -> Result<Option<DeclaredSwap>, Veto>` (`solvers/mod.rs:151`). The architecture doc was written mid-PR and the later refactors never came back to it, so the one document explaining the new design describes an interface that no longer exists — and `CLAUDE.md` points readers here for this trait. Maps to the `docs` and `reviewability` directives.

**Suggestion**: Replace the block with the actual one-method signature and the `DeclaredSwap` struct, and rewrite the paragraph at 109–112 so "a solver's parse fills the whole `SwapIntent`" becomes "fills a whole `DeclaredSwap`".

tools/hindsight/CLAUDE.md
```diff
@@ -132,30 +139,28 @@ positive-only USD histogram (`hindsight_positive_slippage_usd`) whose sum is the
 hypothetical revenue. Absent when the top was unsolved or the re-execution failed (e.g. a pool
 vanished at N).
 
-### Calldata-first Relay decoding
+### The declared decode
 
-`RelayCalldata` reads `token_in`/`token_out`/`amount_in` from the settling solver frame's own
+`declared_flow` reads `token_in`/`token_out`/`amount_in` from the settling solver frame's own
 `SwapIntent` and recovers the settled `amount_out` as the gross amount of `token_out` received by
```
carloszanella: **[DOCS `#2`]**

**Description**: Three separate passages describe the declared decode in terms of the settling solver's `SwapIntent`, including the "Adding a solver" instructions someone will follow next. The type is now `DeclaredSwap` (`decoder/solvers/mod.rs:39`); `SwapIntent` matches nothing in the tree, so a reader grepping for it finds nothing and has to reverse-engineer the rename. Maps to the `docs` directive.

**Suggestion**: Replace all three with `DeclaredSwap`. While there, line 212's "One parse fills the whole `SwapIntent`, including the output recipient" should say `DeclaredSwap`.

tools/hindsight/README.md
```diff
@@ -14,233 +14,179 @@ measurable value of adding Fynd to a venue.
...
 - **provider integrator tag** (`[venue_integrators]`) — a provider's event carried an integrator
-  string mapped to a venue (LiFi frontends). The tag is read by that provider's
-  `SolverKnowledge::integrator`, so `venue_attribution.rs` stays provider-agnostic.
+  string mapped to a venue (LiFi frontends), read by `solvers::lifi::integrator`.
+
+The solver label comes from its own evidence tiers, most- to least-trusted: the venue-declared
+calldata id (MetaMask's `aggregatorId`, normalized through the alias list in `solvers/`), the entry
+point itself, the solver frame in the trace, the largest external call (a guess, for unknown
+routers), and the entry-point label as the honest "don't know". The tier is recorded on the
+record (`solver_source`).
```
carloszanella: **[DOCS `#3`]**

**Description**: The list describes five tiers starting with "the venue-declared calldata id (MetaMask's `aggregatorId`, normalized through the alias list in `solvers/`)". `attribution::solver` (`decoder/attribution.rs:54–74`) has four tiers — `EntryPoint`, `TraceMatch`, `LargestCall`, `Fallback` — and its own doc says a venue's claim "is not consulted". The `solver_aliases` tables were deleted from the address books in the same diff.

**Suggestion**: Drop the first tier from the list so it matches `AttributionSource`, and state the new rule the code documents: the router in the trace is preferred over anything a venue declares.

tools/hindsight/src/decoder/mod.rs
```diff
@@ -82,27 +87,11 @@ pub(crate) struct DecodedTrade {
     /// Gross swap output — a venue fee taken from the output (see `venue_fee_out`) is added
     /// back, so the settled amount is the full swap proceeds, comparable to Fynd's gross output.
```
carloszanella: **[DOCS `#4`]**

**Description**: `amount_in`'s doc says a venue fee "(see `venue_fee_in`) is already subtracted" and `amount_out`'s says a fee "(see `venue_fee_out`) is added back". Both fields were removed from the struct in this diff. Only the `amount_out` gross-back still happens, in `attribution::venue`, and for a netted record the input fee is no longer subtracted at all — `netting.rs` says the fee "sits inside the netted amounts", so the docs are both dangling and contradicted by a sibling module. Maps to the `docs` directive.

**Suggestion**: Drop the two cross-references and say plainly which tier each statement holds for: a declared `amount_in` is already past an input-side fee and nothing is subtracted, a netted one is not; point `amount_out` at `attribution::venue`.

tools/hindsight/src/decoder/registry/ethereum.toml
```diff
@@ -100,47 +103,17 @@ entry_points = [
...
-kyber = "kyberswap"
-paraswap = "paraswap"
-airswap = "airswap"
-openocean = "openocean"
-hashflow = "hashflow"
 
 [venues.rabby]
 # Rabby SwapProxy — the app-owned router `tx.to` for its Uniswap-routed swaps. Rabby also routes
 # through shared solver routers (0x, 1inch, Sushi), where `tx.to` is the solver's own contract and
 # the only Rabby fingerprint is the fee leg below; those are not keyed here (see rabby.rs).
```
carloszanella: **[DOCS `#5`]**

**Description**: All seven books carry "(see rabby.rs)" and "(see rainbow.rs)" comments — also `arbitrum.toml:105,110`, `base.toml:83,88`, `bsc.toml:123,128`, `polygon.toml:105,110`, `unichain.toml:84`, `robinhood.toml:131,137`. Both files were deleted with the whole `venues/` directory, and the comments describe behavior that is gone: Rainbow's calldata fee read has no code behind it any more. `base.toml:5–7` also still promises "Every venue fee collector below was verified on-chain" above a file that now has no collectors. Maps to the `docs` directive.

**Suggestion**: Sweep the seven books for comments describing fee handling and the deleted venue modules; a venue section is now just entry points, so each comment should say only how those addresses were verified.

tools/hindsight/CLAUDE.md
```diff
@@ -132,30 +139,28 @@ positive-only USD histogram (`hindsight_positive_slippage_usd`) whose sum is the
...
-across the 165 trades both paths could decode.
+the output recipient the same calldata declares (falling back to the transaction sender) — the
+one field calldata can never carry. Two guards protect the recipient-receipt query: the recovered
+output must clear the intent's `min_amount_out` floor, and any declared quote must sit within
+`plausible_quote`'s band of it; either failure falls through to the netting fallback. The
+declared `amount_in` is already on the solver-task basis: an input-side venue fee left before the
+solver frame. Venue fees are not modelled otherwise (see the README). See
+`.claude/plans/calldata-first-decoding.md` for the empirics behind calldata-first ordering: on a
+315-transaction Base sample, coverage rises from 60.0% (netting alone) to 91.4% (calldata-first
+union), with zero divergences across the 165 trades both paths could decode.
```
carloszanella: **[DOCS `#6`]**

**Description**: The declared-decode section defers the empirics to `.claude/plans/calldata-first-decoding.md`. There is no `.claude/plans/` directory in the worktree, so the coverage numbers quoted right after it (60.0% → 91.4% on a 315-transaction Base sample) have no source a reader can check. The diff rewrote this paragraph and kept the link.

**Suggestion**: Either commit the plan document or drop the reference and keep only the claim the code supports — declared reads first, netting as the marked fallback.

tools/hindsight/CLAUDE.md
```diff
@@ -54,42 +54,49 @@ Any other name needs `--registry`.
...
-names), and `[venues.<name>]` (entry points, fee collectors, and — for venues that declare
-their solver in calldata — `solver_aliases`).
+tiers its chain has; each book's header says what was checked. An address also moves per chain:
+Robinhood Chain's LiFi Diamond, 0x Settler, OKX router and MetaMask router all sit at
+chain-specific addresses, re-derived rather than copied. Sections: `wrapped_native`,
+`infrastructure` (Permit2 etc.), `usd_stablecoins` (USD anchors for reporting), `batch_settlers`,
+`[solvers]` (router address → name; the name joins to a `SolverDecoder` at load), `[labels]`
+(display-only names), `[venues.<name>]` (entry points), and the venue fingerprints
+(`[venue_owners]`, `[venue_fees]`, `[venue_integrators]`, `[venue_appdata]`). Venue fees are not
+modelled: a fee is charged whichever solver fills, so it cancels out of the comparison.
```
carloszanella: **[DOCS `#7`]**

**Description**: Line 98 says flatly "Venue fees are not modelled: a fee is charged whichever solver fills, so it cancels out of the comparison", while line 158 still describes `DecodedTrade` as having "amounts are venue-fee-adjusted so re-solve compares like-for-like". The code sits between the two: `attribution::venue` (`decoder/attribution.rs:105–107`) still grosses an output-side fee back into `amount_out`. The README states this correctly ("One correction survives"); `CLAUDE.md` states it twice, wrongly, in both directions.

**Suggestion**: Make line 98 say "venue fees are not modelled, except that an output-side fee paid to a `[venue_fees]` wallet is grossed back into `amount_out`", and drop "venue-fee-adjusted" from the `DecodedTrade` bullet.

tools/hindsight/src/decoder/solvers/oneinch.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+}
+
+/// The 1inch solver.
+pub(crate) struct OneInch;
+
+impl SolverDecoder for OneInch {
+    /// The trader's swap terms from a `swap` call's `SwapDescription`. `minReturnAmount` is passed
+    /// through as declared, including a zero — the router's per-hop checks can leave the top-level
+    /// floor at zero, and the terms are still worth recording. The hint is unused: every field is
```
carloszanella: **[DOCS `#8`]**

**Description**: Each doc still explains that "`amount_in_hint` is unused" — the parameter was dropped from the trait in this diff, so the sentences explain the absence of something the reader will never see. `zeroex.rs` compounds it by leaning on "the fillable/margin judgment already treats a zero floor sanely"; no fillable/margin judgment exists anywhere in the crate.

**Suggestion**: Delete the `amount_in_hint` sentence from all four, and in `zeroex.rs` either name the code that actually tolerates a zero floor or cut that clause too, leaving the plain statement that a zero `minAmountOut` is passed through as-is.

tools/hindsight/src/decoder/veto.rs
```diff
@@ -47,19 +55,45 @@ pub(crate) enum Veto {
     FeeOnTransfer,
```
carloszanella, line 55: **[DESIGN `#9`]**

`BridgeOrder` and `OutputNotFound` are only ever produced by a solver's `declared` and by `declared_flow`, never by `check`, and the doc comments have to say so twice; splitting the pre-decode rejections from the post-decode shape checks would let each site's return type say which vetoes are possible.

tools/hindsight/src/decoder/veto.rs
```diff
@@ -73,12 +107,11 @@ pub(crate) fn check(
...
+    flow: &SettledSwap,
     transfer_ledger: &TransferLedger,
     registry: &Registry,
 ) -> bool {
-    let sides =
-        [(flow.swap.token_in, flow.swap.amount_in), (flow.swap.token_out, flow.swap.amount_out)];
+    let sides = [(flow.token_in, flow.amount_in), (flow.token_out, flow.amount_out)];
     for (token, trade_amount) in sides {
         for (recipient, total) in transfer_ledger.sink_payments(token) {
             if registry.is_fee_collector(recipient) || registry.is_infrastructure(recipient) {
```
carloszanella: also looks valid

---

**[IMPL `#2`]**

---

**Category**: bug

**Description**: `fee_on_transfer` treats any pure sink receiving ≥1% of `amount_out` as a token tax. Under netting the tracked trader always both sent and received, so the output recipient was never a pure sink. A declared decode can now anchor `amount_out` on an `output_recipient` that is neither the trader nor a sender — KyberSwap's `dstReceiver`, which `kyberswap.rs:121–122` documents as being paid directly by the inner pool — and `sink_payments` then returns that recipient with the full output, tripping the veto.

**Plain English**: Trades whose proceeds are delivered to an address other than the person who sent the transaction are silently discarded as if the token charged a transfer tax. These are exactly the trades the new calldata-first reading was meant to unlock, since balance netting could never see them at all. The operator sees no record and no explanation, only lower coverage than the sample runs promised. A recipient that the solver's own calldata named as the payee should never count as a fee sink.

**Suggestion**: Exempt `flow.tracked` and the declared `output_recipient` from `sink_payments`' candidates (pass the anchor address into `fee_on_transfer`), or skip the output-side sink test entirely for declared records, whose `amount_out` is already anchored on a named payee.

tools/hindsight/src/decoder/solvers/mod.rs
```diff
@@ -88,165 +168,64 @@ impl SwapIntent {
...
-
-    /// Drop the declared quote, keeping the ABI-enforced terms. Used when the settled amount
-    /// shows the quote was self-reported garbage (see [`plausible_quote`]) — the ABI fields stay
-    /// trustworthy either way.
-    pub(crate) fn clear_quote(&mut self) {
-        self.quoted_amount_out = None;
+    pub(crate) fn promised_amount_out(&self) -> Option<U256> {
+        self.declared_quote
+            .or(self.min_amount_out)
     }
```
carloszanella: **[SIMPLIFY `#3`]**

**Root cause**: The method is annotated `#[cfg_attr(not(test), expect(dead_code, reason = "only called from tests in this PR; its production caller is the reverted-swap path in the stacked follow-up PR"))]`. It ships dead, and the suppression that hides that fact points at a branch that is not in this tree. Maps to the `yagni` directive.

- **Complexity** *(was SIMPLIFY `#5`)*: 14 lines and one speculative concept with no caller, `tools/hindsight/src/decoder/solvers/mod.rs:126–139`
- **Tests** *(was TESTS `#6`)*: its four assertions in `fly.rs:155` and `kyberswap.rs:229, 247, 265` test code no production path reaches, and could read `declared_quote`/`min_amount_out` directly, `tools/hindsight/src/decoder/solvers/kyberswap.rs:229–265`
- **Documentation** *(was DOCS `#9`)*: the `reason` string narrates the PR stack and is unreadable once this merges; suppression reasons should describe the code, not the review process, `tools/hindsight/src/decoder/solvers/mod.rs:128–135`

**Proposed fix**: Drop the method, the `cfg_attr` block, and the four test assertions that only exist to call it, asserting `declared_quote` and `min_amount_out` directly instead. Deleting the method removes the need for the suppression and its reason string at the same time, and the follow-up PR can add it back next to its real caller.

tools/hindsight/src/decoder/solvers/mod.rs
```diff
@@ -1,61 +1,84 @@
...
+/// - `amount_out` absent means the source stated no output (exact-input calldata does this), so the
+///   caller recovers it from `output_recipient`'s receipt in the transfer ledger.
+/// - `amount_in` absent means the source fixed the output and only bounded the input (exact-output
+///   calldata), so the caller recovers it from the payer's net payment.
+/// - `tracked` absent means the source did not name the trader, so the caller falls back to the
+///   transaction sender.
+///
+/// One of the two amounts is always present: calldata fixes one side and bounds the other, and an
+/// event states both.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Serialize)]
```
carloszanella: **[SIMPLIFY `#8`]**

`DeclaredSwap` never reaches any serialized struct — it is consumed into three `Option` fields on `DecodedTrade` — so the derive can go.

carloszanella (commented):
Re-anchoring four findings that landed in the body of my previous review because their lines had moved. I checked each against the current head — all four still apply, so they are inline here on the code they actually describe.

tools/hindsight/src/decoder/attribution.rs
tamaralipows: You were right, and it was wrong. Fixed in d3cc78a23.

An input-side fee was never subtracted. Coinbase Base App is the only venue in the books that takes one, and its 0.95% cut is paid by the LiFi Diamond and the 0x Settler — the same contracts the declared decode reads `amount_in` from. Those state what the trader authorized, before the cut, so Fynd was re-solved on more input than reached the pools and the extra output read as savings. The doc claiming a declared `amount_in` was already past the fee was wrong; the address books saying the cut is netted out of the input were right.

Both sides are corrected now, on either tier, with the input-side test restored.

tools/hindsight/src/decoder/attribution.rs
tamaralipows: The payload has a reader now — d3cc78a23 subtracts `VenueFee::Input` from `amount_in`, so both variants do work. See the reply on the other input-fee thread for why it was wrong to drop it. The address-book comments were already correct and stay as they are.

tools/hindsight/src/decoder/attribution.rs
tamaralipows: Not addressed by the two pushes so far — leaving this one open. `venue` still takes `&mut SettledSwap`, and the fee correction still sits inside the fingerprint chain, so it is reached only when the fee-wallet match is the one that wins. d3cc78a23 made the input side symmetric with the output side, which means the ordering problem now applies to both rather than one. The split you propose (`venue -> Option<String>` pure, plus an `output_fee`/`input_fee` lookup the caller applies unconditionally) is the actual fix.

tools/hindsight/src/decoder/attribution.rs
tamaralipows: Not addressed yet — leaving open.

tools/hindsight/src/decoder/mod.rs
tamaralipows: Done in ef1ef7dc6, taking your first option. `SolverDecoder` gains a defaulted `venue_fingerprint(&self, input, logs) -> Option<VenueTag>`, resolved through `Registry::solver` on the address the solver attribution already produced:

```rust
let tag = attribution.address
    .and_then(|address| registry.solver(address))
    .and_then(|solver| solver.decoder.venue_fingerprint(&root.input, logs));
```

`mod.rs` names no solver module now. `attribution::venue` takes one `VenueTag` in place of the two extracted arguments and picks the map from the variant.

Two things worth your eye:
- `Attribution` gained `address: Option<Address>`. The dispatch needs the router address, and re-deriving it would duplicate the tier precedence `attribution::solver` already implements.
- CoW's `is_batch_settler` guard is gone: the settle decode is the guard, because calldata that is not a `settle` call fails to decode.

I picked the trait over the `FINGERPRINTS` table because the table keys on the solver *name*, which re-introduces the name matching `decoder_for` exists to do once at load. That also answers the DESIGN #4 question -- the trait now has a second real method. If you would rather delete the trait for fn pointers, this becomes a two-field struct instead; same dispatch either way.

New end-to-end test `test_venue_fingerprint_on_a_venue_wrapped_entry` -- an Infinex swap through the LiFi Diamond through `decode_block`. Verified it fails without the dispatch (`left: "lifi", right: "infinex"`).

tools/hindsight/src/report/mod.rs
tamaralipows: Done in df71df665. `decoder` is now `DecodeSource` (`DeclaredFromCalldata`, `DeclaredFromLogs`, `VenueNetting`, `SenderNetting`, `IntentNetting`) with a `tier()` method, and `decode` is `DecodeTier`. Serde renames hold the JSONL column values, so the wire format and the Grafana queries reading it are unchanged. `decode` is written only as `decoder.tier()`, so the two columns can no longer disagree.

On your flag for the human: the reader stays `decode: Option<String>`. It reads historical JSONL, and typing it would turn an unrecognised old value into a deserialize error instead of a kept record. It compares against `DecodeTier::Netted.wire()` rather than a literal of its own, and `test_decode_column_wire_strings` pins every wire string so a variant rename cannot move a dashboard column.

I did not rename the columns to `decode_source` / `decode_tier`. Renaming the wire names while the reader must still parse files written under the old names would need serde aliases on both, which costs more than the readability it buys. Happy to do it if you want it.

tools/hindsight/src/decoder/solvers/okx.rs
tamaralipows: Done in 28967be36. One `NATIVE_TOKEN_SENTINEL` and one `normalize_native` in `solvers/mod.rs`; all six per-solver constants and their normalizers are gone. The name drops both misleading parts -- no solver prefix, and no `_ETH` suffix that was wrong on the BSC and Polygon books.

`normalize_native`'s test moved to `solvers/mod.rs` with the function, and it now also asserts the checksummed spelling maps to `Address::ZERO`, since the six copies were split between the lowercase and checksummed forms of the same address.

tools/hindsight/src/decoder/solvers/mod.rs
tamaralipows: Done in 28967be36. `promised_amount_out`, the `cfg_attr`, and the reason string are gone. The four tests that only existed to call it assert `declared_quote` directly -- which also made two of them sharper, since `promised_amount_out` returning the floor could not distinguish "no quote declared" from "quote equals the floor".

The reverted-swap PR can add it back next to its real caller.

tools/hindsight/src/decoder/solvers/mod.rs
tamaralipows: Done in 28967be36 -- `DeclaredSwap` loses the `Serialize` derive.

tools/hindsight/src/decoder/solvers/oneinch.rs
tamaralipows: The `amount_in_hint` sentences were already gone from all four modules by the time I got here -- nothing in `src/` mentions it now.

The `zeroex.rs` half was real and is fixed in 32e8d62d1: there is no fillable/margin judgment anywhere in the crate. It now names `declared::recover_output`, which is the code that actually tolerates a zero floor (it compares the recovered output against the floor, and every output clears zero).

tools/hindsight/CLAUDE.md
tamaralipows: Both halves fixed. The "Venue fees are not modelled" line went in d3cc78a23, when the input-side correction came back -- it now says a `[venue_fees]` wallet's cut is corrected out on whichever side it was paid, and every other venue fee is left inside the amounts.

The `DecodedTrade` bullet is fixed in 32e8d62d1: "venue-fee-adjusted" is replaced by the same specific statement, so the file no longer says it twice in two directions.

tools/hindsight/CLAUDE.md
tamaralipows: Dropped in 32e8d62d1, along with the coverage numbers that had no checkable source. What is left is the claim the code supports: declared reads first, netting as the marked fallback.

If you would rather have the empirics, the plan document needs committing -- happy to do that instead if you have it.

tools/hindsight/src/decoder/registry/ethereum.toml
tamaralipows: Swept all seven books in 32e8d62d1. Nothing points at `rabby.rs` or `rainbow.rs` any more, and the header accuracy note names `[venue_fees]` instead of a per-venue collector list -- including `base.toml`'s promise about collectors it no longer has.

Two things the sweep turned up beyond what you flagged:
- Rainbow's comments described a calldata fee read with no code behind it. They now state the mechanism instead: Rainbow keeps its input-side fee in the router with no transfer, so there is no wallet to match and the cut stays inside `amount_in`.
- Relay's comments in six books said its fee is backed out. Relay is not in `[venue_fees]`, so nothing backs it out. They now record the fee legs as the evidence that confirmed the deployment, which is what they were actually for.

tools/hindsight/src/decoder/mod.rs
tamaralipows: Already fixed in d3cc78a23. Both dangling cross-references are gone, and each field now says what happens and where: a fee paid to a `[venue_fees]` wallet out of the input is subtracted, out of the output is added back, both by `attribution::venue`.

One difference from your suggestion, worth knowing: rather than documenting that a netted input fee is not subtracted, that commit made it subtracted. The input-side correction had been dropped and the address books still said it happened -- see the two input-fee threads.

tools/hindsight/src/decoder/attribution.rs
tamaralipows: Done in f8c93777a, as you proposed.

`venue` is read-only now and returns just a label. `venue_fee` reports the cut -- venue, side, amount -- and `decode_transaction` applies it via `apply_venue_fee` for every trade with a fee-wallet match, so the correction no longer depends on which fingerprint named the venue. `attribution.rs` only reads.

The double count is fixed too, and moving the correction to the caller is what made it possible: the caller knows which decoder produced `amount_out`, so it can tell when the figure already contains the cut. Two cases, both of which used to have the fee added on top:
- the solver's own event stated the gross output outright (LiFi's `toAmount`, documented and verified as gross in `solvers::lifi`), so nothing was measured from a receipt;
- the receipt was measured at the router named as the declared output recipient, which received the gross amount and forwarded the cut -- your exact case. Detected from the ledger: the anchor sent the fee token onward.

Four tests on `apply_venue_fee` cover add-back, both skips, and the input side. Verified the two skip tests fail without the guard (`left: 10085, right: 10000`). One test in `attribution.rs` pins the ordering fix: a kpk trade (owner match) that also paid a Phantom fee leg still reports the cut.

README updated -- the contradiction you found between lines 138-141 and 158 was already resolved in 32e8d62d1; both passages now also carry the skip condition.

carloszanella (approved):
giving early approval to not block

tools/hindsight/src/decoder/declared.rs
tamaralipows: Fixed in 7563e7f40 -- the walk is breadth-first now, so the outermost frame wins and the code matches what `declared_flow` documents.

Being straight about one thing first: I could not think of a real-world case where this bug bites. I went looking for two *different* routers on separate branches and could not construct a plausible transaction that does that, so I measured instead.

What the traces actually show is different from the premise, and more interesting. Over 200 mainnet blocks (25833100-25833300, 2262 trades) exactly two records change, and in both the branches hold the **same** router entered several times at different depths -- one transaction swapping several legs through Uniswap's universal router. One entered it at depth 4 and depth 6; the other at depth 7 twice and depth 10. Arbitrage contracts, unrecognised entry points, floors of 0 and 1, several legs into native ETH.

Which means such a transaction has no single settling frame, and the walk order only decided which leg got recorded as the whole trade. So the walk fix was not enough: 6b610e628 declines the declared read on those transactions outright, and 96205fb5e writes the limitation down -- **batch transactions are not supported**, in either form, and the tool gives up the coverage rather than invent the amounts. `hindsight_several_legs_total` counts what that costs.

One thing that fell out of measuring: the dropped transaction was dropped silently -- `declared_flow`'s `Err(veto)` went through `.ok()?` in `decode_transaction` with no log. Your BIZ #4 point, hit for real, and fixed in 796d73e7e.

Tests pin the three cases that already resolved correctly (nested solvers, a router at the root, a reverted branch beside a live one) plus the multi-leg case that did not; I checked the first three pass under both walks, so they are real regression pins.

tamaralipows: Follow-up on the above -- 6b610e628 supersedes what the walk fix achieved, and I want to correct my own framing.

The walk fix made the code match its documented contract, but it did not make those records correct: it changed *which* leg of a multi-leg transaction gets recorded, not whether recording one leg is right. Tamara pushed on that and was right to.

So the declared read now declines a multi-leg transaction outright. `trace::solver_frames` returns every outermost solver frame -- a solver reached from inside another solver's frame is a step in that route, not a leg of its own -- and `declared_flow` returns `Ok(None)` when there is more than one. Netting still runs, because it reads the trader's own balances rather than one frame, so it either finds a single net swap or declines. That is the same "coverage miss, never wrong amounts" rule `netting` and CoW's multi-order batches already follow.

Measured over the same 200 blocks:

| | trades |
|---|---|
| before | 2261 |
| after | 2258 |

12 records move from the declared tier to `intent-netting`, where the amounts are the trader's own and the netted marker states the tier honestly. 8 are dropped -- netting found no single swap either. `hindsight_several_legs_total` counts the declined reads so the coverage given up is visible.

One thing worth knowing, since it changed my design: I first put the guard before the whole decode, and that discarded 8 records whose amounts netting had already read correctly. Measuring caught it. The guard belongs on the declared read alone.

The walk fix stays -- `find_solver_frame` still decides the solver *label* for attribution, where ordering matters and amounts do not.

tools/hindsight/src/decoder/registry.rs
tamaralipows: Looks like this is suggesting to get rid of the trait totally. Unfortunately your other comment added a second trait method. Afraid of creeping complexity here. Might decide otherwise in a future PR but keeping out of scope for now.

tools/hindsight/src/decoder/veto.rs
tamaralipows: Confirmed and fixed in 4bca918b4. `check` now takes the anchor address -- the payee the calldata named, or the tracked trader when it named none -- and `fee_on_transfer` exempts it.

Your reasoning about why netting never hit this is exactly right: the tracked trader always both sent and received, so no anchor could be a pure sink. Only the declared read can anchor on an address that sends nothing.

Measured over 200 mainnet blocks (25833100-25833300): **6 trades recovered**, 4 KyberSwap and 2 Uniswap, all `solver-calldata`. So the loss was falling on precisely the different-receiver coverage the declared read exists to add, as you said.

The test asserts both directions -- exempt when anchored on the payee, still vetoed when anchored on the trader -- because the veto is right in the second case and the anchor is the only thing that distinguishes them.

tools/hindsight/src/decoder/netting.rs
tamaralipows: Fixed in 4bca918b4. `fallback_flow` returns `None` when the entry point is a batch settler and no clean-net trader was found, which is what the Intent role's decoder list did before the refactor.

You were right that no commit body mentioned dropping the guard. `Registry::is_batch_settler`'s own doc still stated the reason, which is what made it checkable.

One note on measurement: no record in the 200-block sample changes, so this occurs less often than the payee bug above. I would not read that as low priority -- it fabricates a trade rather than losing one, and a fabricated trade goes into the savings aggregates.

tools/hindsight/src/decoder/veto.rs
tamaralipows: Skipping this one, deliberately.

The observation is right: `BridgeOrder`, `OutputNotFound` and `InputNotFound` can only come from a solver's `declared` or from `declared_flow`, and the rest only from `check`. Two kinds of rejection in one type.

But no bug is reachable through it. No caller matches on a variant it cannot receive, and if one did, splitting the type would not catch it -- it would just be a dead match arm either way. The doc comments already state which site produces which veto, so the information a reader needs is there; the change would move it into the type system at the cost of touching every veto site.

Not worth it against the rest of this PR. Happy to revisit if a third producer appears, which is when the doc comments would start being the wrong place to keep it.

tools/hindsight/src/decoder/declared.rs
tamaralipows: Both halves fixed in 796d73e7e.

**The docs were wrong, not the code.** `CLAUDE.md` said both guards "fall through to the netting fallback". Neither does: a recovered output below the declared floor drops the transaction, and a quote outside `plausible_quote`'s band drops only the quote and keeps the trade. It now says which does what, and why the floor guard does not fall through -- once a solver has stated the trade, netting answers a different question.

**The veto was silent, and it cost me a real trade.** `Err(veto)` reached `decode_transaction` through `.ok()?`, so nothing was logged and nothing counted. I hit this for real while measuring an unrelated change: a transaction vanished from the output with no line explaining it, and I had to diff two decode runs to find out why. Both veto sites now log the variant and increment `hindsight_vetoed_transactions_total`, labelled by veto -- so a misparsed floor in some solver's decoder shows up as a counter climbing rather than as coverage quietly falling.

tools/hindsight/src/decoder/netting.rs
tamaralipows: Done in 796d73e7e -- both are plain `fn` now. Confirmed neither is reached from outside `netting.rs`.

tools/hindsight/README.md
tamaralipows: Already fixed in 32e8d62d1. The block now shows the real two-method trait (`declared` plus `venue_fingerprint`, added in ef1ef7dc6 for DESIGN #1) and the `VenueTag` enum, and the paragraph below says `DeclaredSwap`.

I corrected it as part of that commit rather than this thread, because I had just rewritten the paragraph directly underneath it and leaving a wrong signature above my own edit would have been worse than the original.

tools/hindsight/CLAUDE.md
tamaralipows: Fixed in 32e8d62d1 -- `SwapIntent` matches nothing in the tree now, including line 212's "One parse fills the whole `SwapIntent`". I grepped the whole crate to confirm.

tools/hindsight/README.md
tamaralipows: Fixed in 796d73e7e. The list is the four `AttributionSource` tiers, and it states the rule the code documents: the router in the trace beats anything a venue claims.

Two things beyond what you flagged, in the same file:
- the **pipeline diagram** carried the same phantom tier ("venue-declared id -> entry point -> trace frame -> guess"), so it is corrected too;
- the venue-fingerprint bullet said the integrator tag is "read by `solvers::lifi::integrator`", which stopped being true in ef1ef7dc6 -- it is dispatched through `SolverDecoder::venue_fingerprint` now, and naming the module was the thing DESIGN #1 was about.

I also dropped a third stale claim while there: the declared tier being "already on the solver-task basis" because an input-side fee left before the solver frame. That was the premise the input-fee fix disproved.

tools/hindsight/src/decoder/attribution.rs
tamaralipows: Done in 0c9b1087f. `venue_fee` calls `received_by_address` once per token and the wide `received_by` is deleted, along with its test and a `HashSet` import in two files.

Your overflow note checked out -- the deleted method summed with `+=` where `received_by_address` saturates, and that only differs on a `U256` overflow real token amounts cannot reach. I confirmed it in the source rather than taking it on trust, since it was the one thing that could have made the substitution wrong.

Its test went too: the surviving `received_by_address` test already covers summing several senders into one recipient, native ETH keyed on the zero address, and exclusion by both recipient and token. Nothing lost.

Verified on a decode of 200 mainnet blocks (25833100-25833300): the same 2254 records, none added, removed, or changed -- which is what a pure refactor should look like. First comparison showed 10 missing, all from one block whose receipts the RPC failed to serve, not a code difference.

That closes the last open thread on this PR.

PR #433, opened 2026-08-18
feat: Waterfill - Slim Shady version

carloszanella:
## Waterfill Slim Shady Version

I treated the slow performance of this algorithm as if it was an offense
to my mother, so I dig deeply in how to improve it. Compared to current
main, it is like a 3x~5x speed. We will see once this is in dev for sure.

The code was quite the AI spaghetti, with a lot of bad names, weird
design and separation, questionable abstractions, etc. I had to work
on these to even understand what the code was doing, before I could
work on performance. 

What you see in this commit is a whole mix, I decided not to separate the 
AI clean-up from actual logic changes because it would take me
another week to organize that in proper commits. This is far from ideal,
so in an effort to make you focus on the important things to review I will
describe the actual FEAT changes, and point at the code for you to look at.

**Here we go:**
1. `swap_cache.rs` - new file with a pool swap cache for repeated swap amounts can also interpolate. Interpolating was key here, for very small changes in input amount in relation to the cached value
2. **interpolation** - caching can also interpolate amounts smartly. If amounts are close to the cached value and a boolean flag is used, a linear interpolation will be used to save from a pool simulation. errors and are also handed smartly. this also lives in `swap_cache.rs`. waterfill has solve "stages" and each of them decides whether interpolation can be used or not. the idea is to use interpolation when you dont need exact precision.
3. we skip paths that use the same pool twice (at `rank_at_full_amount`). This is an edge case (very few >=3 token pools) and I dont think it ever makes sense to swap twice on the same pool, when you could just go to the final token directly. 
4. waterfill's **single-path baseline** is now decided based on this cache+interpolation ranking. results might change slightly but from running the benchmark there was no change at all in WF for 10k orders.
5. **Reuse some swap results at the chunking stages** of waterfill. Here it is harder to use caching because amounts are quite different, we cannot interpolate, and state changes with each iteration. But still could take away some repetition from it.
6. Timeout refactoring allows for **finer resolution on solve times**
7. New `swap-metrics` feature, disabled by default, adds a `MeteredProtocolSim` which collects metrics in a thread-local cache for measuring specific protocol usage and solve time. Useful for local debugging. By default it is not activated.

### DIFF
--- a/fynd-core/src/algorithm/most_liquid.rs
+++ b/fynd-core/src/algorithm/most_liquid.rs
@@ -19,13 +19,17 @@ use rustc_hash::{FxHashMap, FxHashSet};
 use smallvec::SmallVec;
 use tracing::{debug, instrument, trace};
 use tycho_simulation::{
-    tycho_common::simulation::protocol_sim::{GetAmountOutResult, Price, ProtocolSim},
+    tycho_common::simulation::protocol_sim::{Price, ProtocolSim},
     tycho_core::models::{token::Token, Address},
 };
 
 use super::{Algorithm, AlgorithmConfig, NoPathReason};
 use crate::{
-    algorithm::{paths, sim_guard::GuardedProtocolSim},
+    algorithm::{
+        paths,
+        sim_guard::GuardedProtocolSim,
+        swap_cache::{PoolDirection, Refusal, SwapCache, SwapResult},
+    },
     derived::{computation::ComputationRequirements, types::TokenGasPrices, SharedDerivedDataRef},
     feed::market_data::{MarketData, MarketState, StateLabel},
     graph::{
@@ -34,6 +38,10 @@ use crate::{
     types::{ComponentId, Order, Route, RouteResult, Swap},
     AlgorithmError,
 };
+
+/// What the simulation report calls the pass that picks a route. Most-liquid has one: every swap
+/// it makes is ranking candidates, and the winner is re-simulated by `build_route` afterwards.
+const SELECTION: &str = "selection";
 
 /// What can go wrong settling one token sequence.
 ///
@@ -249,12 +257,6 @@ struct HopResult {
     amount_out: BigUint,
     /// What that swap costs in gas, in wei.
     gas: BigUint,
-}
-
-impl HopResult {
-    fn new(pool_ix: usize, result: GetAmountOutResult) -> Self {
-        HopResult { pool_ix, amount_out: result.amount, gas: result.gas }
-    }
 }
 
 /// Algorithm that selects routes based on expected output after gas.
@@ -601,7 +603,8 @@ impl MostLiquidAlgorithm {
         ctx: &SolveContext,
     ) -> Result<RouteResult, AlgorithmError> {
         let mut best_route: Option<(&TokenPath, SolvedRoute)> = None;
-        let mut cache = PoolSwapsCache::new(self.cache_pair_swaps);
+        let mut winners = PairWinners::new(self.cache_pair_swaps);
+        let mut swaps = SwapCache::new();
         let timeout_ms = self.timeout.as_millis() as u64;
 
         for (token_path, _) in scored_paths {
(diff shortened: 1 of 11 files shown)


fynd-core/src/algorithm/most_liquid.rs
```diff
@@ -716,27 +715,37 @@ impl MostLiquidAlgorithm {
...
+                    || {
+                        let Some(state) = market.get_simulation_state(component_id) else {
+                            return Err(Refusal::Failed);
+                        };
+                        state
+                            .get_amount_out_guarded(current_amount.clone(), token_in, token_out)
+                            .map(|result| SwapResult { amount_out: result.amount, gas: result.gas })
+                            .map_err(|error| Refusal::of(&error))
+                    },
+                    true,
```
louise-poole, line 736: Interpolation is enabled here. This I rate is fine for an initial pass, but I see on water fill you properly re-simulate the top 8 ranked paths to get a true ranking of them. Maybe we should do the same here?

fynd-core/src/algorithm/water_fill/mod.rs
```diff
@@ -825,107 +820,142 @@ impl WaterFillAlgorithm {
...
-            let mut best: Option<(usize, BigInt, StepResult)> = None;
+            let mut best: Option<(usize, BigInt)> = None;
             for (i, &path_idx) in subset.iter().enumerate() {
-                let Some(step) = Self::simulate_step(
-                    &ctx.ordered[path_idx],
-                    ctx.market,
-                    &committed[i],
-                    chunk.clone(),
-                ) else {
+                if marginals[i].is_none() {
```
louise-poole, line 874: This could be `None` if the chunk failed to simulate no? Then we'd potentially retry that same path multiple times and fail each time.... Would be good if we could differentiate between 'failed' and 'haven't tried yet'.

fynd-core/src/algorithm/swap_cache.rs
```diff
@@ -0,0 +1,557 @@
...
+
+impl Refusal {
+    /// Reads a failed simulation. Only the pool rejecting the input says anything about size;
+    /// a fatal error is what a caught panic arrives as, and a recoverable one is transient.
+    pub(crate) fn of(error: &SimulationError) -> Self {
+        match error {
+            SimulationError::InvalidInput(_, _) => Refusal::OverLimit,
+            SimulationError::FatalError(_) | SimulationError::RecoverableError(_) => {
+                Refusal::Failed
+            }
```
louise-poole: Not sure I'd collapse these two into the same refusal type.... Fatals should not be retried and caching their failure is valid. Recoverable errors should be retried and probably not stored in the cache?

fynd-core/src/algorithm/water_fill/mod.rs
```diff
@@ -316,6 +232,182 @@ impl WaterFillAlgorithm {
...
+        // fatal — the pre-ranked set already guarantees a route.
+        let anchor_tokens = derive_anchor_tokens(graph);
+
+        // Discovery and ranking both swap against untouched state, so they share one cache: every
+        // frontier edge discovery simulates is an answer ranking would otherwise pay for again.
+        // It outlives setup because the allocation passes that read untouched state reuse it too.
+        let mut cache = SwapCache::new();
+        sim_meter::start_solve();
+
+        let (discovered_paths, _) = discover_paths(
```
louise-poole: Is this the only callsite for `discover_paths`? If so - we're dropping half the returned data. Why even return the dropped part to begin with? We could just return the paths.

louise-poole (approved):
Veeeeeery nice! 🔥 Just commented a couple smaller things, have approved in the meantime.

fynd-core/src/algorithm/water_fill/mod.rs
carloszanella: I tried this suggestion. for some reason it got a bit slower overall.. so not applying

fynd-core/src/algorithm/most_liquid.rs
carloszanella: yeah I think it is fine, ML is doing well without this extra simulation

PR #437, opened 2026-08-18
feat: accept exclusive legs that wrap into the output token

tamaralipows:
## What

`has_valid_exclusive_route` required the exclusive leg to produce the route's output token. An exclusive pool quoting the **native** token could therefore never serve a request denominated in the **wrapped** token: tycho streams the ETH/WETH wrap as an ordinary component, so the wrap leg sits after the exclusive leg and makes it non-terminal. `best_exclusive_candidate` dropped the candidate, `combine_with_surplus` returned the public ranking unchanged, and no exclusive commitment was ever made.

This accepts one additional route shape — an exclusive leg followed by a single native wrap leg that produces the route's output token.

## Why it is safe

A wrap converts 1:1. `pin_commitment` attributes captured surplus to a leg with the ratio `path_final_out / leg_out`, which a wrap leaves at exactly 1, so the leg gets the same arithmetic as a terminal leg and needs no inverse simulation.

Every other non-terminal shape stays rejected. The trailing leg must consume the exclusive leg's output, produce the route's output token, and be a native wrap; each of those three conditions has a test that fails without it.

## Notes

- `Chain` is threaded to the validator so the native/wrapped pair comes from tycho's chain config rather than a local constant. The `try_` accessors are used because the plain ones panic on an unregistered custom chain, which the crate's lints deny; an unresolved pair falls back to requiring a terminal leg.
- Effect on prod: the Ekubo exclusive ETH/USDC pool is denominated in native ETH, while observed exclusive demand asks for WETH.

### DIFF
--- a/fynd-core/src/worker_pool_router/mod.rs
+++ b/fynd-core/src/worker_pool_router/mod.rs
@@ -45,13 +45,13 @@ use tycho_execution::encoding::{
     evm::gas_estimator::estimate_gas_usage,
     models::{Solution, Strategy},
 };
-use tycho_simulation::tycho_common::Bytes;
+use tycho_simulation::tycho_common::{models::Chain, Bytes};
 
 use crate::{
     encoding::encoder::Encoder, feed::exclusivity::is_exclusive, price_guard::guard::PriceGuard,
     worker_pool::task_queue::TaskQueueHandle, BlockInfo, EncodingOptions, Order, OrderQuote,
     OrderSide, Quote, QuoteOptions, QuoteRequest, QuoteStatus, SolveError, SolveParams,
-    SurplusInfo,
+    SurplusInfo, Swap,
 };
 
 /// Environment variable overriding [`DEFAULT_USER_IMPROVEMENT_SHARE_BPS`]. Read once, on the
@@ -372,6 +372,7 @@ impl WorkerPoolRouter {
                         request.options(),
                         public_ranked,
                         *USER_IMPROVEMENT_SHARE_BPS,
+                        self.encoder.chain(),
                     )
                 } else {
                     self.rank_quotes(responses, request.options())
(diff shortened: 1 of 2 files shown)


fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -1051,12 +1059,23 @@ fn has_valid_exclusive_route(quote: &OrderQuote) -> bool {
...
+        // Only the immediately following leg is considered: a wrap run that ends at the output
+        // token is always a single leg, and requiring adjacency keeps this validator in step with
+        // the chaining rule `pin_commitment` uses to find a path's final output.
+        let reaches_output = *swap.token_out() == output_token ||
+            swaps
+                .get(index + 1)
+                .is_some_and(|next| {
+                    next.token_in() == swap.token_out() &&
+                        next.token_out() == &output_token &&
+                        is_native_wrap(next, chain)
```
Troshchk, line 1076: Does it work for split routes?
Is wrapping always the next one?

A -(exclusive)> wETH -> ETH
A -(other)> wETH -----> ETH

I think in this case if exclusive is the first leg it will be rejected?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -2707,6 +2762,76 @@ mod tests {
...
+
+    #[test]
+    fn test_exclusive_leg_wrap_of_another_path() {
+        let (native, wrapped, other) = wrap_tokens();
+        let quote = make_route_quote_for_tokens(&[
+            ("vm:exclusive", other, make_address(0x08)),
+            ("uniswap_v2", native, wrapped),
+        ]);
+
+        assert!(!has_valid_exclusive_route(&quote, SimChain::Ethereum));
```
Troshchk, line 2832: This can be parametrized I think?

louise-poole (approved):
Ah yes of course! Good catch, looks good to me!

Troshchk (commented):
Thank you @tamaralipows! Left a couple of qs

Troshchk (approved):

fynd-core/src/worker_pool_router/mod.rs
tamaralipows: Talked about this in-call: it will group the ETH amounts before wrapping once

PR #446, opened 2026-08-20
feat: accept exclusive legs anywhere in the route

tamaralipows:
Sketch: what it takes to drop the rule that an exclusive leg must produce the route's output token.

## The rule and why it was there

A candidate was dropped unless its exclusive leg produced the output token, itself or through a trailing native wrap. The surplus is denominated in the output token, and a mid-route leg withholds a different token, so there was no way to say how much the leg should hold back.

## How the conversion works

`fynd-core/src/worker_pool_router/exclusive_capture.rs` replays the route against the states its swaps were quoted at. Every `Swap` already carries its own `protocol_state`, so asking what the route would have produced had the leg emitted less needs no market read and no new data.

```
baseline = replay(withheld = 0)
target   = baseline - surplus
withheld = largest d where replay(d) >= target        // bisection, 24 steps
```

Route output falls monotonically as the leg withholds more, which is what makes the search sound. The search keeps the largest withheld amount that still clears the target, so stopping early leaves output with the user and costs the protocol capture.

## What this also fixes

The terminal case used `surplus * leg_out / path_final_out` — the average price. Average price overstates what the next unit is worth, so it under-captured whenever the downstream pool was thin. `test_solved_amount_exceeds_the_average_price_estimate` pins that difference.

`is_native_wrap` is gone. A wrap is now an ordinary downstream swap that the replay prices like any other, so the special case has nothing left to do.

## What stays rejected

Two exclusive legs. The route does not say how the surplus divides between them, so there is no single amount to solve for.

## Cost

Solving runs twice per exclusive candidate — once to validate, once to pin — at up to 25 route replays each. That is cheap for native pools and not cheap for VM pools, where a replay is a revm call. Worth a latency read before this goes anywhere near production.

The router now needs token decimals, which the pool simulations take instead of addresses. `WorkerPoolRouter::with_market_data` supplies them and `FyndBuilder` wires it; a router built without it makes no commitment and orders answer from public liquidity.

## Open

The committed amount reaches the chain as a Q32 fee rate on the leg (`derive_fee_q32`), applied to whatever the leg produces at execution time. For a terminal leg a fraction of leg output is the same fraction of user output at any market state. For a mid-route leg the fraction maps to the final output through pools whose state has moved since the quote, so the realized capture drifts from the solved one. Sizing that drift is the next question, and it is a property of the on-chain encoding rather than of this solver.

### DIFF
--- a/fynd-core/CLAUDE.md
+++ b/fynd-core/CLAUDE.md
@@ -141,6 +141,14 @@ beats the committed amount and records the difference as `SurplusInfo`
 (`OrderQuote::surplus_amount()`, `committed_amount_out()`, `Swap::committed_amount_out()`). All are
 `#[serde(skip)]` — internal, not on the wire.
 
+A route can carry a commitment when it has exactly one exclusive leg and replays against the states
+its swaps were quoted at. Where the leg sits does not matter.
+`worker_pool_router/exclusive_capture.rs` solves how much the leg withholds: for a leg that ends the
+route the answer is the surplus itself, and for a leg with swaps after it the answer is what those
+swaps turn into the surplus, found by bisecting replays of the route. The router reads token
+decimals for those replays from `MarketData` (`WorkerPoolRouter::with_market_data`); without it no
+commitment is made and orders answer from public liquidity.
+
 Enable by setting the scope per pool via `PoolConfig::with_liquidity_scope()` or
 `liquidity_scope = "include_exclusive"` in `worker_pools.toml`. A deployment where every pool sets it
 fails the build (`SolverBuildError::NoPublicPool`) — there would be no pool left to establish the
--- a/fynd-core/src/solver.rs
+++ b/fynd-core/src/solver.rs
@@ -862,7 +862,8 @@ impl FyndBuilder {
         let router_config = WorkerPoolRouterConfig::default()
             .with_timeout(self.router_timeout)
             .with_min_responses(self.router_min_responses);
-        let mut router = WorkerPoolRouter::new(solver_pool_handles, router_config, encoder);
+        let mut router = WorkerPoolRouter::new(solver_pool_handles, router_config, encoder)
+            .with_market_data(market_data.clone());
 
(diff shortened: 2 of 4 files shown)


tamaralipows:
Closing. This was a sketch to answer how hard it is to drop the rule that an exclusive leg must produce the route's output token. It answered that, and the answer says don't merge this shape of it.

**The conversion is easy.** Every `Swap` already carries the `protocol_state` it was quoted against, so the route can be replayed without reading the market. Route output falls monotonically as the leg withholds more, so the withheld amount is a bisection. That part works and is tested: the solved amount lands the route exactly on its committed amount, and it beats the average-price formula the terminal case used to rely on.

**The cost is the problem.** The search replays the route 24 times. When a `vm:*` pool sits after the exclusive leg, those are 24 revm calls that cannot be cached — they are precisely the simulations whose input changes each iteration. The router budget is 1000 ms per order. Nobody has measured the real latency yet, so this is structure rather than a number, but the shape is wrong enough not to merge on it.

What a mergeable version needs:

1. A secant search instead of bisection. The function is smooth and monotone, so 3-5 iterations should do what 24 does here.
2. A gate on what sits downstream: accept a non-terminal leg whose downstream pools are all native, reject when a `vm:*` pool follows it.
3. A histogram on the solve duration and replay count first, so the budget question gets answered from production rather than from argument.

One finding worth keeping regardless of what happens to the rest: the existing terminal-leg path converts surplus with `surplus * leg_out / path_final_out`, the average price. Average price overstates what the next unit is worth, so it under-captures whenever the downstream pool is thin. That is a live under-capture on the current code, not something this PR introduced.

Also of note for whoever picks this up: the exclusive test fixtures fabricate `amount_out` values no simulation produces, and the split fixture never set split fractions, so it is not replayable at all. Any approach that re-simulates a candidate route has to fix those first.

Branch `tnl/exclusive-mid-path-leg` stays if anyone wants the code.

PR #447, opened 2026-08-20
fix(api): bound experimental list query limits

tvinagre:
## Summary
- Reject `limit > 1000` with JSON `400 BAD_REQUEST` on experimental `GET /v1/tokens` and `GET /v1/prices`, before derived-data/cache reads.
- Document the `maximum: 1000` contract and `400` response in OpenAPI and generated TypeScript declarations.
- Add boundary coverage for omitted/default limits, explicit maximum, zero, token offsets, and malformed/negative/overflow query values.

### DIFF
--- a/clients/openapi.json
+++ b/clients/openapi.json
@@ -7,7 +7,7 @@
       "name": "MIT",
       "identifier": "MIT"
     },
-    "version": "0.99.1"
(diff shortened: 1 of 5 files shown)


brunoguerios (approved):
LGTM ✅

PR #448, opened 2026-08-20
feat(hindsight): add Robinhood Chain address book

tamaralipows:
Adds a built-in decoder address book for Robinhood Chain (chain-id 4663), so `--chain robinhood` works without `--registry`.

Robinhood Chain is an Arbitrum Orbit rollup with ETH as gas and 1-second blocks. Several contracts sit at chain-specific addresses rather than their mainnet ones, so each was re-derived on chain:

- 0x Settler `0x39b38686…` ("RobinHoodSettler"), read from 0x's registry `ownerOf(2)`. It paid 632 of Robinhood Wallet's 659 fee legs in the sample.
- LiFi Diamond `0xb477751b…`, not the cross-chain `0x1231deb6…`.
- OKX router `0xe58b3089…`, identified by its `smartSwapByOrderId` / `unxswapByOrderId` ABI. Neither published OKX address is deployed.
- MetaMask Swap Router `0xf5307745…`, identified by MetaMask's `swap(string,address,uint256,bytes)` selector and its fee legs.
- Rialto router `0xc94135b6…`, the chain's PropAMM aggregator, read from Rialto's own router registry.
- 1inch v6 is deployed twice, at its usual address and at `0x5a705de8…`.

Every venue was confirmed by its own fee legs in a 100k-block sample (~28 hours) rather than assumed. USDG is the chain's only USD anchor. CoW, UniswapX, 1inch v5 and Sushi are not deployed, so those tiers are absent, and the book's header says so.

The two ERC-4337 EntryPoints are listed as infrastructure: most wallet swaps here are user operations, and without them two unrelated 4337 swaps in one block share the EntryPoint's logs and read as a sandwich bracket.

A `decode` run over blocks 41701320-41701650 returned 689 trades: uniswap 624, relay 29, okx 16, metamask 4, arcus 3, 1inch 3, lifi 3, 0x 3, phantom 2, fly, kyberswap — solved by uniswap, 0x, okx, kyberswap, 1inch, rialto, lifi, fly and relay. The run named one unknown entry point, Arcus' SwapShell, which is in `[labels]` rather than `[solvers]`: its swaps are relayer-sent, and a solver entry replaces the intent path with sender netting and loses the trade.

Two things to know before running it:

- The public RPC serves no `debug_traceTransaction`. Use a tracing endpoint.
- Robinhood Wallet's and Coinbase's own swaps enter through an ERC-4337 EntryPoint or a smart wallet, which matching cannot key on, so those never decode. 41% of Robinhood's fee legs in the sample entered that way. Both venues still decode when they route through a shared router. Closing that gap needs a 4337 decoder, not an address book.

### DIFF
--- a/tools/hindsight/src/decoder/registry/robinhood.toml
+++ b/tools/hindsight/src/decoder/registry/robinhood.toml
@@ -1,0 +1,188 @@
+# Robinhood Chain (chain-id 4663) address book for trade decoding. Same structure as ethereum.toml —
+# see it and registry.rs for what each section drives. Addresses gathered 2026-08-20.
+#
+# Accuracy note: a wrong or missing *solver* address only costs coverage (the swap fails to
+# match), but a wrong *venue fee collector* produces wrong records (the fee stays inside the
+# amounts). Every contract below was checked to be deployed on Robinhood Chain, and every venue was
+# confirmed live by its own on-chain fee legs in a 100k-block sample (~28 hours at 1-second blocks)
+# ending block 41578851.
+#
+# Robinhood Chain is an Arbitrum Orbit (Nitro) rollup with ETH as the gas token, so several
+# contracts sit at chain-specific addresses rather than their mainnet ones. Each was re-derived
+# rather than assumed:
+#   - 0x: the Settler is 0x39b38686… ("RobinHoodSettler"), read from 0x's registry
+#     (0x00000000000004533Fe15556B1E086BB1A72cEae, `ownerOf(2)`). AllowanceHolder is at its usual
+#     cross-chain address.
+#   - OKX: neither published DEX Router address is deployed. The router here is 0xe58b3089…,
+#     identified by its `smartSwapByOrderId` / `unxswapByOrderId` ABI.
+#   - LiFi: the Diamond is at 0xb477751b…, not the cross-chain 0x1231deb6… address.
+#   - MetaMask: the mainnet Swap Router has no code. Its router here is 0xf5307745…, identified by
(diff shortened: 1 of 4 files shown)


tools/hindsight/src/decoder/registry/robinhood.toml
```diff
@@ -0,0 +1,188 @@
...
+#   - 0x: the Settler is 0x39b38686… ("RobinHoodSettler"), read from 0x's registry
+#     (0x00000000000004533Fe15556B1E086BB1A72cEae, `ownerOf(2)`). AllowanceHolder is at its usual
+#     cross-chain address.
+#   - OKX: neither published DEX Router address is deployed. The router here is 0xe58b3089…,
+#     identified by its `smartSwapByOrderId` / `unxswapByOrderId` ABI.
+#   - LiFi: the Diamond is at 0xb477751b…, not the cross-chain 0x1231deb6… address.
+#   - MetaMask: the mainnet Swap Router has no code. Its router here is 0xf5307745…, identified by
+#     MetaMask's `swap(string,address,uint256,bytes)` selector and its fee legs.
+#   - 1inch: v6 is deployed twice, at its usual address and at 0x5a705de8…; both are verified
+#     AggregationRouterV6. v5 is not deployed.
```
Troshchk, line 22: The addresses are omitted here, is it expected?

Troshchk (approved):
Thank you @tamaralipows! ✨ 
Approving with one small q

tools/hindsight/src/decoder/registry/robinhood.toml
tamaralipows: These shortened addresses appear in full further down the file!

PR #451, opened 2026-08-21
feat: serve pAMMs from the Titan price level stream

carloszanella:
Serves a proprietary AMM from Titan's pAMM price level stream instead of simulating it in the EVM. Titan publishes a quote ladder per pair every block; quoting interpolates those levels, which costs a fraction of a VM simulation. Built on tycho-simulation 0.360's `price_level_stream` module and the `PropAMMExecutor` that encodes the fills.

## Using it

```
--protocols all_onchain,exclude:vm:fermiswap,pricelevelstream:fermiswap
```

A `pricelevelstream:{venue}` entry serves that venue from the stream. The served venues are `fermiswap`, `kipseli`, `metric`, `bebop` and `taurusfi`; the stream is Ethereum-only. A venue outside that set is a config error rather than a warning, because these entries are always hand-written and a typo would otherwise stream nothing.

An `exclude:` entry drops a protocol from the resolved list. It is needed here rather than optional: `vm:fermiswap` and `pricelevelstream:fermiswap` price the same maker inventory, so streaming both double-counts it. The remainder goes through `ProtocolSpec::parse`, so `exclude:exclusive:ekubo_v3` and `exclude:ekubo_v3` both name system `ekubo_v3`. An exclusion that names nothing, matches no streamed protocol, or repeats a requested protocol is an error — a typo there would otherwise quietly ship the market the entry exists to prevent.

## How it is wired

`open_price_level_stream` opens the stream for the named venues, and the three feed loops select on it alongside the Tycho and RFQ sources, handing every update to the same `handle_tycho_message`. The stream reconnects internally, so unlike the RFQ clients it needs no supervising task — and it ending is a feed error rather than a clean stop. Every loop opens it before answering its `pending_tx` / `controller_tx` handshake, so a configuration error reaches the caller as an error instead of as a handle to a feed that dies on the next line.

Encoding needs nothing new: tycho-execution resolves any `pricelevelstream:{venue}` through its `pricelevelstream` family entry, and the executor is already deployed.

## Benchmark

The live capture reads only the Tycho stream, so no `--protocols` list could bring a price level venue in. `--include-protocols` adds entries to the streamed list before the capture opens.

```bash
./scripts/bench.sh --market live --include-protocols pricelevelstream:fermiswap \
  --exclude-protocols vm:fermiswap
```

One frame is taken after the Tycho snapshot — last, so the ladders sit as close to the captured block as they get — and merged into it, so every config in a run solves the same frozen ladder. An included name that is already streamed, sends no frame before the capture timeout, or brings no component into the market stops the run.

## Worth knowing

`get_amount_out` rejects amounts below the smallest ladder level instead of extrapolating, and returns a partial fill for amounts above the largest. A quote executes only in the block it was quoted for.

Titan pushes several frames per second, and each one runs the full `handle_tycho_message` path — market write lock, `MarketUpdated` broadcast, one `info!` line. That takes those from roughly once per block to several times per second for any deployment that enables a venue. Left as-is deliberately: coalescing frames would trade away the freshness the stream exists for. Worth watching on the first deployment.

https://claude.ai/code/session_01ACx21TznY4u89nq92GS5da


tamaralipows:
`--include-protocols` will stop the run once propeller-heads/tycho#1335 lands.

That PR changes the protocol system of whitelisted pAMM venues from `pricelevelstream:{venue}` to `propammfallback:{venue}`. Same components, same prices — only the label changes, so tycho-execution routes the swap through Titan's PropAMMRouter.

`capture_market` checks each `--include-protocols` name against the `protocol_system` of the captured components. After the bump, `--include-protocols pricelevelstream:fermiswap` brings in components labelled `propammfallback:fermiswap`, so the name matches nothing and the run stops with "brought no component into the market".

It stops only when `RPC_URL` is set. The whitelist read is what selects the new family; without it the stream keeps every venue on the direct path. So the same command passes or stops depending on the environment.

Fix: accept either family for an included `pricelevelstream:` venue.

tamaralipows (approved):
Thank you!

PR #467, opened 2026-08-26
fix: draw exclusive swap nonces from a random per-process prefix

tamaralipows:
## Background

For exclusive swaps, Fynd signs a short authorization that travels with the
transaction. Each authorization carries a nonce. The contract remembers every nonce it has
accepted and rejects a repeat, so Fynd must never hand out the same nonce twice.

## The problem

The nonce was seeded from the clock, in whole seconds, when the process started, then incremented
once per signed authorization. That stays unique only while the process signs fewer than one
authorization per second. Two ways it breaks:

- **After a restart.** Signing faster than one per second pushes the counter ahead of the clock.
  The next start reseeds from the clock, which is now behind nonces already spent on chain, and
  signs them a second time.
- **With more than one replica.** Every replica seeds from the same clock, so their nonce ranges
  overlap.

Either way the swap reverts with `NonceAlreadyUsed()`.

## The fix

A nonce is now two halves: a random 32-bit value drawn once per process in the high bits, and a
count of the authorizations that process has signed in the low bits. A restart or a second replica
draws its own random half, so it cannot land on a range that is already spent. Nothing needs to be
stored between restarts and replicas need no coordination.

The counter returns an error at its maximum instead of wrapping, because wrapping would reissue
the range this process already signed. That takes 4.3 billion authorizations in one process
lifetime, and a restart resets it with a new random half.

## Not covered

A client resubmitting the same signed transaction still reverts with `NonceAlreadyUsed()`. That is
the replay protection doing its job, and this change does not alter it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3546,6 +3546,7 @@ dependencies = [
  "num-traits",
  "num_cpus",
  "petgraph",
+ "rand 0.9.4",
  "reqwest 0.12.28",
  "rstest",
  "rustc-hash",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -180,6 +180,7 @@ chrono = "0.4"
 itertools = "0.14.0"
 hex = "0.4"
 sha2 = "0.10"
+rand = "0.9"
 
 # Internal crates
 fynd-core = { path = "fynd-core", version = "0.99.16" }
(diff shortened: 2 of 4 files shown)


Troshchk (approved):
Thank you @tamaralipows! ✨

PR #472, opened 2026-08-27
fix!: bind exclusive swap signatures to the Tycho router as locker

tamaralipows:
## What was wrong

For exclusive swaps Fynd signs a short authorization that travels with the
transaction. The Ekubo extension that checks it supports naming one contract as the only caller
allowed to use the authorization. Fynd left that field empty, which means "any caller".

An empty field makes the signed bytes a bearer instrument for as long as they are valid: anyone who
reads them can run the swap themselves. Twice this week a third party did exactly that — took the
signed bytes from a pending transaction, executed the same swap through its own contract two blocks
earlier, and wrapped it in arbitrage. The authorization is single-use, so the original transaction
then arrived at an already-spent authorization and reverted with `NonceAlreadyUsed()`.

## The fix

The signature now names the Tycho router as the only caller allowed to use it. The address comes
from the router address the encoder already resolves for the chain, so nothing new is configured
and no address is hardcoded. A copycat can no longer execute the payload from its own contract; the
only remaining way to use it is through the router itself, which pays the receiver named in the
original calldata.

## Scope

Every fill of these swaps observed on chain takes the Ekubo lock through that same router, so this
binds to the caller already in use rather than changing how swaps execute. Deployments with no
controller key configured are unaffected — they never sign these payloads.

## Not covered

Two related contributors stay open, both outside this change:

- Routes carrying one of these authorizations are submitted through the public mempool with a low
  priority fee, which is what gives a copycat the two blocks it needs. They should go through a
  private route, or pay a competitive tip.
- A route whose authorization is close to expiry should be re-quoted rather than submitted.

### DIFF
--- a/fynd-core/src/encoding/exclusive_swap.rs
+++ b/fynd-core/src/encoding/exclusive_swap.rs
@@ -30,7 +30,7 @@ use tycho_simulation::tycho_common::{models::protocol::ProtocolComponent, Bytes}
 use crate::{SolveError, Swap};
 
 /// Environment variable holding the pool controller's private key (hex, with or without `0x`).
-const ENV_CONTROLLER_KEY: &str = "EXCLUSIVE_SWAP_CONTROLLER_KEY";
+pub(crate) const ENV_CONTROLLER_KEY: &str = "EXCLUSIVE_SWAP_CONTROLLER_KEY";
 
 /// Default validity window for a signed quote, in seconds. Well under the extension's 30-day cap.
 const DEFAULT_DEADLINE_WINDOW_SECS: u32 = 120;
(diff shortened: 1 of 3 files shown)


fynd-core/src/encoding/exclusive_swap.rs
```diff
@@ -580,12 +599,27 @@ mod tests {
...
+        let signer = ExclusiveSwapSigner::new(CONTROLLER_KEY.parse().unwrap(), 1, 0, 120, LOCKER);
+
+        let user_data = signer
+            .build_user_data(&signed_swap(Some(990_000)))
+            .unwrap();
+
+        // meta starts after the fee(8) prefix; its low 128 bits carry the locker's last 16 bytes.
+        let meta = B256::from_slice(&user_data.as_ref()[8..40]);
+        assert_eq!(&meta.as_slice()[16..32], &LOCKER.as_slice()[4..20]);
+        assert_ne!(&meta.as_slice()[16..32], &[0u8; 16], "an unbound payload is usable by anyone");
```
Troshchk: Isn't line 613 implied by line 612?

Troshchk (approved):
Thank you! 
Just one question

fynd-core/src/encoding/exclusive_swap.rs
tamaralipows: Yes — the locker in the test is non-zero, so the equality above already covers it. Removed.

PR #474, opened 2026-08-28
feat(rpc): route override hook and staged quote pipeline

kayibal:
This PR adds extension points to `fynd-rpc` and `fynd-core`. A binary that embeds `fynd-rpc` can replace one endpoint and keep all other endpoints. The hosted service is the first such binary.

## Route override

`FyndRPCBuilder::configure_routes` accepts a closure with the signature `Fn(actix_web::Scope, &AppState) -> actix_web::Scope`. The builder calls the closure on the `/v1` scope before it adds the default routes. A route that the closure adds for a path and method takes precedence over the default route. All other default routes stay active. The closure cannot add a route outside `/v1`. A replaced route does not change the OpenAPI document at `/docs/`.

## Quote pipeline stages

`WorkerPoolRouter::quote` is now a sequence of three public stages:

- `WorkerPoolRouter::solve` returns a `RankedQuotes` value. It holds the ranked candidates of each order. When the price guard is active, each order has one candidate.
- `encode_quotes` encodes a set of order quotes and records the encoding metrics.
- `finalize_quote` adds the gas estimates and sets the solve time.

`RankedQuotes::new` returns an error when an order has no candidates. Represent an order without a route with an `OrderQuote` that has a status other than `Success`.

## Worker pool allowlist

`QuoteOptions::with_worker_pools` limits a request to the named worker pools. The field is not on the wire, because serde skips it. The router checks the allowlist one time for each request. An empty allowlist and an unknown pool name cause `SolveError::InvalidWorkerPools`. The HTTP layer reports it with the code `INVALID_WORKER_POOLS`.

## Public handler parts

The `quote` handler now uses these public functions:

- `validate_quote_request` rejects a request without orders and converts the request to the core type.
- `ReplayRequest::capture` copies the request for the replay log before the conversion.
- `log_quote_outcome` writes the failure and slow-solve log lines.

The handlers, the `AppState` accessors, `HealthTracker`, and `exclusive_access::from_headers` are public. `fynd-rpc` re-exports `actix_web`, so an embedder uses the same actix types.

## Compatibility

The behavior of `fynd serve` and the OpenAPI document do not change. The existing tests pass without changes.

### DIFF
--- a/fynd-core/src/worker_pool_router/mod.rs
+++ b/fynd-core/src/worker_pool_router/mod.rs
@@ -31,7 +31,7 @@ use std::{
 };
 
 pub use allocation::ExclusiveAccess;
-use allocation::{allocate, Allocation, OrderClass};
+use allocation::{allocate, validate_pool_allowlist, Allocation, OrderClass};
 use comparison_log::{log_quote_comparison, solver_error_label};
 use config::WorkerPoolRouterConfig;
 use futures::stream::{FuturesUnordered, StreamExt};
@@ -424,9 +424,12 @@ impl WorkerPoolRouter {
         // trade size joins `OrderClass` they will differ.
         let class = OrderClass::new(access);
         let pool_allowlist = request.options().worker_pools();
+        if let Some(allowlist) = pool_allowlist {
+            validate_pool_allowlist(&self.solver_pools, allowlist)?;
+        }
         let mut allocations: Vec<Allocation<'_>> = Vec::with_capacity(request.orders().len());
         for _order in request.orders() {
-            allocations.push(allocate(&self.solver_pools, class, pool_allowlist)?);
+            allocations.push(allocate(&self.solver_pools, class, pool_allowlist));
         }
 
         if allocations
--- a/fynd-core/src/worker_pool_router/mod.rs
+++ b/fynd-core/src/worker_pool_router/mod.rs
@@ -248,6 +248,88 @@ pub struct WorkerPoolRouter {
     price_guard: Option<PriceGuard>,
 }
 
+/// Ranked, unencoded candidates for every order of a request — the output of
+/// [`WorkerPoolRouter::solve`].
+///
+/// One inner list per request order, in request order, best candidate first. An order with no
+/// route yields a single `NoRouteFound`/`Timeout` placeholder, so every list is non-empty.
+#[must_use]
+#[derive(Debug)]
+pub struct RankedQuotes {
+    per_order: Vec<Vec<OrderQuote>>,
+    started: Instant,
+}
+
+impl RankedQuotes {
+    /// Wraps already-ranked candidates and starts the solve clock now.
+    pub fn new(per_order: Vec<Vec<OrderQuote>>) -> Self {
+        Self { per_order, started: Instant::now() }
+    }
+
+    fn started_at(per_order: Vec<Vec<OrderQuote>>, started: Instant) -> Self {
+        Self { per_order, started }
+    }
+
+    /// Ranked candidates per order, best first.
+    pub fn per_order(&self) -> &[Vec<OrderQuote>] {
+        &self.per_order
+    }
+
+    /// Consumes the ranking, returning the candidates per order.
+    pub fn into_per_order(self) -> Vec<Vec<OrderQuote>> {
+        self.per_order
+    }
+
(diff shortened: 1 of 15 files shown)


fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -248,6 +248,88 @@ pub struct WorkerPoolRouter {
...
+/// HTTP service reports.
+pub async fn encode_quotes(
+    encoder: &Encoder,
+    order_quotes: Vec<OrderQuote>,
+    encoding_options: &EncodingOptions,
+) -> Result<Vec<OrderQuote>, SolveError> {
+    let encode_start = Instant::now();
+    let encoded = encoder
+        .encode(order_quotes, encoding_options.clone())
+        .await;
```
kayibal, line 337: This should happen in parallel.

fynd-rpc/src/api/handlers.rs
```diff
@@ -137,31 +147,32 @@ pub(crate) async fn quote(
...
+}
 
-    let core_quote = result?;
-
+/// Serializes a core quote as the `200 OK` wire response.
+pub fn quote_response(core_quote: fynd_core::Quote) -> HttpResponse {
     let dto_quote: dto::Quote = core_quote.into();
-
-    Ok(HttpResponse::Ok().json(dto_quote))
+    HttpResponse::Ok().json(dto_quote)
```
kayibal: Too shallow, keep it inlined

kayibal: Inlined in 79ed0c41; helper, its test and doc mentions removed.

fynd-core/src/worker_pool_router/mod.rs
kayibal: Agreed to do this in a separate PR: `encode_quotes` only moves the existing sequential batch call out of `quote`; parallelizing encoding across orders is a fynd-core change on its own.

fynd-rpc/src/api/handlers.rs
```diff
@@ -60,41 +72,57 @@ pub(crate) fn configure_routes(cfg: &mut web::ServiceConfig) {
...
 ) -> Result<HttpResponse, ApiError> {
     let access = exclusive_access::from_headers(http_request.headers());
-    let dto_request = request.into_inner();
-
-    // Validate request
-    if dto_request.orders().is_empty() {
-        return Err(ApiError::BadRequest("no orders provided".to_string()));
-    }
+    let capture = ReplayRequest::capture(&request, access);
+    let core_request = validate_quote_request(request.into_inner())?;
```
tamaralipows: We can validate before capturing to avoid doing unnecessary work?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -248,6 +248,114 @@ pub struct WorkerPoolRouter {
...
+    }
+
+    /// Consumes the ranking, keeping only the best candidate of every order.
+    #[must_use]
+    pub fn into_best(self) -> Vec<OrderQuote> {
+        self.per_order
+            .into_iter()
+            .map(|mut candidates| candidates.swap_remove(0)) // O(1); discarded order is irrelevant.
+            .collect()
+    }
```
tamaralipows, line 318: AFAIK `swap_remove(0)` on an empty Vec panics.

tamaralipows (approved):
Thanks, approving not to block, but can you please have a look in particular at the panic?

fynd-rpc/src/api/handlers.rs
kayibal: Done in 8b02b4b8: the handler validates first and `ReplayRequest::capture` now takes the validated `fynd_core::QuoteRequest` (same allowlisted fields, identical JSON — the replay tests pin the format).

fynd-core/src/worker_pool_router/mod.rs
kayibal: Correct — the only guard was a `debug_assert` in the private constructor. In 8b02b4b8 the emptiness check moved into `started_at`, which now returns `Result`; `new` delegates to it and `solve` propagates the error, so every `RankedQuotes` is checked at construction in all profiles and `swap_remove(0)` cannot see an empty list.

PR #475, opened 2026-08-28
feat: attribute quotes and settled volume to the winning algorithm

carloszanella:
Two changes that make it visible which algorithm is doing the work.

## Return the algorithm in the quote response

`fynd_core::OrderQuote` already carried the algorithm that produced it, stamped
by the worker and marked internal with `#[serde(skip)]`. The router ranks quotes
from several worker pools and returns the best one, so that field already named
the winner — it was never exposed.

`algorithm` is now on the wire type, threaded through the TypeScript and Rust
clients. It is absent rather than empty on a quote no algorithm produced, such
as the no-route placeholder the router builds when every pool fails.

The field names the algorithm, not a solver: a solver is an algorithm plus its
config, so two solvers running the same algorithm return the same value here.

## Record the volume each winning algorithm settled

`worker_router_best_quote_pool` counts wins, which hides size: an algorithm can
win few orders and settle most of the flow, or win many small ones and settle
little. It now carries an `algorithm` label alongside `pool`, so wins and volume
aggregate by the same key.

`worker_router_won_volume_gas_token` records the output amount of every winning
quote, labelled by pool and by algorithm. Dividing it by the win count gives the
average size an algorithm wins.

The amount is converted to the gas token first. Output-token base units do not
add up across tokens, so a total over them means nothing. This is the same
conversion `record_gas_token_amount` already uses, and it needs no price source
beyond what the quote itself carries. It is a gauge rather than a counter
because the amounts are fractional.

A quote the conversion cannot price increments
`worker_router_unpriced_won_quotes_total` instead of recording zero, which would
read as an algorithm that won nothing.

## Attribution follows the quote the router returns

Both metrics record in `quote`, once the returned quotes are final — not at
ranking. Three things happen after ranking: the exclusive overlay replaces the
head of the list, the price guard can pick a lower-ranked fallback, and a failed
encoding returns no quote at all.

The worker pool name lived on `WorkerPoolQuote`, which `rank_quotes` drops when
it returns `Vec<OrderQuote>`. So `fynd_core::OrderQuote` gains an internal
`worker_pool`, `#[serde(skip)]` like `algorithm`, stamped by `solve_order` as
the quote arrives. Every later stage moves or mutates quotes rather than
rebuilding them, so the stamp survives to the end.

`worker_router_best_quote_pool` counted the ranking head before this change, so
its numbers shift for exclusive and price-guard traffic.

## Notes for review

Volume is denominated in gas tokens, not USD. `PriceGuard` holds real USD prices
and the router already owns one, but it is opt-in per request, so a USD figure
would cover only price-guard-enabled traffic and skew the share. The ratio
between win share and volume share is the same in either unit; converting to USD
is a multiply on the dashboard.

The only dashboard query on `worker_router_best_quote_pool` is
`sum(increase(worker_router_best_quote_pool[5m])) by (pool)`
(`monitoring/grafana/dashboards/fynd.json:214`). The `sum ... by (pool)`
collapses the new `algorithm` label, so the panel is unchanged.

The OpenAPI regeneration also bumps the spec version from 0.99.11 to 0.99.18,
catching up with a spec that was stale on main.


clients/typescript/client/src/types.ts
louise-poole, line 1: Does the rust client maybe not need updating too?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -647,6 +647,7 @@ impl WorkerPoolRouter {
             let best = valid_quotes[0];
             counter!("worker_router_best_quote_pool", "pool" => best.worker_pool.clone())
                 .increment(1);
+            record_won_volume(best);
```
louise-poole: You're recording this where we rank the quotes, not when we choose the final solution. Sometimes the price guard picks a lower rank solution, or encoding fails. 
Also doing it here means we miss out on correctly considering exclusive liquidity's wins (misattribute it to another algo actually) as its selection is handled after the ranking phase.

fynd-rpc-types/src/lib.rs
```diff
@@ -851,6 +851,12 @@ pub struct OrderQuote {
     /// Fee breakdown (populated when encoding options are provided).
     #[serde(skip_serializing_if = "Option::is_none")]
     fee_breakdown: Option<FeeBreakdown>,
+    /// Routing algorithm that produced this quote.
+    ///
+    /// Absent on a quote no solver produced, such as a no-route placeholder.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[cfg_attr(feature = "openapi", schema(example = "bellman_ford"))]
+    solver: Option<String>,
```
louise-poole: Could you do a quick sweep of our doc strings/claude docs/gitbook docs? I recall when this was added we emphasised it was for internal use only so that will need to be updated and this field needs to be added to public interface docs.

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -647,6 +647,7 @@ impl WorkerPoolRouter {
             let best = valid_quotes[0];
             counter!("worker_router_best_quote_pool", "pool" => best.worker_pool.clone())
```
louise-poole: If we add an algorithm label to the counter here it'll be easier to aggregate by algorithm vs the new volume metric you added.

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -876,6 +877,31 @@ fn to_gas_token_amount(quote: &OrderQuote, amount: &BigUint) -> Option<BigUint>
...
+/// cannot price increments `worker_router_unpriced_won_volume_total` rather than counting as zero,
+/// which would read as a solver that won nothing.
+fn record_won_volume(best: &WorkerPoolQuote) {
+    let pool = best.worker_pool.clone();
+    let algorithm = best.quote.algorithm().to_string();
+    let volume = to_gas_token_amount(&best.quote, best.quote.amount_out())
+        .and_then(|amount| amount.to_f64());
+    let Some(volume) = volume else {
+        counter!(
+            "worker_router_unpriced_won_volume_total",
```
louise-poole: Technically this isn't recording volume no? It's a count.

louise-poole (commented):
Thanks! Just a few quick comments!

fynd-rpc-types/src/lib.rs
```diff
@@ -1806,6 +1817,7 @@ mod conversions {
                 .fee_breakdown()
                 .cloned()
                 .map(Into::into);
+            let solver = (!core.algorithm().is_empty()).then(|| core.algorithm().to_string());
```
Troshchk: `algorithm()` only returns the algo name and does not identify the solver(afaiu solver is identified by algo+config; like number of hops for example)
I think in this case it might be better to name this field into `algorithm` instead of `solver`.
What do you think?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -876,6 +877,31 @@ fn to_gas_token_amount(quote: &OrderQuote, amount: &BigUint) -> Option<BigUint>
...
 
+/// Records the volume a winning quote settled, in whole gas tokens, by pool and algorithm.
+///
+/// Counting wins alone hides size: a solver can win few orders and still settle most of the flow,
+/// or the reverse. Output-token base units do not add up across tokens, so the amount is converted
+/// to the gas token first, the same way `record_gas_token_amount` does it. A quote the conversion
+/// cannot price increments `worker_router_unpriced_won_volume_total` rather than counting as zero,
+/// which would read as a solver that won nothing.
+fn record_won_volume(best: &WorkerPoolQuote) {
+    let pool = best.worker_pool.clone();
```
Troshchk: Can we rename to `worker_pool`?
we have recently made a rename in fynd, to clearly distinguish between pools(liquidity vs worker/solver pools), so I think it would be nice to stay explicit here too. Only talking about the variable, doesn't apply to the metric field

Troshchk (commented):
Thank you @carloszanella

fynd-rpc-types/src/lib.rs
carloszanella: Renamed to `algorithm` in fc8e2d0c — the wire field, the accessor, the OpenAPI spec and the TypeScript client. You are right that it is the algorithm name alone, and the metric label added alongside it was already `algorithm`, so the two now agree.

clients/typescript/client/src/types.ts
carloszanella: Good catch — added in 1b113f21. `Quote::algorithm` plus its accessor, populated in `order_quote_to_quote`. The field is `pub(crate)` and set after construction, the same way `solve_time_ms` is: `Quote::new` is public and already takes fourteen arguments. Mapping tests cover both a response that names an algorithm and one that does not.

fynd-core/src/worker_pool_router/mod.rs
carloszanella: Renamed in dd7b868c. The metric label stays `pool`, so no dashboard query changes.

fynd-core/src/worker_pool_router/mod.rs
carloszanella: You are right. Renamed to `worker_router_unpriced_won_quotes_total` in 05fa690f and reworded the doc comment to say the quote is counted.

fynd-core/src/worker_pool_router/mod.rs
carloszanella: Added in 137e19b4. The only dashboard query on this counter is `sum(increase(worker_router_best_quote_pool[5m])) by (pool)` (monitoring/grafana/dashboards/fynd.json:214), which collapses the new label, so the panel is unchanged.

fynd-core/src/worker_pool_router/mod.rs
carloszanella: Confirmed on all three counts, and the exclusive case was worse than a miss — it was a misattribution. Fixed in 5c565c35.

The three paths after ranking:

- **Exclusive.** When `exclusive_routing_active()`, `quote` ranks `responses.public_only(..)` and `combine_with_surplus` then puts the exclusive quote at the head. The win went to the best public pool while the exclusive pool settled the order.
- **Price guard.** `validate` returns the first candidate that passes, which can be any entry in the fallback chain.
- **Encoding.** A failed `encode` returns an error and no quote at all, after a win was already counted.

What blocked recording later: `rank_quotes` returns `Vec<OrderQuote>` and drops `WorkerPoolQuote.worker_pool`, so nothing downstream knew which pool a quote came from. So the fix adds an internal `worker_pool` to `fynd_core::OrderQuote`, `#[serde(skip)]` like `algorithm`, stamped by `solve_order` as the quote arrives. It survives the rest of the pipeline because every later stage moves or mutates quotes rather than rebuilds them — `pin_commitment` clones, `select_first_valid` moves, `Encoder::encode` takes `mut quotes` and mutates in place.

`record_won_volume` becomes `record_win`, records both the win and the volume, and runs in `quote` once the returned quotes are final. It skips a quote that is not `Success` and a quote with no worker pool — the `NoRouteFound` placeholder is both, since the router builds it and there is no winner to credit.

This changes `worker_router_best_quote_pool` too: it counted the ranking head before, so its numbers shift for exclusive and price-guard traffic.

fynd-rpc-types/src/lib.rs
carloszanella: Swept in 4d8b5e17.

- `fynd_core::OrderQuote::algorithm` no longer says "(internal use only)". That line moved in 5c565c35, which touched the same field.
- `docs/get-started/hosted-api.md`: added to the response example and the response-field table, with a note that the names can change between releases so callers treat it as a label rather than an API.
- `fynd-rpc-types/CLAUDE.md`: added to the `OrderQuote` field list.

The rest of `docs/` describes ranking rather than the response shape, so nothing there claimed the algorithm was hidden. `no_route_cause` is the other `#[serde(skip)]` field and stays internal.

Troshchk (approved):
Thanks, approving

louise-poole (approved):
Thanks! LGTM!

PR #476, opened 2026-08-28
fix: leave a pAMM out of the graph when its fallback cannot be priced

carloszanella:
## What this changes

A `propammfallback:` leg reaches the chain through Titan's PropAMMRouter, which falls back to a Uniswap V3 pool when the maker's quote goes stale. `min_amount_out` still describes the pAMM quote, so a fallback below it reverts the route — the worker therefore refuses to quote a pAMM leg whose fallback pool this market does not hold.

It made that call in `fallback_amount_out`, after the algorithm had assembled a route, and discarded the whole route. The decision is fixed per pair and fee tier, so the algorithm rebuilt the same doomed routes on every order, and it fell hardest on the algorithms that split: the more venues a route touches, the likelier one leg fails.

The graph now leaves such a component out, so the algorithm routes around it.

## Evidence

On a 5000-order live run against block 25852983, with five pAMM venues streamed, **all 2017 dropped routes were `propammfallback:fermiswap` on USDT/cbBTC**. Its resolved tier is the router's global `fallbackFee` of 3000, and `getPool(USDT, cbBTC, 3000)` is `0x3696Ed1d5A75694dd9779976bC9E7c22e555f934`, which exists with `liquidity() = 0`. Tycho correctly excludes an empty pool, so the market holds no fallback and the refusal is right.

Every drop reported `pair_absent_from_market=true`; none were the native-ETH case, and none were a tier the market held a pool at but the router did not resolve to.

Cost of the old placement, same run: aperture settled $9-11M less order volume than `bellman_ford`, concentrated in three orders at d2 — one of them $7.0M — all lost to this check.

## Shape

`lacks_fallback_pool` lives in `propamm_fallback`, beside the fee tiers and the pool index that answer it. The worker asks; it does not restate the rule.

`drops_component_on_build` is separate from `drops_component`. `drops_component` holds for the life of the worker, which is what lets `filter_event` apply it to state updates and removals as well as additions. The pAMM answer expires — both the fee tiers and the fallback pools change — so filtering an update with it would freeze a component's price and filtering a removal would strand it in the graph.

`FeeTierFetcher` runs as its own task, so a worker usually builds its first graph before the tiers arrive and admits every pAMM. The worker records that and rebuilds once the tiers land.

`fallback_amount_out` stays authoritative. The graph filter saves the work of assembling a route that cannot be quoted; it does not decide whether one can.

## Upstream

`fermiswap` is quoting a pair whose resolved fallback pool has no liquidity. Either that pair should not be quoted, or the router needs a `getPairFee` override naming a tier with a real pool. That would recover the routes rather than skip them.

### DIFF
--- a/fynd-core/src/propamm_fallback/mod.rs
+++ b/fynd-core/src/propamm_fallback/mod.rs
@@ -35,6 +35,32 @@ pub fn has_pamm_leg(route: &Route) -> bool {
     })
 }
 
+/// Whether `component` is a pAMM with no Uniswap V3 pool to fall back on.
+///
+/// A `propammfallback:` leg only reaches the chain through its fallback, so a component whose fee
+/// tier has no pool in this market can never produce a quotable route. Returns `false` for any
+/// other protocol system, and for a component that does not name exactly two tokens: the fallback
+/// resolves per leg from the swap's own pair, so a wider component is left to
+/// [`fallback_amount_out`].
+pub fn lacks_fallback_pool(
+    component: &tycho_simulation::tycho_common::models::protocol::ProtocolComponent,
+    fee_tiers: &FeeTiers,
+    index: &FallbackPoolIndex,
+) -> bool {
+    if !component
+        .protocol_system
+        .starts_with(PROPAMM_FALLBACK_PREFIX)
+    {
+        return false;
+    }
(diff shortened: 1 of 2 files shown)


tamaralipows (approved):
Thank you! Looks good

PR #487, opened 2026-08-31
feat!: Misc work extracted from developing and required to add upcoming algorithms

carloszanella:
Commits should be self contained, ran 2 iterations of review skill on top. Here's a summary (maps to the commits):

1. refactor: share some code from ML (will be used on new algos)
2. bug fix: Fix some buggy cases on execute_split_plan, new algos surfaced it
3. bug fix: Fix circular order tracking of amounts on the same function
4. refactor: Add some abstractions to execute_split_plan, it was quite big and did a lot
5. feat: (not prod) Allow for some metadata on metered protocol sim 
6. feat: Add more info on errors from pAMMs
7. feat: (not prod) Allow for live benchmark to use pAMMs
8. feat: (not prod) Allow for live benchmark to use exclusive liquidity
9. feat: (not prod) A bunch of benchmark improvements

 The main logic change that needs more reviewing should be commit #2. I've tested it as much as I could.


fynd-core/src/solver.rs
```diff
@@ -1443,7 +1491,13 @@ impl Solver {
...
+                .task_queue_capacity(pool_cfg.task_queue_capacity())
+                .liquidity_scope(
+                    pool_cfg
+                        .liquidity_scope()
+                        .unwrap_or_default(),
+                )
+                .fallback_fee_tiers(fallback_fee_tiers.clone());
             let (worker_pool, task_handle) = algorithms
                 .configure(pool_cfg.algorithm(), named)?
                 .build(market_data.clone(), Arc::clone(&derived_data), pool_event_rx, derived_rx)?;
```
tamaralipows, line 1509: Seems slightly problematic for testing if this drift is true?

---

**[DESIGN `#1`]**

**Root cause**: `from_recording_with` re-implements the `PoolConfig` → `WorkerPoolBuilder` → `SolverPoolHandle` mapping instead of sharing it with `FyndBuilder::build` (`solver.rs:884–928`). Every field added to one path has to be remembered in the other, and three have not been.

- **Architecture**: the builder gets `liquidity_scope` at `solver.rs:1495–1499`, but the handle is built bare — `SolverPoolHandle::new(worker_pool.name(), task_handle)` at `solver.rs:1505`, where the production path adds `.with_liquidity_scope(pool_scope)` at `solver.rs:925–928`. `SolverPoolHandle::liquidity_scope` is what `serves` and `exclusive_routing_active` read (`worker_pool_router/allocation.rs:63–69, 103–105`), so in a recording every pool is allocated as `PublicOnly`: the exclusive pool answers `ExclusiveAccess::Denied` and no surplus overlay runs. The branch's new exclusive-liquidity bench feature only produces a non-identical result *because* of this omission — `algorithm_bench.rs:446` sends `QuoteOptions::default()`, so with the scope set the exclusive config would score exactly like its twin, contradicting `benches/configs/README.md` and the `build_public_twin` doc.
- **Behavior**: `WorkerPoolBuilder::exclude_protocols` — called by the production path at `solver.rs:889` — is still not called here, so `PoolConfig::exclude_protocols` is read from the TOML and discarded. An operator who configures a pool to exclude a protocol, then replays it in the benchmark, gets a solver that still routes through that protocol, with no warning. `fynd-core/src/solver.rs:1490–1500`
- **Test coverage**: the recording path has no test at all. `test_build_all_exclusive_pools` (`solver.rs:1744`) covers only the `FyndBuilder` twin of the new all-exclusive rejection, and the newly-honoured `connector_tokens` line (`solver.rs:1483–1485`) silently changes what a recording-based pool routes through. `fynd-core/src/solver.rs:1393–1398, 1462–1469, 1483–1485`

**Proposed fix**: Extract one function that maps a `PoolConfig` onto a `WorkerPoolBuilder` and returns the fully-configured `SolverPoolHandle`, and call it from both paths. That closes the scope gap and the `exclude_protocols` gap in one place and makes the next added field impossible to apply to only one path. Then add a recording-path test asserting the handle carries the configured scope and that an excluded protocol is absent from the routed components — one test now covers what three separate omissions produced. Decide separately whether the bench request should grant exclusive access, since fixing the scope makes the exclusive config score like its twin until it does.

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -352,7 +375,7 @@ pub(crate) fn normalize_fractions(fractions: &mut [f64]) -> Result<(), SplitMath
 /// # Errors
 ///
 /// Returns [`SplitMathError::EmptyFractions`] if `fractions` is empty.
-pub(crate) fn fractions_to_amounts(
+pub fn fractions_to_amounts(
     total: &BigUint,
     fractions: &[f64],
 ) -> Result<Vec<BigUint>, SplitMathError> {
```
tamaralipows, line 392: This underflowing looks like a real bug - please have a look?

---

**[IMPL `#1`]**

**Category**: bug

**Description**: The old `assign_splits_and_amounts` divided each hop's fraction by the branch total, so the fractions it handed `fractions_to_amounts` always summed to 1.0 and `total - &running_sum` (line 396) could not go negative. `splits_from_amounts` now derives each fraction as a raw `share_of(swap.amount_in, total_available)` with no normalization, so if the amounts standing at a token sum above the balance at that token, `running_sum` exceeds `total` and the `BigUint` subtraction panics rather than returning an error. Two ways in: `PathAllocation` and `build_split_route` are now `pub`, and `PathLedger::new` (lines 726–729) states the permissive contract explicitly — the amounts "are not required to sum to the order exactly" — so an out-of-crate algorithm that over-allocates reaches it directly. Within the crate the amounts do sum exactly, leaving only the `f64` rounding margin: `share_of` is accurate to about one part in 10^16, so a branch whose smallest swap carries less than that share of the balance can still tip `running_sum` over.

**Plain English**: An operator running the solver with a routing algorithm of their own can crash the worker thread instead of getting an error back. If that algorithm hands the split machinery a set of paths that together ask for slightly more than the order is selling, the solve panics rather than reporting that the allocation is invalid — and the documentation for that machinery tells them the amounts do not have to add up. A bad allocation should come back as a routing error naming what was wrong, not take the thread down.

**Suggestion**: Either normalize the fractions before calling `fractions_to_amounts`, or make `fractions_to_amounts` use `total.checked_sub(&running_sum)` and return a new `SplitMathError` variant when the fractions oversubscribe. Then correct the `PathLedger::new` docstring to state whichever contract you land on.

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -633,17 +710,258 @@ fn build_in_degree(hops_by_token: &FxHashMap<Bytes, Vec<SplitSwap>>) -> FxHashMa
...
+        .into_iter()
+        .map(|split_swap| {
+            let fed = standing
+                .get(&hop_key(&split_swap.hop))
+                .cloned()
+                .unwrap_or_default();
+            (split_swap, fed)
+        })
+        .collect())
+}
```
tamaralipows, line 956: Perhaps will help the performance? Not sure if significantly though

---

**[IMPL `#3`]**

**Category**: perf

**Description**: `HopKey` owns a `ComponentId` and two `Bytes`, so every `hop_key` call is three heap allocations. `standing_at` calls it once per path per branch, and `amounts_for_branch` calls it twice per swap — once to size the swap and once again after the sort, where it also clones the feeder `Vec<usize>`. `execute_split_plan` runs once per golden-section evaluation inside `PathFrankWolfeAlgorithm::optimize_step_size`, so this is the hottest allocation site the diff adds; the code it replaced built the tuple once per hop in `merge_shared_hops` only.

**Suggestion**: Key `standing` on a borrowed tuple (`(&ComponentId, &Bytes, &Bytes)`) or on a hash of the triple, and carry the feeder indices out of the sizing pass alongside each `SplitSwap` so `amounts_for_branch` needs one lookup instead of two plus a clone. The second half of that is the same change the double-lookup finding on `amounts_for_branch` proposes.

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -633,17 +710,258 @@ fn build_in_degree(hops_by_token: &FxHashMap<Bytes, Vec<SplitSwap>>) -> FxHashMa
...
+
+    let executed = SimulatedSplitSwap {
+        hop: swap.hop,
+        split: swap.split,
+        amount_in: swap.amount_in,
+        amount_out: result.amount,
+        gas: result.gas,
+        pre_swap_state: sim.clone_box(),
+    };
+    Ok((executed, result.new_state))
```
tamaralipows, line 1006: I imagine an unnecessary clone on an entire pool state with thousands of ticks is not ideal?

---

**[IMPL `#7`]**

**Category**: perf

**Description**: `run_merged_swap` fills `SimulatedSplitSwap::pre_swap_state` with `sim.clone_box()` on every swap, but only `build_split_route` reads that field; `evaluate_total_output` discards it. The golden-section line search calls `evaluate_total_output` once per evaluation, so each solve deep-clones every pool state on the route tens of times for nothing, and a `vm:*` state is not a cheap clone.

**Suggestion**: Make the field `Option<Box<dyn ProtocolSim>>` and have `execute_split_plan` take a flag (or two entry points) so the scoring pass skips the clone.

**Scope**: The clone predates this branch; the diff moved it into `run_merged_swap` without changing it. Flagged because the extraction is the natural place to fix it.

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -541,20 +576,17 @@ fn merge_shared_hops(
                 desc.token_in.address.clone(),
                 desc.token_out.address.clone(),
             );
```
tamaralipows: **[SIMPLIFY `#3`]**

The three-field tuple built here is exactly `hop_key(desc)`, the helper the same commit added at line 535 — call it.

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -525,12 +561,11 @@ struct SplitExecution {
 }
 
 /// Merge shared hops across paths, summing their flow fractions, and return
-/// them collected by `token_in` (sorted by fraction descending within each
+/// them collected by `token_in` (sorted by amount descending within each
 /// branch collection).
 fn merge_shared_hops(
     paths: &[PathAllocation],
 ) -> Result<FxHashMap<Bytes, Vec<SplitSwap>>, AlgorithmError> {
```
tamaralipows: **[SIMPLIFY `#4`]**

The rewritten body has no `?` and one exit, so the signature could drop `Result` and the `?` at line 976 with it.

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -633,17 +710,258 @@ fn build_in_degree(hops_by_token: &FxHashMap<Bytes, Vec<SplitSwap>>) -> FxHashMa
...
+                .ok_or_else(|| {
+                    AlgorithmError::Other(format!(
+                        "no path feeds the swap through {} at this point in the plan",
+                        split_swap.hop.component_id,
+                    ))
+                })?;
+            split_swap.amount_in = ledger
+                .fed_amounts(fed)
+                .into_iter()
+                .sum();
```
tamaralipows: **[SIMPLIFY `#7`]**

`ledger.fed_amounts(fed).into_iter().sum()` clones every feeder's `BigUint` into a `Vec` to add them up; a `PathLedger::fed_total(&self, fed) -> BigUint` summing references avoids both.

tamaralipows (commented):
Thanks!

fynd-core/benches/common/mod.rs
```diff
@@ -500,6 +553,9 @@ pub struct Market {
     /// capture. Used only when `--gas-price-gwei` is absent, and reported either way. `None` when
     /// the fixture carried none, or no `--rpc-url` was given.
     pub market_gas_price: Option<BigUint>,
+    /// The node the market was captured through, so a solver built on it can read the
+    /// PropAMMRouter's fee tiers. `None` for a fixture, which holds no pAMM component.
+    pub rpc_url: Option<String>,
```
tamaralipows, line 565: **[DOCS `#4`]**

**Description**: The table says `--rpc-url` is "read for the live gas price". The diff adds `Market::rpc_url` (`fynd-core/benches/common/mod.rs:558`), passes it to `Solver::from_recording`, and uses it to read the PropAMMRouter's fee tiers once — without which every route holding a `propammfallback:` leg is dropped (`fynd-core/src/solver.rs:1345`, `fynd-core/src/solver.rs:1461–1467`).

**Suggestion**: Extend the `--rpc-url` cell and the paragraph at 110–112 to say the node is also read once for the PropAMMRouter fee tiers, and that an offline run has none, so a fixture holding pAMM components cannot be routed through.

fynd-core/src/algorithm/path_scoring.rs
```diff
@@ -0,0 +1,570 @@
...
+    pub(crate) pair: (NodeIndex, NodeIndex),
+    pub(crate) pools: &'g [EdgeData<D>],
+    pub(crate) data: T,
+}
+
+/// What an unsplit route through a token path paid, and the pools it went through.
+///
+/// The output is gross. Turning it into a figure net of gas needs the price of gas in the path's
+/// output token, which is the caller's to hold.
+pub(crate) struct WalkedPath {
```
tamaralipows, line 105: I'm also a fan of more concrete language than "Walked"

---

**[DOCS `#8`]**

The crate calls this outcome "simulated" everywhere else (`SimulatedHop`, `SimResult`, `simulate_token_path`, `SolvedRoute`); rename it `SimulatedPath`.

fynd-core/src/algorithm/path_scoring.rs
```diff
@@ -0,0 +1,570 @@
...
+    pub(crate) fn new(reuse_winners: bool) -> Self {
+        Self { winner_by_pair: FxHashMap::default(), reuse_winners }
+    }
+
+    /// Swaps one pair: which pool to go through, and what it pays.
+    ///
+    /// Tries the pool that won this pair before falling back to every pool.
+    ///
+    /// Returns `None` when no pool on the pair can trade the amount.
+    fn swap_pair<'g, D>(
```
tamaralipows: You can decide if you want this change but I tend to agree the name implies actually doing a swap

---

**[DOCS `#12`]**

It chooses which pool on the pair to go through and hands the choice to the caller's `simulate` closure; `best_pool_for_pair` says that without the reader opening it.

tamaralipows (approved):
Approving to not block. Didn't find anything problematic with my own eyes

fynd-core/src/solver.rs
carloszanella: I think this refactor is valid but I didn't want to make the PR bigger - the from_recording_with is just used for dev purposes right now (local tests, benchmark)

fynd-core/src/algorithm/split_primitives.rs
carloszanella: I think I came across this review item and when checked it required a bigger refactoring not making it worth it

fynd-core/src/worker_pool/worker.rs
```diff
@@ -443,7 +466,10 @@ where
                             order_id = %order.id(),
                             "route missing swaps for sell order"
                         );
-                        SolveError::no_route_found(order.id())
+                        SolveError::no_route_found_with_reason(
+                            order.id(),
+                            NoPathReason::RouteMissingSwaps,
```
louise-poole: 🤖 This same diff introduces `RouteRejection` specifically to separate "facts about this deployment: the market held a route and the worker could not price it" from NoPathReason, which is documented (in the same diff) as "what the graph search found." `NoPathReason::RouteMissingSwaps` is itself documented as "a fault in the algorithm rather than a fact about the market" — an algorithm returned a route, and the worker rejected it because it carried no swaps to read an output token from. That is precisely the "route was found, then rejected" shape `RouteRejectio`n exists to hold, not "the graph search found nothing," yet it was added to NoPathReason instead, in the very diff that drew this line. `worker_pool_router/mod.rs's cause_tier` has to carry an explicit comment explaining why this `NoPathReason` variant is ranked with algorithm faults rather than graph facts — a sign the variant is fighting its enum.

**Suggestion**
Move `RouteMissingSwaps` to `RouteRejection` (e.g. `RouteRejection::RouteMissingSwaps`) alongside the Pamm* variants, and raise it via `SolveError::route_rejected` the same way they are.

fynd-core/benches/common/mod.rs
```diff
@@ -318,27 +318,80 @@ fn recording_path() -> PathBuf {
...
+    };
+
+    match build_public_twin(&fields) {
+        // The twin splits the run's worker budget rather than adding to it, so a config carrying
+        // one is compared on the same threads as every other. An odd budget gives the exclusive
+        // pool the extra worker, and one worker each is the floor.
+        Some(public) => {
+            let public_workers = (workers / 2).max(1);
+            HashMap::from([
+                ("bench".to_string(), build(fields, (workers - public_workers).max(1))),
```
louise-poole, line 355: If `workers` is 1 and we have a floor here of 1 worker per worker pool then building a public twin will lead to 2 workers even though 1 was configured? I think we should at least document this.

fynd-core/benches/common/mod.rs
```diff
@@ -318,27 +318,80 @@ fn recording_path() -> PathBuf {
...
+    if fields
+        .get("liquidity_scope")
+        .and_then(toml::Value::as_str)? !=
+        "include_exclusive"
+    {
+        return None;
+    }
+    let mut public = fields.clone();
+    // Absent is public: `PoolConfig::liquidity_scope` defaults to `PublicOnly`.
+    public.remove("liquidity_scope");
```
louise-poole: These str matching leaves room for potential drift if we rename/adjust any config strings. Might be better to deserialise into `PoolConfig` and match on the `LiquidityScope` enum?

fynd-core/benches/bench-analyze.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+def load_routes(run_dir):
+    rows = []
+    with open(run_dir / "routes.jsonl") as handle:
+        for line in handle:
+            line = line.strip()
+            if line:
+                rows.append(json.loads(line))
+    return rows
```
louise-poole: This looks very similar to bench-setdiff.py's `load_jsonl`. Maybe we could move it to be a util fn instead of have duplicate code?

fynd-core/benches/bench-setdiff.py
```diff
@@ -0,0 +1,104 @@
...
+    parser.add_argument("--top", type=int, default=20)
+    args = parser.parse_args()
+
+    losses = 0
+    missing_pool_losses = 0
+    same_pool_losses = 0
+    # How often each pool shows up in a route that beat ours while we never used it. A pool at the
+    # top of this is one whole class of loss, not one order's bad luck.
+    unseen_pools = collections.Counter()
+    unseen_pool_bps = collections.Counter()
```
louise-poole: 🤖 `unseen_pool_bps = collections.Counter()` is an integer-counting type; `collections.defaultdict(float)` says what the values actually are without borrowing a counting collection for summed floats.

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -633,17 +710,258 @@ fn build_in_degree(hops_by_token: &FxHashMap<Bytes, Vec<SplitSwap>>) -> FxHashMa
...
+            .0
+            .entry(token.address.clone())
+            .or_default();
+        if *standing < *amount {
+            return Err(AlgorithmError::Other(format!(
+                "the swap through {component_id} spends {amount} {}, more than the {standing} \
+                 standing at it",
+                token.symbol,
+            )));
+        }
```
louise-poole, line 897: Can you add a test for this new error/check?

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -633,17 +710,258 @@ fn build_in_degree(hops_by_token: &FxHashMap<Bytes, Vec<SplitSwap>>) -> FxHashMa
...
+    let sized: Vec<SplitSwap> = branch
+        .into_iter()
+        .map(|mut split_swap| {
+            let fed = standing
+                .get(&hop_key(&split_swap.hop))
+                .ok_or_else(|| {
+                    AlgorithmError::Other(format!(
+                        "no path feeds the swap through {} at this point in the plan",
+                        split_swap.hop.component_id,
+                    ))
```
louise-poole, line 938: And test for triggering this?

fynd-core/src/worker_pool/worker.rs
```diff
@@ -399,22 +402,42 @@ where
...
                                 order_id = %order.id(),
                                 %reason,
                                 "dropping pAMM route: the Uniswap V3 fallback could not be simulated"
                             );
-                            return Err(SolveError::no_route_found(order.id()));
+                            return Err(SolveError::route_rejected(
+                                order.id(),
+                                RouteRejection::PammFallbackUnpriceable,
+                            ));
                         }
```
louise-poole, line 453: Looks like this match branch is also untested - could you add a test for it?

fynd-core/src/algorithm/split_primitives.rs
```diff
@@ -1730,70 +2039,194 @@ mod tests {
...
+        .expect("the round trip simulates");
+
+        // 1000 A buys 2000 B, which buys 2000 A back. The 1000 that was spent is gone.
+        assert_eq!(total_out, BigUint::from(2000u64), "the order's own input is not output");
+    }
+
+    /// A path set carrying nothing describes no split, and guessing an allocation for it would
+    /// misprice the quote.
+    #[test]
+    fn test_execute_split_plan_rejects_paths_that_carry_nothing() {
```
louise-poole: I am very much not a fan of these super verbose test names. I would rather they name the scenario only and not the assertions/expected outcomes.
```suggestion
    fn test_execute_split_plan_zero_amount_paths() {
```

fynd-core/src/algorithm/path_scoring.rs
```diff
@@ -0,0 +1,570 @@
...
+//!
+//! * the per-leg choice made on output *net of that pool's gas*, not on gross output;
+//! * a pool the path already crossed withheld from its later legs, so one pool's liquidity cannot
+//!   be spent twice inside a single path's score;
+//! * the winner of a pair remembered, so paths sharing a leg do not each rescan it.
+//!
+//! What is deliberately *not* here is the swap itself. The caller passes a closure that answers
+//! what one pool pays for an amount, so whether that reaches the pool or is served from amounts
+//! already asked stays the caller's business — `most_liquid` answers it through a
+//! [`SwapCache`](crate::algorithm::swap_cache::SwapCache), and does not have to say so here.
```
louise-poole, line 18: I find these doc links (square brackets + import path) are nice for AI, but not for humans reading the code. And if I have to make a choice I would lean towards making our codebase nicer for humans as that makes it easier for us to maintain in the long run. Just having 'SwapCache' is enough to indicate it's a code struct we are referring to.
```suggestion
//! `SwapCache`, and does not have to say so here.
```

fynd-core/src/algorithm/path_scoring.rs
```diff
@@ -0,0 +1,570 @@
...
+    use num_bigint::BigInt;
+    use tycho_simulation::tycho_core::models::Address;
+
+    use super::*;
+    use crate::{
+        algorithm::test_utils::fixtures::{addrs, linear_graph},
+        graph::GraphManager,
+    };
+
+    // ==================== heuristic_score Tests ====================
```
louise-poole: I would rather you use nested modules to divide tests than commented banners like this. It reflects nicer on the test reports etc.

fynd-core/src/algorithm/most_liquid.rs
```diff
@@ -57,6 +59,12 @@ enum MostLiquidError {
     TokenMissing(NodeIndex),
 }
 
+/// One leg's two tokens, and the price of gas in the token it pays out.
+///
+/// `None` where the market holds no gas price for that token, which leaves the leg scored on gross
+/// output.
+type LegTokens<'a> = (&'a Token, &'a Token, Option<&'a Price>);
```
louise-poole, line 66: This is not just tokens, but token price too no? Would be nice if the name reflected that somehow.

fynd-core/benches/bench-analyze.py
```diff
@@ -0,0 +1,172 @@
...
+    """How many parallel paths the route splits into, and how many swaps it took."""
+    edges = route.get("edges") or []
+    if not edges:
+        return (0, 0)
+    # A leg's `split` is the fraction of its input token that this swap took. A fresh parallel
+    # path shows up as a swap consuming the order's input token, so counting those counts the
+    # top-level fan-out.
+    first_token = edges[0]["token_in"]
+    top_level = sum(1 for edge in edges if edge["token_in"] == first_token)
+    return (top_level, len(edges))
```
louise-poole: The names used here feel a bit ambiguous. What is 'top_level'? What is 'shape'? I feel like this fn could be named better so that you do not need to rely on the docstring to guess what it does.

louise-poole (commented):
Looks good! Started using your review tool (marked its findings that I copied over with a 🤖 - I hit rate limits so had to manually finish off posting this review lol)

fynd-core/benches/bench-analyze.py
carloszanella: This is fully AI'd script used in debugging. I didn't put production code quality standards on it. This goes back to the question of whether we should add these things to fynd repository or somewhere else.

fynd-core/src/worker_pool/worker.rs
carloszanella: in the end I removed RouteMissingSwaps since it turned out to be unreachable

PR #488, opened 2026-08-31
feat(core)!: let a deployment bring its own routing algorithm

carloszanella:
## What

An algorithm implemented outside `fynd-core` could be written but not really used:

- A pool configuration names its algorithm, and only the built-in names resolve, so a downstream crate could not **serve** one.
- `Solver::from_recording` dispatches the same way, so it could not be **benchmarked or profiled**.
- `serve` lived in `main.rs`, so reusing Fynd's CLI meant reimplementing ~390 lines of argument handling that would then drift.
- Every `AlgorithmError` variant carrying fields is `#[non_exhaustive]`, so from outside the crate **two of the eight** could be constructed. An external algorithm could not report why it failed.

This adds the smallest interface that closes those four gaps.

## Changes

| area | change |
|---|---|
| `algorithm/registry.rs` | New `AlgorithmRegistry` — algorithms held by the name a pool configuration asks for. Empty by default. |
| `solver.rs` | `FyndBuilder::with_algorithms`; configured pools resolve through the registry before falling back to the built-in of that name. |
| `solver.rs` | `Solver::from_recording_with`, for a recorded-market solver whose algorithm is not built in. `from_recording` delegates and is unchanged. |
| `algorithm/mod.rs` | Constructors for `AlgorithmError`: `no_path`, `timeout`, `simulation_failed`, `data_not_found`, `invalid_configuration`. |
| `fynd-rpc/builder.rs` | `with_algorithms` passthrough, plus `with_algorithm` for the single-pool shorthand `FyndBuilder` already had. |
| `src/serve.rs`, `src/lib.rs` | `serve` moves out of `main.rs` into the library, taking a registry. `fynd::cli` is exposed. |
| `src/main.rs` | Keeps only `fn main`, passing an empty registry. |

## Notes

Nothing here is specific to any algorithm — the registry is a name-to-factory map, and the built-ins do not go through it. The shipped binary behaves exactly as before: it passes an empty registry, so every pool resolves by name as it does today.

The `main.rs` → `serve.rs` diff is large but almost entirely a move; the only edits are the added registry parameter and two helpers made `pub` for `commands::derive_connector_tokens`.


target-tree.txt
carloszanella, line 1: Forgot to remove this. I will do it after the PR review assuming it already started

fynd-core/CLAUDE.md
```diff
@@ -69,9 +69,24 @@ pending-block support). `Solver::subscribe_market_events()` returns a broadcast
...
+   ```
+   `FyndRPCBuilder::with_algorithms` does the same for the HTTP server, and
+   `Solver::from_recording_with` for a benchmark or profiler. A name that shipping code already
+   uses is refused, so a registration cannot silently replace a built-in.
+3. Report failures with the `AlgorithmError` constructors (`no_path`, `timeout`,
+   `simulation_failed`, ...); the variants themselves cannot be built from outside the crate.
+4. No changes to fynd-core required
+
+`FyndBuilder::with_algorithm("name", factory)` is the older single-pool shorthand: it adds a pool
+no configuration declares, using default hop limits and worker count. Prefer the registry.
```
kayibal: Can we please deprecate this? I think we should avoid having to ways of registering algorithms.

kayibal (approved):

PR #491, opened 2026-09-01
feat: add disable-slippage-taking encoding

tamaralipows:
## What

Adds a per-request opt-out of positive slippage capture, selected by API key.

## Why

Not all clients can be identified by `tx.origin`, and not all clients will want to sign client fee params. This eliminates both of those necessities. 

## How it works

The Tycho Router identifies a fee client by the signed client fee params in the calldata: whoever's address they name (and whose signature they carry) is the client the on-chain FeeCalculator applies fee configuration for. This change uses that mechanism for attribution without charging any fee.

1. The authenticating proxy maps a designated API key to a request header (`x-disable-slippage-taking: true`). Similar concept to the exclusive-access header.
2. When the header is present, Fynd builds zero-fee client fee params naming a deployment-held signing key's address as the client, signs them over the exact swap calldata, and embeds the signature in the returned transaction. **The caller signs nothing and patches nothing.**
3. On-chain, the FeeCalculator resolves that address as the client and exempts the address from positive slippage capture.

A request that combines the header with its own client fee params is rejected, since both fill the same calldata field. A request with the header but no signing key configured on the deployment fails at encode time.

## Configuration

- `DISABLE_SLIPPAGE_TAKING_SIGNER_KEY`: the deployment's signing key. Its address is the client identity; with a zero fee and zero vault contribution, no funds ever move to it.
- `x-disable-slippage-taking: true`: proxy-injected request header that turns the feature on for one request.


fynd-core/src/encoding/disable_slippage_taking.rs
```diff
@@ -0,0 +1,255 @@
...
+        let signer = key
+            .parse::<PrivateKeySigner>()
+            .map_err(|e| {
+                SolveError::FailedEncoding(format!(
+                    "invalid {ENV_DISABLE_SLIPPAGE_TAKING_KEY}: {e}"
+                ))
+            })?;
+        let router = crate::rpc::to_address(router_address, "router address")
+            .map_err(SolveError::FailedEncoding)?;
+        Ok(Some(Self::new(signer, chain_id, router, DEADLINE_WINDOW_SECS)))
```
louise-poole: Should we document somewhere that there is a deadline window for these quotes/solutions now? Without a signed fee (what was served before) these would never expire so this is a behaviour change...

fynd-core/src/encoding/encoder.rs
```diff
@@ -460,12 +522,36 @@ impl Encoder {
...
+            // are final; nothing is left for the caller to patch.
+            let signer = self
+                .disable_slippage_taking_signer
+                .as_ref()
+                .ok_or_else(|| {
+                    EncodingError::FatalError(format!(
+                        "disable-slippage-taking encoding requested but no signing key is \
+                         configured (set {ENV_DISABLE_SLIPPAGE_TAKING_KEY})"
+                    ))
+                })?;
```
louise-poole, line 539: This same check and error appears in the code twice - could we maybe centralise it so that we don't need up update multiple places if we fix something related to this error?

fynd-rpc/src/api/handlers.rs
```diff
@@ -111,6 +115,52 @@ pub fn validate_quote_request(
...
+        .get(DISABLE_SLIPPAGE_TAKING_HEADER)
+        .and_then(|value| value.to_str().ok())
+        .is_some_and(|value| value.eq_ignore_ascii_case("true"))
+}
+
+/// Marks the request's encoding options for disable-slippage-taking encoding.
+///
+/// Rejects a request that also carries `client_fee_params` — both configure the router's
+/// client fee calldata. A request without encoding options passes through unchanged.
+pub fn apply_disable_slippage_taking(
```
louise-poole: Should this really be pub? 
I actually feel like we're leaking a lot of hosted Fynd specifics in this core Fynd repo now...

fynd-core/src/types/quote.rs
```diff
@@ -350,6 +350,11 @@ pub struct EncodingOptions {
     /// Client fee configuration. When absent, no client fee is charged.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     client_fee_params: Option<ClientFeeParams>,
+    /// Attach server-signed zero-fee client params so the FeeCalculator applies the deployment
+    /// signer's positive-slippage exemption. Defaults to `false`. Mutually exclusive with
+    /// `client_fee_params`.
+    #[serde(default)]
+    disable_slippage_taking: bool,
```
louise-poole, line 357: Since these are mutually exclusive could we not use a cleverer design to make sure it's impossible to set them both? Or I guess we don't want to break backwards compatibility so maybe we should have more explicit checks/order of preference documented?

fynd-core/src/encoding/encoder.rs
```diff
@@ -328,6 +346,47 @@ impl Encoder {
...
+            if encoding_options
+                .client_fee_params()
+                .is_some()
+            {
+                return Err(SolveError::FailedEncoding(
+                    "disable_slippage_taking cannot be combined with client_fee_params; both \
+                     set the router's client fee calldata"
+                        .to_string(),
+                ));
+            }
```
louise-poole: Here we are enforcing this check, but in `encode_tycho_router_call` we simply use `client_fee_params` if both are given. I think it would be best to centralise these checks and then apply them consistently/at an earlier point in the code.

fynd-core/src/encoding/disable_slippage_taking.rs
```diff
@@ -0,0 +1,255 @@
...
+        uint256 deadline;
+        uint256 amountIn;
+        address tokenIn;
+        address tokenOut;
+        uint256 expectedAmountOut;
+        uint256 minAmountOut;
+        address receiver;
+        bytes swaps;
+    }
+}
```
louise-poole, line 59: Do we not define this already elsewhere? I feel like this sol! def is repeated. Maybe we can import instead of redefine? (only if it's simple - ignore this if my hunch here is wrong pls)

fynd-core/src/encoding/disable_slippage_taking.rs
```diff
@@ -0,0 +1,255 @@
...
+
+/// The swap-intent fields the `ClientFee` signature must cover, taken verbatim from the encoded
+/// calldata: the router recomputes the signing hash from its calldata arguments, so any difference
+/// (e.g. the native-ETH sentinel address) invalidates the signature.
+pub struct SwapIntent<'a> {
```
louise-poole: Does everything in this module need to be pub? Probably would be better to set them to `pub(crate)` no?

fynd-core/src/encoding/disable_slippage_taking.rs
```diff
@@ -0,0 +1,255 @@
...
+    /// The `clientFeeReceiver` every payload from this signer carries: the signing key's own
+    /// address.
+    pub fn receiver(&self) -> Address {
+        self.signer.address()
+    }
+
+    /// The deadline for a payload signed now.
+    pub fn deadline(&self) -> u64 {
+        now_unix_secs().saturating_add(self.deadline_window_secs)
+    }
```
louise-poole: The deadline handling in this process feels a bit disjointed. Would it not be better to calculate the deadline during `sign_client_fee` and return it with the signature? Like right now we have to call `DisableSlippageTakingSigner::deadline` to calculate the deadline then pass it back again to the same struct when we call `DisableSlippageTakingSigner::sign_client_fee`?

louise-poole: 🤖 `now_unix_secs()` maps a `duration_since(UNIX_EPOCH)` failure to `0`, so `deadline()` returns `600` — a timestamp 56 years in the past. `sign_client_fee` happily signs it and the encoder embeds it, producing calldata the router rejects on every submission rather than failing the request.

**Suggestion**: Have the deadline computation return `Result` and fail the encode with a clear "system clock is before the Unix epoch" message instead of falling back to `0`.

fynd-core/src/encoding/encoder.rs
```diff
@@ -436,6 +495,9 @@ impl Encoder {
             (None, vec![])
         };
 
+        let fn_sig = encoded_solution.function_signature();
```
louise-poole: Why did you move this earlier? I don't see it used in this new code block and feels weird to take it further away from where it's actually used...

fynd-rpc/src/api/handlers.rs
```diff
@@ -79,6 +79,10 @@ pub async fn quote(
 ) -> Result<HttpResponse, ApiError> {
     let access = exclusive_access::from_headers(http_request.headers());
     let core_request = validate_quote_request(request.into_inner())?;
+    let core_request = apply_disable_slippage_taking(
+        core_request,
+        disable_slippage_taking_requested(http_request.headers()),
+    )?;
     let capture = ReplayRequest::capture(&core_request, access);
```
louise-poole: I wonder if we should include this in the ReplayRequest capture maybe? 🤔

fynd-rpc/src/api/handlers.rs
```diff
@@ -577,6 +631,68 @@ mod tests {
...
+        fn dto_request_with_options(options: serde_json::Value) -> dto::QuoteRequest {
+            serde_json::from_value(serde_json::json!({
+                "orders": [dto_order()],
+                "options": options
+            }))
+            .expect("valid request json")
+        }
+
+        #[test]
+        fn test_apply_disable_slippage_taking_sets_encoding_flag() {
```
louise-poole: Test names are soooo looooong 😭 Be better claude! Don't need to be so verbose and describe the expected outcomes, only what is tested.

fynd-rpc/src/api/handlers.rs
louise-poole: This overrides all other fee-based mapping we use (like tx.origin etc). I think we need to clearly document this for ourselves somewhere.

fynd-core/src/encoding/encoder.rs
```diff
@@ -460,12 +522,36 @@ impl Encoder {
...
+                    min_amount_out,
+                    receiver,
+                    swaps,
+                })
+                .map_err(|e| EncodingError::FatalError(e.to_string()))?;
+            (0u32, signer.receiver(), U256::ZERO, U256::from(deadline), signature.to_vec())
         } else {
             (0u32, Address::ZERO, U256::ZERO, U256::MAX, vec![])
         };
-
```
louise-poole, line 477: I miss this blank line 👀 😂 Can we add it back for readability?

fynd-core/src/encoding/encoder.rs
```diff
@@ -203,6 +215,17 @@ impl Encoder {
...
+    /// Overrides the disable-slippage-taking signer, replacing whatever was read from the
+    /// environment.
+    #[must_use]
+    pub fn with_disable_slippage_taking_signer(
+        mut self,
+        signer: DisableSlippageTakingSigner,
+    ) -> Self {
+        self.disable_slippage_taking_signer = Some(signer);
+        self
+    }
```
louise-poole, line 230: Looks like this is only used in tests?

fynd-core/src/encoding/encoder.rs
```diff
@@ -1384,6 +1469,140 @@ mod tests {
...
+    const DISABLE_SLIPPAGE_TAKING_KEY: &str =
+        "0x3333333333333333333333333333333333333333333333333333333333333333";
+
+    fn disable_slippage_taking_signer() -> DisableSlippageTakingSigner {
+        DisableSlippageTakingSigner::new(
+            DISABLE_SLIPPAGE_TAKING_KEY
+                .parse()
+                .unwrap(),
+            1,
+            alloy::primitives::Address::repeat_byte(0x99),
```
louise-poole: Not a fan of these full path in-line imports

fynd-rpc/src/api/handlers.rs
```diff
@@ -111,6 +115,52 @@ pub fn validate_quote_request(
...
+/// Header through which the authenticating proxy requests disable-slippage-taking encoding.
+const DISABLE_SLIPPAGE_TAKING_HEADER: &str = "x-disable-slippage-taking";
+
+/// Reads whether the authenticating proxy requested disable-slippage-taking encoding.
+fn disable_slippage_taking_requested(headers: &actix_web::http::header::HeaderMap) -> bool {
+    headers
+        .get(DISABLE_SLIPPAGE_TAKING_HEADER)
+        .and_then(|value| value.to_str().ok())
+        .is_some_and(|value| value.eq_ignore_ascii_case("true"))
+}
```
louise-poole: Can you add a test for this?

fynd-rpc/src/api/handlers.rs
```diff
@@ -111,6 +115,52 @@ pub fn validate_quote_request(
...
+
+/// Marks the request's encoding options for disable-slippage-taking encoding.
+///
+/// Rejects a request that also carries `client_fee_params` — both configure the router's
+/// client fee calldata. A request without encoding options passes through unchanged.
+pub fn apply_disable_slippage_taking(
+    request: fynd_core::QuoteRequest,
+    disable_slippage_taking: bool,
+) -> Result<fynd_core::QuoteRequest, ApiError> {
```
louise-poole: Can you test that disable_slippage_taking= false behaves as expected?

fynd-core/src/types/quote.rs
louise-poole, line 357: This must only ever be set internally... I know it's omitted from the DTO struct. Can we add a test ensuring when we convert from the dto to this internal struct we have this automatically set to false? Then if anyone every updates the dto and adds this field unaware of the consequences will cause a test to fail at least?

fynd-rpc/src/api/handlers.rs
```diff
@@ -111,6 +115,52 @@ pub fn validate_quote_request(
...
+    if !disable_slippage_taking {
+        return Ok(request);
+    }
```
louise-poole: We have nothing here making sure the request options weren't already hacked and slippage taking disabled... maybe we should force/ensure with_disable_slippage_taking is false here? Not sure how necessary this is but the safer we are the better...

fynd-core/src/encoding/encoder.rs
```diff
@@ -1384,6 +1469,140 @@ mod tests {
...
+            receiver,
+            client_fee,
+            swaps,
+        ) = <SingleSwapCalldata as SolValue>::abi_decode_params(&tx.data()[4..]).unwrap();
+        let (fee_units, fee_receiver, max_contribution, deadline, signature) = client_fee;
+
+        assert_eq!(fee_units, 0, "the params must carry no fee");
+        assert_eq!(fee_receiver, signer_address);
+        assert_eq!(max_contribution, U256::ZERO);
+        assert_eq!(signature.len(), 65);
```
louise-poole: 🤖 `assert_eq!(signature.len(), 65)` is implied by the later `assert_eq!(signature.as_ref(), resigned)` against a `[u8; 65]`; you can drop it here.

fynd-core/src/encoding/encoder.rs
```diff
@@ -1384,6 +1469,140 @@ mod tests {
...
+
+        // Re-signing the decoded calldata values must reproduce the embedded signature
+        // (ECDSA signing is deterministic), proving the signature binds this exact calldata.
+        let resigned = disable_slippage_taking_signer()
+            .sign_client_fee(&crate::encoding::disable_slippage_taking::SwapIntent {
```
louise-poole: Can you remove the full path inline import here?

fynd-core/src/encoding/disable_slippage_taking.rs
```diff
@@ -0,0 +1,255 @@
...
+
+use crate::SolveError;
+
+/// Environment variable holding the disable-slippage-taking signing key (hex, with or without
+/// `0x`).
+pub(crate) const ENV_DISABLE_SLIPPAGE_TAKING_KEY: &str = "DISABLE_SLIPPAGE_TAKING_SIGNER_KEY";
+
+/// Validity window for a signed `ClientFee`, in seconds. The router rejects the params after
+/// `deadline`, so the window must outlive quote delivery, user wallet signing, and submission.
+const DEADLINE_WINDOW_SECS: u64 = 600;
```
Troshchk: I remember, that for exclusive swap the signature is valid for 120s
I think this deadline is slightly misaligned with it, coz if we have a client who is exempted from the positive slippage and has an exclusive route in the route, will not be able to execute the quote in 120s, not 600s

Troshchk: `DEFAULT_DEADLINE_WINDOW_SECS` this is the constant

Troshchk: So I think we should align these 2 values

Troshchk (commented):
Thank you!

louise-poole (approved):

PR #493, opened 2026-09-01
fix(fynd-client): resolve types for CommonJS consumers

kayibal:
## Problem

#492 made `require('@kayibal/fynd-client')` work at runtime. Two gaps remain for the CommonJS consumers it targeted, and nothing in CI would have caught either one.

**1. Types do not resolve under `moduleResolution: node`.** That mode is the default for `module: commonjs`, and it is what NestJS projects use. It ignores the `exports` map entirely and reads the top-level `main` and `types` fields, which the package does not have:

```
error TS2307: Cannot find module '@kayibal/fynd-client' or its
corresponding type declarations.
```

**2. The published declarations have a dangling import.** `tsc` does not copy hand-written `.d.ts` sources into `outDir`, so `src/schema.d.ts` never reaches `dist/`. The shipped `dist/autogen.d.ts` imports it anyway. Consumers that do not set `skipLibCheck` see:

```
dist/autogen.d.ts(10,28): error TS2307: Cannot find module './schema.js'
```

## Fix

Add the top-level `main` and `types` fields, and copy `src/schema.d.ts` into `dist/` as part of the build.

## Test

`scripts/check-client-packaging.sh` runs in the TypeScript client CI job. It packs the client with `npm pack`, installs the tarball into a throwaway fixture, and checks that:

- a CommonJS consumer can `require()` the package and gets the named exports
- an ESM consumer can `import` it
- `require()` and `import` share one module instance, so a consumer that mixes them does not get two copies of the module state
- types resolve under `moduleResolution: node` and under `moduleResolution: nodenext`
- the shipped declarations have no unresolved imports

Each check was verified against the corresponding broken state: reverting the #492 exports map fails the `require()` check, dropping `main`/`types` fails the `moduleResolution: node` check, and dropping the `schema.d.ts` copy fails the self-containment check.

## Known limitation

A consumer that sets `module: node16` and emits CommonJS still gets TS1479, because the package is `"type": "module"` and ships a single ESM `index.d.ts`. Such projects need `module: nodenext` with TypeScript 5.8 or later. Closing that gap needs a flattened `index.d.cts`, which needs a declaration bundler; a plain copy of `index.d.ts` does not work, because its relative imports then resolve in CJS mode and fail. The README documents this, along with the Node floor for `require(esm)` (20.19, 22.12, 23.0).

## Note

The fix reaches consumers on the next release tag. `release.yaml` sets the client version from the tag and skips the publish when that version is already on npm, so merging alone publishes nothing.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -147,6 +147,8 @@ jobs:
       - run: pnpm --dir clients/typescript --filter @kayibal/fynd-client run typecheck
       - run: pnpm --dir clients/typescript --filter @kayibal/fynd-client run lint
       - run: pnpm --dir clients/typescript --filter @kayibal/fynd-client run test
+      - name: Check published package shape (CJS + ESM + type resolution)
+        run: ./scripts/check-client-packaging.sh
 
   integration_tests:
     name: Integration Tests
--- a/clients/typescript/client/README.md
+++ b/clients/typescript/client/README.md
@@ -10,6 +10,19 @@ For documentation, guides, and API reference visit **<https://docs.fynd.xyz/>**.
 npm install @kayibal/fynd-client
 ```
 
+## Module systems
+
+The package ships a single ESM build. ESM consumers need Node 18 or later.
+
+CommonJS consumers load that build through `require(esm)`, which needs Node
+20.19, 22.12, or 23.0 and later. On earlier versions `require()` fails with
+`ERR_REQUIRE_ESM`, and the workaround is a dynamic `import()`.
+
+TypeScript finds the types with `moduleResolution: node` and with
+`moduleResolution: nodenext`. One case does not work: a project that sets
+`module: node16` and emits CommonJS gets error TS1479, because the declarations
+are ESM. Set `module: nodenext` and use TypeScript 5.8 or later instead.
+
 ## Quick start
 
 See the [quickstart guide](https://docs.fynd.xyz/get-started/quickstart) and the
--- a/clients/typescript/client/package.json
+++ b/clients/typescript/client/package.json
@@ -7,6 +7,8 @@
     "url": "https://github.com/propeller-heads/fynd"
   },
   "type": "module",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
@@ -19,7 +21,7 @@
     "dist"
   ],
   "scripts": {
-    "build": "tsc",
+    "build": "tsc && cp src/schema.d.ts dist/schema.d.ts",
     "typecheck": "tsc --noEmit",
     "lint": "oxlint src/",
     "test": "vitest run"
--- a/scripts/check-client-packaging.sh
+++ b/scripts/check-client-packaging.sh
@@ -1,0 +1,151 @@
+#!/usr/bin/env bash
+# Verify the published shape of @kayibal/fynd-client.
+#
+# Packs the client the same way `npm publish` would, installs the tarball into a
+# throwaway fixture, then checks that consumers can actually load it:
+#   - CommonJS `require()` (Node >= 22.12 loads the ESM build synchronously)
+#   - ESM `import`
+#   - TypeScript type resolution under `moduleResolution: node` (the default for
+#     `module: commonjs`, still used by NestJS and other CJS toolchains)
+#   - TypeScript type resolution under `moduleResolution: nodenext`
+#
+# Known limitation, deliberately not asserted here: a consumer on
+# `module: Node16` that emits CommonJS gets TS1479, because the package is
+# `"type": "module"` and ships a single ESM `index.d.ts`. Such consumers need
+# `module: nodenext` with TypeScript >= 5.8. Fixing that needs a flattened
+# `index.d.cts`, which needs a declaration bundler.
+#
+# Usage:
+#   ./scripts/check-client-packaging.sh
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+TS_ROOT="$REPO_ROOT/clients/typescript"
+CLIENT_DIR="$TS_ROOT/client"
+TSC="$CLIENT_DIR/node_modules/.bin/tsc"
+
+WORK_DIR="$(mktemp -d)"
+trap 'rm -rf "$WORK_DIR"' EXIT
+
+fail() {
+    echo "FAIL: $*" >&2
+    exit 1
+}
+
+echo "==> Building the client..."
+pnpm --dir "$TS_ROOT" install --frozen-lockfile --silent
+pnpm --dir "$TS_ROOT" --filter @kayibal/fynd-client --silent run build
+[[ -x "$TSC" ]] || fail "tsc not found at $TSC"
+
+echo "==> Packing the client..."
+tarball="$(cd "$CLIENT_DIR" && npm pack --silent --pack-destination "$WORK_DIR")"
+[[ -f "$WORK_DIR/$tarball" ]] || fail "npm pack produced no tarball"
+echo "    $tarball"
+
+echo "==> Installing the tarball into a fixture..."
+cd "$WORK_DIR"
+echo '{"name":"packaging-fixture","version":"0.0.0","private":true}' > package.json
+npm install --silent --no-audit --no-fund "$WORK_DIR/$tarball"
+
+# The exports map must resolve for both module systems, and the named exports
+# must survive the require(esm) boundary.
+cat > require.cjs <<'EOF'
+const pkg = require("@kayibal/fynd-client");
+const expected = ["FyndClient", "FyndError", "createFyndClient", "swapSigningHash", "withPermit2"];
+for (const name of expected) {
+  if (typeof pkg[name] !== "function") {
+    throw new Error(`require(): expected ${name} to be a function, got ${typeof pkg[name]}`);
+  }
+}
+new pkg.FyndClient({ baseUrl: "http://127.0.0.1:9" });
+EOF
+
+cat > import.mjs <<'EOF'
+import { FyndClient, FyndError, createFyndClient } from "@kayibal/fynd-client";
+for (const [name, value] of Object.entries({ FyndClient, FyndError, createFyndClient })) {
+  if (typeof value !== "function") {
+    throw new Error(`import: expected ${name} to be a function, got ${typeof value}`);
+  }
+}
+new FyndClient({ baseUrl: "http://127.0.0.1:9" });
+EOF
+
+# One module instance must serve both entry conditions. Two copies would give
+# two sets of module state to a consumer that mixes require() and import.
+cat > single-instance.mjs <<'EOF'
+import { createRequire } from "node:module";
+import { FyndClient } from "@kayibal/fynd-client";
+const required = createRequire(import.meta.url)("@kayibal/fynd-client");
+if (required.FyndClient !== FyndClient) {
+  throw new Error("require() and import gave different FyndClient classes");
+}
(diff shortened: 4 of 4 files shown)


scripts/check-client-packaging.sh
```diff
@@ -0,0 +1,151 @@
...
+  },
+  "files": ["consumer.ts"]
+}
+EOF
+echo "==> Checking that the shipped declarations are self-contained..."
+own_errors="$("$TSC" -p tsconfig.libcheck.json 2>&1 | grep -F "node_modules/@kayibal/fynd-client/" || true)"
+if [[ -n "$own_errors" ]]; then
+    echo "$own_errors" >&2
+    fail "the shipped declarations have unresolved imports"
+fi
```
mp-web3, line 148: This doesn't check whether the compiler crashed. A clean run and a crashed run both produce no matching output, so both print "All packaging checks passed."

Check the exit code instead. A healthy run here exits 0 with no errors, so the grep isn't needed:

```bash
"$TSC" -p tsconfig.libcheck.json || fail "the shipped declarations have unresolved imports"
```

If the grep was there to filter out viem and ox errors, they don't produce any under `nodenext`. I checked.

scripts/check-client-packaging.sh
```diff
@@ -0,0 +1,151 @@
...
+trap 'rm -rf "$WORK_DIR"' EXIT
+
+fail() {
+    echo "FAIL: $*" >&2
+    exit 1
+}
+
+echo "==> Building the client..."
+pnpm --dir "$TS_ROOT" install --frozen-lockfile --silent
+pnpm --dir "$TS_ROOT" --filter @kayibal/fynd-client --silent run build
```
mp-web3, line 37: `pnpm --filter` reports success when the filter matches nothing. I ran it with a package name that doesn't exist and it still returned success.

So if this name ever stops matching, after a rename for example, the build silently does nothing. The script then bundles up whatever was left in `dist/` from the last build, validates that, and says everything passed.

Target the package directly and check the output exists:

```bash
pnpm --dir "$CLIENT_DIR" run build
[[ -f "$CLIENT_DIR/dist/index.js" && -f "$CLIENT_DIR/dist/schema.d.ts" ]] || fail "build produced no dist"
```

scripts/check-client-packaging.sh
```diff
@@ -0,0 +1,151 @@
...
+[[ -x "$TSC" ]] || fail "tsc not found at $TSC"
+
+echo "==> Packing the client..."
+tarball="$(cd "$CLIENT_DIR" && npm pack --silent --pack-destination "$WORK_DIR")"
+[[ -f "$WORK_DIR/$tarball" ]] || fail "npm pack produced no tarball"
+echo "    $tarball"
```
mp-web3, line 43: Nothing checks that `dist/schema.d.ts` actually ended up in the file `npm pack` produced. That copy is half of what this PR fixes, and the only thing covering it is the check at line 144, which can pass on a broken run.

One line here, and unlike that check it can't quietly stop testing anything:

```bash
tar -tzf "$WORK_DIR/$tarball" | grep -qx 'package/dist/schema.d.ts' \
  || fail "the packed tarball is missing dist/schema.d.ts"
```

clients/typescript/client/package.json
```diff
@@ -7,6 +7,8 @@
     "url": "https://github.com/propeller-heads/fynd"
   },
   "type": "module",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
```
mp-web3, line 11: You only need one of these two.

I dropped each in turn and re-ran your own `moduleResolution: node` check. Dropping `main` still passes. Dropping `types` still passes. Only dropping both fails. TypeScript falls back from `main` to the `.d.ts` sitting next to it, so either field alone does the job.

That means the check can't tell you if one of them gets deleted later. Keep one, or keep both and say why in a comment.

scripts/check-client-packaging.sh
```diff
@@ -0,0 +1,151 @@
...
+WORK_DIR="$(mktemp -d)"
+trap 'rm -rf "$WORK_DIR"' EXIT
+
+fail() {
+    echo "FAIL: $*" >&2
+    exit 1
+}
+
+echo "==> Building the client..."
+pnpm --dir "$TS_ROOT" install --frozen-lockfile --silent
```
mp-web3, line 36: `--silent` also hides the failures. I ran `npm --silent run build` on a script that fails on purpose. It returned the failure code and printed nothing at all, not one byte, not even the error.

So a failed install in CI gives you a red step with an empty log. Line 48 has no `|| fail` either, so it stops with no message.

Drop `--silent` from the two installs and the build. Keep it on `npm pack`, where you read the output. Add `|| fail "could not install the packed tarball into the fixture"` to line 48.

scripts/check-client-packaging.sh
```diff
@@ -0,0 +1,151 @@
...
+cat > require.cjs <<'EOF'
+const pkg = require("@kayibal/fynd-client");
+const expected = ["FyndClient", "FyndError", "createFyndClient", "swapSigningHash", "withPermit2"];
+for (const name of expected) {
+  if (typeof pkg[name] !== "function") {
+    throw new Error(`require(): expected ${name} to be a function, got ${typeof pkg[name]}`);
+  }
+}
+new pkg.FyndClient({ baseUrl: "http://127.0.0.1:9" });
+EOF
```
mp-web3, line 61: `require()` on this package returns everything or throws. I checked against the real packed file and `Object.keys()` gives all 14 exports. There's no case where `FyndClient` comes through and `withPermit2` doesn't, so five names test exactly what one name tests.

The list is also a second copy of `src/index.ts`. Rename one of those five on purpose and this check fails saying the packaging is broken, which sends you looking in the wrong place.

```js
if (typeof pkg.FyndClient !== "function") throw new Error("require(): FyndClient is not a function");
```

scripts/check-client-packaging.sh
```diff
@@ -0,0 +1,151 @@
...
+
+cat > import.mjs <<'EOF'
+import { FyndClient, FyndError, createFyndClient } from "@kayibal/fynd-client";
+for (const [name, value] of Object.entries({ FyndClient, FyndError, createFyndClient })) {
+  if (typeof value !== "function") {
+    throw new Error(`import: expected ${name} to be a function, got ${typeof value}`);
+  }
+}
```
mp-web3, line 69: This loop never runs. If the package doesn't export one of those three names, Node fails on line 64 before any code executes. I tested it and the line after a bad import never printed.

The import itself is already the check. Delete lines 65 to 69.

scripts/check-client-packaging.sh
```diff
@@ -0,0 +1,151 @@
...
+# One module instance must serve both entry conditions. Two copies would give
+# two sets of module state to a consumer that mixes require() and import.
+cat > single-instance.mjs <<'EOF'
+import { createRequire } from "node:module";
+import { FyndClient } from "@kayibal/fynd-client";
+const required = createRequire(import.meta.url)("@kayibal/fynd-client");
+if (required.FyndClient !== FyndClient) {
+  throw new Error("require() and import gave different FyndClient classes");
+}
+EOF
```
mp-web3, line 82: This can't fail today. All three entries in the `exports` map point at the same `dist/index.js`, so `require` and `import` are loading one file and must get the same class back.

It would start meaning something the day `require` points at a separate CommonJS build. Either delete it, or say in the comment that it's there for that future case, because right now the comment reads like the problem exists already.

scripts/check-client-packaging.sh
```diff
@@ -0,0 +1,151 @@
...
+  "compilerOptions": {
+    "module": "nodenext",
+    "moduleResolution": "nodenext",
+    "strict": true,
+    "noEmit": true,
+    "skipLibCheck": false
+  },
+  "files": ["consumer.ts"]
+}
+EOF
```
mp-web3, line 142: This config is the same as the `tsconfig.nodenext.json` written above, except for `skipLibCheck`. Pass it as a flag instead and delete the file:

```bash
"$TSC" -p tsconfig.nodenext.json --skipLibCheck false
```

I tested it against the broken state and the flag overrides the setting in the file. Saves 12 lines.

clients/typescript/client/README.md
```diff
@@ -10,6 +10,19 @@ For documentation, guides, and API reference visit **<https://docs.fynd.xyz/>**.
 npm install @kayibal/fynd-client
 ```
 
+## Module systems
+
+The package ships a single ESM build. ESM consumers need Node 18 or later.
+
+CommonJS consumers load that build through `require(esm)`, which needs Node
+20.19, 22.12, or 23.0 and later. On earlier versions `require()` fails with
+`ERR_REQUIRE_ESM`, and the workaround is a dynamic `import()`.
```
mp-web3, line 19: This contradicts line 6 of the new script, which says `require()` needs Node 22.12 or later. Here it says 20.19, 22.12, or 23.0. The README is right, so fix the script comment.

Separately, "Node 18 or later" has no source I can find in the repo. There's no `engines` field, no `.nvmrc`, the tsconfig targets ES2020, and CI only ever runs Node 22. Add `"engines": { "node": ">=18" }` so the number lives somewhere npm can act on, or drop the sentence.

clients/typescript/client/package.json
```diff
@@ -19,7 +21,7 @@
     "dist"
   ],
   "scripts": {
-    "build": "tsc",
+    "build": "tsc && cp src/schema.d.ts dist/schema.d.ts",
```
mp-web3, line 24: `dist/` is now only complete if this script ran. A plain `tsc` from an editor build task skips the copy, and `npm publish` will ship the result.

Add `"prepack": "pnpm run build"` so packing can't produce a half-built `dist/`.

.github/workflows/ci.yaml
```diff
@@ -147,6 +147,8 @@ jobs:
       - run: pnpm --dir clients/typescript --filter @kayibal/fynd-client run typecheck
       - run: pnpm --dir clients/typescript --filter @kayibal/fynd-client run lint
       - run: pnpm --dir clients/typescript --filter @kayibal/fynd-client run test
+      - name: Check published package shape (CJS + ESM + type resolution)
+        run: ./scripts/check-client-packaging.sh
```
mp-web3, line 151: This step downloads viem fresh from npm on every run, roughly 48 MB unpacked. The throwaway project it installs into has no lock file, and the job caches the pnpm store rather than npm's, so nothing is reused.

`check_ts_client` also has no `timeout-minutes`, so if npm hangs the job sits there for the 360 minute default.

Add `timeout-minutes: 15` to the job, cache `~/.npm`, and pass `--prefer-offline` to that install.

mp-web3 (changes requested):
The fix works. I packed the client and confirmed `dist/schema.d.ts` reaches the tarball, and that a consumer on `moduleResolution: node` resolves the types.

Requesting changes on the new script rather than the fix. Two of its five checks can pass on a broken build, and the one thing this PR actually fixes is only covered by the check that can go silent. Details inline. The first three are the ones I'd want fixed before merge; the rest are smaller.

Two things not anchored to the diff:

**`/run-ci` won't run this check.** The TypeScript chain in `.claude/skills/run-ci/SKILL.md:89` stops at `run test`, so running it locally reports green and CI then fails. Same gap in `.claude/knowledge/typescript.md` and `clients/CLAUDE.md`, which both list the checks as typecheck, lint and test.

**This tightens types for every consumer, not just CommonJS ones.** Before the fix the shipped declarations couldn't resolve `./schema.js`, so under `skipLibCheck: true` the path types quietly became `any` and `createFyndClient` accepted any path string. I confirmed it on a fixture: a call with a bogus path compiles clean without `dist/schema.d.ts` and fails TS2345 with it. So downstream code that was passing a wrong path will start failing to compile on the next release. Probably what you want, but worth checking the docs quickstart still compiles before you tag.

PR #496, opened 2026-09-02
feat: Optional quote simulation

carloszanella:


fynd-core/src/simulation/simulator.rs
```diff
@@ -0,0 +1,358 @@
...
+            .parse()
+            .map_err(|error| format!("invalid Permit2 address: {error}")),
+    }
+}
+
+fn simulation_funding_value() -> U256 {
+    // Leaving the high bit clear avoids node-side arithmetic overflow while still funding any
+    // practical quote.
+    U256::MAX >> 1
+}
```
kayibal: why use a function for a constant value?

fynd-core/src/simulation/simulator.rs
```diff
@@ -0,0 +1,358 @@
...
+        to: Some(TxKind::Call(router)),
+        value: Some(value),
+        input: Bytes::copy_from_slice(data).into(),
+        ..Default::default()
+    };
+    let payload = SimulatePayload::default().extend(
+        SimBlock::default()
+            .with_state_overrides(overrides)
+            .call(call),
+    );
```
kayibal, line 339: I would suggest to simulate more realistically here. Meaning setting a sensible gas limit, e.g. what we estimated *2, setting a gas price (means the caller eoa needs ETH, which you can give it via overrides), also setting some other environment variables randomly such as the caller, coinbase, etc.

This is a small change to add that will save us some pain on some more edge case pools. Also don't we have to add pAMM overrides here as well?

kayibal: Small except from Claude:

>   Principle: set everything a contract can read to what the real tx will see.
> 
>   Transaction envelope
>   - tx.origin / from — real, funded EOA with nonce > 0; never 0x0/0x1/0x2 (this family explicitly checks tx.origin ∈ {0x0, 0x1, 0x2})
>   - Gas limit — the router's actual gasleft() at that hop, ~1–3 M, not 8–30 M. This was the actual exploit vector: Tycho simulates hooks with 8 M gas, and
>     the hook treats gasleft() > ~4.04 M as "simulation" → quotes 0% fee, then charges a random 2.5/5/7% fee on the real swap
>   - gasprice ≥ basefee
> 
>   Block environment
>   - block.number = head + 1
>   - block.timestamp = parent + 12
>   - prevrandao = random non-zero
>   - blockhash populated for the last 256 blocks
>   - basefee = next-block value
>   - coinbase = a real builder address
>   - block.gaslimit ≈ 45 M
>   - chainid, blobbasefee correct
> 
>   State overrides
>   - Realistic ETH balances on the sender
> 

I'd ideally set all of those even if the values are not 100% realistic here.

fynd-core/src/simulation/simulator.rs
```diff
@@ -0,0 +1,358 @@
...
+    ) -> SimulationResult {
+        self.simulate_attempt(quote, encoding_options)
+            .await
+            .into_result()
+    }
+
+    pub(crate) async fn simulate_attempt(
+        &self,
+        quote: &OrderQuote,
+        encoding_options: &EncodingOptions,
```
kayibal: I find it weird to require encoding options if you only use `.transfer_type()`later on

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -379,6 +387,12 @@ impl WorkerPoolRouter {
         self
     }
 
+    /// Makes encoded quote simulation available for this router.
+    pub fn with_simulator(mut self, simulator: QuoteSimulator) -> Self {
+        self.simulator = Some(simulator);
+        self
+    }
+
```
kayibal, line 389: I am not sure I appreciate having another async request in the hot path. Ideally this would be done on the request handler imo. IIUC, worker pool threads are limited and we shouldn't do any longer I/O here but I am not 100% sure since the functions is async. Worth doble checking.

carloszanella: apparently this is awaited directly in the router handle, so moving this up to the caller would not change anything in terms of CPU or threads

> WorkerPoolRouter::quote is an async fn awaited directly by the actix handler at fynd-rpc/src/api/handlers.rs:91. The solver pools are reached by queue.enqueue(order, params).await over an async_channel (worker_pool_router/mod.rs:704); the dedicated OS threads run Algorithm::find_best_route and nothing else. So the simulation await sits on a tokio task, not  a solver thread — no worker thread is held. 
> 
> That makes the proposed move a no-op for the stated concern: handlers::quote awaits router.quote(), so simulating after it returns is the same task, the same runtime, the same wall-clock. It relocates code without changing scheduling.

kayibal (approved):

PR #497, opened 2026-09-02
fix: improve protocol resolution handling

louise-poole:
Removing a protocol from Tycho took Fynd down at startup:

      `setup error: failed to resolve protocols: excluded protocol 'vm:fermiswap' is not in the resolved list`

  Two cases in `resolve_protocols` now warn and continue instead of failing: an `exclude:` entry that matches nothing, and a requested protocol Tycho no longer serves (previously it stayed in the list and the stream burned the full startup timeout waiting on a system nothing publishes).

  Malformed input stays fatal — an `exclude:` naming nothing, a protocol both requested and excluded, an empty resolved list.

  To check availability, the protocol-systems list is now fetched whenever any entry names a Tycho system, not only for `all_onchain`/`native_onchain`. Adds `wiremock` as a dev-dependency of `fynd-rpc` to test this against a mock Tycho.


zizou0x (approved):
LGTM!

Troshchk (approved):
Thank you!

PR #500, opened 2026-09-02
refactor(bench): move the benchmark into a crate an outside algorithm can run

carloszanella:
Moves the algorithm benchmark out of `fynd-core/benches/` into a `fynd-bench-harness` crate, and
gives its two programs a registry parameter, so a crate outside this workspace can benchmark its own
routing algorithm against the built-in ones and get the same report.

This is what the aperture port needs: the algorithm now lives in `fynd-hosted-service`, and no
benchmark here can reach it.

## What it does now

`bench::run(&algorithms)` and `profile::run(&algorithms)` are the crate's interface. Each builds its
solvers through `Solver::from_recording_with`, so a pool naming an algorithm in the registry is
served by it and every other pool by the built-in of that name. The two `benches/` targets here pass
an empty registry, which is every built-in.

Three paths a caller outside this repository has to name:

| flag | why |
|---|---|
| `--configs-dir` | Where the caller's config files are. Read on top of the built-in ones, which still run. A name held by two directories stops the run |
| `--fixture` | The market. This repository's copy is in Git LFS, so a checkout cargo makes for a git dependency holds the pointer file |
| `--trades` | The order dataset, gitignored here and so not in that checkout either |

The configs, the token table and the blocked list ship inside the crate and resolve against it, so a
caller reads the same ones without copying anything.

Inside this repository nothing changes: `./scripts/bench.sh`, `./scripts/profile.sh` and
`./scripts/bench-viewer.sh` take the same flags and write the same `bench-results/<run>/` files.

## Behavior changes worth knowing

- A run that cannot start now exits non-zero instead of printing and exiting 0.
- `run.json` records the fixture that was replayed and the file behind each config name. With
  `--configs-dir` a label no longer identifies one file, and two runs whose `BF_d2` came from
  different files are not comparable.
- `fynd-bench-harness` depends on `fynd-core` with `test-utils`, and cargo unifies features, so a
  whole-workspace build compiles `fynd-core` with it on. The feature is additive and the Docker
  build names its packages explicitly, so the image is unaffected.

## Verified

`./check.sh` passes (1701 tests). Both scripts run against the recorded market, and aperture in
`fynd-hosted-service` was benchmarked through this crate as an outside caller: it reports alongside
`BF_d2` in one run, at +30.0 mean bps over 8 orders.

https://claude.ai/code/session_018tNrcXBusq9sGHvuEDE3kY


Troshchk (approved):
Thanks!

bench-harness/src/lib.rs
```diff
@@ -258,64 +291,103 @@ pub fn block_components(updates: &mut [Update], blocked: &HashSet<Address>) -> u
...
-                .map(|stem| stem.to_string_lossy().into_owned())
+impl ConfigCatalog {
+    /// The built-in configurations, plus every `.toml` in `extra`.
+    ///
+    /// # Errors
+    ///
+    /// If a directory cannot be read, or if two directories hold the same stem. Both are the
+    /// caller's mistake, and both would otherwise shrink or silently change a run: a mistyped
+    /// `--configs-dir` would run the built-ins alone, and a duplicate would decide by order.
+    pub fn new(extra: &[PathBuf]) -> Result<Self, String> {
```
louise-poole: Unless we need these as pub? Then just update the README?

---

`ConfigCatalog::new`/`load`/`available` and `ConfigFlags::catalog` are `pub fn` on `pub(crate)` structs, and `Timings` mixes `pub p50_us`/`pub p95_us` with `pub(crate) slowest_us`. `README.md:376` states the opposite ("Only the two `run` functions are public; everything else is `pub(crate)`") — mark all five `pub(crate)` so the claim holds.

bench-harness/src/bench.rs
```diff
@@ -894,10 +914,17 @@ fn run_json(outcome: &BenchOutcome<'_>) -> String {
...
             .results
             .iter()
-            .map(|(config, _)| config.label.clone())
+            .map(|(config, _)| {
+                serde_json::json!({
+                    "label": config.label.clone(),
+                    "path": config.path.display().to_string(),
+                })
+            })
             .collect::<Vec<_>>(),
```
louise-poole: `run_json` used to write `"configs"` as an array of label strings; it now writes an array of `{label, path}` objects. `bench-harness/viewer/index.html` was not updated to match: `CONFIGS = facts.configs` (index.html:1642) and everything downstream — `RIVALS = CONFIGS.filter(c => c !== BASELINE)` (1643), `CONFIGS.includes(state.algo)` (1645), the `byConfig[c]` / `R[c]` / `p.m[c]` lookups (1689, 1876, 2387, 2801), the `c === BASELINE` comparisons (2390, 2809), and the `${c}` interpolations (2129, 2001–2002) — still treat each entry as the bare label. `orders.csv` and `routes.jsonl` still key by `config.label`, so nothing matches once `run.json` hands back an object.

Keep `"configs"` a flat label list and put the file mapping under a separate key (`"config_files": {label: path}`), which leaves the viewer untouched. Updating every `CONFIGS`/`RIVALS` consumer in `index.html` to read `.label` works too, but it's ~20 sites for a field the viewer never displays.

bench-harness/src/bench.rs
```diff
@@ -894,10 +914,17 @@ fn run_json(outcome: &BenchOutcome<'_>) -> String {
...
         "skipped": outcome.skipped,
         "gas_price_gwei": run.gas_price_gwei,
```
louise-poole, line 935: `run_json` writes the contract the viewer and the analysis scripts read, and nothing exercises it — `run_json`, `write_run` and `BenchOutcome` appear nowhere outside `bench.rs`. That's how the `"configs"` field went from an array of strings to an array of objects without the viewer noticing.

Add a test that builds a minimal `BenchOutcome` and asserts the shape of what `run_json` emits. Asserting `configs[0]` is a string would have caught this one.

bench-harness/src/lib.rs
```diff
@@ -419,42 +514,54 @@ pub enum MarketMode {
 ///
 /// Returns a message when a live capture cannot be made. The offline path panics instead -- a
 /// missing fixture is a broken checkout, not a run-time condition.
```
louise-poole: The doc says the offline path panics because "a missing fixture is a broken checkout, not a run-time condition". That held when the fixture path was hardcoded; `--fixture` (lib.rs:543–547) now lets an outside caller point it anywhere, so a typo'd path is exactly a run-time condition for them, and `load_market` (lib.rs:686–693) panics either way.

Rewrite the `# Errors` note to say the offline path panics on an unreadable or non-market fixture. The panic message itself is already actionable, so the doc is what needs fixing; having `load_market` return a `Result` that `build_market` propagates with `?`, the way the `Live` branch already does, is the fuller change if you'd rather external callers handle it.

bench-harness/src/lib.rs
```diff
@@ -523,9 +630,12 @@ impl LiveFlags {
 /// Serialized straight into `run.json`, which is what the viewer's run picker filters on.
 #[derive(Clone, Debug, serde::Serialize)]
 #[serde(tag = "source", rename_all = "snake_case")]
-pub enum MarketSource {
+pub(crate) enum MarketSource {
     /// Replayed from the recorded fixture.
     Offline {
+        /// The recording that was replayed. Two offline runs only compare when this matches, and
+        /// `--fixture` means it is no longer always the same file.
```
louise-poole: "`--fixture` means it is no longer always the same file" says what used to be true; the preceding sentence already carries the point. Cut the clause and keep "Two offline runs only compare when this matches."

bench-harness/src/bench.rs
```diff
@@ -171,20 +172,25 @@ struct Args {
...
     /// is charging, and an offline run at the default, since the fixture carries no gas price.
-    #[arg(long, value_parser = common::parse_gas_price_gwei)]
+    #[arg(long, value_parser = crate::parse_gas_price_gwei)]
     gas_price_gwei: Option<f64>,
 
-    /// Market flags: `--market`, and the Tycho settings a live capture needs.
+    /// Market flags: `--market`, the fixture an offline run replays, and the Tycho settings a live
+    /// capture needs.
+    #[command(flatten)]
+    market: MarketFlags,
```
louise-poole, line 181: Up to you.

---

Both `Args` structs (`bench.rs`, and `profile.rs:69–99`) name the market flags after the struct but name the config flags after its inner `Vec`. Rename it `configs` to match its sibling.

louise-poole (commented):
The move is clean and the registry seam does what the aperture port needs. One thing to fix before this lands: `run.json`'s `configs` field changed shape and the viewer wasn't updated with it, so the results page breaks for every run written from here on. The rest is small.

---

**Stale dev-dependencies** left in `fynd-core/Cargo.toml`** — `fynd-core/Cargo.toml:55,57,58`
`clap`, `tempfile` and `toml` were only used by the benches this PR moved into `bench-harness`; nothing under `fynd-core/src` or `fynd-core/tests` references them now. Drop the three lines.

PR #504, opened 2026-09-02
feat: metrics for quote simulation, and a log of the winning quote's protocols

carloszanella:
Two things, both so the quote path is readable from Grafana and Loki.

## Quote simulation metrics

| Metric | Type | Labels | Meaning |
|---|---|---|---|
| `quote_simulations_total` | counter | `outcome` = `success` / `reverted` / `failed`, `pool`, `algorithm` | One count per simulated quote |
| `quote_simulation_deviation_bps` | histogram | `pool`, `algorithm` | Gap between the simulated amount and what the quote promised, in basis points |

The router returns the output after it takes the router and client fees, so the
deviation is measured against the quoted amount less those same fees, taken from
the quote's fee breakdown. Comparing against the raw `amount_out` would record a
standing gap the size of the fee rather than simulation drift. A negative value
means the simulation returned less than the quote promised.

`reverted` and `failed` are counted apart: a revert means the route the solver
priced does not execute, while a failure means the simulation itself did not run,
so it says nothing about the route. Both are recorded inside the simulator, so
every caller is measured; `QuoteSimulator::simulate`, which had no callers left,
is gone.

The histogram gets its own Prometheus buckets, symmetric to 1000 bps on both
sides, because `histogram_quantile` cannot interpolate past the outermost bound.

```promql
# success rate
sum(rate(quote_simulations_total{outcome="success"}[5m]))
  / sum(rate(quote_simulations_total[5m]))

# median deviation, by solver
histogram_quantile(0.50, sum by (le, pool) (rate(quote_simulation_deviation_bps_bucket[5m])))
```

Both panels are added to the local `monitoring/grafana/dashboards/fynd.json`.

## Winning-quote protocol log

One line per returned route, on the `fynd::winning_protocols` target at TRACE:

```
winning_protocols order_id=abc block=23481902 pool=aperture_d3 algorithm=aperture swaps=3 protocols={"uniswap_v2":2,"vm:balancer_v2":1}
```

The counts are swaps, not distinct pools: a split route that crosses two
Uniswap V2 pools reports 2, which is what the solution executes. Only a
successful quote with a route gets a line.

TRACE, matching the comparison line beside it, so a plain `RUST_LOG=info` leaves
it off. propeller-heads/helm-configuration#997 turns it on for Ethereum dev and
staging.

Loki:

```logql
{namespace="dev-fynd"} |= "winning_protocols"
  | pattern "<_>protocols=<protocols>"
  | line_format "{{.protocols}}" | json
```

`comparison_log.rs` becomes `instrumentation.rs` and holds both router logs, with
the capturing test subscriber they share in `log_capture`.

https://claude.ai/code/session_016WAK3N3QW3TUZTQ2TeTJLt

### DIFF
--- a/fynd-core/src/worker_pool_router/mod.rs
+++ b/fynd-core/src/worker_pool_router/mod.rs
@@ -28,6 +28,7 @@ pub mod config;
 mod log_capture;
 
 use std::{
+    collections::BTreeMap,
     sync::LazyLock,
     time::{Duration, Instant},
 };
@@ -42,7 +43,7 @@ use num_bigint::BigUint;
 use num_traits::{CheckedSub, ToPrimitive};
 use rustc_hash::{FxHashMap, FxHashSet};
 use serde::{Deserialize, Serialize};
-use tracing::{debug, warn};
+use tracing::{debug, info, warn, Level};
 use tycho_execution::encoding::{
     evm::gas_estimator::estimate_gas_usage,
     models::{Solution, Strategy},
@@ -358,6 +359,7 @@ pub async fn encode_quotes(
 pub fn finalize_quote(order_quotes: Vec<OrderQuote>, solve_time_ms: u64) -> Quote {
     for quote in &order_quotes {
         record_win(quote);
+        log_winning_protocols(quote);
     }
 
     let total_gas_estimate = order_quotes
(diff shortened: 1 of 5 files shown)


monitoring/grafana/dashboards/fynd.json
```diff
@@ -310,6 +310,58 @@
         },
         "overrides": []
       }
+    },
+    {
```
tvinagre, line 314: pls add this to terraform as well to show up on grafana

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -602,8 +609,9 @@ impl WorkerPoolRouter {
                 .iter_mut()
                 .map(|quote| async move {
                     if quote.status() == QuoteStatus::Success {
-                        let result = simulator.simulate(quote).await;
-                        quote.set_simulation_result(result);
+                        let attempt = simulator.simulate_attempt(quote).await;
```
tvinagre, line 612: unrelated to this PR,  but was there a reason why u preferred to simulate individually async instead of a single multicall?

tvinagre (approved):
LGTM!

fynd-core/src/worker_pool_router/mod.rs
carloszanella: no specific reason

PR #505, opened 2026-09-02
fix(feed): tolerate delayed Robinhood Tycho updates

tvinagre:
## Summary

- expose an optional `TYCHO_LATENCY_BUFFER_SECS` / `--tycho-latency-buffer-secs` override through Fynd's CLI, RPC builder, core builder, and Tycho feed configuration
- use a 5-second latency buffer by default only for Robinhood, producing a 6-second effective window with Robinhood's 1-second block time
- preserve Tycho's native chain-specific defaults for every other chain unless an operator explicitly configures an override
- apply the configured builder consistently to normal, pending-block, and step-controlled feed modes

## Context

Intermittent Robinhood upstream delivery pauses of approximately 4.14–4.88 seconds exceeded Tycho's previous 3-second effective readiness window. The resulting empty feed message was decoded as `Missing block!`, causing Fynd's feed supervisor to exit and Kubernetes to restart the service.

This mitigates those finite pauses in Fynd without changing tycho-client's shared empty-`FeedMessage` interface. Genuine pauses beyond the configured window still follow the existing failure path.

## Validation

- [x] `cargo +nightly fmt --all -- --check`
- [x] targeted chain-selection, CLI, and core configuration tests
- [x] `cargo +nightly clippy --locked --all --all-features --all-targets -- -D warnings -A deprecated`
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --locked --package fynd-core --package fynd-rpc-types --package fynd-rpc --package fynd-client`
- [x] `cargo nextest run --workspace --locked --all-targets --all-features --bin fynd -E 'not binary(algorithm_bench) and not binary(profile)'` — 1,731 passed, 10 skipped
- [x] independent Claude/Fable 5.1 pre-commit review — APPROVE, no blockers

`./check.sh` currently stops on the untouched `Atomic::fetch_update` deprecation in `fynd-core/src/encoding/exclusive_swap.rs`. The identical failure was reproduced from clean `origin/main`; clippy passes with only that confirmed baseline deprecation allowed.


tvinagre:
Superseded by propeller-heads/tycho#1383, which adjusts the Robinhood timeout at the shared chain-default source.

PR #507, opened 2026-09-03
feat!: Improvements/fixes to simulation metrics

carloszanella:
5 changes TLDR:
- refactor to simplify interfaces
- fix to compare simulated amount versus amount less fees
- add algorithm label to these metrics so we can break down this if needed
- tweak to histogram bins
- move the new metric log closer to the other one and away from core code


fynd-core/src/simulation/simulator.rs
```diff
@@ -297,23 +297,37 @@ impl SimulationAttempt {
...
+    let pool = quote.worker_pool().to_string();
+    let algorithm = quote.algorithm().to_string();
     let outcome = match attempt {
         SimulationAttempt::Success { amount_out, .. } => {
             if let Some(deviation) = deviation_bps(quote, amount_out) {
-                histogram!("quote_simulation_deviation_bps").record(deviation);
+                histogram!(
+                    "quote_simulation_deviation_bps",
+                    "pool" => pool.clone(),
+                    "algorithm" => algorithm.clone()
```
kayibal, line 315: It would be good to also somehow know the protocols used here. Not sure if this is the right place but sth like if bps deviation if one ekubo pool was in the route.

kayibal (approved):

fynd-core/src/simulation/simulator.rs
carloszanella: Good call — done, in d55eca88.

Not in the metric, though: a `protocol` label would record one route once per protocol it crosses, so a 3-hop route would enter the histogram three times and the series count would grow by the size of the protocol set. It also still wouldn't tell you which combination a given deviation came from.

Instead the deviation rides on the `fynd::winning_protocols` line, which already names the route's protocols, one line per returned quote:

```
winning_protocols order_id=abc block=25895760 pool=aperture_d3 algorithm=aperture \
  swaps=3 simulated=success deviation_bps=-3.2140 protocols={"ekubo_v3":1,"uniswap_v3":2}
```

`simulated=` is `success`/`failure`, empty when the request asked for no simulation; `deviation_bps=` is empty unless the call returned an amount. Both keys stay on every line so the shape is constant.

Your Ekubo question in Loki:

```logql
{namespace="dev-fynd"} |= "winning_protocols" |= "ekubo"
  | pattern "<_>deviation_bps=<bps> protocols=<protocols>"
  | unwrap bps
```

The deviation formula is shared with the metric rather than derived twice.

PR #509, opened 2026-09-03
feat(ts-client)!: support client fee signing

louise-poole:
The TypeScript client couldn't finish the client-fee flow. The fee signature has to cover the quoted swap, but the mapper dropped the two fields needed to build it — so there was no way to sign after quoting, and the published guide showed a flow that didn't work.

  It's one quote per swap now: quote with unsigned fee params, compute the hash, sign it, patch the signature into the returned calldata with the new patchClientFeeSignature.

  The guide's TypeScript tab is rewritten against a runnable example, so check-doc-snippets.sh keeps it honest. Also drops the orphaned clients/typescript/autogen/ package, which nothing built and four docs still pointed at.


clients/typescript/client/src/viem.ts
```diff
@@ -23,6 +23,7 @@ export interface ViemPublicClient {
   }>;
   call(
     args: {
+      account: Address;
       to: Address;
       data: Hex;
       value: bigint;
```
carloszanella, line 29: now sure how this client is versioned but this is a breaking change and the PR doesnt state so

---

**[BIZ `#2`]**

**What changed**: `viemProvider.call` now passes `account: sender` to viem. This is not limited to
the client fee flow: it changes `swapPayload({ simulate: true })`, `executeSwap` dry runs, and the
`eth_call` replay used to fetch a revert reason. Those calls previously ran from the zero address,
so any route pulling funds with `transferFrom(msg.sender, ...)` reverted. Some callers were getting
a simulation failure that the swap itself would not have produced.

**Where**: `clients/typescript/client/src/viem.ts:24–34, 95–108`; call sites in
`clients/typescript/client/src/client.ts:353–357, 478–481, 619–621`

- The `account` field is now required on `ViemPublicClient.call`. Check whether any consumer passes
  a hand-written object for that interface rather than a viem `PublicClient`.
- `getRevertReason` replays the failed transaction through this same `call`. Confirm the reason text
  it reports for a real revert is still the one you expect now that the caller is the sender.

clients/typescript/client/src/client-fee.ts
```diff
@@ -119,19 +128,68 @@ export function clientFeeSigningHash(
...
+  assertSignatureLength(signature, 'Client fee');
+  const calldata = toBytes(tx.data, 'Quote calldata');
+  if (offset < 0 || offset + SIGNATURE_BYTES > calldata.length) {
+    throw FyndError.config(
+      `Client fee signature at offset ${String(offset)} does not fit ${String(calldata.length)}-byte calldata`
+    );
+  }
+  calldata.set(toBytes(signature, 'Client fee signature'), offset);
+  return { ...quote, transaction: { ...tx, data: bytesToHex(calldata) } };
+}
```
carloszanella, line 196: sounds like good validation for safety

---

**[IMPL `#1`]**

**Category**: bug

**Description**: In `clients/typescript/client/src/client-fee.ts:175` the only check on `clientFeeSignatureOffset` is that 65 bytes fit inside the calldata. Nothing checks that the 65 bytes at that offset are the placeholder the server wrote. If the server and the client ever disagree about the offset — an older server, a router encoding change — the patch overwrites 65 bytes of an adjacent ABI field and returns a quote that still looks valid. The doc comment on line 148 says the function overwrites "the placeholder", which is stronger than what the code verifies.

**Plain English**: An integrator asks for a quote, signs the fee hash, patches the signature in, and gets back what looks like a normal transaction. If the reported position of the signature slot is wrong, the patch quietly writes over a different part of the transaction instead — the amounts, the receiver, or the route. Nothing complains at that point. The transaction is sent, the chain rejects it or, worse, executes with corrupted values, and the integrator pays gas for a failure with no hint of where it came from. The client should notice that the bytes it is about to replace are not the placeholder and refuse to patch.

**Suggestion**: Before writing, check that the 65 bytes at `offset` equal the placeholder — all zeros when `params.signature` was left unset, which is the documented flow. Throw a `FyndError.config` naming the offset when they do not.

clients/typescript/client/src/viem.ts
```diff
@@ -92,7 +93,10 @@ export function viemProvider(
       };
     },
     async call(tx) {
+      // Without `account` the node runs the call from the zero address, and any router that
+      // pulls funds with `transferFrom(msg.sender, ...)` reverts.
       const result = await client.call({
+        account: sender,
         to: tx.to,
         data: tx.data,
         value: tx.value,
```
carloszanella, line 104: ok this looks like unwanted behaviour, we should use the provided `sender`

---

**[IMPL `#3`]**

**Category**: bug

**Description**: `clients/typescript/client/src/viem.ts:99` now sets `account: sender` on `call`, where `sender` is fixed when `viemProvider` is built. But `clients/typescript/client/src/client.ts:306` resolves the sender per call as `hints.sender ?? this.options.sender`. When a caller passes `SigningHints.sender`, the transaction is built for one address while the simulation and the gas estimate run from another. Before this change `call` ran from the zero address for everyone, so the mismatch is new for `call` and pre-existing for `estimateGas` (line 111).

**Plain English**: An integrator who serves several wallets from one client asks to quote and simulate a swap on behalf of a second wallet. The simulation runs against the first wallet's balances and approvals instead. It can pass when the real wallet would revert, or fail when the real wallet would succeed. Either way the answer describes the wrong account, and nothing says so. The simulation should run from the same address the transaction will be sent from.

**Suggestion**: Carry the sender on the transaction the provider receives, or take it as an argument on `EthProvider.call` and `EthProvider.estimateGas`, so `client.ts` can pass the sender it resolved.

clients/typescript/examples/swap-client-fee/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@fynd/swap-client-fee",
+  "version": "0.0.0",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "start": "tsx main.ts",
+    "typecheck": "tsc --noEmit"
```
carloszanella, line 8: maybe should run this on CI but since its an example not big deal

---

**[IMPL `#6`]**

CI runs typecheck with `--filter @kayibal/fynd-client` only, and `run-all-examples.sh` executes the example through `tsx`, which strips types without checking them. Add the examples to the typecheck job so a type error in the snippet the guide publishes fails the build.

clients/typescript/client/src/mapping.ts
```diff
@@ -159,7 +159,9 @@ function toWireClientFeeParams(p: ClientFeeParams): WireClientFeeParams {
         receiver: p.receiver,
         max_contribution: p.maxContribution.toString(),
         deadline: p.deadline,
-        signature: p.signature ?? '',
+        // The wire field is required; empty bytes tell the server to encode a placeholder
+        // the client patches after signing.
+        signature: p.signature ?? '0x',
     };
```
carloszanella, line 180: yeah seems like we dont need this optional parameter before the quote

---

**[SIMPLIFY `#1`]**

**Description**: The PR establishes that the client fee signature covers the quoted swap, so it can only exist after the quote comes back — and by then the way to apply it is `patchClientFeeSignature`. That leaves nothing that can legitimately set `ClientFeeParams.signature` before a quote. `clients/typescript/client/src/types.ts:65–66` says "Set `signature` only if you already hold a signature valid for this exact swap", but the same paragraph explains why a caller cannot hold one. Three pieces of code exist only for that path: the optional field at `clients/typescript/client/src/types.ts:78`, the conditional check in `withClientFee` at `clients/typescript/client/src/client-fee.ts:141–143`, and the `?? '0x'` fallback in `toWireClientFeeParams` at `clients/typescript/client/src/mapping.ts:164`. Neither the example nor the guide uses it, and `docs/guides/client-fees.md:71-73` now tells the reader not to re-quote after signing, which closes the last way to hold one. Maps to the `yagni` directive.

**Suggestion**: Drop `signature` from `ClientFeeParams`, drop the branch in `withClientFee`, and send a plain `'0x'` in `toWireClientFeeParams`. That removes one optional field, one branch and one test from the surface. This is a public API break: it takes an exported field away and makes `withClientFee` a pure spread, so leave the call to the human. Nothing on the wire changes, since the only value the field can now carry is the one the fallback already sends.

clients/typescript/client/src/signing.ts
```diff
@@ -72,6 +73,23 @@ export interface SettleOptions {
...
+ * `label` names the signature in the error message, e.g. `'Permit2'`.
+ */
+export function assertSignatureLength(signature: Hex, label: string): void {
+  const hexChars = SIGNATURE_BYTES * 2 + 2;
+  if (signature.length !== hexChars) {
+    throw FyndError.config(
+      `${label} signature must be exactly ${String(SIGNATURE_BYTES)} bytes (${String(hexChars)} hex chars), got ${String(signature.length)} chars`
+    );
+  }
+}
```
carloszanella: I like things in the right places

---

**[PERSONAL `#1`]**

`SIGNATURE_BYTES` and `assertSignatureLength` sit between `DEFAULT_SETTLE_TIMEOUT_MS` and `ExecutionReceipt`, which belong together. Move the two new declarations above `SettleOptions`, next to the other signature helpers. Maps to the `placement` directive.

clients/typescript/client/src/types.ts
```diff
@@ -152,6 +162,13 @@ export interface FeeBreakdown {
   maxSlippage: bigint;
   /** Minimum amount the user receives on-chain (the min_amount_out in the tx). */
   minAmountReceived: bigint;
+  /**
+   * keccak256 of the ABI-encoded swap bytes.
+   *
+   * Present only when `clientFeeParams` was set on the quote request. Feed it to
+   * `clientFeeSigningHash` as `ClientFeeSwapContext.swapsHash`.
+   */
+  swapsHash?: Hex;
```
carloszanella, line 185: **[DOCS `#3`]**

**Description**: The new doc on `FeeBreakdown.swapsHash` says "keccak256 of the ABI-encoded swap bytes". The server computes `keccak256(swaps)` over the packed swaps blob (`fynd-core/src/encoding/encoder.rs:567`), which is not ABI encoding. The same diff describes the same value correctly twice elsewhere: `clients/typescript/client/src/client-fee.ts:53` says "keccak256 of the encoded swaps bytes", and `clients/rust/src/types.rs:171` uses that same wording.

**Suggestion**: Use the established wording — "keccak256 of the encoded swaps bytes" — in `clients/typescript/client/src/types.ts:166`, so all three docs say one thing.

clients/typescript/client/src/client-fee.ts
```diff
@@ -119,19 +128,68 @@ export function clientFeeSigningHash(
...
+
+/**
+ * Converts a hex string to bytes, rejecting anything `hexToBytes` would mangle.
+ *
+ * viem throws its own error type on invalid characters and silently left-pads an odd number
+ * of digits, which would shift every byte of the calldata by a nibble.
+ */
+function toBytes(value: Hex, label: string): Uint8Array {
```
carloszanella: yeah the name doesnt say what it does tbh. although I dont like the suggestion

---

**[DOCS `#4`]**

The function's job is to reject hex that `hexToBytes` would mangle; the name says only "convert". `toBytes` is also the name viem gives its own conversion helper, and this module imports from viem. Rename it `hexToBytesStrict`. Maps to the `naming` directive.

carloszanella (commented):
Review of the client-fee signing flow. Accepted findings below — 8 inline, 4 that have no addressable line in the diff collected here.

---

**[IMPL `#5`]**

**Description**: The diff deletes `clients/typescript/autogen/` and repoints `clients/README.md`, `scripts/update-openapi.sh` and the run-ci skill, but `fynd-rpc-types/CLAUDE.md:46` still says `./scripts/update-openapi.sh` regenerates `autogen/src/schema.d.ts`. That directory no longer exists, so the instruction sends the next reader to a path that is gone. Maps to the `docs` directive.

**Suggestion**: Change the path to `clients/typescript/client/src/schema.d.ts`, which is what the script and the CI drift check now write.

---

**[IMPL `#8`]**

`keccak256(stringToHex('TychoRouter'))` and `keccak256(stringToHex('1'))` are constants recomputed on each `clientFeeSigningHash` call. Lift them to module constants beside `EIP712_DOMAIN_TYPEHASH`.

---

**[DOCS `#1`]**

**Description**: The checklist in `.claude/knowledge/typescript.md:43–46` lists two steps: create `main.ts`, add the name to `TS_EXAMPLES`. Adding `swap-client-fee` in this PR also required a `package.json`, a `tsconfig.json`, an entry in `clients/typescript/pnpm-workspace.yaml`, and — because the guide embeds the snippet — an entry in `scripts/check-doc-snippets.sh`. Someone who follows the checklist gets a package pnpm never installs.

**Suggestion**: Extend the checklist to name all five steps, using `clients/typescript/examples/swap-client-fee/` as the reference layout.

---

**[DOCS `#2`]**

Maps to the `docs` directive. The comment on `autogen.ts` still reads "(replaces @fynd/autogen)", but this diff deletes `clients/typescript/autogen/` and the package no longer exists anywhere in the repo. Drop the parenthetical.

carloszanella (commented):
I accepted all AI SLOP items. you can just tell Claude to address them all. they are just better passes at writing stuff using codex trying to remove Claude brain rot

clients/typescript/client/src/client-fee.ts
```diff
@@ -119,19 +128,68 @@ export function clientFeeSigningHash(
...
+ *
+ * The server encodes the transaction with a zeroed 65-byte placeholder and reports its byte
+ * offset as `transaction.clientFeeSignatureOffset`, so one quote request is enough: sign the
+ * hash from `clientFeeSigningHash`, patch it in here, and submit the transaction. The Rust
+ * client calls this same step `Quote::with_client_fee_signature`.
+ *
+ * Returns a new quote; the input is left untouched. Throws when the quote carries no encoded
+ * transaction (set `encodingOptions` on the request), when it carries no signature offset
+ * (set `clientFeeParams` too), when `signature` is not 65 bytes, when either the signature or
+ * the calldata is not valid hex, or when the offset does not fit the calldata.
```
carloszanella: **[SLOP `#1`]**

**Before**: Returns a new quote; the input is left untouched. Throws when the quote carries no encoded
 * transaction (set `encodingOptions` on the request), when it carries no signature offset
 * (set `clientFeeParams` too), when `signature` is not 65 bytes, when either the signature or
 * the calldata is not valid hex, or when the offset does not fit the calldata.
**After**: Returns a new quote and leaves the input untouched.
 *
 * Throws when:
 * - the quote carries no encoded transaction — set `encodingOptions` on the request;
 * - the quote carries no signature offset — set `clientFeeParams` on the request too;
 * - `signature` is not 65 bytes;
 * - the signature or the calldata is not valid hex;
 * - the offset does not fit the calldata.
**Why**: clarity — a 45-word sentence listing five cases needs a second read; the list keeps every fact.

clients/typescript/client/src/client-fee.ts
```diff
@@ -119,19 +128,68 @@ export function clientFeeSigningHash(
 
 /**
  * Attach client fee configuration to encoding options.
- * Validates that signature is present and exactly 65 bytes (130 hex chars + '0x' prefix).
+ *
+ * `params.signature` is optional: the signature covers the quoted swap, so the usual flow is
+ * to quote with unsigned params and patch the signature into the returned calldata with
+ * `patchClientFeeSignature`. When a signature is set it must be exactly 65 bytes
+ * (130 hex chars + '0x' prefix).
```
carloszanella: **[SLOP `#2`]**

**Before**: `params.signature` is optional: the signature covers the quoted swap, so the usual flow is
 * to quote with unsigned params and patch the signature into the returned calldata with
 * `patchClientFeeSignature`. When a signature is set it must be exactly 65 bytes
 * (130 hex chars + '0x' prefix).
**After**: `params.signature` is optional. Leave it unset, quote, then patch the signature into the
 * returned calldata with `patchClientFeeSignature`. A signature that is set must be exactly
 * 65 bytes (130 hex chars + '0x' prefix).
**Why**: simplicity — "the usual flow is to" and the passive "when a signature is set" both drop out without losing a fact.

clients/typescript/client/src/types.ts
```diff
@@ -59,8 +59,11 @@ export interface PermitSingle {
 /** Client fee configuration for the Tycho Router.
  *
  * When provided, the router charges a fee in basis points on the swap output.
- * The `signature` must be an EIP-712 signature by the `receiver` over the
- * `ClientFee` typed data — compute the hash with `clientFeeSigningHash`.
+ *
+ * The signature covers the quoted swap, so it can only be produced after the quote comes
+ * back. Send these params unsigned, then patch the signature into the returned calldata with
+ * `patchClientFeeSignature` — no second quote request needed. Set `signature` only if you
+ * already hold a signature valid for this exact swap.
```
carloszanella: **[SLOP `#3`]**

**Description**: "The signature covers the quoted swap, so quote unsigned and patch the signature into the calldata" is written out six times, each in different words: the `ClientFeeParams` doc (`clients/typescript/client/src/types.ts:63–66`), the `Transaction.clientFeeSignatureOffset` doc (`clients/typescript/client/src/types.ts:102–108`), the `FeeBreakdown.swapsHash` doc (`clients/typescript/client/src/types.ts:165–171`), the `withClientFee` doc (`clients/typescript/client/src/client-fee.ts:132–135`), the `patchClientFeeSignature` doc (`clients/typescript/client/src/client-fee.ts:150–153`), and `docs/guides/client-fees.md`. Six wordings of one rule drift apart on the next change, and a reader cannot tell whether they say the same thing. Maps to the `docs` directive.

**Suggestion**: Keep the full flow in the guide and on `patchClientFeeSignature`, which owns the step. Cut the three `types.ts` field docs to what each field is, and let them point at `patchClientFeeSignature`.

carloszanella: **[SLOP `#4`]**

**Before**: The signature covers the quoted swap, so it can only be produced after the quote comes
 * back. Send these params unsigned, then patch the signature into the returned calldata with
 * `patchClientFeeSignature` — no second quote request needed. Set `signature` only if you
 * already hold a signature valid for this exact swap.
**After**: The signature covers the quoted swap, so you can only produce it after the quote comes
 * back. Send these params unsigned and patch the signature into the returned calldata with
 * `patchClientFeeSignature`. Set `signature` only if you already hold one valid for this swap.
**Why**: simplicity — "no second quote request needed" repeats what "patch the returned calldata" already says.

clients/typescript/client/src/types.ts
```diff
@@ -152,6 +162,13 @@ export interface FeeBreakdown {
   maxSlippage: bigint;
   /** Minimum amount the user receives on-chain (the min_amount_out in the tx). */
   minAmountReceived: bigint;
+  /**
+   * keccak256 of the ABI-encoded swap bytes.
+   *
+   * Present only when `clientFeeParams` was set on the quote request. Feed it to
+   * `clientFeeSigningHash` as `ClientFeeSwapContext.swapsHash`.
```
carloszanella: **[SLOP `#5`]**

**Before**: Present only when `clientFeeParams` was set on the quote request. Feed it to
   * `clientFeeSigningHash` as `ClientFeeSwapContext.swapsHash`.
**After**: Present only when `clientFeeParams` was set on the quote request. Pass it to
   * `clientFeeSigningHash` as `ClientFeeSwapContext.swapsHash`.
**Why**: consistency — the surrounding docs say "pass" (`clients/typescript/client/src/client-fee.ts:60`, `clients/typescript/client/src/types.ts:105`).

clients/typescript/client/src/types.ts
```diff
@@ -71,7 +74,7 @@ export interface ClientFeeParams {
   maxContribution: bigint;
   /** Unix timestamp after which the signature is invalid. */
   deadline: number;
-  /** 65-byte EIP-712 ECDSA signature by `receiver`. Set after signing. */
+  /** 65-byte EIP-712 ECDSA signature by `receiver`. Usually left unset. */
```
carloszanella: **[SLOP `#6`]**

**Before**: /** 65-byte EIP-712 ECDSA signature by `receiver`. Usually left unset. */
**After**: /** 65-byte EIP-712 ECDSA signature by `receiver`. Set it only for a signature you already hold. */
**Why**: concreteness — "usually" leaves the reader guessing which case is the exception.

clients/typescript/client/src/client-fee.ts
```diff
@@ -119,19 +128,68 @@ export function clientFeeSigningHash(
...
+    throw FyndError.config(
+      `Client fee signature at offset ${String(offset)} does not fit ${String(calldata.length)}-byte calldata`
+    );
+  }
+  calldata.set(toBytes(signature, 'Client fee signature'), offset);
+  return { ...quote, transaction: { ...tx, data: bytesToHex(calldata) } };
+}
+
+/**
+ * Converts a hex string to bytes, rejecting anything `hexToBytes` would mangle.
```
carloszanella: **[SLOP `#7`]**

**Before**: Converts a hex string to bytes, rejecting anything `hexToBytes` would mangle.
**After**: Converts a hex string to bytes, rejecting input `hexToBytes` would misread.
**Why**: clarity — "mangle" is slang, and the next paragraph already names the two failures.

clients/typescript/client/src/client-fee.ts
```diff
@@ -119,19 +128,68 @@ export function clientFeeSigningHash(
...
+ *
+ * viem throws its own error type on invalid characters and silently left-pads an odd number
+ * of digits, which would shift every byte of the calldata by a nibble.
+ */
+function toBytes(value: Hex, label: string): Uint8Array {
+  if (!isHex(value, { strict: true }) || value.length % 2 !== 0) {
+    throw FyndError.config(`${label} is not valid hex: ${value.slice(0, 12)}`);
```
carloszanella: **[SLOP `#8`]**

`` `${label} is not valid hex: ${value.slice(0, 12)}` `` prints the first 12 characters of a long calldata string with nothing marking the cut, so the reader can read it as the whole value. Append an ellipsis.

clients/typescript/client/src/mapping.ts
```diff
@@ -159,7 +159,9 @@ function toWireClientFeeParams(p: ClientFeeParams): WireClientFeeParams {
         receiver: p.receiver,
         max_contribution: p.maxContribution.toString(),
         deadline: p.deadline,
-        signature: p.signature ?? '',
+        // The wire field is required; empty bytes tell the server to encode a placeholder
+        // the client patches after signing.
```
carloszanella: **[SLOP `#9`]**

**Before**: // The wire field is required; empty bytes tell the server to encode a placeholder
        // the client patches after signing.
**After**: // The wire field is required. Empty bytes tell the server to encode a placeholder that
        // the client patches after signing.
**Why**: clarity — the missing "that" reads as a second independent clause on first pass.

clients/typescript/client/src/viem.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    estimateGas:           vi.fn().mockResolvedValue(150000n),
+    sendRawTransaction:    vi.fn().mockResolvedValue('0xtxhash' as Hex),
+    getTransactionReceipt: vi.fn().mockResolvedValue(null),
+    readContract:          vi.fn().mockResolvedValue(0n),
+  } as unknown as ViemPublicClient & { call: ReturnType<typeof vi.fn>; estimateGas: ReturnType<typeof vi.fn> };
+}
+
+describe('viemProvider', () => {
+  // A call without `account` runs from the zero address, so routers that pull funds with
+  // `transferFrom(msg.sender, ...)` revert during simulation.
```
carloszanella: **[SLOP `#10`]**

`clients/typescript/client/src/viem.ts:96–97` already explains why `call` passes `account`, and the test repeats the same fact in different words. Delete the copy in the test; the test name already says what it checks.

clients/typescript/client/src/client-fee.test.ts
```diff
@@ -132,9 +132,10 @@ describe('withClientFee', () => {
     expect(result.transferType).toBe('transfer_from');
   });
 
-  it('throws when signature is missing', () => {
+  it('accepts params without a signature', () => {
```
carloszanella: **[SLOP `#11`]**

**Before**: it('accepts params without a signature', () => {
**After**: it('leaves the signature unset when params carry none', () => {
**Why**: the project rule — name the test after what is tested, not the expected outcome (`.claude/knowledge/rust.md:53`).

clients/typescript/client/src/signing.test.ts
```diff
@@ -179,3 +185,23 @@ describe('approvalSigningHash', () => {
     expect(approvalSigningHash(baseApprovalPayload)).not.toBe(approvalSigningHash(other));
   });
 });
+
+describe('assertSignatureLength', () => {
+  const VALID = `0x${'ab'.repeat(65)}` as Hex;
+
+  it('accepts a 65-byte signature', () => {
```
carloszanella: **[SLOP `#12`]**

**Before**: it('accepts a 65-byte signature', () => {
**After**: it('checks a 65-byte signature', () => {
**Why**: the project rule — name the test after what is tested, not the expected outcome (`.claude/knowledge/rust.md:53`).

clients/typescript/client/src/mapping.test.ts
```diff
@@ -382,6 +401,60 @@ describe('fromWireQuote', () => {
...
+          },
+        },
+      ],
+    };
+    const quote = fromWireQuote(wire, TOKEN_OUT, SENDER);
+    expect(quote.transaction?.clientFeeSignatureOffset).toBe(68);
+    expect(quote.feeBreakdown?.swapsHash).toBe(`0x${'11'.repeat(32)}`);
+  });
+
+  it('client fee signing fields are undefined without a client fee', () => {
```
carloszanella: **[SLOP `#13`]**

**Before**: it('client fee signing fields are undefined without a client fee', () => {
**After**: it('maps client fee signing fields when the wire values are null', () => {
**Why**: the project rule — "are undefined" is the assertion, not the case under test (`.claude/knowledge/rust.md:53`).

clients/typescript/client/src/client-fee.test.ts
```diff
@@ -157,3 +158,80 @@ describe('withClientFee', () => {
...
+    expect(patched.amountOut).toBe(2000n);
+  });
+
+  it('patches at offset 0', () => {
+    const patched = patchClientFeeSignature(quoteWithOffset(0), SIG);
+    expect(patched.transaction?.data).toBe(`0x${SIG.slice(2)}0000${SUFFIX}`);
+  });
+
+  it('patches at the last offset that fits', () => {
+    // Calldata holds 69 bytes, so a 65-byte signature starting at byte 4 ends exactly at the end.
```
carloszanella: **[SLOP `#14`]**

**Before**: // Calldata holds 69 bytes, so a 65-byte signature starting at byte 4 ends exactly at the end.
**After**: // The calldata holds 69 bytes, so a 65-byte signature at byte 4 fills it to the last byte.
**Why**: clarity — "ends exactly at the end" repeats itself, and the missing article reads as clipped.

clients/typescript/examples/swap-client-fee/main.ts
```diff
@@ -0,0 +1,134 @@
...
+// known once the server has encoded the transaction. The server encodes a zeroed 65-byte
+// signature placeholder and reports its byte offset, so one quote request is enough:
+//
+//   1. Quote with unsigned client fee params.
+//   2. Sign the 11-field EIP-712 hash using `swapsHash` from the response.
+//   3. Patch the signature into the calldata with `patchClientFeeSignature`.
+//   4. Execute.
+//
+// Two keys are used: the dev key as the sender, and a random ephemeral key as the fee
+// receiver (in production this is the integrator's key). The fee receiver needs no funds —
```
carloszanella: **[SLOP `#15`]**

**Before**: // Two keys are used: the dev key as the sender, and a random ephemeral key as the fee
// receiver (in production this is the integrator's key).
**After**: // The example uses two keys: the dev key as the sender, and a throwaway key generated per
// run as the fee receiver (in production this is the integrator's key).
**Why**: active voice, and "random ephemeral" is two words for one idea.

clients/typescript/examples/swap-client-fee/main.ts
```diff
@@ -0,0 +1,134 @@
...
+const USDC = '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48';
+const SELL_AMOUNT = parseUnits('1', 18); // 1 WETH
+const SLIPPAGE = 0.01; // 1%
+const FEE_BPS = 50; // 0.5% client fee
+
+const fyndUrl = process.env['FYND_URL'] ?? 'http://localhost:3000';
+const rpcUrl = process.env['RPC_URL'] ?? 'http://localhost:8545';
+const account = privateKeyToAccount(process.env['PRIVATE_KEY'] as `0x${string}`);
+
+// Separate fee receiver key — in production this is the integrator's key.
```
carloszanella: **[SLOP `#16`]**

The header comment (lines 13–14) already says the fee receiver is a separate key and that in production it is the integrator's. Delete the inline repeat at line 49.

clients/typescript/examples/swap-client-fee/main.ts
```diff
@@ -0,0 +1,134 @@
...
+});
+
+const info = await client.info();
+if (info.routerAddress === null) {
+  throw new Error('example requires a chain with encoding support');
+}
+const routerAddress = info.routerAddress;
+const chainId = info.chainId;
+
+// Approve the router to spend WETH if the current allowance is insufficient.
```
carloszanella: **[SLOP `#17`]**

**Before**: // Approve the router to spend WETH if the current allowance is insufficient.
**After**: // Approve the router to spend WETH when the allowance is too small.
**Why**: simplicity — `checkAllowance: true` on the next line already says "current"; "insufficient" is "too small".

docs/guides/client-fees.md
```diff
@@ -109,30 +117,52 @@ client library helpers sign the scaled value rather than the raw bps.
 {% tabs %}
 {% tab title="TypeScript" %}
 ```typescript
-// Build fee params (without signature).
+// Step 1: request a quote using unsigned client fee params. The server encodes the full
+// calldata with a 65-byte signature placeholder and returns `swapsHash` in the fee breakdown
+// plus `clientFeeSignatureOffset` in the transaction, so the client can patch the real
+// signature in.
```
carloszanella: **[SLOP `#18`]**

**Before**: // plus `clientFeeSignatureOffset` in the transaction, so the client can patch the real
// signature in.
**After**: // plus `clientFeeSignatureOffset` in the transaction. Patch the real signature in at that
// offset.
**Why**: clarity — the trailing "patch ... in" splits the verb across a line break and names no place; the doc snippet carries the same text.

clients/typescript/examples/swap-client-fee/main.ts
```diff
@@ -0,0 +1,134 @@
...
+
+// Step 3: patch the real signature into the calldata — no second quote request.
+const signed = patchClientFeeSignature(quote, signature);
+// [doc:end client-fee-typescript]
+
+console.log(`amount_in:  ${signed.amountIn}`);
+console.log(`amount_out: ${signed.amountOut}`);
+console.log(`client_fee: ${feeBreakdown.clientFee}`);
+
+// Sign and execute the swap.
```
carloszanella: **[SLOP `#19`]**

The lines call `swapPayload`, `sign` and `executeSwap`. Delete the comment.

docs/guides/client-fees.md
```diff
@@ -60,9 +60,17 @@ min_amount_received  = 994,990 - 9,949                   = 985,041
 ## Setting up client fees
 
 1. Set a fee in basis points (e.g. `50` = 0.5%), a receiver address, and a `maxClientContribution`.
-2. Have the fee receiver sign an EIP-712 `ClientFee` message authorizing these parameters.
-3. Attach the signed params to `EncodingOptions.clientFeeParams`.
-4. The router verifies the signature on-chain and deducts the fee. Fees go to the receiver's vault balance.
+2. Request a quote with those params and **no signature**, via `EncodingOptions.clientFeeParams`. The
```
carloszanella: **[SLOP `#20`]**

**Before**: 2. Request a quote with those params and **no signature**, via `EncodingOptions.clientFeeParams`. The
**After**: 2. Put those params in `EncodingOptions.clientFeeParams` with **no signature** and request a quote. The
**Why**: clarity — the trailing "via ..." makes the reader hold the instruction until the end of the sentence.

docs/guides/client-fees.md
```diff
@@ -60,9 +60,17 @@ min_amount_received  = 994,990 - 9,949                   = 985,041
...
+   response holds the encoded transaction with a zeroed 65-byte signature placeholder, plus
+   `fee_breakdown.swaps_hash` and `transaction.client_fee_signature_offset`.
+3. Have the fee receiver sign the EIP-712 `ClientFee` message, which binds the fee params to the
+   quoted swap.
+4. Patch the signature into the calldata at `client_fee_signature_offset` and submit the transaction.
+5. The router verifies the signature on-chain and deducts the fee. Fees go to the receiver's vault balance.
+
+One quote request per swap is enough — the signature is patched into the calldata you already
+have. Do not re-quote after signing: the signature covers `expectedAmountOut` and the quoted
+swaps, so a fresh quote invalidates it.
```
carloszanella: **[SLOP `#21`]**

**Before**: One quote request per swap is enough — the signature is patched into the calldata you already
have. Do not re-quote after signing: the signature covers `expectedAmountOut` and the quoted
swaps, so a fresh quote invalidates it.
**After**: Do not re-quote after signing: the signature covers `expectedAmountOut` and the quoted
swaps, so a fresh quote invalidates it.
**Why**: simplicity — steps 2 to 4 already say one quote is enough; only the warning is new.

clients/typescript/client/src/viem.ts
louise-poole: Addressed in edd33abb, together with IMPL #3.

On "breaking": adding `account` to the *argument* type of `ViemPublicClient.call` isn't breaking for implementors — parameter contravariance keeps a hand-written object that omits it assignable. What genuinely changed is behaviour: simulations, `executeSwap` dry runs and the `eth_call` revert replay now run from the sender rather than the zero address.

You're right that the PR doesn't state it though, and it now carries three deliberate API breaks: `viemProvider` loses its `sender` argument, `Eip1559Transaction` gains a required `from`, and `ClientFeeParams.signature` is gone. Updating the description to call those out.

clients/typescript/client/src/client-fee.ts
louise-poole: Done in 3aaa98cd, and it got cheaper than expected: with `ClientFeeParams.signature` gone (SIMPLIFY #1) the placeholder is always 65 zero bytes, so the check is just "are these bytes zero".

Also ported it to the Rust client, where `Quote::with_client_fee_signature` was in worse shape than the TS one: no length check at all, and `tx.data[offset..offset + signature.len()]` **panicked** on a bad offset rather than returning an error — in a public API, against the panic-prevention lints. It now returns `FyndError::Protocol` for wrong length, out-of-bounds offset and a non-zero placeholder.

It had no test coverage whatsoever, so I added five. Each asserts its specific error message rather than just `is_err()`, so they pin distinct branches — worth doing, because my first cut of the out-of-bounds test used an offset that was actually *in* bounds and was being caught by the placeholder check instead.

clients/typescript/client/src/viem.ts
louise-poole: Agreed, fixed in edd33abb — though not quite the way you suggested.

Taking the sender as an argument on `EthProvider.call`/`estimateGas` doesn't actually work: `getRevertReason` (`client.ts:478`) and `dryRunExecute` (`client.ts:619`) only ever receive an `Eip1559Transaction`, so they'd have to fall back to `options.sender` and the bug would survive on the dry-run path.

So it's your first option — the sender rides on the transaction. `Eip1559Transaction` gains a required `from`, set from the resolved sender in both `fyndSwapPayload` and `approval`. It travels inside `FyndPayload` → `SignedSwap`, so all five call sites get the right address without changing any method signature. `serializeTransaction` lists its fields explicitly in both places, so the signed bytes are unchanged.

That left `viemProvider`'s captured sender as a redundant second source of truth — exactly what caused this — so it's gone: `viemProvider(publicClient)`.

clients/typescript/examples/swap-client-fee/package.json
louise-poole: Not done, and it turns out it isn't the small change it looks like — trying it surfaced a real bug.

Neither example typechecks today, and both failures are pre-existing on `main`:

1. `@types/node` is missing from both examples' devDependencies, so `process` is unresolved.
2. A real viem `PublicClient` is **not** assignable to `ViemPublicClient`. The optional `request?<T>(args: { method: string; params?: readonly unknown[] })` is wider than viem's `EIP1193RequestFn<PublicRpcSchema>`, so `viemProvider(publicClient)` is a type error. The doc comment on `viem.ts` claims "Any viem `PublicClient` created via `createPublicClient` satisfies this interface" — it doesn't, and every integrator copying the guide hits this.

`tsx` strips types without checking them, which is why `run-all-examples.sh` never caught it.

Louise's call was to keep this out of scope here. Turning the CI job on needs both fixed first — happy to open a follow-up issue.

clients/typescript/client/src/mapping.ts
louise-poole: Agreed, done in 3aaa98cd — on **both** clients. The Rust client had the same `signature: Option<Bytes>` (`clients/rust/src/types.rs:137`) plus a `with_signature` builder, so dropping it on the TS side alone would have split the two surfaces.

TypeScript: field gone, `withClientFee` is a pure spread, `mapping.ts` sends a literal `'0x'`. Rust: field, `with_signature` and the `unwrap_or_default()` in `mapping.rs` all gone.

Nothing changes on the wire, as you say. It also unlocked IMPL #1 — the placeholder is now always zeroed, so `patchClientFeeSignature` can assert that before overwriting.

clients/typescript/client/src/signing.ts
louise-poole: Moved in 9aba5e3b — they now sit right after `PrimitiveSignature`, above `SignedSwap`.

clients/typescript/client/src/types.ts
louise-poole: Fixed in 9aba5e3b — now "keccak256 of the encoded swaps bytes", matching `client-fee.ts:53` and `clients/rust/src/types.rs:171`.

clients/typescript/client/src/client-fee.ts
louise-poole: Renamed in 3aaa98cd. Went with `parseHexBytes` rather than `hexToBytesStrict`, since you didn't like that one — "parse" already implies it can reject its input, and it doesn't read as a variant of viem's `toBytes`.

clients/typescript/client/src/client-fee.ts
louise-poole: Applied in 3aaa98cd — the list rode along with the signature removal, which rewrote the same doc block to add the placeholder case.

clients/typescript/client/src/client-fee.ts
louise-poole: Superseded in 3aaa98cd. With `ClientFeeParams.signature` dropped (SIMPLIFY #1) there's no optional signature left to describe, so the doc is now just: "Sign after quoting — see `patchClientFeeSignature`."

clients/typescript/client/src/types.ts
louise-poole: Applied in 9aba5e3b. Cut the three `types.ts` field docs to what each field is and pointed them at `patchClientFeeSignature`, which owns the step, leaving the full flow in the guide.

Where this overlapped with SLOP #4 I took this one, since it's the structural fix and #4's rewrite would have been cut down again anyway.

louise-poole: Folded into SLOP #3 in 9aba5e3b — that one cuts this doc down to the field description, so the longer rewrite didn't survive.

clients/typescript/client/src/types.ts
louise-poole: Applied in 9aba5e3b — "Pass it to `clientFeeSigningHash`".

clients/typescript/client/src/types.ts
louise-poole: Moot as of 3aaa98cd — the `signature` field is gone entirely (SIMPLIFY #1), so there's no doc left to sharpen.

clients/typescript/client/src/client-fee.ts
louise-poole: Applied in 3aaa98cd — "rejecting input `hexToBytes` would misread".

clients/typescript/client/src/client-fee.ts
louise-poole: Applied in 3aaa98cd — the truncated value now ends in an ellipsis.

clients/typescript/client/src/mapping.ts
louise-poole: Applied in 3aaa98cd, and the same comment now sits on the Rust side (`clients/rust/src/mapping.rs`), which sends the same empty bytes.

clients/typescript/client/src/viem.test.ts
louise-poole: Deleted in edd33abb.

clients/typescript/client/src/client-fee.test.ts
louise-poole: Moot as of 3aaa98cd — with `ClientFeeParams.signature` gone this test had nothing left to assert, so it's deleted rather than renamed.

clients/typescript/client/src/signing.test.ts
louise-poole: Renamed in 9aba5e3b.

clients/typescript/client/src/mapping.test.ts
louise-poole: Renamed in 9aba5e3b.

clients/typescript/client/src/client-fee.test.ts
louise-poole: Applied in 3aaa98cd. The surrounding fixtures changed too: with the placeholder check in place (IMPL #1), the offset-0 and offset-4 cases needed calldata actually zeroed at those offsets, rather than patching over the prefix and suffix.

clients/typescript/examples/swap-client-fee/main.ts
louise-poole: Applied in 9aba5e3b.

clients/typescript/examples/swap-client-fee/main.ts
louise-poole: Deleted in edd33abb.

clients/typescript/examples/swap-client-fee/main.ts
louise-poole: Applied in 9aba5e3b.

docs/guides/client-fees.md
louise-poole: Applied in 9aba5e3b, in both the example and the guide snippet.

clients/typescript/examples/swap-client-fee/main.ts
louise-poole: Deleted in 9aba5e3b.

docs/guides/client-fees.md
louise-poole: Applied in 9aba5e3b.

docs/guides/client-fees.md
louise-poole: Applied in 9aba5e3b — only the warning remains.

louise-poole:
Replied inline on all 29 threads. Three commits:

- 9aba5e3b — wording, naming, placement and stale-path fixes
- edd33abb — simulate from the sender the transaction will use
- 3aaa98cd — drop the pre-quote client fee signature

The four review-body items have no inline anchor, so they're here:

- **IMPL #5** — 9aba5e3b. `fynd-rpc-types/CLAUDE.md:46` now points at `clients/typescript/client/src/schema.d.ts`.
- **IMPL #8** — 9aba5e3b. Both domain hashes are module constants beside `EIP712_DOMAIN_TYPEHASH`.
- **DOCS #1** — 9aba5e3b. The example checklist now names all four steps (`package.json`/`tsconfig.json`, `pnpm-workspace.yaml`, `TS_EXAMPLES`, `check-doc-snippets.sh`), with `swap-client-fee/` as the reference layout.
- **DOCS #2** — 9aba5e3b. Parenthetical dropped.

Two things worth flagging beyond what you raised:

**The Rust client had the same bug class as IMPL #1, worse.** `Quote::with_client_fee_signature` had no length check and `tx.data[offset..offset + signature.len()]` **panicked** on a bad offset instead of erroring — public API, against the panic-prevention lints — with no test coverage at all. Fixed and tested in 3aaa98cd.

**IMPL #6 surfaced a pre-existing public-API bug.** A real viem `PublicClient` is not assignable to `ViemPublicClient`: the optional `request?<T>(args: { method: string; ... })` is wider than viem's `EIP1193RequestFn<PublicRpcSchema>`, so `viemProvider(publicClient)` is a type error on `main` today. `tsx` strips types without checking, which is why the examples never caught it. Left out of scope here; the CI typecheck job needs it fixed first.

Verification: `tsc` clean, oxlint 0 warnings, 188 TS tests; `cargo +nightly clippy -D warnings` clean, 111 `fynd-client` tests, `cargo check --workspace --all-targets` clean, rustdoc builds, `check-doc-snippets.sh` all OK. Each of the three commits typechecks standalone.

Still to do: the PR description needs updating for the zero-address→sender behaviour change and the three API breaks (`viemProvider` arity, `Eip1559Transaction.from`, `ClientFeeParams.signature`).

tamaralipows (approved):
Thank you! It seems everything was already caught by carlos.

PR #510, opened 2026-09-03
chore: reduce claude pre-push checks to format and relevant tests only

louise-poole:
Claude running the full CI before every commit/push was tedious and unnecessary. Now it only runs formatting checks and relevant tests only.


Troshchk (approved):
Niiiice, thanks!

PR #512, opened 2026-09-03
feat(core)!: Solver-side pAMM fallback venue selection

louise-poole:
### What this does

  A pAMM only fills in the block its maker quoted for. If the transaction lands late it reverts, also simulating it on a node fails. Tycho's `TychoFallbackRouter` handles this by running the leg on a fallback pool instead, but it needs the solver to say which pool.

  This PR does that. For each `fallback:` leg it picks the candidate pool with the best simulated output, prices the route through the chosen pools, and drops the route before ranking if that price misses `min_amount_out`. The pool goes to tycho as JSON on the swap's `user_data`.

### Breaking changes

  - Module `propamm_fallback` is now `fallback`
  - `RouteRejection::PammFallback*` are now `Fallback*`
  - Components arrive as `fallback:{venue}` instead of `propammfallback:{venue}`
  - Capture slugs change from `route/pamm_fallback_*` to `route/fallback_*`

  The third one is easy to miss. Anything excluding `propammfallback:`, either in `route_filter.exclude_protocols` or in a pool's` exclude_protocols`, has to be updated. An old name matches nothing and there's no warning.

### Removed

  The fee tier code the old `PropAMMRouter path` needed: `FeeTierFetcher`, `FeeTiers`, `SharedFeeTiers`, `RouteRejection::PammFeeTiersUnread`, `SolverBuildError::FeeTierFetcher`, and the `rpc_url` parameter on `Solver::from_recording`.

###  Before merging

  This needs tycho #1384 (router contracts) and #1434 (encoder). Neither is merged yet, and the tycho crates we pin don't emit `fallback:` components, so none of this runs until we bump them.

  One thing to check: the Uniswap encodings match tycho's own test fixtures, but I haven't verified Curve's `pool_type`/`i`/`j` or Fluid's `zero2one` against real components.

  🤖 Generated with Claude Code (https://claude.com/claude-code)


fynd-core/src/propamm_fallback/mod.rs
```diff
@@ -1,251 +1,180 @@
...
+/// `NoFallbackPool` when a pAMM leg's pair has no candidate in `index`; `NotPriceable` when no
+/// candidate could be simulated.
+pub fn select_fallbacks(
     route: &Route,
     market: &MarketDataView<'_>,
-    fee_tiers: &FeeTiers,
     index: &FallbackPoolIndex,
-) -> FallbackAmountOut {
-    let mut substituted = Vec::with_capacity(route.swaps().len());
+) -> Result<Vec<Option<FallbackLeg>>, FallbackSelectionError> {
```
tamaralipows: Why exactly is this an `Option`? I think that's avoidable

tamaralipows (commented):
Thank you, I like this solution! I'm relieved to see we can just adapt the existing fallback infra and don't need to start from scratch

fynd-core/src/fallback/user_data.rs
```diff
@@ -0,0 +1,350 @@
...
+/// Reads `component` into the variant its protocol system calls for.
+fn fallback_protocol(
+    component: &ProtocolComponent,
+    token_in: &Address,
+) -> Result<FallbackProtocol, FallbackError> {
+    let pool = pool_address(component)?;
+    match component.protocol_system.as_str() {
+        "uniswap_v2" => Ok(FallbackProtocol::UniswapV2 { pair: pool, fee_bps: UNISWAP_V2_FEE_BPS }),
+        "uniswap_v3" => Ok(FallbackProtocol::UniswapV3 { pool }),
+        UNISWAP_V4_SYSTEM => uniswap_v4(component),
```
tamaralipows: why is this one a const but the other ones are strings?

fynd-core/src/fallback/user_data.rs
```diff
@@ -0,0 +1,350 @@
...
+            dex: pool,
+            zero2one: component.tokens.first() == Some(token_in),
+        }),
+        other => Err(unencodable(component, format!("{other} is not a fallback protocol"))),
+    }
+}
+
+/// A Uniswap V4 pool identifies itself by its key, not by an address, so its fee, tick spacing and
+/// hook travel instead of a pool address.
+fn uniswap_v4(component: &ProtocolComponent) -> Result<FallbackProtocol, FallbackError> {
```
tamaralipows: can we name these fns more descriptively

fynd-core/src/fallback/user_data.rs
```diff
@@ -0,0 +1,350 @@
...
+    token_in: &Address,
+) -> Result<FallbackProtocol, FallbackError> {
+    let pool_type = u8::try_from(attribute_u32(component, CURVE_POOL_TYPE_ATTRIBUTE)?)
+        .map_err(|_| unencodable(component, "pool type does not fit a byte".to_string()))?;
+    let coins = curve_coins(component);
+    let token_out = component
+        .tokens
+        .iter()
+        .find(|token| *token != token_in)
+        .ok_or_else(|| unencodable(component, "names only one token".to_string()))?;
```
tamaralipows: What do we do for curve pools with three tokens?

fynd-core/src/fallback/user_data.rs
```diff
@@ -0,0 +1,350 @@
...
+
+/// The fee every canonical Uniswap V2 pool charges, in basis points.
+///
+/// `FALLBACK_PROTOCOL_SYSTEMS` admits `uniswap_v2` alone, not its forks, so every candidate pool
+/// charges this. It is also the most the router accepts — it reverts above 30 — so admitting a
+/// fork here would need the fork's own fee read off the component and checked against that cap.
+const UNISWAP_V2_FEE_BPS: u8 = 30;
+
+/// Curve's coin list, in the pool's own index order.
+const CURVE_COINS_ATTRIBUTE: &str = "coins";
```
tamaralipows: I know we've adopted this pattern in many places but I honestly don't see the utility of putting these index names in consts at the top of the file. I feel like it'd be more straightforward to write the string directly in the fn where its called? Feel free to keep it if you disagree though

fynd-core/src/fallback/user_data.rs
```diff
@@ -0,0 +1,350 @@
...
+        return Err(unencodable(component, format!("{name} is wider than four bytes")));
+    }
+    Ok(bytes
+        .iter()
+        .fold(0u32, |value, byte| (value << 8) | u32::from(*byte)))
+}
+
+fn unencodable(component: &ProtocolComponent, reason: String) -> FallbackError {
+    FallbackError::MissingPoolData { component_id: component.id.clone(), reason }
+}
```
tamaralipows: remove passthrough function

fynd-core/src/fallback/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+    algorithm::sim_guard::GuardedProtocolSim,
+    feed::{component_filter::protocol_matches, events::MarketEvent, market_data::MarketDataView},
+    replay::replay_route,
+    types::{ComponentId, FallbackLeg, Route, RouteExclusionFilter, RouteRejection, Swap},
+};
+
+/// Marks a component the `TychoFallbackRouter` executes, e.g. `fallback:fermiswap`. Must match
+/// `tycho-execution`'s `FALLBACK_PREFIX`.
+///
+/// tycho-simulation gives a venue on the router's on-chain whitelist this family instead of
```
tamaralipows, line 34: lets not call it venue (actually I think you were the one who commented that on my PR that we shouldn't call it that?)

fynd-core/src/fallback/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+
+/// Protocol systems a pAMM leg may fall back to.
+///
+/// Must match the systems `TychoFallbackRouter` supports, which Tycho calls venues. A component
+/// under any other protocol system — a Uniswap V2 fork included — has no venue byte the router
+/// understands, so it is never a candidate pool.
+pub(crate) const FALLBACK_PROTOCOL_SYSTEMS: &[&str] =
+    &["uniswap_v2", "uniswap_v3", "uniswap_v4", "vm:curve", "fluid_v1"];
+
+/// The one system in `FALLBACK_PROTOCOL_SYSTEMS` whose pools carry conditions of their own.
```
tamaralipows: conditions of their own is pretty vague

fynd-core/src/fallback/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+}
+
+/// Whether `swap` is a leg the `TychoFallbackRouter` executes.
+///
+/// Named for the family, not for a stamped fallback: a leg is one of these before selection has
+/// run, whereas `Swap::fallback` says whether a pool has been chosen for it.
+fn is_fallback_leg(swap: &Swap) -> bool {
+    swap.protocol()
+        .starts_with(FALLBACK_PREFIX)
+}
```
tamaralipows: unnecessary passthrough

fynd-core/src/fallback/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+
+/// Stamps a fallback pool on every pAMM leg of `route` and returns what the route delivers
+/// through them.
+///
+/// Each `fallback:` swap gets the `FallbackLeg` that `select_fallback` picks for it, so the
+/// route leaves this call both priced and encodable; every other leg is untouched. Stamping and
+/// pricing are one pass, so a route cannot reach the replay unstamped.
+///
+/// The replay goes through `replay_route`, so split fractions and shared-pool depletion behave
+/// exactly as they do for any other route: a substituted leg's smaller output feeds the next one,
```
tamaralipows: Too many implementation/irrelevant details

fynd-core/src/fallback/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+    let mut best: Option<FallbackLeg> = None;
+    let mut last_failure: Option<String> = None;
+    for candidate in admitted {
+        let (Some(component), Some(state)) =
+            (market.get_component(candidate), market.get_simulation_state(candidate))
+        else {
+            last_failure = Some(format!("{candidate} left the market before it could be priced"));
+            continue;
+        };
+        // A pool the router cannot be told how to run is no use however well it prices.
```
tamaralipows: That cannot be encoded?

fynd-core/src/fallback/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+                    ));
+                }
+            }
+            Err(error) => last_failure = Some(format!("{candidate}: {error}")),
+        }
+    }
+
+    best.ok_or_else(|| FallbackError::SimulationFailed {
+        component_id: leg.to_string(),
+        reason: last_failure.unwrap_or_else(|| "no candidate pool could be simulated".to_string()),
```
tamaralipows: It's a bit weird to just give the failure reason for the last candidate and ignore the previous ones. Maybe it would be less confusing if we just said no candidate pool could be simulated and that's all? And then for each individual failure we could put it in a debug log

fynd-core/src/fallback/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+    {
+        return false;
+    }
+    if component.protocol_system != UNISWAP_V4_SYSTEM {
+        return true;
+    }
+    let hooked = component
+        .static_attributes
+        .get(HOOKS_ATTRIBUTE)
+        .is_some_and(|hooks| !is_zero(hooks.as_ref()));
```
tamaralipows: We would also like to filter out angstrom no since it requires attestations and we didn't include support for it on the contract side

fynd-core/src/worker_pool/worker.rs
```diff
@@ -526,85 +456,26 @@ where
...
-                            "dropping pAMM route: the router's fee tiers are not read yet"
-                        );
-                        return Err(SolveError::route_rejected(
-                            order.id(),
-                            RouteRejection::PammFeeTiersUnread,
-                        ));
-                    };
-                    // The same view the algorithm solved against, so the fallback is priced on the
-                    // requested overlay rather than the base state.
+                // A pAMM route is ranked on what its fallbacks pay against `min_amount_out`, so a
```
tamaralipows, line 459: I'm not sure I understood this - why are we ranking them based on the fallback amount? Wouldn't this tie with another solution that is basically exactly the fallback? I think we should make sure the fallback surpasses the min amount - but not rank by it? Did I misunderstand something

fynd-core/src/fallback/user_data.rs
```diff
@@ -0,0 +1,350 @@
...
+///
+/// # Errors
+///
+/// `MissingPoolData`, with what the component is missing.
+pub(super) fn check_encodable(
+    component: &ProtocolComponent,
+    token_in: &Address,
+) -> Result<(), FallbackError> {
+    fallback_protocol(component, token_in).map(|_| ())
+}
```
tamaralipows: passthrough

fynd-core/src/encoding/encoder.rs
```diff
@@ -115,15 +116,22 @@ fn solution_from_quote(
...
                 token_in,
                 token_out,
                 s.gas_estimate().clone(),
             )
             .with_split(*s.split())
             .with_protocol_state(Arc::from(s.protocol_state().clone_box()))
-            .with_estimated_amount_in(s.amount_in().clone()))
+            .with_estimated_amount_in(s.amount_in().clone());
+            // `TychoFallbackRouter` refuses a swap that does not name where the leg retries, so a
+            // stamped leg carries its pool across as `user_data`.
```
tamaralipows: alien language. it refuses a swap that doesn't name the fallback, the fallback pool info is in user_data

fynd-core/src/fallback/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+/// currency: `TychoFallbackRouter` runs V4 without hook data and without value.
+fn is_fallback_candidate(component: &ProtocolComponent) -> bool {
+    if !FALLBACK_PROTOCOL_SYSTEMS.contains(&component.protocol_system.as_str()) ||
+        component.tokens.len() < 2
+    {
+        return false;
+    }
+    if component.protocol_system != UNISWAP_V4_SYSTEM {
+        return true;
+    }
```
tamaralipows: UNIV4 isn't the only one accepting native eth, I think here you will return true if it's a fluid pool but some of them also work with native eth. Also curve too I think? Can you double check please

fynd-core/src/fallback/user_data.rs
```diff
@@ -0,0 +1,350 @@
...
+        assert_eq!(
+            one_to_zero,
+            format!(r#"{{"protocol":"fluid_v1","dex":"{dex}","zero2one":false}}"#)
+        );
+    }
+
+    /// A system the router has no venue byte for is rejected rather than encoded as something else.
+    #[test]
+    fn test_unsupported_protocol_system() {
+        let error = fallback_user_data(&leg("0x1234", "sushiswap_v2", &[]), &address(USDC))
```
tamaralipows: gg we actually want to support sushiswap lol. Sushi is a uniswapv2 clone so it should map nicely to the uniswapv2 executors. Could you please update the PR logic in general so it handles clones correctly?

fynd-core/src/fallback/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+/// whitelisted venue could use it; any pAMM qualifies for this one.
+pub const FALLBACK_PREFIX: &str = "fallback:";
+
+/// Protocol systems a pAMM leg may fall back to.
+///
+/// Must match the systems `TychoFallbackRouter` supports, which Tycho calls venues. A component
+/// under any other protocol system — a Uniswap V2 fork included — has no venue byte the router
+/// understands, so it is never a candidate pool.
+pub(crate) const FALLBACK_PROTOCOL_SYSTEMS: &[&str] =
+    &["uniswap_v2", "uniswap_v3", "uniswap_v4", "vm:curve", "fluid_v1"];
```
tamaralipows: I think we should probably have these in a dict keyed by chain

tamaralipows (commented):
Thank you! A few nitpicks but also several quite big concerns - let me know if you want to discuss them further

fynd-core/src/types/quote.rs
```diff
@@ -2058,6 +2059,64 @@ pub struct Swap {
...
+/// Chosen by the worker among the pools the market holds for the leg's pair, by best simulated
+/// amount out for the leg's `amount_in`. `TychoFallbackRouter` runs it in place of the pAMM swap.
+/// The component and state let the route be replayed through it and let the encoder derive the
+/// router's `[venue][venue data]`. In-process only: it never enters the wire format.
+#[derive(Debug, Clone)]
+pub struct FallbackLeg {
+    /// Identifier of the fallback pool.
+    component_id: ComponentId,
+    /// The fallback pool's component, which the encoder turns into the router's venue data.
+    protocol_component: ProtocolComponent,
```
Troshchk: Doesn't `ProtocolComponent` contain `ComponentId`?
Why do we need both?

fynd-core/src/fallback/user_data.rs
Troshchk: I agree -- I expected `UNISWAP_V4_SYSTEM` to store smth else, not uniswap_v4 string

fynd-core/src/fallback/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+                self.pools.remove(&pair);
+            }
+        }
+    }
+}
+
+/// Whether `component` can serve as a fallback pool.
+///
+/// Its protocol system must be one of `FALLBACK_PROTOCOL_SYSTEMS` and it must hold at least two
+/// tokens. A Uniswap V4 pool must also have a zero or absent hook and no native-ETH (zero address)
```
Troshchk: why absent hook? we don't want to support angstrom, but what about other hooks?

fynd-core/src/fallback/user_data.rs
```diff
@@ -0,0 +1,350 @@
...
+    fallback_protocol(component, token_in).map(|_| ())
+}
+
+/// Reads `component` into the variant its protocol system calls for.
+fn fallback_protocol(
+    component: &ProtocolComponent,
+    token_in: &Address,
+) -> Result<FallbackProtocol, FallbackError> {
+    let pool = pool_address(component)?;
+    match component.protocol_system.as_str() {
```
Troshchk, line 89: We expect to match all of the protocol in FALLBACK_PROTOCOL_SYSTEMS here, but I think if you add smth only to `FALLBACK_PROTOCOL_SYSTEMS` and won't update match here, than we can receive Err when we expect a FallbackProtocol. Can we add a test?

fynd-core/src/fallback/manager.rs
```diff
@@ -117,29 +91,24 @@ impl PammManager {
...
-    /// An unbacked pAMM drops out of the additions. A pAMM whose fallback pool has left joins the
-    /// removals. A pAMM withheld earlier joins the additions once the market backs it: the event
-    /// names the pool that moved, never the pAMMs that fall back to it.
+    /// An unbacked pAMM drops out of the additions. A pAMM whose last candidate pool has left joins
+    /// the removals. A pAMM withheld earlier joins the additions once the market backs it: the
+    /// event names the pool that moved, never the pAMMs that fall back to it.
     ///
     /// Takes the event the market broadcast, before the caller's own filter touches it. The pool
     /// index describes the market rather than one worker's graph, so a pAMM falls back to a pool
     /// its worker excludes just as well, and the pAMM and its pool can arrive in one block.
```
Troshchk, line 106: Is it correct, that we fallback to the pool which worker excludes?

fynd-core/src/lib.rs
```diff
@@ -28,16 +28,16 @@ pub mod algorithm;
 pub mod derived;
 /// Encodes solved routes into ABI-encoded on-chain calldata via Tycho's router contracts.
 pub mod encoding;
+/// Computes the amount out a route delivers when its pAMM legs fall back to Uniswap V3, so the
```
Troshchk: It can now fallback to multiple protocols, not only `Uniswap V3`

bench-harness/src/bench.rs
```diff
@@ -431,12 +431,9 @@ fn error_label(error: &SolveError) -> &'static str {
...
         SolveError::RouteRejected { reason, .. } => match reason {
-            RouteRejection::PammFallbackExcluded => "route rejected: pAMM fallback excluded",
-            RouteRejection::PammFeeTiersUnread => "route rejected: pAMM fee tiers not read",
-            RouteRejection::PammFallbackPoolMissing => "route rejected: pAMM fallback pool missing",
-            RouteRejection::PammFallbackUnpriceable => {
-                "route rejected: pAMM fallback not simulatable"
-            }
+            RouteRejection::FallbackExcluded => "route rejected: fallback excluded",
+            RouteRejection::FallbackPoolMissing => "route rejected: fallback pool missing",
+            RouteRejection::FallbackUnpriceable => "route rejected: pAMM fallback not simulatable",
```
Troshchk: Why is it `Unpriceable` for "not simulatable"?

fynd-core/src/fallback/manager.rs
```diff
@@ -97,7 +72,6 @@ impl PammManager {
         caller_drops: &dyn Fn(&ProtocolComponent) -> bool,
     ) -> FxHashSet<ComponentId> {
         self.fallback_pools = FallbackPoolIndex::build(market);
-        let fee_tiers = self.fee_tiers.snapshot();
```
Troshchk, line 75: A general q: why don't we need fee tiers?

fynd-core/src/fallback/user_data.rs
```diff
@@ -0,0 +1,350 @@
...
+        .ok_or_else(|| unencodable(component, "names only one token".to_string()))?;
+    let (Some(i), Some(j)) = (coin_index(&coins, token_in), coin_index(&coins, token_out)) else {
+        return Err(unencodable(component, "coins do not name both tokens of the pair".to_string()));
+    };
+    Ok(FallbackProtocol::Curve { pool, pool_type, i, j })
+}
+
+/// The pool's coins in index order, falling back to its token list when it carries no `coins`
+/// attribute — the two agree for a two-coin pool, and only a wider pool can order them differently.
+fn curve_coins(component: &ProtocolComponent) -> Vec<Address> {
```
Troshchk: why coins and not tokens?

fynd-core/src/feed/protocol_registry.rs
```diff
@@ -147,8 +139,8 @@ pub(crate) fn has_tycho_protocols(protocols: &[String]) -> bool {
 /// Most entries name their own label. An `exclusive:{system}` entry selects the system's
 /// exclusive-liquidity stream variant, and the prefix is stripped before registration, so its
 /// components arrive under the bare system name. A `pricelevelstream:{venue}` entry names the venue
-/// to stream, and its components arrive labelled `propammfallback:{venue}` when that venue is on
-/// the PropAMMRouter whitelist, so both prefixes answer for the same entry.
+/// to stream, and its components arrive labelled `fallback:{venue}` when that venue is on the
```
Troshchk: venue here as well

Troshchk (commented):
Thank you! Leaving some comments

fynd-core/src/fallback/mod.rs
Troshchk: I actually like this one for readability

tamaralipows: I think the same readability can be achieved with a variable name in one of the places where it's called (below), and in the other place (which is literally right above this), the function name is already called `has_fallback_leg` so it's pretty obvious what the line would be doing. So not really sure what readability this adds - I'm not a fan of crowding the mods with too many functions and obscuring the main important ones

fynd-core/src/fallback/user_data.rs
louise-poole: Inlined it. All the arms are plain strings now.

fynd-core/src/fallback/user_data.rs
louise-poole: Renamed to `uniswap_v4_fallback` and `curve_fallback`.

fynd-core/src/fallback/user_data.rs
louise-poole: This was a bug, thanks. `token_out` was picked as the first token that isn't `token_in`, which gives the wrong coin index on any pool with more than two coins. It now takes the leg's own `token_out`, threaded through from the swap.

fynd-core/src/fallback/user_data.rs
louise-poole: Fair, inlined them.

fynd-core/src/fallback/user_data.rs
louise-poole: Removed. Selection calls `fallback_protocol` directly.

fynd-core/src/fallback/user_data.rs
louise-poole: Curve indexes `exchange` by coin position, and the component's `tokens` make no promise about that order. `tokens` is only the fallback when a pool carries no `coins` attribute. Expanded the doc to say so.

fynd-core/src/fallback/user_data.rs
louise-poole: Added `test_every_admitted_system_encodes`, which walks `FALLBACK_PROTOCOL_SYSTEMS` and asserts each one encodes.

fynd-core/src/fallback/mod.rs
louise-poole: Dropped. The `venue` left in that file refers to the pAMM, which I think is the right use of it.

fynd-core/src/fallback/mod.rs
louise-poole: The const is gone, so the doc went with it.

fynd-core/src/fallback/mod.rs
louise-poole: Trimmed to what the caller needs.

fynd-core/src/fallback/mod.rs
louise-poole: Yes. Reworded to "A pool we cannot encode is no use however well it prices."

fynd-core/src/fallback/mod.rs
louise-poole: Agreed, changed. Each candidate logs its own reason at debug as it's skipped, and the error just says no pool could be used.

fynd-core/src/encoding/encoder.rs
louise-poole: Reworded: "`TychoFallbackRouter` refuses a swap that does not name its fallback, and the fallback pool goes in `user_data`."

fynd-core/src/types/quote.rs
louise-poole: It does. Dropped the field, and the accessor reads it off the component.

fynd-core/src/lib.rs
louise-poole: Fixed.

fynd-core/src/feed/protocol_registry.rs
louise-poole: Fixed, calls it the pAMM now.

bench-harness/src/bench.rs
louise-poole: A remanent from a previous iteration - well spotted. Thanks!

fynd-core/src/fallback/manager.rs
louise-poole: This was very specific to the old fallback limitations we had with the usv3-only approach. We had to use the fee tiers titan specified for that propAMM's fallback and so needed this here to correctly choose the fallback pool. This is no longer needed now that we have out own fallback logic.

fynd-core/src/fallback/mod.rs
louise-poole: Agreed on Angstrom. The catch is where the address comes from: tycho keeps it in `PROTOCOL_SPECIFIC_CONFIG`, which is `pub(crate)` in tycho-execution, so fynd can't read it at any version. Rather than hardcode a copy on our side I'd like tycho to expose it. Until then the rule stays strict and rejects every hooked pool; once the address is readable this becomes "reject Angstrom, allow the rest".

fynd-core/src/fallback/mod.rs
louise-poole: Other hooks should be allowed, agreed. We reject them all today only because the one we must reject, Angstrom, has its address in tycho's `pub(crate)` config, so we can't tell it apart from the rest. I think we should make tycho expose this address; then this becomes an Angstrom-only check.

fynd-core/src/worker_pool/worker.rs
louise-poole: It isn't ranked on that. `fallback_amount_out` is only read in `worker_pool_router/mod.rs`, inside `fallback_clears_min_amount_out`, to drop a quote whose fallback misses the floor. Ranking stays on `amount_out_net_gas`. So the behaviour is the one you describe wanting; nothing changed here.

fynd-core/src/fallback/mod.rs
louise-poole: Base is planned, though there's no fallback encoder for it yet and the current one rejects every chain but Ethereum at construction. I'd rather key this by chain once the Base encoder exists and says which protocols it supports than guess the shape now. Left flat for this PR.

fynd-core/src/fallback/mod.rs
louise-poole: You were right, and the check was wrong in two ways: it only tested the zero address, where Curve and Fluid use the all-`0xee` sentinel, and it judged the pool's whole token list, which would have thrown out a three-coin Curve pool holding ETH even for its USDC/WETH pair. It's now per pair, in `FallbackPoolIndex::insert`: a pair with native ETH on either side under either address is never filed, and the pool's other pairs still serve. 4bcb346c — https://github.com/propeller-heads/fynd/commit/4bcb346c

fynd-core/src/fallback/user_data.rs
louise-poole: Done. The V2 and V3 forks are in `FALLBACK_PROTOCOL_SYSTEMS` and encode as their base protocol. The fee comes off the pool's own `ProtocolSim::fee()` rather than a table, so PancakeSwap's 25 bps rides through, and anything over the router's 30 bps cap is refused at selection instead of reverting on chain. The fork names are hardcoded for now with a note to import `UNISWAP_V2_FORKS`/`UNISWAP_V3_FORKS` from tycho-execution once the encoder is released, so the two lists can't drift. c5a1e70b — https://github.com/propeller-heads/fynd/commit/c5a1e70b

fynd-core/src/fallback/manager.rs
louise-poole: Good catch. The request's filter was honoured but the worker pool's own `exclude_protocols` wasn't, so a pool told never to route through `uniswap_v3` would still settle its pAMM legs there. Both now rule candidates out in `select_fallback`, with a test. 4bcb346c — https://github.com/propeller-heads/fynd/commit/4bcb346c

bench-harness/src/bench.rs
louise-poole: Fair. Renamed to `FallbackNotSimulatable`, through the capture slug and the bench label too. 1c1f8a41 — https://github.com/propeller-heads/fynd/commit/1c1f8a41

fynd-core/src/fallback/mod.rs
louise-poole: Removed. `has_fallback_leg` and the stamp loop test the prefix directly. 1c1f8a41 — https://github.com/propeller-heads/fynd/commit/1c1f8a41

fynd-core/src/fallback/user_data.rs
louise-poole: Removed, the struct literal is written out at each site. 1c1f8a41 — https://github.com/propeller-heads/fynd/commit/1c1f8a41

fynd-core/src/fallback/mod.rs
louise-poole: `FALLBACK_PROTOCOL_SYSTEMS` is now `fallback_protocol_systems(chain)`, a `match` over `Chain` so the compiler checks it: Ethereum returns its list, anything else returns nothing, which withholds every pAMM on a chain with no router. Base gets an arm when its encoder exists and we know what it supports. `is_fallback_candidate` reads the chain off the component, so nothing upstream changed. 34ef017f — https://github.com/propeller-heads/fynd/commit/34ef017f

fynd-core/src/fallback/mod.rs
Troshchk: Ok, thanks, that is clear
So can it be part of this change?

Troshchk (approved):
Thank you! ✨

tamaralipows (approved):
Thank you! All looks good! Please just double check on removing hook support like we discussed today

PR #514, opened 2026-09-03
fix!: Simulate non erc20 tokens and expose revert reasons

carloszanella:
The breaking change is just on the simulation tool.

This adds support for simulating almost all tokens. Shared the dashboard on slack.

Tested overnight and dashboard is live


fynd-core/src/simulation/token_layout.rs
```diff
@@ -0,0 +1,447 @@
...
+/// Its `balanceOf` multiplies a holder's shares by the pooled-ETH rate, so tracing it finds the
+/// arithmetic rather than the mapping. The shares view is traced instead. The address is mainnet's
+/// and matches nothing on another chain, which leaves every other token on the ordinary path.
+const STETH: Address = address!("ae7ab96520de3a18e5e111b5eaab095312d7fe84");
+/// OpenZeppelin v5's standard ERC-7201 balance namespace.
+const OZ_V5_BALANCES_NS: B256 =
+    B256::new(alloy::hex!("52c63247e1f47db19d5ce0460030c497f067ca4cebf71ba98eeadabe20bace00"));
+/// Field one of OpenZeppelin v5's ERC-20 storage struct.
+const OZ_V5_ALLOWANCES_NS: B256 =
+    B256::new(alloy::hex!("52c63247e1f47db19d5ce0460030c497f067ca4cebf71ba98eeadabe20bace01"));
```
louise-poole, line 50: Yes - removing code duplication is preferable.

---

🤖 `token_layout` re-declares `PROBE_SENTINEL`, byte-identical `OZ_V5_BALANCES_NS`/`OZ_V5_ALLOWANCES_NS`, its own `slot_matches`, its own `IERC20` probe interface, and a second enum also called `MappingPosition` with different variants — all sitting beside the `erc20_slots` originals. `solidity_mapping` is `erc20_slots::balance_slot_at_b256` written out again and the Solidity arm of `allowance_slot` is `allowance_slot_at_b256`. The ERC-7201 namespace, the mapping formula and the sentinel-probe protocol are now encoded twice: fix a slot-derivation bug and you fix it twice, and nothing fails to compile when you miss one. A reader landing on either `MappingPosition` also has to work out which module they're in.

Make `token_layout` the single implementation and rewrite `erc20_slots`' three tool-facing entry points (`find_balance_slot`, `find_allowance_slot`, `state_override_single` — the only things `tools/fynd-gas-audit`, `tools/fynd-swap-cli` and `tools/common` use) as thin wrappers over it, deleting the parallel probe path. If that's too much for this PR, at minimum have `token_layout` call `erc20_slots`' keccak helpers and import the two namespace constants rather than restating them, and give one of the two `MappingPosition` enums a name saying which convention set it covers. Worth deciding here too whether the tool crates move onto the traced path — they're still on the sentinel probe this PR just demonstrated can't place USDT.

---

fynd-core/src/simulation/token_layout.rs
```diff
@@ -0,0 +1,447 @@
...
+/// The base of one balance or allowance mapping.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum MappingPosition {
+    /// A small integer base slot under Solidity or Vyper mapping layout.
+    Direct { base: u16, key_order: KeyOrder },
+    /// The canonical OpenZeppelin v5 ERC-7201 balances namespace.
+    OpenZeppelinV5Balances,
+    /// The canonical OpenZeppelin v5 ERC-7201 allowances namespace.
+    OpenZeppelinV5Allowances,
+}
```
louise-poole: 🤖 `OpenZeppelinV5Balances` and `OpenZeppelinV5Allowances` encode which mapping the caller wants, but the caller already picked that when it chose `mapping_slot` or `allowance_slot`. That leaves four combinations, two of which compute slots nothing reads: `allowance_slot(owner, spender, OpenZeppelinV5Balances)` nests a mapping over the balances namespace, and `mapping_slot(holder, OpenZeppelinV5Allowances)` returns a balance slot in the allowances namespace. The recovery functions only ever produce the matching pair.

Collapse to one `OpenZeppelinV5` variant and let `mapping_slot` use `OZ_V5_BALANCES_NS` and `allowance_slot` use `OZ_V5_ALLOWANCES_NS` — the shape `erc20_slots::balance_slot_for_position`/`allowance_slot_for_position` already has. That deletes two match arms and removes the impossible states from the type.

---

fynd-core/src/simulation/token_layout.rs
```diff
@@ -0,0 +1,447 @@
...
+/// that touches more than this many slots is a token whose layout this module does not resolve.
+const MAX_SLOTS_TO_VERIFY: usize = 32;
+/// A value that survives common packed-balance flags and narrow integer casts.
+const PROBE_SENTINEL: U256 = U256::from_limbs([0xdead_beef_cafe_babe, 0, 0, 0]);
+/// Mainnet Lido stETH.
+///
+/// Its `balanceOf` multiplies a holder's shares by the pooled-ETH rate, so tracing it finds the
+/// arithmetic rather than the mapping. The shares view is traced instead. The address is mainnet's
+/// and matches nothing on another chain, which leaves every other token on the ordinary path.
+const STETH: Address = address!("ae7ab96520de3a18e5e111b5eaab095312d7fe84");
```
louise-poole: Yes - let's keep this chain agnostic?

---

🤖 `balance_calldata` branches on one hard-coded mainnet address to trace `sharesOf` instead of `balanceOf`. The chain is a deployment parameter everywhere else in this module, and the next share-based rebasing token — wstETH, or the same pattern on another chain — falls out as `Unsupported` with nothing pointing at why.

Drop the constant and make it a fallback: when `recover_balance_position` can't place the slot the `balanceOf` trace observed, retry with `sharesOf`. Pros — no address list, works on any chain, picks up the whole share-accounted family. Cons — an extra trace plus probe round for tokens that are genuinely unrecoverable, and a second failure path to report. If the single address is the right call for now, say in the doc that the next one needs the fallback rather than another constant.

---

fynd-core/src/simulation/simulator.rs
```diff
@@ -218,11 +230,35 @@ impl QuoteSimulator {
...
+                    timeout(self.request_timeout, traced_revert_reason(&self.provider, *replay))
+                        .await
+                        .ok()
+                        .flatten();
+                SimulationAttempt::Reverted {
+                    reason: format!("simulation reverted: {}", traced.unwrap_or(message)),
+                }
+            }
         }
     }
```
louise-poole: 🤖 `simulate_with_overrides` boxes the request, the state overrides and the block overrides into a `Replay` and hands them back up, and `simulate_within_timeout` immediately passes them to `traced_revert_reason`. That's a continuation dressed as an outcome: the `Replay` struct, the `Box`, and one enum variant exist only to move the second RPC across a function boundary. The stated reason for the split — the trace gets its own budget — is a parameter, not a structural constraint.

Pass the trace budget into `simulate_with_overrides` and let it run `traced_revert_reason` itself, returning a plain `Reverted { reason }`. That removes `Replay`, the box, and the `RevertedUntraced` variant, and leaves `simulate_within_timeout` mapping three outcomes instead of four. If the variant stays, rename it `RevertedWithoutReason` — what happened is that the revert carried no decodable payload, not that a trace hasn't run yet.

---

louise-poole (commented):
Read through this one carefully — the traced-layout approach is the right call and the revert decoding is a genuine improvement to what we can see. Two things I'd want resolved before merge: an RPC error response during discovery is currently recorded as "wrong slot" and then cached permanently, so one throttled burst can disable simulation for a token until restart; and the new discovery path is only covered by an `#[ignore]`d live test. The rest is mostly duplication between `token_layout` and `erc20_slots`, plus metric naming.


---

**Constant still says "slot"** — `fynd-core/src/solver.rs:106`

🤖 `SIMULATION_SLOT_DISCOVERY_TIMEOUT` guards layout discovery now, and the error message it feeds already reads "token layout discovery timed out" — rename to `SIMULATION_LAYOUT_DISCOVERY_TIMEOUT`? The file isn't in this diff, so it's your call whether it rides along.

---

**The checked-in Grafana dashboard gains no panel for the three new metrics** — `monitoring/grafana/dashboards/fynd.json:1`

Actually why do we have a grafana dashboard living in this repo?

fynd-core/src/worker_pool_router/instrumentation.rs
```diff
@@ -279,6 +282,72 @@ pub(super) fn log_winning_protocols(quote: &OrderQuote) {
     }
 
     let (outcome, deviation) = simulation_fields(quote);
+    count_swaps_per_protocol(&swaps_per_protocol, outcome);
+    if let Some(bps) = simulated_deviation_bps(quote) {
+        record_deviation_per_protocol(&swaps_per_protocol, bps);
+    }
+    log_winning_protocols(quote, route, &swaps_per_protocol, outcome, &deviation);
```
louise-poole: 🤖 `record_winning_protocols` calls `simulation_fields`, which runs `deviation_bps` and `format!`s the result, then calls `simulated_deviation_bps`, which runs `deviation_bps` again on the same quote. The formatted string is only used by `log_winning_protocols`, which returns early because `fynd::winning_protocols` is off outside dev — so in production every winning quote pays for a formatted string nothing reads, and a change to one baseline has to be made in both places.

Have `simulation_fields` return `(&'static str, Option<f64>)`, feed that `Option<f64>` to `record_deviation_per_protocol`, and move the `format!("{bps:.4}")` inside `log_winning_protocols` after its `enabled!` gate. `simulated_deviation_bps` goes away.

---

fynd-core/src/worker_pool_router/instrumentation.rs
```diff
@@ -279,6 +282,72 @@ pub(super) fn log_winning_protocols(quote: &OrderQuote) {
...
+    }
+}
+
+fn log_winning_protocols(
+    quote: &OrderQuote,
+    route: &Route,
+    swaps_per_protocol: &BTreeMap<&str, usize>,
+    outcome: &'static str,
+    deviation: &str,
+) {
```
louise-poole, line 355: 🤖 `route` is just `quote.route()`, which the function's first parameter already carries — derive it inside and drop the parameter, along with the `Route` import added at line 25.

---

fynd-core/src/simulation/revert.rs
```diff
@@ -0,0 +1,440 @@
...
+        error UniswapXFiller__AddressZero();
+        error UniswapXFiller__BatchExecutionNotSupported();
+        error Vault__AddressZero();
+        error Vault__AmountZero();
+        error Vault__InsufficientBalance(address user, address token, uint256 requested, uint256 available);
+        error Vault__InvalidInputDelta(address token, int256 expected, int256 actual);
+        error Vault__UnexpectedInputDelta(int256 inputDelta);
+        error Vault__UnexpectedNonZeroCount(uint256 nonZeroCount);
+    }
+}
```
louise-poole, line 132: Should we maybe export this on tycho-execution?

---

🤖 `RouterErrors` is a hand-copied snapshot of the Solidity error set in a crate this workspace pins by revision, under a doc promising "Every custom error the Tycho router, its dispatcher and its executors can revert with". The moment `tycho-execution` adds an executor or renames an error that claim is false and nobody finds out — the selector fallback quietly absorbs it and the error reports as "unknown error, selector 0x…" forever. A mistyped argument type fails the same silent way. The four errors the tests exercise all round-trip through the same declaration, so they can't catch a wrong signature.

Best would be for `tycho-execution` to export the ABI and for this interface to be generated from it, so the errors are defined once by the crate that owns the contracts. Failing that, assert a handful of selectors against the values `tycho-execution` publishes, or add the regeneration step to the tycho version-bump checklist. Either way, soften the doc to "the errors of the `tycho-execution` version this crate pins" and drop the "every" — a promise you can't keep is worse than no promise.

---

fynd-core/src/simulation/simulator.rs
```diff
@@ -251,35 +287,71 @@ impl QuoteSimulator {
...
+    fn remember(&self, token: Address, verdict: Result<TokenLayout, String>) -> Result<(), String> {
+        self.layout_cache
             .lock()
-            .map_err(|_| "simulation slot cache lock poisoned".to_string())?
-            .insert(token, positions);
-        Ok(positions)
+            .map_err(|_| "simulation layout cache lock poisoned".to_string())?
+            .insert(token, verdict);
+        Ok(())
     }
```
louise-poole: 🤖 Three levels of failure in the return type of a map lookup, and its only caller pairs it with a five-line `remember` that repeats the same lock-poisoning arm. Inline both into `cached_layout`, which is where the nesting resolves anyway.

fynd-core/src/simulation/token_layout.rs
```diff
@@ -0,0 +1,447 @@
...
+            Ok(response.len() >= 32 && U256::from_be_slice(&response[..32]) == PROBE_SENTINEL)
+        }
+        // A guarded proxy reverts when its implementation slot is overwritten, which proves the
+        // slot is not the mapping. A node that failed to answer proves nothing, so it is reported
+        // rather than counted as a miss that would end in "could not identify".
+        Err(error) if error.as_error_resp().is_some() => Ok(false),
+        Err(error) => Err(DiscoveryError::Rpc(format!(
+            "sentinel probe for {token:#x} slot {slot:#x} failed: {error}"
+        ))),
+    }
```
louise-poole, line 304: 🤖 `slot_matches` maps every JSON-RPC error response to `Ok(false)`. The comment above that arm says a node that failed to answer is reported rather than counted as a miss, but `as_error_resp()` also catches rate limits (`-32005`), `header not found` on a moving head, and per-request compute limits. When a burst of those lands, every candidate scores false, `find_accessed_slot` returns `DiscoveryError::Unsupported`, and `cached_layout` (`simulator.rs:326–329`) writes that verdict into `layout_cache` with no eviction — so the token is never simulated again for the life of the process. The 32-way concurrent fan-out on the line above makes hitting a provider limit more likely, not less.

An operator runs with simulation enabled behind a rate-limited or busy RPC provider. One burst of throttled responses while a new token is being examined, and from then on every quote touching that token comes back with "token layout discovery failed" instead of a simulation result — including long after the provider has recovered. The only way back is a server restart. A node that refused to answer should leave the token undecided so the next quote tries again, exactly as a transport failure already does.

Distinguish a revert (which does prove the slot is wrong) from every other error response. Match on the payload — treat "execution reverted" / revert data as `Ok(false)` and return `DiscoveryError::Rpc` for the rest, so the verdict isn't remembered.

---

fynd-core/src/simulation/simulator.rs
```diff
@@ -218,11 +230,35 @@ impl QuoteSimulator {
...
+                let traced =
+                    timeout(self.request_timeout, traced_revert_reason(&self.provider, *replay))
+                        .await
+                        .ok()
+                        .flatten();
+                SimulationAttempt::Reverted {
+                    reason: format!("simulation reverted: {}", traced.unwrap_or(message)),
+                }
+            }
         }
```
louise-poole, line 257: Sounds reasonable, but up to you...

---

🤖 `simulate_within_timeout` awaits the simulate call under `request_timeout`, then on `RevertedUntraced` awaits the trace under a second full `request_timeout`. With the shipped default of 3s that's 6s on the quote path, on top of the 2s discovery budget the same request may already have spent. `debug_traceCall` with the call tracer is markedly more expensive server-side than `eth_simulateV1`, so the second budget is the one likely to be consumed.

Give the trace a budget of its own — a fixed 500ms–1s — rather than reusing `request_timeout`, so the diagnostic can never cost the caller more than the simulation itself.

---

fynd-core/src/simulation/simulator.rs
```diff
@@ -251,35 +287,71 @@ impl QuoteSimulator {
...
+            }
+            Err(DiscoveryError::Unsupported(reason)) => {
+                self.remember(token, Err(reason.clone()))?;
+                Err(format!("simulation token layout discovery failed: {reason}"))
+            }
+            Err(DiscoveryError::Rpc(reason)) => {
+                Err(format!("simulation token layout discovery failed: {reason}"))
+            }
+        }
+    }
```
louise-poole, line 341: Not sure how necessary this would be, but sounds nice.

---

🤖 `cached_layout` checks the cache, drops the lock, and starts discovery. Quotes arriving for the same token before the first finishes all miss and all discover: each is one `debug_traceCall` plus up to 32 concurrent `eth_call`s per traced account, twice. Every new token pays that multiplier once, and it lands as a burst on the same endpoint whose error responses get cached as a permanent verdict.

Hold one in-flight entry per token — a `HashMap<Address, Shared<...>>` or a per-token `tokio::sync::OnceCell` — so concurrent callers await the same discovery instead of repeating it.

---

fynd-core/src/simulation/simulator.rs
```diff
@@ -473,35 +579,55 @@ fn native_balance_override(sender: Address) -> StateOverride {
...
+        .await
+    {
+        Ok(frame) => revert::reason_from_frame(&frame),
+        Err(error) => {
+            tracing::debug!(%error, "tracing a reverted simulation failed");
+            None
         }
     }
-    format!("reverted with data 0x{}", alloy::hex::encode(data))
 }
```
louise-poole, line 634: 🤖 `traced_revert_reason` is documented as running "against the same block", and it carries the same `prevrandao`, coinbase and gas limit. But `eth_simulateV1` executes in a block built on top of the head (number N+1, timestamp advanced), while `debug_traceCall` at `BlockNumberOrTag::Latest` executes in block N's own environment. `block_overrides()` (`simulator.rs:437–444`) pins neither `number` nor `timestamp`, so the two differ.

An integrator sees a quote come back saying the swap reverted, with a named reason. For a route whose pools or signatures key off the block height or the clock, that named reason can belong to a different failure than the one the simulation actually hit — or the replay can succeed where the simulation reverted, leaving the generic message instead. The reason attached to a reverted quote should describe the same execution the quote was rejected for.

Set `number` and `timestamp` explicitly in `block_overrides()` to the values `eth_simulateV1` will use, so both calls run in the same environment rather than one head block apart.

fynd-core/src/simulation/token_layout.rs
```diff
@@ -0,0 +1,447 @@
...
+            return Some(direct);
+        }
+        let vyper = MappingPosition::Direct { base, key_order: KeyOrder::Vyper };
+        if allowance_slot(owner, spender, vyper) == slot {
+            return Some(vyper);
+        }
+    }
+    let oz = MappingPosition::OpenZeppelinV5Allowances;
+    (allowance_slot(owner, spender, oz) == slot).then_some(oz)
+}
```
louise-poole: 🤖 `recover_balance_position` and `recover_allowance_position` are the same loop over `0..=MAX_BASE_SLOT` trying Solidity, then Vyper, then the OpenZeppelin namespace; they differ only in which slot function they call and which OZ variant they end on. 33 lines carrying one idea.

One `fn recover_position(slot: B256, oz: MappingPosition, slot_for: impl Fn(MappingPosition) -> B256) -> Option<MappingPosition>`, which is the shape `erc20_slots::probe_position` already uses. Saves roughly 18 lines and one place to forget when a third convention is added.

---

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -601,21 +605,19 @@ impl WorkerPoolRouter {
     /// # Errors
     ///
     /// Returns [`SolveError::Internal`] when the request asks for simulation and the server was
-    /// built without a simulator.
+    /// started without `--enable-simulation`.
     pub async fn simulate_quotes(
         &self,
-        mut order_quotes: Vec<OrderQuote>,
+        order_quotes: Vec<OrderQuote>,
```
louise-poole: 🤖 The parameter carried `mut` directly before and is now taken immutably and rebound one line later — same thing, one line longer. Put the `mut` back on line 611 and delete the rebinding.

---

fynd-core/src/simulation/deviation.rs
```diff
@@ -0,0 +1,151 @@
...
+pub(crate) mod fixtures {
+    use num_bigint::BigUint;
+
+    use crate::OrderQuote;
+
+    /// Router fee the fixture charges, in output-token units.
+    pub(crate) const FIXTURE_ROUTER_FEE: u64 = 7_000;
+
+    /// Client fee the fixture charges, in output-token units.
+    pub(crate) const FIXTURE_CLIENT_FEE: u64 = 3_000;
```
louise-poole: 🤖 Only `quote_with_fees` crosses a module boundary; drop `FIXTURE_ROUTER_FEE`, `FIXTURE_CLIENT_FEE` and `quote_without_fees` from `pub(crate)` to `pub(super)`.

---

fynd-core/src/simulation/simulator.rs
```diff
@@ -218,11 +230,35 @@ impl QuoteSimulator {
...
+            CallOutcome::RevertedUntraced { message, replay } => {
+                let traced =
+                    timeout(self.request_timeout, traced_revert_reason(&self.provider, *replay))
+                        .await
+                        .ok()
+                        .flatten();
+                SimulationAttempt::Reverted {
+                    reason: format!("simulation reverted: {}", traced.unwrap_or(message)),
+                }
+            }
```
louise-poole: Super minor - up to you.

---

🤖 Both `Reverted` arms wrap their reason in the same prefix; resolve the reason in the match and format once after it.

fynd-core/src/simulation/token_layout.rs
```diff
@@ -0,0 +1,447 @@
...
+        }
+        // A guarded proxy reverts when its implementation slot is overwritten, which proves the
+        // slot is not the mapping. A node that failed to answer proves nothing, so it is reported
+        // rather than counted as a miss that would end in "could not identify".
+        Err(error) if error.as_error_resp().is_some() => Ok(false),
+        Err(error) => Err(DiscoveryError::Rpc(format!(
+            "sentinel probe for {token:#x} slot {slot:#x} failed: {error}"
+        ))),
+    }
+}
```
louise-poole, line 305: 🤖 `discover_layout`, `find_accessed_slot` and `slot_matches` are the new behavior this PR exists for, and the only test that touches them needs `RPC_URL` and is `#[ignore]`d, so CI never runs it. The pure keccak recovery functions are well covered; none of the decision logic is — the error-response-means-"wrong slot" rule, `Rpc` vs `Unsupported` classification, the `MAX_SLOTS_TO_VERIFY` cap, the highest-keys-first candidate ordering, or the "balance and allowance in different contracts" rejection. The crate already drives these paths with `Asserter` elsewhere in the simulation tests.

Add `Asserter`-backed tests for `find_accessed_slot`: a prestate trace whose second candidate matches, one where every probe returns an error response (asserting `Unsupported`), one where a probe fails at the transport (asserting `Rpc`), and one where the allowance trace names a different contract.

---

fynd-core/src/simulation/simulator.rs
louise-poole, line 257: 🤖 `revert::reason_from_frame` is thoroughly unit-tested, but the wiring that reaches it isn't. No test drives `simulate_with_overrides` to a `!status` result with empty `return_data`, so nothing covers that `RevertedUntraced` is produced, that `traced_revert_reason` is called with the same overrides and environment, that its answer replaces the node's message, or that a trace failure falls back to that message. This is the path the PR title advertises.

Add a `simulate_within_timeout` test with a mocked provider that answers the simulate call with a reverted, empty-payload result and the trace call with a `CallFrame` carrying a router error, asserting the reason names the error; add a second where the trace errors and the node's message survives.

---

fynd-core/src/tests/simulation/simulator.rs
```diff
@@ -199,35 +198,58 @@ async fn test_simulated_call_decodes_revert_data_from_mocked_rpc_error() {
...
+    let error = simulator
+        .cached_layout(token, Address::repeat_byte(1), Address::repeat_byte(2))
         .await
-        .expect("cache resolves slots");
-    assert!(asserter.read_q().is_empty());
+        .expect_err("the token stays unresolvable");
+
+    assert!(error.contains("no supported balance mapping"), "{error}");
+    assert!(asserter.read_q().is_empty(), "a remembered verdict makes no RPC call");
 }
```
louise-poole: 🤖 Both tests insert into `layout_cache` directly and assert the entry is read back without an RPC call. The behavior that matters — that an `Unsupported` discovery is written to the cache and an `Rpc` failure deliberately isn't — is never triggered, so a change that starts caching RPC failures (or stops caching unsupported tokens) passes.

Drive `cached_layout` through a mocked discovery returning each `DiscoveryError` variant and assert on the resulting cache contents, then that a second call re-probes for `Rpc` and doesn't for `Unsupported`.

---

fynd-core/src/tests/simulation/simulator.rs
```diff
@@ -109,6 +97,17 @@ fn test_block_overrides_leave_nothing_a_pool_reads_at_zero() {
...
 
+/// The funding value is what makes a simulated sender solvent, and it is bounded on both sides:
+/// too small starves a large trade, too large overflows a rebasing token's balance arithmetic.
+#[test]
+fn test_funding_value_is_ten_to_the_thirty_sixth() {
+    let ten = U256::from(10_u8);
+    assert_eq!(SIMULATION_FUNDING_VALUE, ten.pow(U256::from(36_u8)));
+    // Room left above the value, so a token that packs flags into the balance word still reads
+    // it back unchanged.
+    assert!(SIMULATION_FUNDING_VALUE < U256::MAX >> 128);
```
louise-poole, line 139: 🤖 `assert_eq!(SIMULATION_FUNDING_VALUE, ten.pow(36))` re-spells the literal it checks and fails only when someone deliberately changes it — drop it and keep the headroom bound below, which is the assertion carrying the property. Rename to `test_funding_value_bounds` so the name says what's under test rather than the expected value.

---

fynd-core/src/tests/simulation/simulator.rs
```diff
@@ -272,111 +294,6 @@ async fn test_simulation_times_out_when_the_node_does_not_answer() {
...
-#[test]
-fn test_deviation_bps_amount_beyond_f64() {
-    let huge = BigUint::from(1u8) << 2_000;
-
-    assert_eq!(deviation_bps(&quote_with_fees(1_000_000), &huge), None);
-}
-
 /// Names every metric a run recorded, with its labels and value.
 fn recorded_metrics(
     snapshotter: &metrics_util::debugging::Snapshotter,
```
louise-poole: 🤖 `recorded` and `recorded_metrics` are the same function with the same doc comment. Move one into a shared test helper and delete the other.

---

fynd-core/src/tests/simulation/erc20_slots.rs
```diff
@@ -89,14 +89,3 @@ fn test_oz_v5_allowances_ns_is_balances_ns_plus_one() {
         U256::from_be_bytes(*OZ_V5_BALANCES_NS) + U256::from(1_u8)
     );
 }
```
louise-poole: 🤖 `test_oz_v5_allowances_ns_is_balances_ns_plus_one` covers `erc20_slots`' constants; `token_layout.rs:42–47` now holds an unguarded second copy. Cover it too, or share the constants.

fynd-core/src/worker_pool_router/instrumentation.rs
```diff
@@ -279,6 +282,72 @@ pub(super) fn log_winning_protocols(quote: &OrderQuote) {
...
+    // shortfall so the counter never has to take a negative delta.
+    let shortfall = (-bps * SHORTFALL_SCALE)
+        .max(0.0)
+        .round() as u64;
+    for protocol in swaps_per_protocol.keys() {
+        counter!("winning_quote_shortfall_centibps_sum", "protocol" => protocol.to_string())
+            .increment(shortfall);
+        counter!("winning_quote_shortfall_count", "protocol" => protocol.to_string()).increment(1);
+    }
+}
```
louise-poole: 🤖 `.max(0.0)` floors a positive deviation to zero but `winning_quote_shortfall_count` still increments, so a simulation that came back *above* the quote drags the protocol's mean shortfall down instead of being left out. The comment says a negative deviation is "every case the router's positive-slippage capture allows" — is that still true with `disable_slippage_taking` set, or when the pool moves in the user's favour between the quote and the simulation?

Skip both counters when `bps > 0.0`, so the mean is taken over the routes that actually fell short. If you'd rather keep sight of the favourable ones, give them their own counter — but don't let them ride into the sum as zeroes.

---

fynd-core/src/simulation/simulator.rs
```diff
@@ -251,35 +287,71 @@ impl QuoteSimulator {
...
+                .map_err(|reason| format!("simulation token layout discovery failed: {reason}"));
+        }
+        let discovered = timeout(
+            SIMULATION_SLOT_DISCOVERY_TIMEOUT,
+            discover_layout(&self.provider, token, holder, spender),
+        )
+        .await
+        .map_err(|_| {
+            format!("simulation token layout discovery timed out after {SIMULATION_SLOT_DISCOVERY_TIMEOUT:?}")
+        })?;
```
louise-poole: 🤖 `SIMULATION_SLOT_DISCOVERY_TIMEOUT` (`fynd-core/src/solver.rs:106`) is still the 2s the old sentinel probe used, but discovery now runs two `debug_traceCall`s plus up to 32 `eth_call`s *per traced account*, twice over. On a proxy token whose prestate names several accounts that's a lot of round trips to fit into 2s. And a timeout comes back as an `Rpc`-shaped failure, which `cached_layout` deliberately doesn't remember — so every subsequent quote on that token pays the full 2s again.

Raise the budget to match the new work, and cap the probes across all traced accounts rather than per account so the worst case is bounded. You've had it running overnight — how often did the timeout branch actually fire?

---

fynd-core/src/simulation/revert.rs
```diff
@@ -0,0 +1,440 @@
...
+
+        let reason = reason_from_frame(&top).expect("the frame reverted");
+
+        assert!(reason.contains("TychoRouter__EmptySwaps"), "{reason}");
+    }
+
+    #[test]
+    fn test_reason_from_frame_keeps_a_plain_revert_reason() {
+        let mut top = frame(None, Some("execution reverted"), vec![]);
+        top.revert_reason = Some("execution reverted".to_string());
```
louise-poole, line 521: I am a fan of collapsing appropriate tests into test cases.

---

🤖 `test_reason_from_frame_*` is one function tested eight times over different call trees, each name carrying its outcome to tell it from its neighbours. That's usually the signal for one test with cases.

Collapse them into a single `rstest` taking the frame tree and the substring the reason must contain. The three that assert something genuinely different (no revert, plain reason, hex reason) can stay standalone if they read better that way.

---

fynd-core/CLAUDE.md
```diff
@@ -16,7 +16,7 @@ applications.
...
 | `price_guard/`        | Price guard: external price validation for quotes. Sub-modules: `guard` (validation logic), `binance_ws` (Binance WebSocket price provider), `hyperliquid` (Hyperliquid oracle provider), `provider_registry`, `config`, `utils` |
-| `simulation/`         | ERC-20 storage-slot discovery and `eth_simulateV1` quote simulation with temporary state overrides |
+| `simulation/`         | Encoded-quote simulation, switched on per deployment with `--enable-simulation`. `token_layout` discovers a token's balance and allowance slots by tracing the storage its own view functions read, so a non-standard layout (USDT, stETH shares) can still be funded; `simulator` runs the quote under `eth_simulateV1` with those overrides; `revert` names the error a reverting call carries by replaying it under the call tracer. `erc20_slots` keeps the older sentinel-probe helpers, which the tool crates use. `deviation` is ungated: the quote log reports the same figure the metric does |
```
louise-poole: 🤖 "`erc20_slots` keeps the older sentinel-probe helpers" is a nod to the pre-refactor version rather than a description of what's there, and "`deviation` is ungated: the quote log reports the same figure the metric does" won't parse for anyone who hasn't read this diff.

Describe what each module is, not what changed: `erc20_slots` — sentinel-probe slot lookup used by the tool crates; `deviation` — the quote-versus-simulation bps figure, shared by the log and the metric. The rest of the cell is good.

---

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -3626,7 +3622,7 @@ mod tests {
     /// Runs `log_winning_protocols` and returns the payload of each line it wrote.
```
louise-poole: 🤖 Says "Runs `log_winning_protocols`" but the helper below now calls `record_winning_protocols`.

---

fynd-core/src/simulation/simulator.rs
```diff
@@ -251,35 +287,71 @@ impl QuoteSimulator {
...
+        Ok(token_overrides(sender, router, permit2, layout))
     }
 
-    async fn cached_positions(
+    /// The token's storage layout, discovering it on the first quote that needs it.
+    ///
+    /// A token whose layout this build cannot resolve is remembered as unresolvable: the verdict
+    /// is a property of the token, and rediscovering it would spend a trace and up to 64
+    /// `eth_call`s on every quote that touches it. A node that failed to answer decides nothing,
+    /// so that is not remembered and the next quote tries again.
```
louise-poole: 🤖 `MAX_SLOTS_TO_VERIFY` is 32 slots *per traced account*, and `find_accessed_slot` loops over every account the prestate trace touched — once for the balance probe and once for the allowance probe. A proxy whose `balanceOf` reads three accounts costs up to 192 calls, not 64, so the number understates the cost the cache exists to avoid. The same sentence is repeated in the test.

State the real bound — `MAX_SLOTS_TO_VERIFY` per traced account per probe, in both places — or drop the figure. The point of the sentence is that rediscovery is expensive, which survives without it.

---

fynd-core/src/worker_pool_router/instrumentation.rs
```diff
@@ -246,8 +247,13 @@ fn improvement_bps(baseline_net: Option<f64>, net: Option<f64>) -> Option<f64> {
 /// `RUST_LOG=...,fynd::winning_protocols=trace`.
 const WINNING_PROTOCOLS_TARGET: &str = "fynd::winning_protocols";
 
-/// Logs the protocols the winning quote swaps on, how many swaps it makes on each, and how the
-/// quote simulated when the request asked for it.
+/// Hundredths of a basis point, the unit `winning_quote_shortfall_centibps_sum` accumulates in.
+const SHORTFALL_SCALE: f64 = 100.0;
```
louise-poole: Or switch to use a histogram metric? Or use a gauge to log the values and on the dashboard you can easily sum it over a period to get totals and count it to get the count?

---

🤖 `winning_quote_shortfall_centibps_sum` and `winning_quote_shortfall_count` are plain counters, but those two suffixes are exactly the derived series a histogram named `winning_quote_shortfall_centibps` would export. Anyone reading the `/metrics` page — or a recording rule that pattern-matches on the pair — reads them as a histogram that has lost its `_bucket` series. Every other counter in this crate ends in `_total`.

Rename to `winning_quote_shortfall_centibps_total` and `winning_quote_shortfall_routes_total`, and update the four assertions in `worker_pool_router/mod.rs`.

---

fynd-core/src/worker_pool_router/instrumentation.rs
```diff
@@ -279,6 +282,72 @@ pub(super) fn log_winning_protocols(quote: &OrderQuote) {
...
+fn count_swaps_per_protocol(swaps_per_protocol: &BTreeMap<&str, usize>, outcome: &'static str) {
+    let simulated = if outcome.is_empty() { "none" } else { outcome };
+    for (protocol, swaps) in swaps_per_protocol {
+        counter!(
+            "winning_quote_swaps_total",
+            "protocol" => protocol.to_string(),
+            "simulated" => simulated
+        )
+        .increment(*swaps as u64);
+    }
```
louise-poole, line 308: 🤖 `quote_simulations_total` labels outcomes `success`/`reverted`/`failed`; the new `winning_quote_swaps_total{simulated=...}` labels the same concept `success`/`failure`/`none`. The `failure` spelling was previously only log text — this PR promotes it to a metric label, so a dashboard correlating the two has to translate between them.

Use the established `reverted`/`failed`/`success` set for the `simulated` label. It also distinguishes a route that didn't execute from a simulation that didn't run, which the merged `failure` label throws away.

---

fynd-core/src/simulation/simulator.rs
```diff
@@ -3,38 +3,49 @@
...
-/// practical quote.
-const SIMULATION_FUNDING_VALUE: U256 = U256::MAX.wrapping_shr(1);
+/// `1e36`: a million billion tokens at 18 decimals, comfortably above any practical input.
+///
+/// The ceiling matters as much as the floor. A rebasing token derives its balance arithmetically
+/// -- stETH multiplies a holder's shares by the pooled-ETH rate -- and `U256::MAX >> 1` overflows
+/// that multiplication before the simulated transfer starts. Leaving the high bits clear also
+/// keeps a token that packs flags into its balance word reading the value back unchanged.
+const SIMULATION_FUNDING_VALUE: U256 =
+    uint!(1_000_000_000_000_000_000_000_000_000_000_000_000_U256);
```
louise-poole, line 51: 🤖 `1e36` at 18 decimals is `1e18` tokens — a billion billion, not "a million billion", which would be `1e33`.

---

fynd-core/src/simulation/token_layout.rs
```diff
@@ -0,0 +1,447 @@
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) struct TokenLayout {
+    storage_contract: Address,
+    balance: MappingPosition,
+    allowance: MappingPosition,
+}
+
+impl TokenLayout {
+    /// Creates a layout from known positions. [`discover_layout`] resolves them from a live token;
+    /// this builds one directly, which is what the tests need.
```
louise-poole: 🤖 "this builds one directly, which is what the tests need" — `discover_layout` calls it at line 167 too. Cut the sentence and leave "Creates a layout from known positions."

---

fynd-core/src/worker_pool_router/instrumentation.rs
```diff
@@ -279,6 +282,72 @@ pub(super) fn log_winning_protocols(quote: &OrderQuote) {
...
+/// Totals the shortfall of the routes that crossed each protocol, and how many did.
+///
+/// A sum and a count rather than a histogram: the mean is what the per-protocol view reports, and
+/// bucketing it would multiply the deviation histogram's boundaries across every protocol. The
+/// figure is one per quote, so a route crossing two protocols reports the same shortfall to both.
+///
+/// The sum is in hundredths of a basis point because a counter only takes whole numbers, and a
+/// shortfall under one basis point is the ordinary case: rounding to whole bps would report most
+/// routes as perfect. Divide the two by `SHORTFALL_SCALE` to read a mean in bps.
+fn record_deviation_per_protocol(swaps_per_protocol: &BTreeMap<&str, usize>, bps: f64) {
```
louise-poole: 🤖 The body converts the deviation into a shortfall and both metrics it writes are named `shortfall`; rename it `record_shortfall_per_protocol` so it matches its siblings and its output.

---

fynd-core/src/simulation/token_layout.rs
```diff
@@ -0,0 +1,447 @@
...
+        let vyper = MappingPosition::Direct { base, key_order: KeyOrder::Vyper };
+        if allowance_slot(owner, spender, vyper) == slot {
+            return Some(vyper);
+        }
+    }
+    let oz = MappingPosition::OpenZeppelinV5Allowances;
+    (allowance_slot(owner, spender, oz) == slot).then_some(oz)
+}
+
+fn mapping_slot(holder: Address, position: MappingPosition) -> B256 {
```
louise-poole: 🤖 Its sibling is `allowance_slot` and its callers are `TokenLayout::balance_slot` and `recover_balance_position`; name it `balance_slot` so the pair reads as a pair.

---

fynd-core/src/simulation/token_layout.rs
carloszanella: Done, the full version. `erc20_slots` is deleted and `token_layout` is the single implementation; the constants, the sentinel protocol and the override helper moved across with their doc comments. The tool crates move onto the traced path too, so they gain USDT and proxy support: they now write to `layout.storage_contract()` rather than assuming the token holds its own state.

Addressed in f7cd184b.

fynd-core/src/simulation/token_layout.rs
carloszanella: Collapsed to one `OpenZeppelinV5`. The namespace now follows from the mapping being addressed, so `balance_slot` reads the balances one and `allowance_slot` the allowances one, and the two impossible combinations are gone from the type.

Addressed in f7cd184b.

fynd-core/src/simulation/token_layout.rs
carloszanella: Made it the fallback you describe, so no address list. `balanceOf` is traced first and `sharesOf` retried when the observed slot cannot be placed, which covers the family rather than one member. A refused trace still ends discovery, so a node failure does not send the caller on to a view the token may not have.

Addressed in 78c840fc.

fynd-core/src/simulation/simulator.rs
carloszanella: Agreed, and it subsumes the trace-budget comment. `simulate_with_overrides` takes the request budget and runs both calls itself, so `Replay`, its box and `RevertedUntraced` are gone and the caller maps three outcomes. The call travels as one `SimulatedCall` rather than four parameters.

Addressed in 8465c670.

fynd-core/src/worker_pool_router/instrumentation.rs
carloszanella: Done as described: the outcome function returns `Option<f64>`, that number feeds the shortfall counters, and the `format!` moved inside `log_winning_protocols` after its `enabled!` gate. `simulated_deviation_bps` is gone.

Addressed in 1ffaa55d.

fynd-core/src/worker_pool_router/instrumentation.rs
carloszanella: Dropped, along with the `Route` import. The helper derives the swap count from the quote it already has, after its level check.

Addressed in 1ffaa55d.

fynd-core/src/simulation/revert.rs
carloszanella: Agreed the right fix is upstream, and I have not done that here. What this commit does is soften the doc -- it no longer promises "every" error, and says the set is a snapshot of the pinned version that goes stale silently -- and add `test_router_error_selectors_match_the_contracts`, which pins four selectors against values taken from the contracts' own signatures rather than from this declaration. That catches a drifted argument type. A newly added error still needs the list regenerating on a version bump. Happy to raise the ABI export with the tycho-execution team if you want it chased.

Addressed in 674483f8.

fynd-core/src/simulation/simulator.rs
carloszanella: Both helpers are gone. Holding one `OnceCell` per token collapsed the nesting into the two cases that mean something: `Ok(Err(reason))` remembers an unsupported token, `Err(..)` leaves the cell empty so a node failure is retried.

Addressed in 0abbf822.

fynd-core/src/simulation/token_layout.rs
carloszanella: Confirmed and fixed. `slot_matches` now separates a revert -- geth's code 3, or a message naming a revert -- from every other error response, and reports the rest as `DiscoveryError::Rpc`, which is not remembered. Two mocked tests cover it: all-revert reaches `Unsupported`, a throttled response reaches `Rpc`.

Addressed in 5c499772.

fynd-core/src/simulation/simulator.rs
carloszanella: Taken. The trace gets a fixed 500ms of its own, and it now runs inside `simulate_with_overrides` rather than after it, so the worst case for a reverting quote drops from two request budgets to one plus 500ms.

Addressed in 8465c670.

fynd-core/src/simulation/simulator.rs
carloszanella: Done. One `OnceCell` per token, so concurrent callers await the same discovery. It also replaced the cache's own structure, which is what let `cached_verdict` and `remember` go.

Addressed in 0abbf822.

fynd-core/src/simulation/simulator.rs
carloszanella: Good catch. The envelope now carries the block the quote was solved against, advanced by one block and twelve seconds, and `block_overrides` sets `number` and `time` from it, so both calls report the same environment. A caller with no quote leaves them unset.

Addressed in 04bb1ce9.

fynd-core/src/simulation/token_layout.rs
carloszanella: Folded into one `recover_position` taking the slot function, which is the `probe_position` shape you point at.

Addressed in f7cd184b.

fynd-core/src/worker_pool_router/mod.rs
carloszanella: Put back on the parameter, rebinding deleted.

Addressed in 918d4ae8.

fynd-core/src/simulation/deviation.rs
carloszanella: Dropped to `pub(super)`. `quote_with_fees` is the only one that crosses a boundary and stays `pub(crate)`.

Addressed in c2eee0ca.

fynd-core/src/simulation/simulator.rs
carloszanella: Done -- the reason is resolved in the match and the prefix applied once.

Addressed in 8465c670.

fynd-core/src/simulation/token_layout.rs
carloszanella: Added, all four you list. `find_accessed_slot` is covered by mocked traces for: a second candidate matching, every probe reverting (asserting `Unsupported`), a probe refused (asserting `Rpc`), and a trace failure. The split-storage rejection and the shares fallback are covered too, in f7cd184b and 78c840fc. The live test stays `#[ignore]`d, but it is no longer the only coverage.

Addressed in 5c499772.

fynd-core/src/simulation/simulator.rs
carloszanella: Added both. One mocked run answers the simulate call with a reverted, empty-payload result and the trace with a frame carrying a router error, asserting the reason names it; the other fails the trace and asserts the node's own message survives.

Addressed in 8465c670.

fynd-core/src/tests/simulation/simulator.rs
carloszanella: Rewritten to drive `cached_layout` through mocked discovery instead of inserting into the cache. Three tests now: a resolved layout is reused, an unsupported verdict is remembered, and a node failure is retried on the next call.

Addressed in 0abbf822.

fynd-core/src/tests/simulation/simulator.rs
carloszanella: Dropped the restatement, kept the ceiling with a comment saying why it matters, and renamed to `test_funding_value_bounds`.

Addressed in 117bea17.

fynd-core/src/tests/simulation/simulator.rs
carloszanella: Moved to `src/tests/metrics.rs`, reached by a crate-level `tests` module, and both call sites point at it. The row type is named so `counter_value` does not spell the triple out.

Addressed in f3490bcf.

fynd-core/src/tests/simulation/erc20_slots.rs
carloszanella: Resolved by sharing rather than covering twice: there is one set of constants now, and the ERC-7201 derivation check moved into the `token_layout` tests along with the known-good keccak vectors.

Addressed in f7cd184b.

fynd-core/src/worker_pool_router/instrumentation.rs
carloszanella: Real bug, fixed. Both counters are skipped when the deviation is not a shortfall. To your question: no, the comment no longer holds -- the positive-slippage capture exempts a signer that disables slippage taking, and the pool can move in the user's favour between the quote and the simulation, so a favourable result is reachable. A test asserts it records no shortfall while still counting as a successful simulation.

Addressed in bf1a289f.

fynd-core/src/simulation/simulator.rs
carloszanella: Raised to 5s and renamed to `SIMULATION_LAYOUT_DISCOVERY_TIMEOUT`, and the probes are now capped across all traced accounts rather than per account, so the worst case is bounded. To your question: over six hours on dev the timeout branch fired 6 times in 1168 simulation outcomes, all inside the warm-up window after a restart.

Addressed in 6bc4b4d8.

fynd-core/src/simulation/revert.rs
carloszanella: Collapsed the four that differ only in tree shape into `rstest` cases, each named after the rule it pins. The three that assert something else -- no revert, a plain reason, a hex reason -- stay standalone.

Addressed in 415b5ced.

fynd-core/CLAUDE.md
carloszanella: Rewritten to describe each module rather than the diff. `erc20_slots` is gone entirely, so the cell now names what `token_layout`, `simulator`, `revert` and `deviation` each are.

Addressed in f7cd184b.

fynd-core/src/worker_pool_router/mod.rs
carloszanella: Fixed.

Addressed in 64325b3d.

fynd-core/src/simulation/simulator.rs
carloszanella: Fixed by making the bound true rather than by restating it: the candidates are collected across every traced account and then capped, so the figure is the whole probe's cost. The sentence in the test was rewritten with it.

Addressed in 6bc4b4d8.

fynd-core/src/worker_pool_router/instrumentation.rs
carloszanella: Renamed to `winning_quote_shortfall_centibps_total` and `winning_quote_shortfall_routes_total`, and the four assertions with them. Not switching to a histogram: bucketing per protocol multiplies the deviation histogram's boundaries across every venue, which is why this is a sum and a count. A gauge would lose the aggregation across replicas.

Addressed in 249103ad.

fynd-core/src/worker_pool_router/instrumentation.rs
carloszanella: Adopted the outcome counter's words. One caveat: `reverted` is not available at this point -- `SimulationResult` merges a revert and a failure into one variant, so splitting them is a wire-type change. The label is `success`/`failed`/`none`, which at least removes the translation.

Addressed in 249103ad.

fynd-core/src/simulation/simulator.rs
carloszanella: Fixed -- it reads "a billion billion" now. Disclosure: this one-line doc fix rode into the block-environment commit rather than standing on its own. I noticed too late to split it without rewriting six commits.

Addressed in 04bb1ce9.

fynd-core/src/simulation/token_layout.rs
carloszanella: Cut, leaving "Creates a layout from known positions."

Addressed in f7cd184b.

fynd-core/src/worker_pool_router/instrumentation.rs
carloszanella: Renamed to `record_shortfall_per_protocol`.

Addressed in 249103ad.

fynd-core/src/simulation/token_layout.rs
carloszanella: Renamed to `balance_slot`, so the pair reads as a pair.

Addressed in f7cd184b.

louise-poole (approved):
Looks good! Thank you!

PR #522, opened 2026-09-04
fix(core)!: calculate price impact for split-route quotes

tvinagre:
## Summary

- The worker now calculates `price_impact_bps` for linear routes and split routes.
- The calculation uses token metadata from `Route` and the spot price that each swap state reports for `(token_in, token_out)`.
- The calculation processes split routes as token-flow graphs. It groups swaps by input token and, for validated routes, visits the groups in producer-before-consumer order.
- The calculation does not read shared `MarketState`.
- A calculation failure does not fail the quote. The worker omits `price_impact_bps`, logs the reason at debug level, and increments `worker_pool_price_impact_calculations_total{pool,outcome}`.
- `worker_pool_price_impact_duration_seconds{pool}` records the calculation time.
- Fynd continues to call `ProtocolSim::spot_price(token_in, token_out)` for compatibility. Tycho implementations use different direction and fee conventions, which can add a venue-dependent fee bias to `price_impact_bps`.
- Path Frank-Wolfe uses `estimate_probe_impact` only to decide whether the split search should continue and to size its next probe. The worker does not use this heuristic as the quote metric.
- This breaking change removes `RouteResult::with_price_impact`, the `RouteResult` price-impact field, and its internal accessor from `fynd-core`.
- Tests cover linear and split routes, token decimals, missing token data, spot-price failures, and metric labels.


fynd-core/src/types/quote.rs
```diff
@@ -1331,26 +1331,12 @@ pub struct RouteResult {
     net_amount_out: BigInt,
     /// Effective gas price (in wei) at the time the route was computed.
     gas_price: BigUint,
```
tvinagre, line 1566: @carloszanella this was a conscious decision to remove this from the interface, since it's useless rn. Pls lmk if you think it's a bad decision so I can rollback & mark this as deprecated

fynd-core/src/worker_pool/price_impact.rs
```diff
@@ -1,175 +1,472 @@
...
+    NoIdealOutput,
+}
+
+/// One swap as the price-impact walk sees it.
+pub(crate) struct SpotLeg<'a> {
+    pub token_in: &'a Address,
+    pub token_out: &'a Address,
+    /// Raw units of `token_in` the swap consumed.
+    pub amount_in: f64,
+    /// Human units of `token_out` per human unit of `token_in` before the trade, fee included.
```
tvinagre: Worth noticing that the `spot` calculation *should*, by definition, include fee, but this is not guaranteed and there might be protocols that exclude the fee from the calculation, skewing the price impact calculation. 
If this happens, needs to be fixed on Tycho integration, not here.

fynd-core/src/worker_pool/price_impact.rs
```diff
@@ -1,175 +1,472 @@
...
-    // Reject anything that is not a simple linear chain.
-    for pair in swaps.windows(2) {
-        if pair[0].token_out() != pair[1].token_in() {
-            return None;
-        }
-    }
+    let (Some(first), Some(last)) = (swaps.first(), swaps.last()) else {
+        return Err(PriceImpactError::EmptyRoute);
+    };
+    let view = market
```
tvinagre: Holding the lock here seems a very bad design. I'll take a deeper look

fynd-core/src/worker_pool/worker.rs
```diff
@@ -473,7 +492,7 @@ where
                             self.route_carries_no_swaps()
                         })?
                 };
```
Troshchk, line 626: This was the most concerning for me, pls, check if this is expected

**[IMPL `#1`]**

**Category**: bug

**Description**: For a buy order the worker sets `amount_in_raw` to `route.swaps().first().amount_in()`, then passes it to `route_price_impact` as the route's total input. A split route's first branch collection holds several swaps that all consume the input token, so `.first()` captures only one branch and understates the total. The walk seeds `reference_human_by_token[token_in]` with this understated amount, so the reference output — and the resulting price impact — is wrong. Before this change split routes took their price impact from the algorithm's path allocations, not from this value, so the diff newly routes buy split routes through it. `PathFrankWolfeAlgorithm::optimize_split` does not gate on order side, so a buy order can produce a split route.

**Plain English**: An integrator submits a buy order that the solver fills by splitting the input across several pools in parallel. The quote comes back with a price-impact figure computed as if only one of those parallel legs had been funded, so the number is inflated and does not describe the trade the integrator is actually getting. A split buy should measure impact against the full amount going into all parallel legs.

**Suggestion**: For the buy branch, sum the input across the route's first branch collection (the swaps whose `token_in` is the route input token) rather than taking the first swap. The same value feeds the quote's reported `amount_in`, so this is worth confirming independently of price impact.

**Scope**: The `.first()` logic is pre-existing (the diff only renamed `amount_in` to `amount_in_raw`); the diff makes it worse by feeding it into split-route price impact. I found no `OrderSide::Buy` coverage in `fynd-core/src`, so I could not confirm buy split routes occur in practice.

fynd-core/src/worker_pool/price_impact.rs
```diff
@@ -1,175 +1,775 @@
...
+}
+
+/// Price impact of a route from each swap's reported spot price.
+///
+/// Token decimals come from the route's own token map, which every built-in algorithm fills in
+/// for every swap, so the calculation touches no shared market state.
+pub(crate) fn route_price_impact(
+    route: &Route,
+    amount_in_raw: &BigUint,
+    amount_out_raw: &BigUint,
```
Troshchk, line 250: **[IMPL `#2`]**

The `f64` bound inside the swap loop shares the name of the `&BigUint` parameter used later for `PriceImpactInputs`. Rename the loop binding (e.g. `swap_amount_in_raw`) so the raw-float leg amount and the route-level input are not the same identifier.

fynd-core/src/worker_pool/price_impact.rs
```diff
@@ -1,175 +1,775 @@
...
+
+    for (token_in, collection) in branch_collections(legs, |leg| leg.token_in) {
+        let available_reference_input_human = reference_human_by_token
+            .get(&token_in)
+            .copied()
+            .ok_or_else(|| PriceImpactError::UnfedToken(token_in.clone()))?;
+        let consumed_input_raw_total = consumed_raw_by_token
+            .get(&token_in)
+            .copied()
+            .unwrap_or_default();
```
Troshchk: **[SIMPLIFY `#1`]**

**Description**: `spot_reference_output` builds an `FxHashMap<&Address, f64>` of raw input summed per token in a first pass, then reads it back inside the `branch_collections` loop. But `branch_collections(legs, |leg| leg.token_in)` already groups every leg for a token into one `collection`, so the per-token total is just the sum over that collection. The map is a second, eager copy of the same grouping the walk already has, and the reader must reconcile two ways of grouping by `token_in`.

**Suggestion**: Delete the `consumed_raw_by_token` map and its build loop; compute `let consumed_input_raw_total: f64 = collection.iter().map(|leg| leg.amount_in_raw).sum();` at the top of each collection iteration. Removes a map allocation, a loop, and the `unwrap_or_default()` lookup (~8 lines). The `<= 0.0` guard and `ZeroConsumedInput` behavior are unchanged; nothing is lost.

fynd-core/src/worker_pool/worker.rs
```diff
@@ -492,22 +511,39 @@ where
...
                 let mut quote = OrderQuote::new(
                     order.id().to_string(),
                     QuoteStatus::Success,
-                    amount_in,
-                    amount_out,
+                    amount_in_raw,
+                    amount_out_raw,
                     gas_estimate,
                     amount_out_net_gas,
                     block_info.clone(),
```
Troshchk, line 677: **[TESTS `#1`]**

**Description**: The diff rewrites the worker so a failed price-impact calculation logs, counts the outcome, omits `price_impact_bps`, and still returns a successful quote. Only the success path is asserted (`test_quote_price_impact_for_split_route`, `price_impact_bps() == Some(400)`); no worker test drives a route whose calculation fails and asserts the quote still succeeds with `price_impact_bps() == None`. This is the core guarantee the change establishes.

**Suggestion**: Add a worker test with a mock algorithm returning a route whose token map is empty (or whose spot price errors), then assert `worker.quote(...)` is `Ok` and `quote.order().price_impact_bps()` is `None`.

fynd-core/src/worker_pool/price_impact.rs
```diff
@@ -1,175 +1,775 @@
...
+        if consumed_input_raw_total <= 0.0 {
+            return Err(PriceImpactError::ZeroConsumedInput(token_in));
```
Troshchk, line 181: **[TESTS `#2`]**

**Description**: `ZeroConsumedInput` (a branch collection consuming zero raw input) and `NoReferenceOutput` (the reference output not being positive and finite) are both reachable through `price_impact_from_spot_legs`, which the tests already call directly, but neither has a test.

**Suggestion**: Add two `price_impact_from_spot_legs` cases — one with a leg whose `amount_in_raw` is `0.0` to hit `ZeroConsumedInput`, and one whose legs never reach `inputs.token_out` to hit `NoReferenceOutput` — each destructuring the returned error variant.

fynd-core/src/worker_pool/price_impact.rs
```diff
@@ -1,175 +1,775 @@
...
 
 use num_bigint::BigUint;
+use num_traits::ToPrimitive;
+use rustc_hash::FxHashMap;
+use tycho_simulation::tycho_common::models::{token::Token, Address};
+
+use crate::types::{quote::branch_collections, ComponentId, Route, Swap};
 
-use crate::{feed::market_data::MarketData, types::Route};
+/// Identifies an amount that cannot be represented as a finite `f64`.
```
Troshchk: **[DOCS `#1`]**

The doc reads "Identifies an amount that cannot be represented as a finite `f64`", but the enum (`RouteInput`/`RouteOutput`/`SwapInput`) names *which* amount a value is, and its `Display` output ("route input", …) never mentions `f64`. Reword to the concept, e.g. "Which amount in the price-impact calculation a value refers to."

fynd-core/src/worker_pool/price_impact.rs
```diff
@@ -1,175 +1,775 @@
...
+            },
+        ];
+        let (amount_in, amount_out) = (parse_biguint("1000"), parse_biguint("980"));
+        let impact_inputs = inputs(&a, &c, &amount_in, &amount_out);
+        let impact = price_impact_from_spot_legs(&legs, &impact_inputs).unwrap();
+        assert!((impact - 0.02).abs() < 1e-9, "got {impact}");
+    }
+
+    #[test]
+    fn test_favorable_execution_is_negative() {
```
Troshchk, line 381: **[SLOP `#2`]**

`is_negative` names the assertion, not the scenario; rename `test_favorable_execution_is_negative` to `test_favorable_execution`.

fynd-core/src/worker_pool/price_impact.rs
```diff
@@ -1,175 +1,775 @@
...
     }
 
     #[test]
-    fn absurdly_large_amount_does_not_overflow_to_garbage() {
-        // Rust parses out-of-range floats to Ok(inf), not Err — the is_finite guards
-        // must catch that and yield None rather than a bogus impact.
-        let huge = "1".to_string() + &"0".repeat(400); // ~1e400 > f64::MAX
-        let pi = price_impact_from_spot_product(1.0, &bu(&huge), &bu("1"), 0, 0);
-        assert_eq!(pi, None);
+    fn test_no_slippage_sell_rate_pool_reports_zero_impact() {
```
Troshchk, line 819: **[SLOP `#3`]**

`reports_zero_impact` names the expected result; rename `test_no_slippage_sell_rate_pool_reports_zero_impact` to `test_no_slippage_sell_rate_pool`.

fynd-core/src/worker_pool/worker.rs
```diff
@@ -1702,16 +1827,16 @@ mod tests {
     }
 
     #[test]
-    fn solve_duration_metric_recorded() {
+    fn quote_duration_metric_recorded() {
```
Troshchk: **[SLOP `#4`]**

The rename dropped the project's `test_` prefix that its sibling `test_price_impact_metrics_recorded` keeps; rename `quote_duration_metric_recorded` to `test_quote_duration_metric_recorded`.

Troshchk (commented):
Thank you!

fynd-core/src/worker_pool/worker.rs
```diff
@@ -628,22 +643,39 @@ where
...
+                    Ok(_) => "computed",
+                    Err(err) => err.outcome(),
+                };
+                record_price_impact_metrics(
+                    &self.pool_name,
+                    price_impact_started.elapsed(),
+                    price_impact_outcome,
+                );
+                let price_impact_bps = match price_impact_result {
+                    Ok(impact) => Some((impact * 10_000.0).round() as i32),
```
tamaralipows: Maybe double check this is intended pls?

---

**[IMPL `#1`]**

**Category**: bug

**Description**: `spot_reference_output` only checks that the reference is finite and `> 0.0`, so a reference of, say, `1e-300` (a venue reporting a near-zero spot price, or a token whose decimals are wrong) yields `impact = 1.0 - executed/reference` in the range of `-1e300`. `(impact * 10_000.0).round() as i32` is a saturating cast, so the quote carries `price_impact_bps = -2147483648` with outcome `computed`, and nothing downstream clamps or rejects it (`fynd-rpc-types/src/lib.rs:1862` passes it straight through).

**Plain English**: An integrator asks for a quote and gets back a price impact of minus two billion basis points, which reads as an impossibly favorable trade. Nothing in the response marks the number as suspect, and the operator's dashboards count the calculation as successful, so the only way to notice is to eyeball the number. The same condition on the other side is harmless, because a bad reference in that direction tops out at ten thousand basis points. A price impact outside a plausible range should be treated the same as a calculation failure: the field is left off the quote and the outcome is counted under its own reason.

**Suggestion**: Bound the result before the cast — reject anything outside a plausible range (for example below `-1.0` or above `1.0`) with a new `PriceImpactError` variant so it gets its own `outcome` label, instead of publishing a saturated `i32`.

fynd-core/src/worker_pool/price_impact.rs
```diff
@@ -1,175 +1,804 @@
...
+//!
+//! # Limitation: the reference is the reported spot price, not a marginal output rate
+//!
+//! Tycho documents `ProtocolSim::spot_price(base, quote)` as a fee-marked-up buy price, but its
+//! implementations do not use a consistent direction or fee convention. Fynd deliberately keeps the
+//! executed-direction call it has always made, `spot_price(token_in, token_out)`, so this module
+//! extends the existing behaviour to split routes without reversing it for any venue. The result
+//! can therefore carry a venue-dependent fee bias until Tycho exposes a consistent
+//! execution-direction marginal-output rate; [`reported_spot_price`] is the one place to adjust
+//! when it does.
```
tamaralipows: **[DOCS `#2`]**

**Description**: "Fynd deliberately keeps the executed-direction call it has always made", "so this module extends the existing behaviour to split routes without reversing it for any venue", and "`reported_spot_price` is the one place to adjust when it does" describe the history of the decision and a hypothetical future edit, not what the code does. The project rule bans narration and references to prior implementations in doc-comments.

**Suggestion**: Replace with a plain statement of current behavior: "Fynd calls `ProtocolSim::spot_price(token_in, token_out)` in the executed direction. Tycho implementations differ in direction and fee convention, so `price_impact_bps` can carry a venue-dependent fee bias." Drop the sentence naming which function to change later.

fynd-core/src/worker_pool/price_impact.rs
```diff
@@ -1,175 +1,804 @@
...
-pub(crate) fn spot_price_impact(
-    route: &Route,
-    amount_in: &BigUint,
-    amount_out: &BigUint,
-    market: &MarketData,
-) -> Option<f64> {
-    let swaps = route.swaps();
-    if swaps.is_empty() {
-        return None;
+/// The walk seeds the input token with the route's input and visits the legs one branch
```
tamaralipows: **[DOCS `#6`]**

`SpotLeg`'s doc (line 97), `price_impact_from_spot_legs`'s doc (line 121) and `branch_collections`'s doc (`quote.rs:1771`) all refer to "the price-impact walk", but no identifier is called a walk — the function is `spot_reference_output`. Name the function instead.

tamaralipows (approved):
Thank you! Only minor things so approved

EDIT: Ugh I just realized it posted the first docs comment in this main comment. Sorry about that. Your claude will know where it is. 

---

**[DOCS `#1`]**

**Description**: The diff renames the Frank-Wolfe heuristic to `estimate_probe_impact`/`probe_impact` precisely so it stops reading as the quote's price impact, but the published algorithm doc still says "price impact estimate" (line 26), "This is `gas_cost / price_impact`" (line 42) and "price impact exit" (line 81). The public doc now names a quantity the code no longer has.

**Suggestion**: Change those three lines to "probe impact estimate", `gas_cost / probe_impact`, and "probe-impact exit", and add one sentence saying this estimate is not `price_impact_bps`.

fynd-core/src/worker_pool/worker.rs
tvinagre: Hmm - not sure if we should make assumptions of a range, but probably prevent from having this saturation, if it comes from an overflow or a calculation bug.

PR #523, opened 2026-09-05
fix!: propAMM graph admission

carloszanella:
A pAMM reaches the chain only through a Uniswap V3 fallback pool. With no such pool in the market,
the worker routed through the pAMM anyway, could not price the leg, and rejected the whole order — a
worker returns one route, not a list. With a single worker pool that order came back
`NO_ROUTE_FOUND`, so a user saw "no route" for an order the market could serve. The capture logged
`route/pamm_fallback_pool_missing`.

Two bugs behind it.

### The guard ran only when the graph was built

`drops_component_on_build` was applied in one place. Market events were filtered by a rule that
carries no pAMM check, and the graph was rebuilt once, the first time the router's fee tiers
arrived. So an unbacked pAMM reached the graph four ways: arriving in a later block, its fallback
pool leaving, a fee tier moving, or the window before the first tier read, when the check was
skipped and every pAMM was admitted.

Now the event path applies the same rule as the build, a pAMM whose fallback pool leaves is removed,
no pAMM is admitted before the tiers are read, and a tier change rebuilds the graph — a tier decides
which pool backs a pAMM.

### The rule only ever removed

A pAMM kept out because its fallback pool did not exist yet was then forgotten. The market names a
component in `added_components` once, so the pool arriving later put nothing back: that pAMM stayed
out until the process restarted. Worse than before the rule existed, when it started routing as soon
as the pool appeared.

The pAMMs kept out are now tracked beside the ones let in, and one is let back when its pool
arrives. Each decision is counted on `propamm_admissions_total{outcome,pool}`.

### Refactor

The rule was spread over five worker fields and two methods, with a call order nothing enforced: the
pool index had to read a market event before the admission pass did, and only a comment said so.
`PammAdmission` now owns the tiers, the pool index and the two sets, and decides pAMMs and nothing
else — one method per step. No behaviour change and no public API change.

### Breaking

`propamm_fallback::lacks_fallback_pool` is now `is_unbacked_pamm` and takes `Option<&FeeTiers>`
instead of `&FeeTiers`, so an unread tier answers "unbacked" rather than skipping the check. The
module is public; nothing in this workspace called it.


fynd-core/src/propamm_fallback/admission.rs
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    /// Updates the pool index from one market event, as the market broadcast it.
+    ///
+    /// Unfiltered, because the index describes the market rather than one worker's graph, and
+    /// before [`select_pamm_updates`](Self::select_pamm_updates), so a pAMM arriving in the same
+    /// block as its pool is judged against an index that holds it.
+    pub(crate) fn update_pools(&mut self, market: &MarketDataView<'_>, event: &MarketEvent) {
+        self.pools.apply_event(market, event);
+    }
```
louise-poole: This idea feels cleaner.

---

🤖 `PammAdmission` still has to be driven in a sequence nothing enforces: `rebuild_pools` → `remove_components` → `record_build` on the build path, and `update_pools` → `filter_event` → `select_pamm_updates` on the event path, ordered only by the doc comments at `admission.rs:86`, `120–122` and `134–136`. That is the same contract the PR body says the refactor removed, moved from worker comments into `PammAdmission`'s. The test helper `admit` (`worker.rs:1374–1392`) hand-rolls the event sequence and leaves `filter_event` out, so the order the tests exercise and the order production runs can drift apart silently. The middle step also matters: the early return at `admission.rs:146` reads the *filtered* event, so a worker whose own filter empties both lists skips readmission for that block even though the pool index already moved. Maps to the `design` and `document-assumptions` directives.

Let `PammAdmission` own both whole steps and take the worker's predicate as a `&dyn Fn(&ProtocolComponent) -> bool`: `filter_topology(market, topology, drop) -> kept` (index rebuild + filter + record) and `apply_event(market, event, drop) -> MarketEvent` (index update + filter + select). Six `pub(crate)` methods become two, no caller can misorder them, and the tests drive the real sequence. If that's more surgery than you want here, then either way the PR description shouldn't claim the ordering problem is gone.

fynd-core/src/propamm_fallback/manager.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    /// The tiers as they stand, `None` before `FeeTierFetcher` reads the router.
+    pub(crate) fn fee_tiers(&self) -> Option<FeeTiers> {
+        self.fee_tiers.snapshot()
+    }
+
+    /// The pools a pAMM leg can fall back to, for pricing a route that holds one.
+    pub(crate) fn pools(&self) -> &FallbackPoolIndex {
+        &self.pools
+    }
```
louise-poole, line 72: Up to you...

---

🤖 `fee_tiers()` and `pools()` exist for the quote path, not for admission — `quote` reads both to price a fallback. The type therefore owns the worker's pAMM market state as well as the admission rule, which is not the "decides pAMMs and nothing else" the PR body describes.

Name and document it for what it holds — the tiers, the pool index and the admitted/withheld record — rather than for the one decision it makes. `PammFallbackState` with an `admission` role reads truthfully; holding the tiers and the index together is worth keeping, since both must describe the same market read.

fynd-core/src/propamm_fallback/admission.rs
```diff
@@ -0,0 +1,247 @@
...
+/// hold different graphs, because they exclude different components.
+pub(crate) struct PammAdmission {
+    /// Fee tiers the PropAMMRouter falls back on, read from chain by `FeeTierFetcher`.
+    fee_tiers: SharedFeeTiers,
+    /// Uniswap V3 pools the PropAMMRouter can fall back to, kept current from market events.
+    pools: FallbackPoolIndex,
+    /// The pAMMs the graph holds.
+    admitted: FxHashSet<ComponentId>,
+    /// The pAMMs the graph left out, because the market did not back them.
+    withheld: FxHashSet<ComponentId>,
```
louise-poole: 🤖 `admitted` and `withheld` encode one per-pAMM state across two collections and nothing keeps them disjoint. The `retain` at line 181 inserts into `admitted` without removing from `withheld`, so a withheld pAMM re-announced in `added_components` once it is backed lands in both, and the readmission pass at 186–200 finds it again and increments `outcome=admitted` a second time for one decision.

One `FxHashMap<ComponentId, PammState>` with `Admitted` / `Withheld` variants. That matches the project's enums-over-parallel-collections rule and makes double membership unrepresentable.

fynd-core/src/propamm_fallback/admission.rs
```diff
@@ -0,0 +1,247 @@
...
+            }
+            let set = if kept.contains_key(component_id) {
+                &mut self.admitted
+            } else {
+                &mut self.withheld
+            };
+            set.insert(component_id.clone());
+        }
+        self.built_with_fee_tiers = fee_tiers;
+    }
```
louise-poole: Important finding - we'd potentially re-add a pool that was initially excluded due to another rule.

---

🤖 `record_build` sorts every pAMM in the market by whether it survived into `kept`, but `kept` was filtered by `should_drop_component` as well as the pAMM rule (`worker.rs:225–228`). A pAMM excluded because the pool sets `exclude_protocols = ["propammfallback:"]` — the commented example at `worker_pools.toml:37` — or because it is exclusive in a `PublicOnly` pool therefore lands in `withheld`. On the next event carrying any added or removed component, the readmit pass finds it backed, inserts it into `added_components` and hands it to the graph manager, past the `filter_event` that had already removed it. The worker then routes through liquidity its configuration forbids. Maps to the `correctness` and `design` directives.

An operator configures a worker pool to keep a whole venue family out of its routing, or restricts a pool to public liquidity only. The pool starts up correctly and skips those venues. Then on the next block that adds or removes any pool anywhere in the market, those venues quietly reappear in that pool's routing and start showing up in its quotes, with nothing in the logs saying so. An exclusion set in configuration should hold for the life of the pool regardless of what later blocks contain.

Make `withheld` mean only "kept out by the pAMM rule". Cheapest route: have the filter closure in `initialize_graph` record the decision as it goes — it already knows which of the two rules dropped the component — and drop `record_build`'s second pass over the topology entirely, which also resolves IMPL `#3`. Failing that, pass the build's `should_drop_component` predicate into `record_build` so those pAMMs end up in neither set.

fynd-core/src/propamm_fallback/admission.rs
```diff
@@ -0,0 +1,247 @@
...
+        for component_id in readmitted {
+            withheld.remove(&component_id);
+            admitted.insert(component_id.clone());
+            count("admitted");
+            let tokens = market
+                .get_component(&component_id)
+                .map(|component| component.tokens.clone())
+                .unwrap_or_default();
+            added_components.insert(component_id, tokens);
+        }
```
louise-poole: 🤖 The `unbacked` closure uses `is_some_and`, so it answers `false` for a withheld id the market no longer holds. That id passes the `!unbacked` filter, is counted `admitted`, and is inserted into `added_components` with the empty token list `unwrap_or_default` supplies — which `add_components` rejects as `InvalidComponents`. The market can advance between the broadcast and the worker's `read()`, so this is reachable. Each readmitted id also costs two `get_component` lookups. Maps to the `correctness` and `error-handling` directives.

A venue is held out of routing because the market holds nothing to execute it against. The venue then disappears from the market entirely, and the worker reads that disappearance as the venue becoming routable again. It tries to add a venue that no longer exists, the graph refuses it, and the operator gets an invalid-component error on that block while the admission metric records a success that never happened. A venue the market has dropped should stay held out.

Read the component once per withheld id and readmit only when it is both present and backed, which removes the empty-token fallback with it:

```rust
let readmitted: Vec<(ComponentId, Vec<Bytes>)> = withheld
    .iter()
    .filter_map(|component_id| {
        let component = market.get_component(component_id)?;
        (!is_unbacked_pamm(component, fee_tiers, pools))
            .then(|| (component_id.clone(), component.tokens.clone()))
    })
    .collect();
```

This changes behaviour — a withheld id the market has forgotten stops being readmitted — so confirm that's the intent.

fynd-core/src/propamm_fallback/manager.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    /// Whether `fee_tiers` has moved since the graph was filtered.
+    ///
+    /// A tier decides which pool a pAMM falls back to, so a change can make one routable or
+    /// unroutable, and only a rebuild can add a component back.
+    pub(crate) fn needs_rebuild(&self, fee_tiers: Option<&FeeTiers>) -> bool {
+        fee_tiers != self.built_with_fee_tiers.as_ref()
+    }
```
louise-poole, line 80: I wonder which is worse in this case... failing to re-fetch fee tiers = stop solving over propAMMs until they fetch again, or keep solving assuming the old tiers hold up.

---

🤖 `needs_rebuild` compares whole `FeeTiers` values, and `fetch_default_tier_and_pairs` skips a venue whose `getPairs` decode fails (`fee_tier_fetcher.rs:210`) while still returning `Ok` and publishing the result through `set`. So one flaky RPC round writes a `FeeTiers` missing that venue's per-pair overrides, which now forces a full topology rebuild on every worker thread and evicts that venue's pAMMs wherever the default tier names no pool; the next successful round flips both back. Before this diff a changed tier did nothing beyond route pricing, so that warn was harmless — it is load-bearing now. Maps to the `correctness` directive.

One venue's on-chain read fails for a round — a transient RPC hiccup. Every worker then throws away and rebuilds its whole routing graph, and that venue's pairs drop out of routing until the next successful read a timer later, so a user quoting one of those pairs sees a worse route or no route for no reason visible anywhere. Under a flapping endpoint this repeats every fetch round. A read that could not reach one venue should leave the previous tiers in place rather than publish an incomplete set as the new truth.

Make the venue skip an error so a partial read never reaches `SharedFeeTiers` and `refresh_once` leaves the stored tiers untouched, as its doc already promises for a failed refresh. If you'd rather not touch the fetcher in this PR, at minimum log at `info` when `needs_rebuild` fires, with the old and new default tier, so this is visible on the cluster.

**Scope**: The partial-read behavior in `fee_tier_fetcher.rs` is pre-existing; this diff is what gives it teeth.

louise-poole (commented):
The admission rework is the right shape, and both bugs it fixes are real. Two things I'd want resolved before this merges: the withheld set currently also holds pAMMs that this worker's own exclusion rules dropped, so the readmit pass can put back liquidity a pool is configured never to route through; and a partial fee-tier read now churns every worker's graph. The rest is smaller.

fynd-core/src/propamm_fallback/admission.rs
louise-poole: 🤖 `market.component_topology()` is not a getter — it allocates a fresh map, cloning every component id and token vector (`market_data.rs:388–393`). `record_build` builds one just to walk the keys, on top of the copy `initialize_graph` already made and the one `FallbackPoolIndex::build` makes, so a rebuild pays three full topology copies instead of two, with a `get_component` per id. Rebuilds are no longer once per startup. Maps to the `performance` directive.

Hand `record_build` the map `initialize_graph` already cloned — collect the pAMM ids from `topology` before it moves into `remove_components`, or take it by reference and filter afterwards.

fynd-core/src/propamm_fallback/admission.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// Whether `component` is a pAMM this market does not back, against the pools indexed here.
+    ///
+    /// `fee_tiers` is passed in so one pass over many components reads it once.
+    pub(crate) fn is_unbacked(
+        &self,
+        component: &ProtocolComponent,
+        fee_tiers: Option<&FeeTiers>,
+    ) -> bool {
+        is_unbacked_pamm(component, fee_tiers, &self.pools)
+    }
```
louise-poole: 🤖 One caller, and it already has `pools()`; call `is_unbacked_pamm(component, fee_tiers.as_ref(), pamm_admission.pools())` and delete the wrapper. Maps to the `simplify` directive.

fynd-core/src/propamm_fallback/admission.rs
```diff
@@ -0,0 +1,247 @@
...
+    ) {
+        self.admitted.clear();
+        self.withheld.clear();
+        for component_id in market.component_topology().keys() {
+            let is_pamm_component = market
+                .get_component(component_id)
+                .is_some_and(is_pamm);
+            if !is_pamm_component {
+                continue;
+            }
```
louise-poole: 🤖 `if !market.get_component(component_id).is_some_and(is_pamm) { continue; }` says the same thing in one line.

fynd-core/src/worker_pool/worker.rs
```diff
@@ -1251,76 +1258,322 @@ mod tests {
...
+        worker.process_event(event).await;
+
+        assert!(
+            worker
+                .pamm_admission
+                .admitted()
+                .contains(PAMM),
+            "the index must see the event before admission decides"
+        );
+    }
```
louise-poole, line 1514: 🤖 `market_for_admission(true)` puts both components in the market before `initialize_graph`, so `rebuild_pools` has already indexed the fallback pool. Delete the `update_pools` call from `process_event` and this test still passes: `update_pools` re-inserting the pool is a no-op, and `select_pamm_updates` would admit the pAMM just the same if the two calls were swapped. The `forget_for_test` in the middle is the tell — it clears the two sets but not the index, which is the thing the ordering claim is about. Maps to the `tests` directive.

Build the worker against a market holding neither component, initialize, then upsert both into the market and fire the event that adds both. Only the `update_pools`-first order admits the pAMM then, and `forget_for_test` is no longer needed.

fynd-core/src/worker_pool/worker.rs
```diff
@@ -1251,76 +1258,322 @@ mod tests {
...
+            .is_empty());
+        assert!(
+            worker
+                .pamm_admission
+                .withheld()
+                .contains(PAMM),
+            "kept, so it can come back"
         );
-        assert!(!worker.drops_component_on_build(&pamm, None));
     }
```
louise-poole, line 1435: 🤖 The test's market already holds the fallback pool, so one `admit(&mut worker, &market, added_component_event(PAMM))` admits the pAMM the way a block would, and the removal event can follow. Seeding `admitted` directly skips the step that proves the pAMM was in the graph in the first place.

Admit the pAMM through an added-component event, then fire the removal, and delete the `admit_for_test` accessor at `admission.rs:237–240`.

fynd-core/src/worker_pool/worker.rs
```diff
@@ -1251,76 +1258,322 @@ mod tests {
...
+            Some(&FeeTiers::new(ADMISSION_TIER))
+        );
+        assert!(
+            worker
+                .pamm_admission
+                .admitted()
+                .contains(PAMM),
+            "the rebuild admits the backed pAMM"
+        );
     }
```
louise-poole: 🤖 Every admission test runs on a worker whose `liquidity_scope` and `exclude_protocols` drop nothing, so nothing exercises the interaction between the two rules — which is where IMPL `#1` lives.

Add a case building a worker with `exclude_protocols = [PROPAMM_FALLBACK_PREFIX]` against a market holding the pAMM and its fallback pool, then assert the pAMM appears in neither the event's additions nor `admitted` after a later event.

louise-poole, line 1547: 🤖 `test_process_event_after_the_first_fee_tier_read` covers the `None` to `Some` transition only. The PR's claim that "a tier change rebuilds the graph" — and the eviction that follows when the new tier names no pool — has no test.

Set a tier the market holds a pool at, initialize, then `set` a different tier and fire an event; assert the rebuild happened and the pAMM moved to `withheld`.

fynd-core/src/worker_pool/worker.rs
```diff
@@ -1251,76 +1258,322 @@ mod tests {
...
+            .admitted()
+            .is_empty());
+        assert!(
+            worker
+                .pamm_admission
+                .withheld()
+                .contains(PAMM),
+            "kept, so it can come back"
         );
+    }
```
louise-poole: 🤖 `added_ids(&event).is_empty()` already proves the unbacked pAMM was not added; `admitted().is_empty()` on the next line is the internal mirror of it. Same pattern in the readmit test, where `admitted().contains(PAMM)` and `withheld().is_empty()` restate the `added_ids` assertion above them. Dropping these plus TESTS `#3`'s seeding leaves all five `#[cfg(test)]` accessors unused — about 25 lines of production surface that exists only for the tests, and that pins the two-set representation DESIGN `#3` wants to change.

Keep the assertion on the rewritten event and the one `withheld` assertion that shows the pAMM can come back; drop the rest, then delete the accessors at `admission.rs:222–246`. `admitted()` earns its place only after a rebuild, where the event is not observable.

fynd-core/src/worker_pool/worker.rs
```diff
@@ -1251,76 +1258,322 @@ mod tests {
...
+
+    /// A market holding the pAMM, plus the Uniswap V3 pool it falls back to at [`ADMISSION_TIER`]
+    /// when `with_fallback` is true.
+    fn market_for_admission(with_fallback: bool) -> MarketData {
+        let market = market_with_pamm();
+        if with_fallback {
+            add_fallback_pool(&market);
+        }
+        market
+    }
```
louise-poole: 🤖 Drop the boolean flag and call `market_with_pamm()` or a `market_with_pamm_and_fallback()` directly at each site.

fynd-core/src/propamm_fallback/manager.rs
```diff
@@ -0,0 +1,247 @@
...
+
+/// Which pAMMs one worker's graph holds, and the market facts that decide it.
+///
+/// Built for one worker, and only that worker reads it: two worker pools facing the same market
+/// hold different graphs, because they exclude different components.
+pub(crate) struct PammAdmission {
+    /// Fee tiers the PropAMMRouter falls back on, read from chain by `FeeTierFetcher`.
+    fee_tiers: SharedFeeTiers,
+    /// Uniswap V3 pools the PropAMMRouter can fall back to, kept current from market events.
+    pools: FallbackPoolIndex,
```
louise-poole: Got PTSD from 'pools' used by too many conflicting concepts lol.

---

🤖 `PammAdmission` has a `pools` field meaning Uniswap V3 fallback pools and a `pool_name` field meaning the worker pool — two senses of "pool" as neighbours, where the old worker field `fallback_pools` didn't have that problem. We've been hyperconscious about bare "pool" in this codebase for exactly this reason. Maps to the `naming` directive.

`fallback_pools` for the field and the accessor; `worker_pool_name` for the label field if you want to go the whole way.

fynd-core/src/worker_pool/worker.rs
```diff
@@ -1251,76 +1258,322 @@ mod tests {
...
+        worker
+            .pamm_admission
+            .select_pamm_updates(&view, fee_tiers.as_ref(), &mut event);
+        event
+    }
+
+    /// A pAMM whose pair this market holds no Uniswap V3 pool for can never produce a quotable
+    /// route, so it must not reach the graph.
+    #[test]
+    fn test_apply_pamm_admission_without_fallback_pool() {
```
louise-poole: 🤖 `test_apply_pamm_admission_without_fallback_pool` / `_with_fallback_pool` / `_before_fee_tiers` are the same three lines of setup with a different `(with_fallback, fee_tiers)` pair and a different expectation — needing the condition in the name to tell them apart is usually the signal for one `rstest` with cases. Maps to the `tests` directive.

One `#[rstest]` over `(with_fallback, fee_tiers, expect_admitted)` for those three. The two that drive a sequence of events (`_after_the_fallback_pool_leaves`, `_after_the_fallback_pool_arrives`) are genuinely different tests — keep them separate.

fynd-core/src/propamm_fallback/mod.rs
```diff
@@ -35,24 +36,40 @@ pub fn has_pamm_leg(route: &Route) -> bool {
...
+/// that the PropAMMRouter executes through a Uniswap V3 fallback pool.
+pub fn is_pamm(
     component: &tycho_simulation::tycho_common::models::protocol::ProtocolComponent,
-    fee_tiers: &FeeTiers,
-    index: &FallbackPoolIndex,
 ) -> bool {
-    if !component
+    component
         .protocol_system
         .starts_with(PROPAMM_FALLBACK_PREFIX)
```
louise-poole: 🤖 `propamm_fallback` is `pub mod` at the crate root, and the only caller of `is_pamm` is `admission.rs`. Make it `pub(crate)` rather than promising it — this PR is already breaking one thing in this module. Maps to the `yagni` directive.

fynd-core/CLAUDE.md
```diff
@@ -9,12 +9,12 @@ applications.
...
-| `worker_pool/`        | `WorkerPool` manages dedicated OS threads. `SolverWorker` runs a prioritized select loop (shutdown > market events > derived events > tasks). `TaskQueue` is `async_channel`-based. `SolverWorker::drops_component` decides what stays out of one worker's graph: exclusive components in a `PublicOnly` pool, plus every protocol system in the pool's `exclude_protocols` |
+| `worker_pool/`        | `WorkerPool` manages dedicated OS threads. `SolverWorker` runs a prioritized select loop (shutdown > market events > derived events > tasks). `TaskQueue` is `async_channel`-based. `SolverWorker::drops_component_or_unbacked_pamm` decides what stays out of one worker's graph: exclusive components in a `PublicOnly` pool, every protocol system in the pool's `exclude_protocols`, and every unbacked pAMM (`is_unbacked_pamm`). The first two hold for the worker's life; the pAMM rule expires, so `apply_pamm_admission` re-decides it on every market event, in both directions, and counts each decision on `propamm_admissions_total{outcome=admitted|dropped|evicted,pool}` |
```
louise-poole: 🤖 The row names `SolverWorker::drops_component_or_unbacked_pamm` and `apply_pamm_admission`. What shipped is `should_drop_component` — a free function in `worker_pool/worker.rs:51`, not a method — and `PammAdmission::select_pamm_updates` in `propamm_fallback/admission.rs:137`. Neither documented name appears anywhere in the workspace, and the row never names `PammAdmission`, which is where the rule and its state now live. A claude doc pointing at symbols you can't grep for is worse than no doc. Maps to the `docs` directive.

Point the row at the real symbols and say which module owns them: the scope and protocol rule is `should_drop_component` in `worker_pool/worker.rs`; the pAMM rule is `propamm_fallback::admission::PammAdmission`, re-decided per event by `select_pamm_updates`. While you're in there, the cell is now a ~90-word paragraph narrating mechanism in a module map, and it spells out `outcome=admitted|dropped|evicted`, which rots the next time an outcome is added — cut it back to what stays out of the graph and where the pAMM rule lives, and let the metric labels live at the `counter!` call.

fynd-core/src/worker_pool/worker.rs
```diff
@@ -31,14 +31,32 @@ use crate::{
...
 
+/// Whether a worker with this scope and exclusion list must keep `component` out of its graph: an
+/// exclusive component in a `PublicOnly` worker pool, or a component of an excluded protocol
+/// system.
+///
+/// Holds for the life of the worker, which is what lets it filter state updates and removals as
+/// well as additions. The pAMM rule in [`PammAdmission`] does not, and is applied beside this one.
+///
+/// A free function so [`PammAdmission`] can take it as the caller's half of the rule without
+/// borrowing the whole worker.
```
louise-poole: 🤖 The doc says the function is free "so `PammAdmission` can take it as the caller's half of the rule without borrowing the whole worker". Nothing passes it to `PammAdmission`; the actual reason is the `let Self { .. }` destructuring the two call sites need at `worker.rs:224` and `268`.

Say that — the two call sites destructure `self` so the closure can read these fields while `pamm_admission` is borrowed — or drop the paragraph.

fynd-core/src/propamm_fallback/mod.rs
```diff
@@ -35,24 +36,40 @@ pub fn has_pamm_leg(route: &Route) -> bool {
...
+/// returns `true` for every pAMM until the tiers arrive.
+///
+/// Returns `false` for any other protocol system, and for a pAMM that does not name exactly two
+/// tokens once the tiers are known: the fallback resolves per leg from the swap's own pair, so a
+/// wider component is left to [`fallback_amount_out`].
+pub fn is_unbacked_pamm(
+    component: &tycho_simulation::tycho_common::models::protocol::ProtocolComponent,
+    fee_tiers: Option<&FeeTiers>,
+    index: &FallbackPoolIndex,
+) -> bool {
```
louise-poole: 🤖 With the tiers unread the function returns `true` for a pAMM the market does back, so the name asserts something about the market that the body does not check. A reader has to reach the third doc paragraph to learn that.

Rename to `must_withhold_pamm`, matching the `withheld` set it feeds, and keep the tier paragraph as the reason.

fynd-core/src/propamm_fallback/admission.rs
louise-poole: 🤖 Use `withheld` for the counter label too, so the metric and the set agree and `evicted` stays the one word for a pAMM that leaves the graph.

fynd-core/src/propamm_fallback/manager.rs
```diff
@@ -0,0 +1,247 @@
...
+    ///
+    /// Expects an event the caller has already filtered, and the pool index to have seen the
+    /// unfiltered one — see [`update_pools`](Self::update_pools). `fee_tiers` is the one the
+    /// caller tested with [`needs_rebuild`](Self::needs_rebuild), so both decisions read one value.
+    pub(crate) fn select_pamm_updates(
+        &mut self,
+        market: &MarketDataView<'_>,
+        fee_tiers: Option<&FeeTiers>,
+        event: &mut MarketEvent,
+    ) {
```
louise-poole: 🤖 Its own doc opens with "Rewrites `event`", and it mutates the caller's additions and removals in place — rename it to `apply_pamm_admission`, which is what `fynd-core/CLAUDE.md` and the five `test_apply_pamm_admission_*` names already reach for.

fynd-core/src/worker_pool/worker.rs
```diff
@@ -1251,76 +1258,322 @@ mod tests {
...
-        let worker = SolverWorker::new(
-            market,
-            DerivedData::new_shared(),
-            MockAlgorithm::new(),
-            0,
-            "test_pool".to_string(),
-        );
-        let fee_tiers = FeeTiers::new(3000);
+    /// The fee tier the router resolves the pAMM's pair to.
+    const ADMISSION_TIER: u32 = 3000;
```
louise-poole: 🤖 Same value, same meaning, same test module as `FALLBACK_FEE_TIER` — use that one.

fynd-core/src/propamm_fallback/admission.rs
carloszanella: I had this as an earlier version but then PammAdmission was doing more than just handling pAMMs so I removed it. I didn't want pAMM admission logic handling exclusive liquidity or other filters

fynd-core/src/propamm_fallback/manager.rs
carloszanella: I renamed it to manager to make it also makes sense that it holds and exposes this information

fynd-core/src/propamm_fallback/manager.rs
carloszanella: outdated state should be worse?

seems out of scope. will not address suggestions that will make the scope of the PR bigger if there's no strong reason

fynd-core/src/propamm_fallback/admission.rs
carloszanella: Ill try to make it better

carloszanella:
@louise-poole about
> and a partial fee-tier read now churns every worker's graph.

I think it is best how it is right now. the alternatives (potentially handling stale state or more complex bookkeeping of pAMMs) are worse IMO. The graph rebuild is not that problematic, we are not loading data from tycho again. and this should be an exception most likely. I added a log for this case

louise-poole (approved):
Looks good!

PR #524, opened 2026-09-06
feat!: route filter on the quote request

carloszanella:
This breaking change introduces a new filter to the API that allows for filtering for protocols and pools on a request level.

A request names what it will not route through in `options.route_filter`: `exclude_pools` (component ids), `exclude_protocols` (protocol systems, named exactly as the market carries them) and `exclude_tokens` (tokens no route may pass through, its own two endpoints excepted). Nothing is excluded unless a request asks. The graph search drops that liquidity while it walks, so an excluded pool is never ranked and never simulated, and the four built-in algorithms honour it.

Line change breakdown:

| | added | removed | net |
|---|---:|---:|---:|
| test modules | 473 | 81 | +392 |
| production code | 523 | 149 | +374 |
| doc comments | 144 | 35 | +109 |
| generated | 72 | 1 | +71 |
| markdown docs | 23 | 7 | +16 |
| **total** | **1235** | **273** | **+962** |

## Commits

Nine, in dependency order — the market index, the core types, the graph, the algorithms, the request shape, the plumbing, the API, the clients, the replay capture.

## Breaking

`RouteFilter::with_components` is `with_pools`, and `SolveRequest::filter` is `SolveRequest::exclusions`. `Algorithm::find_best_route` keeps its single-argument shape from the previous commit.


fynd-core/examples/custom_algorithm.rs
```diff
@@ -63,18 +62,21 @@ impl Algorithm for DirectComponentAlgorithm {
...
         &self,
-        graph: &Self::GraphType,
-        market: MarketData,
-        label: Option<StateLabel>,
-        _derived: Option<SharedDerivedDataRef>,
-        order: &Order,
+        request: SolveRequest<'_, Self::GraphType>,
     ) -> Result<RouteResult, AlgorithmError> {
+        // Everything a solve reads arrives on the request. Honour `request.filter()` if your
+        // algorithm can: nothing enforces it, and a caller that excluded a pool expects it left
```
michcio1234: > nothing enforces it

As a user, I would be very surprised if I requested to exclude a pool but then got a route that goes through that pool. I would consider it a bug. A better outcome from user's perspective is IMO an error in such case; better to get a clear error than data that you didn't expect.

There could be a small function that inspects routes returned from an algorithm for consistency with the filters.

fynd-core/src/algorithm/mod.rs
```diff
@@ -197,29 +197,22 @@ pub trait Algorithm: Send + Sync {
...
     ///
-    /// * `graph` - The algorithm's preferred graph type (e.g., petgraph::Graph)
-    /// * `market` - Shared reference to market data for state lookups (algorithms acquire their own
-    ///   locks)
-    /// * `label` - Optional overlay label; when `Some`, the algorithm reads market state through
-    ///   the named overlay so per-request component overrides are applied during solving
-    /// * `derived` - Optional shared reference to derived data (token prices, etc.)
-    /// * `order` - The order to solve
+    /// Honour [`SolveRequest::filter`]. Nothing enforces it, so an algorithm that ignores it
+    /// returns routes the caller asked not to have.
```
michcio1234: As a user, I would be very surprised if I requested to exclude a pool but then got a route that goes through that pool. I would consider it a bug. A better outcome from user's perspective is IMO an error in such case; better to get a clear error than data that you didn't expect.

There could be a small function that inspects routes returned from an algorithm for consistency with the filters.

fynd-core/src/algorithm/request.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /// The derived data the algorithm may read: token prices, component depths.
+    #[must_use]
+    pub fn with_derived(mut self, derived: Option<SharedDerivedDataRef>) -> Self {
+        self.derived = derived;
+        self
+    }
+
+    /// Liquidity this solve must not route through.
```
michcio1234, line 78: Non-future-proof docstring, assuming that the filter may also hold pools that the algorithm SHOULD route through (it is written somewhere else that that's also intended usage of the filter struct).

fynd-core/src/types/quote.rs
```diff
@@ -73,6 +73,62 @@ impl QuoteRequest {
...
+/// requires something of one belongs here too.
+///
+/// Empty for most quotes. An algorithm is expected to honour it; nothing enforces it, so one that
+/// ignores it returns routes the caller asked not to have.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct RouteFilter {
+    components: FxHashSet<ComponentId>,
+    protocols: FxHashSet<String>,
+    tokens: FxHashSet<Address>,
+}
```
michcio1234: Looking at just the struct, without the docstring, I would assume that the fields specify what the route SHOULD go through, not what it SHOULD NOT go through.

I suggest renaming the fields to `excluded_components`, `excluded_protocols`, `excluded_tokens`.
And rename the `with*` methods analogously.

fynd-core/src/feed/market_data.rs
```diff
@@ -324,9 +324,10 @@ pub struct MarketState {
     /// Block info for the last update (only updated when protocols reported "Ready" status).
     /// None if no block has been processed yet.
     last_updated: Option<BlockInfo>,
-    /// Number of components per protocol system, maintained incrementally on
-    /// upsert/remove so readers never scan the full component map.
-    component_counts: FxHashMap<String, u64>,
+    /// The components of each protocol system, maintained on upsert/remove so a quote that
+    /// excludes a protocol names that system's pools without scanning the component map, and so
+    /// the metrics sampler can count them without one either.
+    components_by_protocol: FxHashMap<String, FxHashSet<ComponentId>>,
```
michcio1234, line 333: Not sure if the whole codebase uses the words "protocol" and "system" interchangeably, but if not, a better name for this would be `components_by_system`.

fynd-core/src/feed/market_data.rs
```diff
@@ -547,11 +563,28 @@ impl MarketState {
...
-            component_counts: FxHashMap::default(), // Not needed for simulation
+            components_by_protocol,
         }
     }
 }
 
+/// Groups component ids by the protocol system their component carries.
+///
+/// A subset built by [`MarketState::extract_subset`] needs the same index the base state keeps, so
+/// a filter naming a protocol system finds that system's pools in either one.
```
michcio1234: I don't understand this comment, and I don't think it's relevant here.

fynd-core/src/algorithm/request.rs
```diff
@@ -67,16 +74,17 @@ impl<'a, G> SolveRequest<'a, G> {
...
-    /// What the caller will accept in a route.
+    /// The pools and tokens this solve must not use.
     ///
-    /// Empty for most quotes. Hop bounds are not here: those are the algorithm's own configuration,
-    /// not the caller's to set.
-    pub fn filter(&self) -> &RouteFilter {
-        &self.filter
+    /// The caller's [`crate::RouteFilter`] is already resolved: a protocol system it named is here
+    /// as that system's pools. Hop bounds stay in the algorithm's own configuration; a caller does
+    /// not set them.
```
michcio1234: Irrelevant comment.

fynd-core/src/types/quote.rs
```diff
@@ -113,13 +116,87 @@ impl RouteFilter {
...
     }
 
-    /// Whether this pool is excluded, by its component id or by its protocol system.
+    /// The pools and tokens this filter excludes, with every protocol system it names replaced by
+    /// that system's pools.
+    ///
+    /// A graph knows a pool by its component id alone, so the expansion happens once per solve
+    /// rather than at every pool a search looks at. The market indexes its components by protocol
+    /// system, so this reads one entry per system named rather than passing over the market. It
+    /// copies the ids it finds.
```
michcio1234: Too detailed comment.

fynd-core/src/types/quote.rs
```diff
@@ -113,13 +116,87 @@ impl RouteFilter {
...
+}
+
+/// The pools and tokens one solve must not use.
+///
+/// What [`RouteFilter::create_exclusions`] makes of a caller's filter: the protocol systems are
+/// gone, and every pool one of them named is in `pools`. This is what a graph query and an
+/// algorithm read.
+///
+/// A token here is excluded as an intermediate. The order's own two tokens stay allowed, because
+/// every route touches them.
```
michcio1234: I can see your claude tried to exclude also order's own tokens, and it must have been very proud of itself for noticing this mistake, because it writes about it in many places now. ^^

fynd-core/src/algorithm/paths.rs
michcio1234, line 1: Commenting on the commit message:

> - `pair_has_allowed_pool` steps a walk over a pair with no usable pool

huh? 🤨

fynd-core/src/graph/token_graph.rs
```diff
@@ -71,6 +72,35 @@ impl<D> PairEdge<D> {
 /// A route as a sequence of tokens, before the pools serving each leg are chosen.
 pub type TokenPath = SmallVec<[NodeIndex; INLINE_TOKENS]>;
 
+/// Pools a leg may go through, held inline up to `INLINE_POOLS`.
+type Leg<'a, D> = SmallVec<[&'a EdgeData<D>; INLINE_POOLS]>;
```
michcio1234: Is this like a candidate list? I don't quite understand what this struct holds from the docstring + the struct name. Maybe I'll understand when I see it used.

fynd-core/src/graph/token_graph.rs
```diff
@@ -71,6 +72,35 @@ impl<D> PairEdge<D> {
...
+
+/// Pools one leg holds before it spills to the heap. Most pairs are traded by fewer.
+const INLINE_POOLS: usize = 8;
+
+/// Whether a token may sit between a route's two endpoints.
+///
+/// The algorithm's connector list and the request's exclusions both answer the same question, so a
+/// search asks it once. The endpoints are the caller's to exempt: every route touches them.
+struct IntermediateRule<'a> {
```
michcio1234: Maybe rename to `IntermediateTokenRule`? But if "intermediate" is understood as "intermediate token" throughout the codebase, then `IntermediateRule` is ok.

fynd-core/src/graph/token_graph.rs
```diff
@@ -98,6 +128,24 @@ impl<D> TopologyGraph<D> {
...
+    /// search steps over it rather than returning a route the caller refused.
+    fn pair_has_allowed_pool(
+        &self,
+        from: NodeIndex,
+        to: NodeIndex,
+        exclusions: &RouteExclusions,
+    ) -> bool {
+        if exclusions.is_empty() {
+            return true;
+        }
```
michcio1234, line 171: So this function returns `true` if there are no exclusions **even if** the graph itself has no pool between the two specified tokens? 🤨 

But if there are some exclusions, but they don't touch the specified tokens, then the function returns `true` **only if** there is at least one pool connecting the tokens.

Looks like the behavior changes depending on `exclusions`. Looks like a bug or at least a potential for future bugs.

fynd-core/src/graph/token_graph.rs
```diff
@@ -410,6 +475,9 @@ impl<D> TopologyGraph<D> {
                     continue;
                 };
                 for neighbor in self.neighbors(last) {
+                    if !self.pair_has_allowed_pool(last, neighbor, rule.exclusions) {
+                        continue;
+                    }
```
michcio1234: I don't know the codebase well enough to judge if this rule has been applied in **all** places that need it. The places that I see here look correct, but that's me measuring only your accuracy, not sensitivity.

docs/guides/custom-algorithm.md
```diff
@@ -11,7 +11,7 @@ Fynd exposes an `Algorithm` trait that lets you plug in custom routing logic wit
 The trait has four methods:
 
 * `name()` — a string identifier used in config and logs
-* `find_best_route()` — given a [`SolveRequest`], return the best route. The request carries the graph, the market, the order, the overlay to read state through, the derived data, and a `RouteFilter` saying what the caller will accept in a route — honour `request.filter()`. It is `#[non_exhaustive]`, so later additions do not break your algorithm. Call `Route::validate()` on each candidate and skip invalid ones (disconnected swaps, repeated tokens, malformed splits): the solver worker rejects an invalid route, which drops the whole solution for that worker pool, so prefer the next-best valid route instead
+* `find_best_route()` — given a `SolveRequest` (graph, market, order, overlay label, derived data, and the pools and tokens the caller excluded), return the best route. Honour `request.exclusions()`: nothing enforces it. Call `Route::validate()` on each candidate and skip invalid ones (disconnected swaps, repeated tokens, malformed splits): the solver worker rejects an invalid route, which drops the whole solution for that worker pool, so prefer the next-best valid route instead
```
michcio1234: Wait, so "nothing enforces it" or "solver worker rejects an invalid route"?

fynd-core/src/algorithm/water_fill/mod.rs
```diff
@@ -214,15 +214,20 @@ impl WaterFillAlgorithm {
...
+    ) -> Result<SetupResult<'a, 'a>, AlgorithmError> {
+        let graph = request.graph();
+        let market = request.market().clone();
+        let label = request.label().cloned();
+        let derived = request.derived().cloned();
+        let order = request.order();
+        let exclusions = request.exclusions().clone();
+        // The hop bounds are this algorithm's, the exclusions are the request's; a search borrows
+        // both.
+        let search = RouteSearch { filter: &self.query, exclusions: &exclusions };
```
michcio1234: Maybe a better named for this struct would be `let bounds = RouteSearchBounds{...}`?

fynd-core/src/algorithm/water_fill/mod.rs
```diff
@@ -216,15 +216,10 @@ impl WaterFillAlgorithm {
...
+        request: SolveRequest<'a, TopologyGraph<DepthAndPrice>>,
         deadline: Deadline,
     ) -> Result<SetupResult<'a, 'a>, AlgorithmError> {
-        let graph = request.graph();
-        let market = request.market().clone();
-        let label = request.label().cloned();
-        let derived = request.derived().cloned();
-        let order = request.order();
-        let exclusions = request.exclusions().clone();
+        let (graph, order, market, label, derived, exclusions) = take_parts(request);
```
michcio1234: I thought `take_parts` would be a method of `SolveRequest`.

fynd-core/src/algorithm/request.rs
```diff
@@ -20,6 +18,24 @@ pub struct SolveRequest<'a, G> {
     exclusions: RouteExclusions,
 }
 
+/// The request taken apart, so an algorithm owns the market and the derived data in one move.
+///
+/// Every built-in algorithm opens by moving the market, the overlay label and the derived data out
+/// of the request. Crate-internal: an algorithm outside this crate reads the request through its
+/// getters.
```
michcio1234: Too detailed.

fynd-rpc/src/api/request_capture.rs
```diff
@@ -32,13 +32,50 @@ struct ReplayOptions {
...
+/// The liquidity a captured request excluded, sorted so two captures of one request read alike.
+#[derive(Debug, Serialize)]
+struct ReplayRouteFilter {
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pools: Vec<String>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    protocols: Vec<String>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    tokens: Vec<String>,
+}
```
michcio1234: Same here, would be clearer if the names contained `excluded_`.

michcio1234 (commented):
LGTM, but - as I said in one of the comments - I can only judge "accuracy", not "sensitivity": can judge correctness of what was changed, but can't say if you missed anything.

There's one comment pointing a possible bug.

carloszanella:
last force pushed just with the rebase

fynd-core/src/algorithm/paths.rs
```diff
@@ -280,14 +280,22 @@ mod tests {
...
     fn hops(min_hops: usize, max_hops: usize) -> GraphQueryFilter {
         GraphQueryFilter { min_hops, max_hops, connector_tokens: None }
     }
 
+    fn search<'a>(
+        filter: &'a GraphQueryFilter,
+        exclusions: &'a RouteExclusions,
+    ) -> RouteSearch<'a> {
+        RouteSearch { filter, exclusions }
+    }
```
tamaralipows: I don't think this helper fn serves any purpose

fynd-core/src/algorithm/request.rs
```diff
@@ -0,0 +1,112 @@
...
+use crate::{
+    derived::SharedDerivedDataRef,
+    feed::market_data::{MarketData, StateLabel},
+    types::{quote::RouteExclusions, Order},
+};
+
+/// One order to solve, and everything the algorithm reads to solve it.
+///
+/// Build one with [`SolveRequest::new`] and the `with_` methods. Read it through the getters or
+/// consume it with [`SolveRequest::into_parts`].
```
tamaralipows: Pretty unnecessary info to mention in docstrings

fynd-core/src/feed/market_data.rs
```diff
@@ -324,12 +327,36 @@ pub struct MarketState {
...
     last_updated: Option<BlockInfo>,
-    /// Number of components per protocol system, maintained incrementally on
-    /// upsert/remove so readers never scan the full component map.
-    component_counts: FxHashMap<String, u64>,
+    /// The components of each protocol system, maintained on upsert/remove so a quote that
+    /// excludes a protocol names that system's pools without scanning the component map, and so
+    /// the metrics sampler can count them without one either.
+    components_by_protocol: FxHashMap<String, FxHashSet<ComponentId>>,
+    /// Changes only when a component is added or removed, not when its state changes.
+    component_generation: u64,
```
tamaralipows, line 335: This doc doesn't really tell me what the param does

fynd-core/src/worker_pool/worker.rs
```diff
@@ -403,6 +477,34 @@ where
...
+                            {
+                                debug!(order_id = %order.id(), fallback_pool = %pool,
+                                    "dropping pAMM route: request excludes its fallback pool");
+                                return Err(SolveError::route_rejected(
+                                    order.id(),
+                                    RouteRejection::PammFallbackExcluded,
+                                ));
+                            }
+                        }
+                    }
```
tamaralipows, line 542: The `quote` function is now absolutely massive and I'm wary of it getting worse. Any way to split this up into smaller fns for organization?

fynd-core/src/types/quote.rs
```diff
@@ -73,6 +76,122 @@ impl QuoteRequest {
...
+///
+/// Algorithms must honour these exclusions during search and simulation. The worker rejects
+/// returned routes that violate the request filter.
+///
+/// Tokens are excluded only as intermediates; the order's input and output remain allowed.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct RouteExclusions {
+    pub(crate) pools: FxHashSet<ComponentId>,
+    pub(crate) tokens: FxHashSet<Address>,
+}
```
tamaralipows, line 159: Not super clear at a glance the difference between `RouteExclusions` and `RouteExclusionFilter` and why we need both, pretty similarly named. Maybe a better docstring or naming?

tamaralipows (approved):
Ran an agent on this and mostly got nonsense so I'm using my mortal eyes...

I didn't find any bugs or blockers so I'll pre-emptively approve, only see my readability/architecture comments. Please be sure to also mark Michal's comments as resolved if you've resolved them. I didn't double-check anything he already mentioned.

fynd-core/src/algorithm/request.rs
```diff
@@ -0,0 +1,105 @@
...
+    ) -> (
+        &'a G,
+        &'a Order,
+        MarketData,
+        Option<StateLabel>,
+        Option<SharedDerivedDataRef>,
+        Arc<RouteExclusions>,
+    ) {
+        (self.graph, self.order, self.market, self.label, self.derived, self.exclusions)
+    }
```
louise-poole: I'm not a fan of tuples in general in rust. They're merely informal/unnamed structs.

---

🤖 `into_parts` is a method now, but it still hands back six positional values that all three call sites destructure in the same order. A new field silently shifts a position.

Return a named struct, or make the fields `pub(crate)` and destructure `SolveRequest` directly at the call sites.

fynd-core/src/graph/mod.rs
```diff
@@ -143,9 +147,12 @@ where
...
 }
 
-/// What a caller will accept from a route search.
+/// The bounds a route search runs under.
 ///
-/// Bundles the bounds every path query carries so they travel as one argument instead of three.
+/// Bundles what an algorithm configures once — how long a route may be, and which tokens it may
+/// pass through — so they travel as one argument instead of three. What a request excludes is not
+/// here: that changes per solve and travels beside this as a [`RouteExclusions`].
+#[derive(Debug, Clone)]
```
louise-poole, line 155: 🤖 `RouteSearch` holds a reference, so it gets `Copy` free and needs only `Debug`. Drop `Clone`.

fynd-core/src/algorithm/most_liquid.rs
```diff
@@ -756,7 +765,7 @@ impl Algorithm for MostLiquidAlgorithm {
 
         // Step 2: Score and sort all paths by estimated output (higher score = better)
         // No lock needed — scoring uses only local graph data.
-        let mut scored_paths = rank_by_heuristic(graph, all_paths);
+        let mut scored_paths = rank_by_heuristic(graph, all_paths, search.exclusions);
```
louise-poole: 🤖 Both reads go through the struct to reach the local it was built from; use `&exclusions`.

fynd-core/src/worker_pool/worker.rs
```diff
@@ -56,6 +63,105 @@ fn should_drop_component(
...
+        if *cached_generation == generation {
+            return exclusions.clone();
+        }
+    }
+    let exclusions = Arc::new(
+        view.base_market_state()
+            .resolve_route_filter(params.route_filter()),
+    );
+    *cache = Some((generation, exclusions.clone()));
+    exclusions
```
louise-poole, line 102: Nice debugging/tracing here would likely make our lives easier later.

---

🤖 `resolve_exclusions` now caches per market generation and returns a shared `Arc`, so a log here would fire once per generation rather than once per worker — cheaper than when I first raised it. Nothing still records what the filter resolved to, so a filtered request coming back `NoPath` leaves an operator unable to tell a disconnected market from a filter that emptied the search.

A `debug!` on the cache-miss path with the resolved pool and token counts.

fynd-core/src/graph/token_graph.rs
```diff
@@ -71,6 +69,12 @@ impl<D> PairEdge<D> {
 /// A route as a sequence of tokens, before the pools serving each leg are chosen.
 pub type TokenPath = SmallVec<[NodeIndex; INLINE_TOKENS]>;
 
+/// Pools a leg may go through, held inline up to `INLINE_POOLS`.
+type Leg<'a, D> = SmallVec<[&'a EdgeData<D>; INLINE_POOLS]>;
+
+/// Pools one leg holds before it spills to the heap. Most pairs are traded by fewer.
+const INLINE_POOLS: usize = 8;
```
louise-poole: Is this an assumption or a verified fact?

---

🤖 "Most pairs are traded by fewer" is a claim I would like a source for, and I doubt it holds for the pairs that matter most — record where the number came from.

louise-poole (commented):
Re-checked after the rebase. Some of my earlier comments are resolved by the new commits. Six left, below. Nothing blocking.

---

🤖 **`most_liquid` snapshots simulation state for excluded pools** — `fynd-core/src/algorithm/most_liquid.rs:490–513`

_**The less we need to clone the better.**_

---

🤖 `snapshot_market_state` collects component ids with no exclusion check, so `extract_subset_with_overlay` clones VM state for pools the solve then refuses at line 639.

Take the exclusions as an argument and skip any pool `excludes_pool` rejects.

louise-poole (approved):
None of my findings above are blocking/that important so have approved! Thanks!

fynd-core/src/algorithm/mod.rs
carloszanella: Addressed on c8c65454. The worker now checks every swap of the returned route, split branches included, and rejects one naming an excluded pool, protocol system or intermediate token; a pAMM leg is also checked against the Uniswap V3 fallback pool its fee tier selects, before the fallback is priced. So an algorithm that ignores the exclusions gets an error rather than a route the caller excluded, which is the outcome you asked for.

fynd-core/src/algorithm/request.rs
carloszanella: Addressed on d4bf3e01 and a2e143a3. The type is now `RouteExclusionFilter` and what a solve reads is `RouteExclusions`, so exclusion-only is the stated contract rather than an assumption in a docstring. If an include-list is ever wanted it would be a separate field, not a reinterpretation of this one.

fynd-core/src/types/quote.rs
carloszanella: Addressed on f701a405. The fields, builders and getters are now `excluded_pools`, `excluded_protocols` and `excluded_tokens`, on the core type and on the wire type. The JSON is untouched: the serde names were already `exclude_*`, and only the Rust accessors moved.

fynd-core/src/feed/market_data.rs
carloszanella: Not changed, deliberately. The codebase already uses "protocol" as the shorthand for protocol system in names that predate this PR: `component_counts_by_protocol` on this same type, plus `exclude_protocols` and `is_excluded_protocol` in `feed/component_filter.rs`. Renaming only the new accessor would make it the odd one out, so I kept the existing vocabulary.

fynd-core/src/feed/market_data.rs
carloszanella: Addressed on cfff283c. The comment is deleted. It was also wrong: it claimed a subset needs the index so a protocol filter finds that system's pools, but nothing resolves a filter against a subset.

fynd-core/src/algorithm/request.rs
carloszanella: Addressed on d4bf3e01. The sentence about hop bounds is gone from the `exclusions()` getter. It described `GraphQueryFilter`, was equally true above every other getter, and was already stated on the type that owns it.

fynd-core/src/types/quote.rs
carloszanella: Addressed on 99f23838 and a2e143a3. That doc went with `create_exclusions`, which moved to `MarketState::resolve_route_filter`, and the replacement says what the call returns and how a protocol system matches rather than explaining the indexing. Two sentences instead of five.

fynd-core/src/types/quote.rs
carloszanella: Addressed on faeaeaf6 and cfff283c. Fair hit. It is down to two places, one for each audience: the doc on `RouteExclusions` and the doc on the wire field an API user reads. The other copies went with `IntermediateRule` and the `get_subgraph` doc.

fynd-core/src/algorithm/paths.rs
carloszanella: No longer valid. The branch history was rebuilt as a first-time ship, so that commit and its message no longer exist. Nothing in the current series uses that phrasing.

fynd-core/src/graph/token_graph.rs
carloszanella: Partly out of scope. Line 73 is `TokenPath`, which predates this PR and is unchanged by it. The two things this PR did add just below it are `Leg` and `INLINE_POOLS`, and Louise has raised the `INLINE_POOLS` doc separately, so I will answer it there.

fynd-core/src/graph/token_graph.rs
carloszanella: Addressed on faeaeaf6. `IntermediateRule` is deleted. Its job moved onto `RouteExclusions::allows_token` and `RouteSearch::allows_token`, so the endpoint rule lives in one predicate instead of a struct plus three hand-written copies.

fynd-core/src/graph/token_graph.rs
carloszanella: Addressed on 337eb073, and you were right. It answered `true` for a pair the graph holds no pool for, but only while nothing was excluded; adding any exclusion flipped the same pair to `false`. It now reads the pair's pools first and answers `false` when there are none, whatever the filter says, keeping the empty-filter shortcut after that check.

fynd-core/src/graph/token_graph.rs
carloszanella: Noted, and the shape has changed since you looked. The rule is now in one predicate, `RouteSearch::allows_token`, which the graph walk, Bellman-Ford's `can_cross` and water-fill's `can_extend_path` all call, so "is it applied everywhere" is a question about call sites rather than about three separate copies agreeing.

docs/guides/custom-algorithm.md
carloszanella: Addressed on faeaeaf6. Both, and the guide no longer says otherwise: an algorithm is asked to honour the exclusions while it searches and simulates, and the worker rejects a returned route that violates them. The contradiction was written before the enforcement existed.

fynd-core/src/algorithm/water_fill/mod.rs
carloszanella: Addressed on f701a405, though not by renaming the struct. `RouteSearch` holds the exclusions as well as the bounds, so instead I renamed the field: `RouteSearch::filter` is now `RouteSearch::bounds`, which is what the `GraphQueryFilter` doc already called it. That removes `search.filter`, which meant hop bounds one field away from the exclusions this PR spent the word "filter" on.

fynd-core/src/algorithm/water_fill/mod.rs
carloszanella: Addressed on faeaeaf6. It is a method now, `SolveRequest::into_parts`, and public, so an algorithm outside the crate can use it too. The free crate-private function is gone.

fynd-core/src/algorithm/request.rs
carloszanella: Addressed on a2e143a3. That doc went with `take_parts`. `into_parts` now carries one line naming the six parts, with no explanation of what it does or does not clone.

fynd-rpc/src/api/request_capture.rs
carloszanella: Addressed on f701a405. The capture now records `exclude_pools`, `exclude_protocols` and `exclude_tokens` — the wire's own names rather than `excluded_`. That also fixes a second problem: a harness can post a captured `options` straight back without translating the three keys.

fynd-core/src/algorithm/paths.rs
carloszanella: Addressed on a2e143a3. Removed; the cases build the `RouteSearch` literal directly. You were right that it only rebuilt a two-field struct.

fynd-core/src/algorithm/request.rs
carloszanella: Addressed on a2e143a3. The sentence telling readers to build one with `new` and the `with_` methods is gone. The signatures already carry that.

fynd-core/src/feed/market_data.rs
carloszanella: Addressed on a2e143a3. It now says what comes back — the pools and tokens the filter excludes, with each protocol system replaced by the components this market holds for it — and how a system matches, including the `:` family form and that an unknown name excludes nothing.

fynd-core/src/worker_pool/worker.rs
carloszanella: Addressed on c73c022e. The two blocks this branch added are now `resolve_exclusions` and `excluded_fallback_pool`, which takes `quote` from 303 lines back to 265. For context it was already 247 before this branch, so this keeps the branch's own growth to 18 lines; splitting the pre-existing bulk felt like it belonged in its own PR, happy to do that separately if you want it.

fynd-core/src/types/quote.rs
carloszanella: Addressed on a2e143a3. Both docstrings now say which side of the resolution they sit on: `RouteExclusionFilter` is the caller's own words and may name protocol systems, `RouteExclusions` is what a search reads once each system has been replaced by its pools. The filter's doc links to the call that converts one into the other.

fynd-core/src/graph/mod.rs
carloszanella: This one does not hold, I think. `RouteSearch` derives `Copy`, and `Copy` has `Clone` as a supertrait, so `#[derive(Debug, Clone, Copy)]` is the only form that compiles — dropping `Clone` is a hard error. Leaving it as is unless I have misread you.

fynd-core/src/algorithm/request.rs
carloszanella: Addressed on 5d0430e3. `into_parts` now returns a named `SolveParts` struct and the three built-ins, the shipped example and the guide destructure it by field name, so adding a field is a compile error at the call sites rather than a silent shift. Agreed on the general point — it was an unnamed struct pretending to be a return type.

fynd-core/src/algorithm/most_liquid.rs
carloszanella: Addressed on 5d0430e3. Both reads now take `&exclusions` directly; they went through `search` to reach the local it had just been built from.

fynd-core/src/worker_pool/worker.rs
carloszanella: Addressed on 8c474699. The resolve path now logs, at debug, how many pools and protocol systems the filter named and how many pools and tokens that resolved to. It sits on the miss path as you suggested, so a request pays for it once per market generation rather than once per order per pool, and a filtered request that comes back `NoPath` can now be told apart from a market that genuinely has none.

fynd-core/src/graph/token_graph.rs
carloszanella: Addressed on f57b6ba5, and it was an assumption. I traced it: `INLINE_POOLS` and `Leg` only exist because this PR filters excluded pools inside `expand_path`, and a filtered subset of a pair's pools is not contiguous so it can no longer be a slice into the graph. The number is an inline capacity that was picked, and that sentence was written afterwards to justify it — so the doc now says what the constant is, that a pair traded by more pools still works on the heap, and why `Leg` exists at all. No measurement, so no claim.

carloszanella:
Louise — the snapshot finding from your review body had no inline thread to reply on, so: addressed on 0d1c934d. `snapshot_market_state` now takes the exclusions and leaves an excluded pool out of the component ids it asks the subset for, so `extract_subset_with_overlay` no longer clones VM state for pools the simulation then refuses.

The other five are answered on their own threads. One I pushed back on: `RouteSearch` derives `Copy`, and `Copy` has `Clone` as a supertrait, so dropping `Clone` does not compile.

fynd-core/src/graph/token_graph.rs
carloszanella: Follow-up on 54354b5e: rather than reword the constant, I removed it. `Leg` is now an enum — the pair's own slice, or the pools that survive the filter — and `TopologyGraph::leg_between` only collects when the request actually excludes a pool on that pair. An unfiltered quote is back to a pointer and a length like before this branch, a filtered one pays only on the legs the filter touches, and the collect is rare enough that a `Vec` does, so there is no capacity left to justify.

PR #526, opened 2026-09-08
fix(core): give a simulated swap four times the estimated gas

carloszanella:
The simulated call was given twice the gas the quote estimated. A swap whose real execution costs more than that reverts out of gas under simulation, and the quote is reported as failing a check it would have passed on chain.

`SIMULATION_GAS_LIMIT_MULTIPLIER` goes from 2 to 4. Four times the estimate leaves room for the variance a real transaction meets, and still reads as a swap rather than as a call with the whole block to spend — which matters because a pool that reads `gasleft()` can tell those apart and answer differently.

The floor and ceiling are unchanged: an estimate under 125k still lands on the 500k floor, and a large one still clamps at 30M.

The test now reads the multiplier instead of restating the product, so the factor is stated in one place.

https://claude.ai/code/session_01CsPfEGfzpCMQWepNM1sJNQ

### DIFF
--- a/fynd-core/src/worker_pool_router/instrumentation.rs
+++ b/fynd-core/src/worker_pool_router/instrumentation.rs
@@ -14,7 +14,7 @@
 
 use std::collections::BTreeMap;
 
-use metrics::counter;
+use metrics::{counter, histogram};
 use num_traits::ToPrimitive;
 use tracing::{trace, Level};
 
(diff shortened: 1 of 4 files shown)


Troshchk (approved):
Thank you! ✨

fynd-core/src/worker_pool_router/instrumentation.rs
```diff
@@ -37,22 +37,23 @@ struct OrderCoverage {
...
+/// deployment gets the per-algorithm comparison without turning the log on. The log stays behind
+/// its TRACE target because it prints one line per pool per order, several times the metric's
+/// cost, and carries the order and token detail the histogram cannot label.
 ///
 /// Covers every pool that answered, whatever its liquidity scope. `is_best` marks the pool whose
 /// quote had the highest output net of gas. The amounts are the ones each pool solved; on
 /// exclusive-liquidity deployments the amount finally quoted to the caller can be lower, because
 /// `combine_with_surplus` withholds part of it.
-pub(super) fn log_quote_comparison(
+pub(super) fn record_quote_comparison(
```
Troshchk, line 52: why did you rename this?

carloszanella: it doesn't just log anymore. it also records metrics in Prometheus

PR #527, opened 2026-09-08
docs: add Stable Swap Rate guide

pistomat:
## Summary

Adds `docs/guides/stable-swap-rate.md`, the Beta guide for fixed-rate swaps, and lists it in `docs/SUMMARY.md` under Guides after "Charge Fees on your Swaps".

The guide covers:

- How a declared rate is settled through the existing client fee and client contribution primitives, with the formula block and two worked examples (market above and below the declared rate).
- The `x_stable_rate` object on the quote response.
- Hosted deployment and how to get Beta access through the Fynd Telegram.

## Test plan

- [ ] Every link resolves: client-fees, router-fees, encoding-options, the `maxclientcontribution` anchor, the Tycho vault page, and the Telegram invite (checked with curl).
- [ ] Rendered locally inside the live GitBook page shell to check hint, list, table, and code block markup.
- [ ] Verify the published page on docs.fynd.xyz after GitBook re-syncs from `main`.


docs/guides/stable-swap-rate.md
```diff
@@ -0,0 +1,97 @@
...
+3. When the market is better than your rate, the difference is collected as a client fee into your vault.
+4. When the market is worse, your vault tops the output up to the declared amount via client contributions.
+5. The router settles atomically and your user receives the declared amount.
+
+```text
+user --approve + swap--> TychoRouter --> pools
+                            |
+                     client vault
+                  fee / contribution
+```
```
louise-poole: Could this be a diagram pls? Makes the docs nicer when they are not inlined as text like this...

docs/guides/stable-swap-rate.md
```diff
@@ -0,0 +1,97 @@
...
+
+Fynd encodes the declared rate into the same `ClientFee` payload that [client fees](client-fees.md) use. The fee moves surplus into your vault, and `maxClientContribution` caps how much the router may take from your vault to cover a shortfall. No new contract sits between your user and the router.
+
+Your vault never funds the swap itself. The pools do. The vault only settles the difference between your declared rate and the onchain rate, so a declared rate that tracks the market needs only a small float.
```
louise-poole: > Your vault never funds the swap itself. The pools do.

Don't really like this line. I don't think anyone starts reading this assuming they're funding the swap no? I would leave out those 2 sentences and just have "The vault settles the difference between your..."

docs/guides/stable-swap-rate.md
```diff
@@ -0,0 +1,97 @@
...
+
+Given the market quote `Q`, the declared output `D`, and your depeg tolerance in basis points:
```
louise-poole: I would have assumed the market quote is the declared output... what's the difference?
EDIT: wait I see! Can you maybe define them nicely here so that you don't need to get to the examples below for things to feel clear?

docs/guides/stable-swap-rate.md
```diff
@@ -0,0 +1,97 @@
...
+
+{% hint style="warning" %}
+**Beta.** Stable Swap Rate is available to selected integrators. To get access, contact us on [Telegram](https://t.me/+B4CNQwv7dgIyYTJl).
+{% endhint %}
+
+## How it works
+
+1. You declare the rate your user sees, for example 1 USDC = 1 USDT.
+2. Fynd finds the best market route, as for any quote.
+3. When the market is better than your rate, the difference is collected as a client fee into your vault.
```
louise-poole: Can we make it clearer that we're not talking about slippage here. But about quoted market amounts?

docs/guides/stable-swap-rate.md
```diff
@@ -0,0 +1,97 @@
...
+```
+max_client_contribution = 1,000,000 * 30 / 10,000 = 3,000
+client_fee              = 0
+client_contribution     = 1,000,000 - 999,700 = 300   <- your vault
+user receives           = 1,000,000
+```
+
+## Quote response
+
+The Stable Swap Rate endpoint is Fynd-compatible. Send the same `POST /v1/{chain}/quote` request with encoding enabled, and read the same `OrderQuote` back. The `fee_breakdown` and the encoded `transaction` are the standard ones, so an existing Fynd client only needs a different base URL. Encoding is required because the fixed rate lives in the transaction's `ClientFee` payload. See [encoding options](encoding-options.md).
```
louise-poole: I don't understand... what endpoint? Are we deploying a new hosted Fynd for stable swaps?

docs/guides/stable-swap-rate.md
```diff
@@ -0,0 +1,97 @@
...
+
+Each quote carries an additional `x_stable_rate` object. Use it to show your user the declared rate next to the market, and to monitor how far the market drifts from your rate and how much your vault subsidizes.
```
louise-poole: Where is this defined? Not in this repo no?

docs/guides/stable-swap-rate.md
louise-poole: Ah - I see this introduced later in the docs. I think we should mention that first so that it's clear a new Fynd service is deployed for this. Then we talk about how it's similar/different to the existing Fynd service.

docs/guides/stable-swap-rate.md
pistomat: Added a diagram 🙏

docs/guides/stable-swap-rate.md
pistomat: Yes, you are right!

docs/guides/stable-swap-rate.md
pistomat: I tried to rephrase this

docs/guides/stable-swap-rate.md
pistomat: Right, I should reorder the info

docs/guides/stable-swap-rate.md
pistomat: No, not in this repo. I removed the weirdly specific `x_stable_rate` reference, we will explain it more when we do the onboarding.

docs/guides/stable-swap-rate.md
```diff
@@ -8,6 +8,8 @@ Fix the output rate of a token pair for your users. Fynd routes the swap at mark
 
 Stable Swap Rate is built from two primitives you may already use: the [client fee](client-fees.md) and the [client contribution](client-fees.md#maxclientcontribution).
 
+Stable Swap Rate runs as a separate Fynd server that we host for you, with its own base URL. It accepts the same requests as the standard Fynd API and returns the same responses, plus an API extension with Stable Swap Rate fields.
```
louise-poole, line 11: Final q here: we have some clients that like to host their own Fynd instance locally. Are there any docs we can have for them to set this up, or are we limited to a hosted-Fynd version for this only?

louise-poole (approved):
Feels much clearer! Thanks! Just one last comment, otherwise I've approved.

PR #528, opened 2026-09-08
feat: support new Robinhood protocols

tvinagre:
## Summary
- Add protocol registry support for SushiSwap V3, RobinSwap V3, and Ramses V3 on Robinhood.
- Raise the Tycho simulation and execution dependency floor to `0.384.0` for the required Robinhood executor mappings.
- Add a regression test that covers all six Robinhood protocol systems.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -158,8 +158,12 @@ metrics-exporter-prometheus = "0.16"
 metrics-util = "0.20"
 
 # Tycho
-tycho-execution = ">=0.382.0"
-tycho-simulation = ">=0.382.0"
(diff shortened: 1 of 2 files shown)


brunoguerios (approved):
Approving so we can get these protocols supported ✅ 

But I found it a bit odd to test against log lines like this.
I'd suggest rethinking the approach in a follow-up PR.
Maybe we should handle unknown protocols as `DataFeedError::Config`, which is the same way we handle unknown pAMMs, then we can simply check if `is_ok()` instead of parsing logs.

PR #531, opened 2026-09-09
fix!: Rerank after quoting RFQs

carloszanella:


tvinagre (approved):

fynd-core/src/worker_pool_router/rerank.rs
```diff
@@ -0,0 +1,452 @@
...
+};
+
+/// How far behind the best candidate the next-best one may sit and still make requoting the best
+/// one worthwhile, in basis points.
+///
+/// A requote moves the best candidate down, never up, so a next-best candidate further behind
+/// than the largest level-to-signed gap cannot overtake it, and the maker request is wasted. 1000
+/// bps sits well above the gaps `rfq_signed_quote_deviation_bps` records: skipping a requote that
+/// would have changed the winner costs the user the difference, while making one that changes
+/// nothing costs a request.
```
tamaralipows, line 37: I think this docstring can be clearer and more direct. Re-quote just means to get the expected amount out from the binding quote? So if the second-best is too far below the price levels quote, we don't bother to get the quote again from the binding quote right?

fynd-core/src/worker_pool_router/rerank.rs
```diff
@@ -0,0 +1,452 @@
...
+    let requoted_pool = best.worker_pool().to_string();
+    candidates.sort_by(|a, b| {
+        b.amount_out_net_gas()
+            .cmp(a.amount_out_net_gas())
+    });
+    let outcome = if candidates[0].worker_pool() == requoted_pool { "held" } else { "changed" };
+    counter!("rfq_rerank_total", "outcome" => outcome, "pool" => requoted_pool).increment(1);
+}
+
+/// Whether requoting this order's best candidate can change which candidate wins.
```
tamaralipows, line 76: Question: in the case where we don't re-rank since there is no second-best option within 1000bps margin, don't we still risk providing a too-high expected and min amount that will revert on-chain? Are we fine with this?

tamaralipows (approved):
Thank you. Code itself looks good - I just have one question about the main assumption. Conditional approve on that not being a blocker for us

PR #535, opened 2026-09-10
chore: remove hindsight

tamaralipows:
Hindsight re-solves settled trades to measure what Fynd would have quoted, so it has to run the algorithms the deployment serves.

The crate arrives there with its history: propeller-heads/fynd-labs#26.

Nothing here imports hindsight. `release-plz.toml` already marked it `release = false`, and the two remaining mentions in `fynd-core` are comments naming it as a consumer, which stays true.

`fynd-tools-common` stays: `tools/benchmark` uses it, and fynd-labs takes it from this repo by git the same way `fynd-bench-harness` is taken.

## Merge order

This removes `/usr/local/bin/hindsight` from the `fynd` image. Merge only after fynd-labs#26 is released and helm-configuration points every hindsight release at the fynd-labs image. Until then the running pods pull an image tag that still carries the binary, so they keep working, but a hindsight release pinned to a newer `fynd` tag would break.


Troshchk (approved):
Thank you!

PR #536, opened 2026-09-10
feat!: encode each order independently

tamaralipows:
An encoding failure discarded every order in the request. `Encoder::encode` sent all solutions to `encode_solutions` in one call and returned one `Result` for the batch, so one order with an RFQ leg that could not be encoded took the calldata of every other order with it.

Each order now prepares and encodes on its own blocking task. An order that fails gets the new `QuoteStatus::FailedEncoding` and no transaction; the other orders keep their calldata.

The failure is always reported on the order, never on the call, so a request whose only order fails to encode answers 200 with `status: "failed_encoding"` where it used to answer 422 `FAILED_ENCODING`. Callers read the per-order status, as they already do for `no_route_found` and `price_check_failed`.

Encoding one order per call takes tycho-execution's single-item path, which skips its `MAX_ENCODING_THREADS` cap, so `MAX_CONCURRENT_ENCODES` applies the same limit of 32 here. The round trips still overlap.


fynd-core/src/encoding/encoder.rs
```diff
@@ -271,83 +290,161 @@ impl Encoder {
...
+            }
+        }
+
+        for (idx, error) in failures {
+            tracing::warn!(
+                order_id = %quotes[idx].order_id(),
+                %error,
+                "encoding failed for this order; it is returned without a transaction"
+            );
+            counter!("encoding_failures_total").increment(1);
```
Troshchk, line 341: there is this in encode_quotes():
```    
if encoded.is_err() {
        counter!("encoding_failures_total").increment(1);
    }
```

Are we expected to count both?

clients/rust/src/types.rs
```diff
@@ -665,6 +665,9 @@ pub enum QuoteStatus {
     NotReady,
     /// The solution failed external price validation.
     PriceCheckFailed,
+    /// A route was found, but it could not be encoded into router calldata. The quote carries no
+    /// transaction. Other orders in the same request keep theirs.
+    FailedEncoding,
```
Troshchk: May be EncodingFailed ? Same strcture as the error above?

Troshchk (commented):
Thank you!

fynd-core/src/encoding/encoder.rs
```diff
@@ -271,83 +290,161 @@ impl Encoder {
...
+        // time; the rest wait for a permit rather than for a thread.
+        futures::stream::iter(tasks)
+            .buffered(MAX_CONCURRENT_ENCODES)
+            .map(|joined| {
+                joined.map_err(|e| {
+                    SolveError::FailedEncoding(format!("the encoding task failed: {e}"))
+                })
+            })
+            .try_collect()
+            .await
```
Troshchk, line 471: Looks like something worth checking

---

**[DESIGN `#1`]**

**Description**: `encode_prepared_solutions` returns `Result<Vec<Result<EncodedSolution, EncodingError>>, SolveError>`, a two-tier error model: an `EncodingError` is kept per order, but a `JoinError` (a panicking blocking task) is mapped to `SolveError::FailedEncoding` and `try_collect` (encoder.rs:444) short-circuits the outer `Result`, so `encode` returns `Err` via `?` and `encode_quotes` discards every other order's calldata. That contradicts the module's stated guarantee that "a failure that belongs to one order never fails the call": a panic inside `tycho-execution`'s single-item path (an `unwrap`, an index) is exactly a one-order failure, yet it takes down the batch. This also leaves `encode`'s `# Errors` doc (encoder.rs:268–270) inaccurate — it lists only `EncodingUnavailable` and claims per-order failures never fail the call.

**Suggestion**: Fold a `JoinError` into the inner per-order `Result` so a panicking order gets `FailedEncoding` and the rest keep their calldata; reserve the call-fatal `SolveError` for the genuinely global `EncodingUnavailable`, which is already handled up front by the let-else guard. If you instead decide a panic *should* be call-fatal, correct the `# Errors` doc to say so rather than leaving the blanket sentence.

Troshchk (commented):
I think it is an important ai finding, can you check it, please?

fynd-core/src/encoding/encoder.rs
tamaralipows: We make a conscious effort not to panic in our code, so I don't think we should count on any subtask possibly panicking here

tamaralipows: To be clear about what the current code does: an encoding error on one order only fails that order, and the other orders keep their calldata. Only a panic is fatal for the whole call, and that is intended.

fynd-core/src/encoding/encoder.rs
tamaralipows: Good catch. The two increments never fired together, but the counter meant two different things, so I removed the one in `encode_quotes()`. `encoding_failures_total` now counts orders only.

clients/rust/src/types.rs
tamaralipows: Agreed, renamed to `EncodingFailed` to match `NoRouteFound` and `PriceCheckFailed`. The wire value is now `encoding_failed`. `SolveError::FailedEncoding` keeps its name.

Troshchk (approved):
Thank you! ✨

.claude/CODEBASE.md
```diff
@@ -85,7 +85,9 @@ See `docs/ARCHITECTURE.md` for the full architecture diagram and detailed compon
 3. Each pool's `TaskQueue` dispatches to a `SolverWorker` on a dedicated OS thread
 4. Worker resolves request exclusions with `MarketState::resolve_route_filter`, then calls `Algorithm::find_best_route` with a `SolveRequest` carrying its local graph, shared market/derived data, and resolved pools/tokens. It rejects returned routes that violate the request filter, including excluded pAMM fallback pools
 5. `WorkerPoolRouter` collects results, ranks candidates by `amount_out_net_gas` descending; if price guard is enabled it validates in rank order
-6. If `EncodingOptions` provided, `Encoder` produces ABI-encoded calldata
+6. If `EncodingOptions` provided, `Encoder` produces ABI-encoded calldata, one order at a time: an
```
louise-poole: Can we refer to them as 'one quote at a time'? Technically all quotes are for 1 order request so the phrasing here sounds weird.

fynd-core/src/encoding/encoder.rs
```diff
@@ -243,12 +253,21 @@ impl Encoder {
 
     /// Encodes order solutions for execution.
     ///
+    /// Each order is encoded on its own: an order that fails keeps its route and amounts, gets
```
louise-poole: Same here - they're multiple quotes we're encoding all for 1 order. The wording needs to reflect that nicer 👀

fynd-core/src/encoding/encoder.rs
```diff
@@ -271,83 +290,161 @@ impl Encoder {
...
+            });
+            match encoded {
+                Ok((gas_estimate, transaction, fee_breakdown)) => {
+                    quotes[idx].set_gas_estimate(gas_estimate);
+                    quotes[idx].set_transaction(transaction);
+                    quotes[idx].set_fee_breakdown(fee_breakdown);
+                }
+                Err(e) => failures.push((idx, SolveError::FailedEncoding(e.to_string()))),
+            }
+        }
```
louise-poole: We build the `encoded` object and then immediately match it. Might be more readable if we handled the Ok and Err parts inline if possible? Skip the match entirely? Let me know if this doesn't make sense 😅

fynd-core/src/encoding/encoder.rs
```diff
@@ -271,83 +290,161 @@ impl Encoder {
...
-                        return Err(SolveError::FailedEncoding(
-                            "quote routes through an exclusive pool but no signing key is \
-                             configured (set EXCLUSIVE_SWAP_CONTROLLER_KEY)"
-                                .to_string(),
-                        ));
-                    }
-                    solution
+            match self.prepare_solution(quote, &encoding_options, &router_fees) {
+                Ok((solution, fee_breakdown, fee_rates)) => {
+                    to_encode.push((i, solution, fee_breakdown, fee_rates));
```
louise-poole: This tuple appears in multiple places. Maybe it should be made a named struct instead to make it more readable and maintainable?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -334,6 +334,10 @@ impl RankedQuotes {
 
 /// Encodes successful order quotes into router calldata, recording the encoding metrics the
 /// HTTP service reports.
+///
+/// `encoding_failures_total` counts orders, and [`Encoder::encode`] increments it for each order
+/// that fails. An error returned here fails the whole call, and `http_requests_total` counts it
+/// under its error status.
```
louise-poole: This comment is misplaced now that you removed `encoding_failures_total` from here...

louise-poole (approved):
Approved with some minor comments (proudly spotted with my own eyeballs lol)

.claude/CODEBASE.md
tamaralipows: Done, the step now says one quote at a time.

fynd-core/src/encoding/encoder.rs
tamaralipows: Done, the docstring and the failure log now say quote.

fynd-core/src/encoding/encoder.rs
tamaralipows: Makes sense. The build-then-match is gone: `attach_transaction` writes the transaction to the quote and uses `?`, and the loop pushes a failure when it errors.

fynd-core/src/encoding/encoder.rs
tamaralipows: Done, the tuple is now `PreparedSolution`.

fynd-core/src/worker_pool_router/mod.rs
tamaralipows: Removed. The docstring now describes the histogram this function records and where a per-quote failure is reported.

PR #537, opened 2026-09-11
feat!: pin exclusive swap payloads to the quoted amounts

louise-poole:
A signed exclusive-swap authorization used to permit a swap of any size, in either
  direction, at the signed fee. `minBalanceUpdate` — the only field in the payload that
  can constrain the trade — was sent as the accept-anything sentinel. On 8 September a
  bot lifted one out of the mempool and spent it through our router at an amount we
  never quoted.

  It now carries two bounds on the pool's balance deltas:

  - **Input floor** — the pool must gain at least the quoted input, less the request's
    own slippage. Deriving it from the request means the leg never reverts on drift the
    router's `min_amount_out` would have accepted anyway.
  - **Output cap** — the pool may give up at most the quoted output plus a tolerance,
    default 5000 bps, set per deployment by `EXCLUSIVE_SWAP_OUTPUT_CAP_BPS`. The default
    is wide because a leg at the end of a route absorbs the drift of every hop before it,
    and a tight cap would revert ordinary trades in thin pools.

  The cap rejects an oversized swap; the floor rejects the reverse direction.

  ## Still open

  Replaying the *exact* quoted trade is still possible. Closing that needs the payload to
  name its taker, and the extension has no field for it. The cap's 5000 bps default is a
  starting guess; `hindsight` can size it from measured drift.


fynd-core/src/encoding/exclusive_swap.rs
```diff
@@ -22,16 +28,82 @@ use alloy::{
...
+    bps,
     encoding::{now_unix_secs, DEFAULT_DEADLINE_WINDOW_SECS},
     SolveError, Swap,
 };
 
 /// Environment variable holding the pool controller's private key (hex, with or without `0x`).
 pub(crate) const ENV_CONTROLLER_KEY: &str = "EXCLUSIVE_SWAP_CONTROLLER_KEY";
 
+/// Environment variable overriding [`DEFAULT_OUTPUT_CAP_TOLERANCE_BPS`], read when the signer is
+/// built from the environment.
```
tamaralipows: Don't think we should put where the thing is used in the docstrings

fynd-core/src/encoding/exclusive_swap.rs
```diff
@@ -152,9 +242,14 @@ impl ExclusiveSwapSigner {
...
-        // signed bytes cannot execute them through its own contract.
+        // signed bytes cannot execute them through its own contract. The locker is the router,
+        // which anyone may call, so the pinned balance minimums do the rest of the work.
         let meta = signed_swap_meta(deadline, fee, nonce, self.authorized_locker);
-        let min_balance_update = min_balance_update_accept_any();
+        let min_balance_update = pinned_min_balance_update(
+            swap,
+            input_floor_factor_bps(slippage),
+            bps::DENOMINATOR.saturating_add(self.output_cap_tolerance_bps),
+        )?;
```
tamaralipows, line 253: I worry this might not be lenient enough. We are already seeing insane positive slippage on chains like Robinhood (though I know this is only on mainnet for now), but also in general for memecoins where slippage could be up to 45%. Like perhaps we should take the max of the slippage and BPS tolerance and cap it at that? That would still already prevent us from going from $20 to 100k at least. Not sure. 

**EDIT:** I see we are just flooring the min, not capping it, so this only floors negative slippage. So maybe we should just raise the output amount or take the max of that and the slippage? 100 BPS is only 1%

fynd-core/src/encoding/exclusive_swap.rs
```diff
@@ -194,12 +289,59 @@ fn signed_swap_meta(deadline: u32, fee: u32, nonce: u64, authorized_locker: Addr
     B256::from(word)
 }
 
-/// Returns the `PoolBalanceUpdate` that accepts any swap output (`delta0 = delta1 = i128::MIN`).
-fn min_balance_update_accept_any() -> B256 {
+/// Builds the `PoolBalanceUpdate` that pins a swap to the amounts it was quoted for.
+///
+/// The extension compares the realized pool balance deltas against these minimums, and applies its
+/// own fee only after that check. The pool gains `token_in`, so a minimum there floors what the
+/// taker pays in; it loses `token_out`, so a negative minimum there caps the pool's gross outflow —
```
tamaralipows: "so a minimum there floors what the taker pays in"
=> "a minimum caps the input amount".

anthropic alien language -_-

fynd-core/src/encoding/exclusive_swap.rs
```diff
@@ -194,12 +289,59 @@ fn signed_swap_meta(deadline: u32, fee: u32, nonce: u64, authorized_locker: Addr
...
+    word[0..16].copy_from_slice(&delta0.to_be_bytes());
+    word[16..32].copy_from_slice(&delta1.to_be_bytes());
+    Ok(B256::from(word))
+}
+
+/// Scales `amount` by `factor_bps` and converts it to an `i128` delta magnitude; the caller applies
+/// the sign the pool's side of the swap calls for. `None` when the result exceeds `i128::MAX`.
+fn pinned_delta(amount: &BigUint, factor_bps: u32) -> Option<i128> {
+    bps::scale_truncating(amount, factor_bps).to_i128()
 }
```
tamaralipows: can we remove this passthrough

fynd-core/src/encoding/exclusive_swap.rs
```diff
@@ -271,9 +413,18 @@ fn pool_id(token0: &[u8], token1: &[u8], config: B256) -> Result<B256, SolveErro
...
 
+/// Whether `token_in` is the pool's `token0`, the lower of the two addresses.
+///
+/// The `poolId` the signature covers and the delta halves it pins both order by this rule, so it
+/// lives in one place: if the two ever disagreed, the extension would check the input floor against
+/// the output token.
+fn token_in_sorts_first(token_in: &[u8], token_out: &[u8]) -> bool {
+    token_in <= token_out
+}
```
tamaralipows: Do we also need an extra fn for an inequality

tamaralipows (commented):
Thank you, code itself looks good. I have some nitpicks. And overall I'm worried about a 1% restriction on output positive slippage. Let's think about raising that default or also using the user-inputted slippage value there (if bigger)?

fynd-core/src/encoding/exclusive_swap.rs
louise-poole: So my thought here is: we're the only one routing through this pool. So it's not like we're competing with tons of flow that will cause slippage to be dramatic. 
Though I am happy to raise the output amount. I made it configurable via an env var so that we could do just that. Do you want the inline default to be bigger too? Maybe we set it to 50%?

tamaralipows: Slippage would rather be dramatic when it's the last pool of a sequence, depending on the slippage of preceding pools - not just because of the slippage this pool alone. I think 50% would be good!

fynd-core/src/encoding/exclusive_swap.rs
Troshchk: Is outflow also an alien language?

fynd-core/src/worker_pool_router/mod.rs
```diff
@@ -86,11 +84,12 @@ fn user_improvement_share_bps_env() -> u32 {
     match parse_user_improvement_share_bps(&raw) {
         Some(bps) => bps,
         None => {
+            let denominator = bps::DENOMINATOR;
```
Troshchk, line 87: Can we just import `bps::DENOMINATOR` on top? Applies to every `bps::DENOMINATOR`

Troshchk (commented):
Thank you!
Just two minor things

fynd-core/src/worker_pool_router/mod.rs
louise-poole: I kind of liked the qualifier here because the word 'DENOMINATOR' is very generic. `bps::DENOMINATOR` is short enough to have inline and make it clear what kind of denominator we're using 👀

fynd-core/src/encoding/exclusive_swap.rs
louise-poole: Dropped, it now just says what the variable overrides. Fixed in ca0f0b62.

fynd-core/src/encoding/exclusive_swap.rs
louise-poole: Good point about the last pool in a sequence, I hadn't thought about the drift accumulating across hops. Default is now 5000 bps and the doc comment gives that reason. `EXCLUSIVE_SWAP_OUTPUT_CAP_BPS` still overrides it per deployment. ca0f0b62

fynd-core/src/encoding/exclusive_swap.rs
louise-poole: Rewritten, and "outflow" is gone too:

```
/// The pool gains `token_in`, so the minimum on that side is the smallest input the swap may pay;
/// it loses `token_out`, so the negative minimum on that side is the largest output the pool may
/// pay out, before the fee.
```

I didn't take "a minimum caps the input amount" literally: the minimum is on the pool's *gain* of `token_in`, so it's a floor on the input, not a cap. Saying "caps" would flip the meaning. ca0f0b62

fynd-core/src/encoding/exclusive_swap.rs
louise-poole: Removed, `bps::scale_truncating(...).to_i128()` is inlined at both call sites now. ca0f0b62

fynd-core/src/encoding/exclusive_swap.rs
louise-poole: No, dropped it. Both sites compare `token_in <= token_out` directly. The note about the two orderings having to agree moved onto `sorted_tokens`. ca0f0b62

fynd-core/src/worker_pool_router/mod.rs
Troshchk: okok, makes sense, lets keep it 👍

tamaralipows (approved):
Thank you!

PR #538, opened 2026-09-11
feat: expose price data status

tvinagre:
## Summary

- Replace the experimental `/v1/prices.blocks` field with required `data_status`.
- Report the selected Tycho head and source age with the same semantics as `/v1/health`.
- Report the persisted block and monotonic update age for token prices, spot prices, and component depths.
- Return `NOT_READY` when a requested computation has never persisted; reserve `STALE_DATA` for data that exists but violates a freshness rule.
- Preserve gas-token-only pricing readiness, optional payload selection, derived-data error precedence, and existing `/v1/health` behavior.
- Regenerate the OpenAPI and TypeScript contracts and align the generation documentation.
- Validate the integrated head with cargo check, strict Clippy except the unchanged `fetch_update` deprecation, rustdoc except current-`main` baseline documentation lints, 1,472 workspace tests, 179 RPC tests, 134 derived tests, 167 TypeScript tests, and idempotent contract generation.
- Keep all plan files outside the PR diff.


fynd-rpc/src/api/handlers.rs
```diff
@@ -292,30 +296,84 @@ pub async fn get_prices(
...
+            .cloned()
+            .ok_or(ApiError::StaleData { age_ms: u64::MAX })?;
+        let spot_prices_status = store
+            .spot_prices_status()
+            .map(|status| ComputationDataStatus {
+                block: status.block(),
+                last_update_ms: status.age_ms_at(now),
+            });
+        if want_spot && spot_prices_status.is_none() {
+            return Err(ApiError::StaleData { age_ms: u64::MAX });
```
tvinagre: Shouldn't this be NotReady instead of StaleData? Since the data is not there, should mean that it was never calculated.
There should be a way to differenciate between StaleData and NotReady

fynd-rpc/src/api/mod.rs
```diff
@@ -120,6 +124,27 @@ pub fn openapi_spec() -> utoipa::openapi::OpenApi {
     openapi
 }
 
+fn block_age_ms_at(timestamp_secs: u64, now_secs: u64) -> u64 {
```
tvinagre: This function is unnecessary and its tests are asserting behavior of a code that is not our implementation. We should test block_age_ms_at_time behaviour directly.

fynd-rpc/src/api/handlers.rs
```diff
@@ -292,30 +296,92 @@ pub async fn get_prices(
...
+            component_depths_data,
+            component_depths_status,
+        )
+    };
+
+    let tycho_status = state
+        .health_tracker()
+        .tycho_head_status()
+        .await
+        .ok_or_else(|| ApiError::NotReady("Tycho head is unavailable".to_string()))?;
```
brunoguerios, line 354: `/v1/prices` now returns 503 when there is no Tycho head, even with token prices computed. The PR description says the head is reported, not required. Is there a startup window where derived data lands before `update_last_updated` runs? If yes, a 200 becomes a 503 in that window. Worth one line in the description either way.

fynd-rpc/src/api/mod.rs
```diff
@@ -120,6 +124,25 @@ pub fn openapi_spec() -> utoipa::openapi::OpenApi {
...
+            now.as_secs()
+                .saturating_sub(timestamp_secs)
+                .saturating_mul(1000)
+        })
+}
+
+fn tycho_head_status_at(head: BlockInfo, now: SystemTime) -> TychoHeadStatus {
+    let last_update_ms = block_age_ms_at_time(head.timestamp(), now);
+    TychoHeadStatus { head, last_update_ms }
+}
```
brunoguerios: `tycho.last_update_ms` is wall clock minus block timestamp, in whole seconds. `computations.*.last_update_ms` is monotonic milliseconds since the store write. Same name, different clocks. Do we expect integrators to compare the two? If not, the `TychoDataStatus.last_update_ms` doc should say it is second-granular and wall-clock based.

fynd-rpc-types/src/lib.rs
```diff
@@ -1075,10 +1075,7 @@ pub enum QuoteStatus {
     PriceCheckFailed,
 }
```
brunoguerios, line 1080: This doc also renders on `QuoteResponse.block` in the stable `/v1/quote` contract. The old text told quote consumers that a quote is valid only for its block. The new one drops that. Suggest moving the dropped sentence to the `OrderQuote.block` field doc; field descriptions survive the `$ref` in the generated spec.

fynd-rpc/src/api/handlers.rs
```diff
@@ -292,30 +296,92 @@ pub async fn get_prices(
...
+            ComputationDataStatus {
+                block: token_prices_status.block(),
+                last_update_ms: token_prices_status.age_ms_at(now),
+            },
+            spot_prices_data,
+            spot_prices_status,
+            component_depths_data,
+            component_depths_status,
+        )
+    };
```
brunoguerios: Non-blocking, fine after the FE launch. The store keeps output and status in one `ComputedSlot`, but this block re-pairs them by hand three times, so the three `Internal("... status exists without output")` branches cannot fire. One store accessor that returns `(output, status)` removes them. A `ComputationDataStatus::at(status, now)` constructor removes the triplicated mapping. A small named struct replaces the 6-tuple. Checking the Tycho head before the clones also avoids the clone cost on that 503 path.

fynd-rpc/src/api/mod.rs
brunoguerios: `tycho_head_status_at` and `tycho_head` have one caller each and exist for the clock injection in `test_tycho_head_status`. That test then never calls `tycho_head_status`. Suggest inlining both into `tycho_head_status` and asserting the head fields through it. `test_block_age_ms_at_time` already covers the arithmetic.

fynd-rpc/src/api/mod.rs
```diff
@@ -319,8 +346,172 @@ pub(crate) fn configure_app(
...
+                "ComputationDataStatuses.{optional_property} must reference the status schema"
+            );
+            assert!(
+                !property
+                    .to_string()
+                    .contains("\"null\""),
+                "ComputationDataStatuses.{optional_property} must be omitted, not nullable"
+            );
+        }
+    }
```
brunoguerios: The CI drift check already compares `openapi_spec()` with the committed `clients/openapi.json`, so the shape assertions here duplicate a reviewed artefact. The test also pins doc phrases (`"Tycho ready synchronizer"`, `!contains("quote")`), so any doc edit fails it. Suggest keeping only the `nullable = false` check on `spot_prices` / `component_depths` and deleting the rest.

fynd-rpc/src/api/handlers.rs
```diff
@@ -767,12 +889,118 @@ mod tests {
...
+        assert_eq!(computations["spot_prices"]["block"], 18_999_999);
+        assert_eq!(computations["component_depths"]["block"], 18_999_998);
+    }
+
     #[cfg(feature = "experimental")]
     #[actix_web::test]
     async fn test_prices_handler_applies_limit_boundaries() {
         use num_bigint::BigUint;
 
+        const MAX_TEST_COMPUTATION_AGE_MS: u64 = 60_000;
```
brunoguerios: `data_status` assertions are spread over the limit and decimal tests, `MAX_TEST_COMPUTATION_AGE_MS` is declared twice, and the status loop is repeated. Suggest one `test_prices_handler_data_status` that owns the head, `blocks`-absence and age assertions, and strip them from the other tests.

fynd-rpc/src/api/handlers.rs
```diff
@@ -684,6 +774,36 @@ mod tests {
         )
         .await;
         assert_eq!(resp.status().as_u16(), 503);
+        let body = body_json(resp).await;
+        assert_eq!(body["code"], "NOT_READY", "body was: {body}");
+        assert_eq!(body["error"], "data not ready: Tycho head is unavailable");
+    }
```
brunoguerios: The four `NOT_READY` tests share the same scaffold, and `..._preserves_token_not_ready_error_precedence` differs from `..._returns_not_ready_without_token_prices` only by `seed_tycho_head`. A table with `(seed_head, seed_token_prices, uri, expected_error)` rows, like the one in `..._for_requested_uncomputed_data`, covers all four.

fynd-core/src/derived/store.rs
```diff
@@ -371,6 +432,131 @@ mod tests {
...
+            10,
+            updated_at,
+        );
+        store.set_output_at(SpotPriceComputation::ID, SpotPrices::default(), 11, updated_at);
+
+        store.clear_token_prices();
+
+        assert!(store.token_prices_status().is_none());
+        assert!(store.spot_prices_status().is_some());
+    }
```
brunoguerios: Status and output live in the same slot, so these two clear tests cannot fail independently of `test_clear_all_resets_all_fields` and `test_clear_token_prices_clears_failure_map`. Suggest dropping them, or adding one `token_prices_status().is_none()` assert to the existing tests.

fynd-rpc/src/api/prices.rs
```diff
@@ -284,6 +321,63 @@ mod tests {
...
+
+        assert_eq!(
+            serde_json::to_value(computations).unwrap(),
+            serde_json::json!({
+                "token_prices": { "block": 10, "last_update_ms": 100 },
+                "spot_prices": { "block": 11, "last_update_ms": 200 },
+                "component_depths": { "block": 12, "last_update_ms": 300 },
+            })
+        );
+    }
```
brunoguerios: All fields are `Some` here, so this only checks serde. The `skip_serializing_if` omission is the behavior worth guarding, and `test_data_status_serialization` already covers it. Suggest dropping this test, or giving the first test one `Some` and one `None` optional.

fynd-rpc/src/api/mod.rs
```diff
@@ -319,8 +346,172 @@ pub(crate) fn configure_app(
         }));
 }
 
+#[cfg(test)]
+mod health_tracker_tests {
+    use std::sync::Arc;
+
+    use fynd_core::types::BlockInfo;
```
brunoguerios: `use super::*;` already imports `BlockInfo`; this line is redundant.

brunoguerios (commented):
Thanks, `data_status` is clearer than the old `blocks` field. Two questions and a set of test cleanups. Nothing here blocks the merge.

**`/v1/tokens` still returns `STALE_DATA` for never-computed prices.** `get_tokens` has two `ok_or(ApiError::StaleData { age_ms: u64::MAX })` sites (`handlers.rs:534`, `handlers.rs:550`). `/v1/prices` now returns `NOT_READY` for the same store state, so the two endpoints disagree on the error code. The fix looks small: switch both sites to `NotReady` and update the tokens tests. Is that in scope here, or a follow-up?

**Handler doc.** The `get_prices` doc still says "Returns 503 until the first token-price solve has landed", and the utoipa 503 entry says "Data not yet available". Please mention the `NOT_READY` code and the new Tycho-head cause in both.

brunoguerios (approved):
Posted a review with some suggestions, but approving anyway since none of them are blockers ✅

fynd-rpc/src/api/handlers.rs
tvinagre: There's no price computation without Tycho head, so I'd say this can't happen. Worth the comment though.

fynd-rpc/src/api/mod.rs
tvinagre: Relevant comment. `last_update_ms` is in milliseconds, but clipped at second precision because the block timestamp is unfortunately in `seconds`. Integrators should not compare them, but the worst deviation is 999ms in this case. Worth a comment

PR #552, opened 2026-09-18
feat(clients): add zero-fee ClientFeeParams constructor

louise-poole:
Adds a helper for clients who want to identify themselves without charging a fee:

  - Rust: `ClientFeeParams::zero(receiver, deadline)`
  - TypeScript: `zeroClientFee(receiver, deadline)`
  
  Why: the router only knows who the client is from the fee receiver address. If you don't send client fee params, it assumes the transaction sender is the client, and any custom fee rates for your address won't apply. Passing `bps: 0` already fixed that, but there were no docs for it.

  This is name made easy with the new zero-fee constructors and the client fees guide has been updated with examples in both languages.


tamaralipows (approved):
Thanks!

